PR #152, opened 2015-01-16
GlassFish Upstream Cherry-Pick 13/01/2015 - 15/01/2015

Pandrex247:
Cherry-Pick of commits from GlassFish upstream.

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -63,7 +63,7 @@
         <jms-api.version>2.0</jms-api.version>
         <jsp-api.version>2.3.2-b01</jsp-api.version>
         <jsp-impl.version>2.3.3-b02</jsp-impl.version>
-        <jstl-impl.version>1.2.3</jstl-impl.version>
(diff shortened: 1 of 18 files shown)


jerrinot:
cherry-picking changes from upstream and merging them directly to the Payara master doesn't sound like a sustainable strategy. It will become a nightmare to maintain eventually. 

If you really have to cherry-pick changes I'd do it to a different branch. This branch would contain all cherry-picked changed from the upstream and Payara master would just merge it in from time to time. This would give you a clear separation of your own development/fixing from the incoming changes from upstream.

Anyway, cherry-picking is only really good if you really have to skip some commits from the upstream branch otherwise there are better ways to propagate upstream changes to Payara.

smillidge:
This is mainly to remove 100s of pom conflicts, during a wholesale upstream branch merge, as our POM maven coordinates are different. If you can think of a better way please help.

I like the idea of the separate cherry pick branch.

PR #158, opened 2015-01-19
Integrate JBatch 1.0.1-b4

smillidge:
Integrate the released JBatch upgrade, now it is on Maven Central and remove dependency on Payara patched JBatch.

Also fixes a bug in the MySQL Persistence where checking the schema exists failed.

### DIFF
--- a/appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/MySqlPersistenceManager.java
+++ b/appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/MySqlPersistenceManager.java
@@ -119,16 +119,12 @@ public class MySqlPersistenceManager extends JBatchJDBCPersistenceManager {
 
 		try {
 			conn = getConnectionToDefaultSchema();
-			dbmd = conn.getMetaData();
-			rs = dbmd.getSchemas();
+                        PreparedStatement stmt = conn.prepareStatement("SHOW DATABASES like ?");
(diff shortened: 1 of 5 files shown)


appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/MySqlPersistenceManager.java
```diff
@@ -119,16 +119,12 @@ private boolean isMySQLSchemaValid() throws SQLException {
 
 		try {
 			conn = getConnectionToDefaultSchema();
-			dbmd = conn.getMetaData();
-			rs = dbmd.getSchemas();
+                        PreparedStatement stmt = conn.prepareStatement("SHOW DATABASES like ?");
```
jerrinot: you should close the statement

appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/MySqlPersistenceManager.java
```diff
@@ -119,16 +119,12 @@ private boolean isMySQLSchemaValid() throws SQLException {
...
+                        stmt.setString(1, schema);
+			rs = stmt.executeQuery();
 
 			while (rs.next()) {
```
jerrinot: perhaps replacing "while" with "if" would make it more readable?

smillidge:
I agree and just pushed a commit to fix

PR #179, opened 2015-01-29
GlassFish upstream cherry pick 29-01-2015

Pandrex247:
Cherrypick of glassfish upstream.


appserver/transaction/jts/src/main/java/com/sun/jts/CosTransactions/RecoveryManager.java
```diff
@@ -219,7 +219,7 @@ static void initialise() {
             try {
                 recoveryInProgress.post(); // BUGFIX (Ram Jeyaraman)
                 resyncComplete(false, false);
-            } catch (Throwable exc) {}
+            } catch (Throwable exc) {exc.printStackTrace();}
```
jerrinot: this is nasty. it's slightly less nasty than silently swallowing the exception, but still nasty:)

PR #210, opened 2015-02-17
https://github.com/payara/Payara/issues/207

davewinters:


appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/DB2PersistenceManager.java
```diff
@@ -201,7 +201,7 @@ protected void createDB2TableNotExists(String tableName,
 			stmt = conn.createStatement(ResultSet.TYPE_SCROLL_INSENSITIVE,
 					ResultSet.CONCUR_READ_ONLY);
 			String query = "select name from sysibm.systables where name ="
-					+ "\'" + tableName + "\'" + "and type = 'T'";
+					+ "\'" + tableName.toUpperCase() + "\'" + "and type = 'T'";
```
jerrinot: I have learnt toUpperCase() is not completely risk-free. This test is failing on e.g. Turkish locale:
```java
@Test
    public void testLocale() {
        String original = "i";
        String expectedOutcome = "I";

        assertEquals(expectedOutcome, original.toUpperCase());
    }
```

```
org.junit.ComparisonFailure: 
Expected :I
Actual   :İ
```

Some languages have weird rules :) I'm not sure what's best solution. Perhaps you deserve it when you are using something else than en_US on your server :)))

davewinters:
he he .. we do have multi-language build which support approx 10 languages but we do not cater for these scenario's in our code. Interesting point though. One could of course fork payara and build on their environment which will then use their Locale in toUpperCase() etc..

jerrinot:
isn't this driven by a default locale of a system Payara is running on? 
If I take your binary and start it on my system then a JVM will use a default locale of my system, won't it?

davewinters:
Yes it is.

Issue #139, opened 2014-12-31
Fix exception handling code in com.sun.enterprise.util.io.FileUtils.java


davewinters:
Exceptions are swallowed in multiple locations throughout this file.

davewinters:
https://github.com/payara/Payara/issues/139

PR #212, opened 2015-02-17
https://github.com/payara/Payara/issues/139

davewinters:


nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -1336,7 +1349,13 @@ public static String resourceToString(String resourceName) {
                     is.close();
                 }
                 catch (Exception ex) {
-                    // ignore...
+                	_utillogger.log(Level.SEVERE, CULoggerInfo.exceptionIO, ex);
+                }finally{
+                	   try {
+						is.close();
```
jerrinot: this is a very weird construct. Why there is is.close() on 3 places? 1 should be enough:)
Why no to add finally to the outer try/catch block and close it there?

PR #250, opened 2015-04-14
Grizzly 2.3.19 update

davewinters:


smillidge:
Dave can you check this as the upstream commit to Grizzly 2.3.19 has a small additional edit

https://github.com/payara/Payara/commit/8310c0e7699baa54fb81909f936e9fc1a4aa8fe4

PR #257, opened 2015-04-20
Grizzly 2.3.19 further updates

davewinters:


smillidge:
Have we already got the changes to  appserver/web/web-core/src/main/java/org/apache/catalina/ssi/SSIMediator.java

davewinters:
we already have the changes from SSIMediator.java so just updated the copyright year and grizzly version in packager .py .. #258

Issue #269, opened 2015-04-28
Nightly build 2015-04-28 does not allow creating jdbc connection pool resources through web interface


matthiasblaesing:
Steps to reproduce:
- get fresh copy of "Payara 4.1.152 (Full Java EE)"
- Open web console
- Go to: Resources->JDBC->JDBC Connection Pools and select "New"
- Pool Name: XY, Resource Type: "javax.sql.DataSource", Database Driver Vendor: "Derby"
- Press next

=> results in HTTP Status 500 - Internal Server Error

The error reporting in the server.log gives not too much information:

```
[2015-04-28T20:20:37.510+0200] [Payara 4.1] [INFO] [] [org.glassfish.admingui] [tid: _ThreadID=107 _ThreadName=admin-listener(6)] [timeMillis: 1430245237510] [levelValue: 800] [[
  Exception Occurred :null]]

[2015-04-28T20:21:20.856+0200] [Payara 4.1] [WARNING] [] [javax.enterprise.web] [tid: _ThreadID=44 _ThreadName=admin-listener(1)] [timeMillis: 1430245280856] [levelValue: 900] [[
  StandardWrapperValve[FacesServlet]: Servlet.service() for servlet FacesServlet threw exception
java.lang.IllegalStateException: getOutputStream() has already been called for this response
	at org.apache.catalina.connector.Response.getWriter(Response.java:777)
	at org.apache.catalina.connector.ResponseFacade.getWriter(ResponseFacade.java:224)
	at com.sun.faces.context.ExternalContextImpl.getResponseOutputWriter(ExternalContextImpl.java:851)
	at com.sun.faces.context.PartialViewContextImpl.createPartialResponseWriter(PartialViewContextImpl.java:504)
	at com.sun.faces.context.PartialViewContextImpl.access$300(PartialViewContextImpl.java:79)
	at com.sun.faces.context.PartialViewContextImpl$DelayedInitPartialResponseWriter.getWrapped(PartialViewContextImpl.java:642)
	at javax.faces.context.PartialResponseWriter.startDocument(PartialResponseWriter.java:120)
	at com.sun.faces.context.AjaxExceptionHandlerImpl.handlePartialResponseError(AjaxExceptionHandlerImpl.java:201)
	at com.sun.faces.context.AjaxExceptionHandlerImpl.handle(AjaxExceptionHandlerImpl.java:126)
	at javax.faces.context.ExceptionHandlerWrapper.handle(ExceptionHandlerWrapper.java:100)
[...]

PR #271, opened 2015-04-29
Issue #269

Pandrex247:
Fixes error and keeps original weighting.
Only removes weighting from @Produces annotations with only 1 media type, which seems to be what causes error.


Pandrex247:
Fixes #269

PR #274, opened 2015-05-01
Increment POM versions and change version suffix to 4.1.153-SNAPSHOT

Pandrex247:
Increment version number for next release


Pandrex247:
ok to test

PR #278, opened 2015-05-07
Fix Issue 275

### LINKED ISSUE #275: Bug on click Enable Secure Admin button
Steps to reproduce:

1. unzip payara-4.1.152.zip
2. cd payara41/bin
3. ./asadmin start-domain domain1
4. Go to browser, http://localhost:4848
5. Set admin password 
6. On server node, click button **Secure Administration**

Actual result: it notify **An error has occurred**, after that click on button **Enable Secure Admin** will redirect to error page

Expected result: no message appear, and click button **Enable Secure Admin** work properly.

Pandrex247:
Fixes #275


Pandrex247:
I've completely removed the weighting, as it seemed to be causing issues.
I re-ordered a few of the MediaTypes to keep the original "weighting" intact, as Jersey defaults on MediaTypes left-to-right.

I also changed `"text/html"` to the MediaType to make things clearer, though this can be removed if it's deemed a superfluous change that may break something.

PR #295, opened 2015-05-25
Fixes #292

### LINKED ISSUE #292: Fix Payara Micro parameters bug
Fix the bug referenced in #291

smillidge:
Fixes #292 parameter bug on parsing alternateDomain.xml on Payara Micro

### DIFF
--- a/nucleus/payara-modules/payara-micro/src/main/java/fish/payara/micro/PayaraMicro.java
+++ b/nucleus/payara-modules/payara-micro/src/main/java/fish/payara/micro/PayaraMicro.java
@@ -592,7 +592,7 @@ public class PayaraMicro {
                 }
                 i++;
             } else if ("--domainConfig".equals(arg)) {
-                alternateDomainXML = new File(args[i] + 1);
+                alternateDomainXML = new File(args[i + 1]);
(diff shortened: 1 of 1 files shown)


nucleus/payara-modules/payara-micro/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -592,7 +592,7 @@ private void scanArgs(String[] args) {
                 }
                 i++;
             } else if ("--domainConfig".equals(arg)) {
-                alternateDomainXML = new File(args[i] + 1);
+                alternateDomainXML = new File(args[i + 1]);
                 if (!alternateDomainXML.exists() || !alternateDomainXML.isFile() || !alternateDomainXML.canRead() || !alternateDomainXML.getAbsolutePath().endsWith(".xml")) {
                     System.err.println(alternateDomainXML.getAbsolutePath() + " is not a valid path to an xml file and will be ignored");
```
jerrinot: I'd go for fail-fast behaviour here. If you explicitly ask for a non-existing file then it should fail to start.

smillidge: Good idea I'll raise it as a separate issue to make all of them failfast

Issue #319, opened 2015-06-07
master-password file is not created with secure file permissions on cluster nodes


phillipross:
When the change-master-password command with the --savemasterpassword parameter is used on the DAS, the master-password file is created with secure file permissions:

`[payara@web01 ~]$ ls -al /home/payara/payara41/glassfish/domains/domain1/config/master-password
-rw------- 1 payara other 508 Jun  7 01:26 /home/payara/payara41/glassfish/domains/domain1/config/master-password`

But when the command is used on a cluster node, the master-password file is not created with secure permissions :broken_heart: :

`[payara@web02 ~]$ ls -al /home/payara/payara41/glassfish/nodes/web02/agent/master-password
-rw-r--r-- 1 payara other 508 Jun  7 00:35 /home/payara/payara41/glassfish/nodes/web02/agent/master-password`

smillidge:
fixed by #331

PR #317, opened 2015-06-06
master-password file locations

phillipross:
Addresses issue #316 for master-password file management.

DAS, local instances colocated with the DAS and remote nodes are all working now.

In cluster configurations, changing the master password on the DAS will update the master passwords on all nodes with the new master password specified, but the master-password files must be manually created/updated on each node (via asadmin change-master-password --savemasterpassword)to adhere to "no master passwords over the wire" guideline.

Once the master-password file is handled on each node... the remote instances can be started normally.

PR #320, opened 2015-06-07
Set secure file permissions on master-password file on cluster nodes

phillipross:
Fix for issue #319

### DIFF
--- a/nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
+++ b/nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
@@ -49,6 +49,7 @@ import com.sun.enterprise.universal.i18n.LocalStringsImpl;
 import com.sun.enterprise.universal.xml.MiniXmlParser;
 import com.sun.enterprise.universal.xml.MiniXmlParserException;
 import com.sun.enterprise.util.HostAndPort;
+import com.sun.enterprise.util.OS;
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.CommandException;
 
@@ -57,6 +58,7 @@ import org.glassfish.hk2.api.PerLookup;
 
 import java.io.File;
 import java.io.FileFilter;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -179,11 +181,26 @@ public  class ChangeNodeMasterPasswordCommand extends LocalInstanceCommand {
             PasswordAdapter p = new PasswordAdapter(pwdFile.getAbsolutePath(),
                 MASTER_PASSWORD_ALIAS.toCharArray());
             p.setPasswordForAlias(MASTER_PASSWORD_ALIAS, newPassword.getBytes());
-            pwdFile.setReadable(true);
-            pwdFile.setWritable(true);
+            chmod("600", pwdFile);
         } catch (Exception ex) {
             throw new CommandException(strings.get("masterPasswordFileNotCreated", pwdFile),
                 ex);
+        }
+    }
+
+
+    protected void chmod(String args, File file) throws IOException {
+        if (OS.isUNIX()) {
+            if (!file.exists()) throw new IOException(Strings.get("fileNotFound", file.getAbsolutePath()));
+
+            // " +" regular expression for 1 or more spaces
+            final String[] argsString = args.split(" +");
+            List<String> cmdList = new ArrayList<String>();
+            cmdList.add("/bin/chmod");
+            for (String arg : argsString)
+                cmdList.add(arg);
+            cmdList.add(file.getAbsolutePath());
+            new ProcessBuilder(cmdList).start();
         }
     }
 
--- a/appserver/packager/glassfish-ha/pom.xml
+++ b/appserver/packager/glassfish-ha/pom.xml
@@ -132,7 +132,7 @@
               <version>${project.version}</version>
           </dependency>
            <dependency>
-              <groupId>org.glassfish.main.ha</groupId>
+              <groupId>fish.payara.ha</groupId>
               <artifactId>ha-hazelcast-store</artifactId>
               <version>${project.version}</version>
           </dependency>


smillidge:
ok to test

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
```diff
@@ -179,8 +181,7 @@ protected void createMasterPasswordFile() throws CommandException {
             PasswordAdapter p = new PasswordAdapter(pwdFile.getAbsolutePath(),
                 MASTER_PASSWORD_ALIAS.toCharArray());
             p.setPasswordForAlias(MASTER_PASSWORD_ALIAS, newPassword.getBytes());
-            pwdFile.setReadable(true);
```
jerrinot: what's side-effect of this on a non-Unix OS?

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
```diff
@@ -188,6 +189,22 @@ protected void createMasterPasswordFile() throws CommandException {
     }
 
 
+    protected void chmod(String args, File file) throws IOException {
```
jerrinot: Glassfish depends on Java 7, doesn't it? There is a native support for Posix permissions in NIO2, no need to fiddle with a new process.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
```diff
@@ -188,6 +189,22 @@ protected void createMasterPasswordFile() throws CommandException {
...
+            if (!file.exists()) throw new IOException(Strings.get("fileNotFound", file.getAbsolutePath()));
+
+            // " +" regular expression for 1 or more spaces
+            final String[] argsString = args.split(" +");
+            List<String> cmdList = new ArrayList<String>();
+            cmdList.add("/bin/chmod");
+            for (String arg : argsString)
+                cmdList.add(arg);
+            cmdList.add(file.getAbsolutePath());
+            new ProcessBuilder(cmdList).start();
```
jerrinot: I think this can hang if the implementation of `chmod` produces an output.

phillipross:
So far, the only changes here are to modify the [ChangeNodeMasterPasswordCommand](https://github.com/payara/Payara/blob/master/nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java#L176) class to function the same way that the  [CreateLocalInstanceCommand] (https://github.com/payara/Payara/blob/master/nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/CreateLocalInstanceCommand.java#L311) class operates with respect to master password files.

Copying the chmod implementation from one class to another is ugly, but a "low-risk" short-term change.  The reason I call it low-risk is that it works just as well (or poorly) as it does in the other class without inducing any new apparent side-effects.  In my opinion, it's a non-optimal hack, but it's also easy to merge (into [Payara](http://www.payara.co.uk/) and [Glassfish](http://glassfish.java.net/) both)... and gets the management of the file more secure than it already is, while achieving parity with the local command.

Sadly, I feel that it appears that I'm kicking the can down the road, but these commands, in my opinion, should be refactored, updated to use Java7 NIO facilities, REDOCUMENTED (:grin:) and the class hierarchy should addressed.  This takes more time, more testing, and is a bigger merge which I feel would be less likely to happen into Glassfish in the short term... so, I threw together this first patch just to get things more secure :wink:

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
phillipross: The side affect is that the file will not be explicitly set to be readable by non-owners, but will only have the default permission.  If the default permissions are such that only the owner can read, then these permissions will not be relaxed for the file.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
phillipross: This is true.  The initial swipe at this was meant to only reach parity with how the master-password file is handled locally, but I'll follow up with some testing on an NIO2 implementation that can be used both locally and with node agents.  But first I want to implement it uniformly across both commands and test to make sure it actually works on multiple platforms.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
phillipross: This chmod implementation is problematic for multiple reasons :disappointed: 
This code just makes and relies upon too many assumptions.

If it's only used to set mode 0600 on files... the same thing can be achieved with Java7 NIO2 or maybe even Java6 setReadable/setWritable method 2-parameter variants which specify to only make it readable and writable for the owner, but neither by non-owners.

smillidge:
Hi Phil,

I'm slighly confused by the two pull requests #320 #317 as they seem to have similar commits in them. Any chance you can rationalise the two changes into a single pull request or split them a little cleaner?

Steve

phillipross:
woops, I just realized there's an extra commit in both of them.  I was not able to build without applying bb0972c first, so that commit shows up in both pulls.  Other than that commit, they are isolated changes that merge cleanly and correspond to the individual issues (with a common theme).  I can combine everything into a single pull if that makes things easier and/or more straightforward.

phillipross:
Reworking this pull requests as #331

PR #328, opened 2015-06-10
Add support for access-logging of session attributes

nlmueng:
Was working on glassfish 3.1.2.2, fails on glassfish 4.1.


smillidge:
payara test please

smillidge:
Could you describe a test case?

nlmueng:
Currently is you change the access log format to:

%client.name% %datetime% %request% %status% %response.length% %header.user-agent%  %session.SomeRandomObjecInSessiont%

it reverts back to the default in the output

After the change it works either displaying the session objects data or NULL-SESSION

Issue #335, opened 2015-06-12
glassfish client password file should not be readable by non-fileowners


phillipross:
Currently, when the glassfish client password ($HOME/.gfclient/pass) is created, it's world readable on UNIX.  This file should be have more secure file permissions set immediatly upon creation.

smillidge:
Fixed by #336

Issue #334, opened 2015-06-12
file permissions should be set on glassfish client password file using NIO2 file attribute api rather than chmod-via-ProcessBuilder


phillipross:
Calling out to a chmod binary with a hardcoded path using ProcessBuilder can be dangerous, unwieldy, and problematic. The NIO2 file attributes api should be used instead.

smillidge:
Fixed by #336

PR #336, opened 2015-06-13
master-password & gfclient pass file permissions via NIO2 file attribute api rather than chmod-via-ProcessBuilder

phillipross:
Addresses issues #333, #334, and #335 

Factors out functionality for setting 0600 file mode on a specific file to a common utility class which uses NIO2 Files api with PosixFilePermission to set the mode.  Various classes have been updated to use this utility class rather than using ProcessBuilder.  Proper detection of PosixFileAttributeSupport has been implemented in the utility class to make it safe for platforms such as Windows using filesystems that do not support posix file permissions.

These changes do NOT replace all places in the source code where ProcessBuilder+chmod are used... only the places where explicitly setting 0600 with chmod were used.

### DIFF
--- a/appserver/packager/glassfish-ha/pom.xml
+++ b/appserver/packager/glassfish-ha/pom.xml
@@ -40,7 +40,7 @@
     holder.
 
 -->
-<!-- Portions Copyright [2014] [C2B2 Consulting Limited] -->
+<!-- Portions Copyright [2015] [C2B2 Consulting Limited] -->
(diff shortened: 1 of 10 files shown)


smillidge:
payara test please

phillipross:
No problem Steve... I've pushed two more changes which add the copyright comment to all files touched in this pull request as well as the previous merge that you were referring to.

PR #356, opened 2015-06-28
PAYARA-260 Enable arbitrary keys in logging.properties

### LINKED ISSUE #321: Support all attribute names in set-log-attributes
The set-log-attributes and set-log-levels is almost enough to manage the logging properties for payara instance. Unfortunately set-log-attributes restricts the allowable keys to a set of known values, thus hindering the ability to use new logging properties (i.e. we route our messages over GELF to Logstash). To set these properties we need to modify the logging properties directly on the filesystem and restart payara. It would be much easier if set-log-attributes could allow the setting of any key. As then the logging could be completely configured via asadmin.

smillidge:
This updates the set-log-attributes command to add an additional command argument --validate=false to switch off validation of the attributes against an allowed set of keys.

This allows arbitrary key-values to be set in the logging.properties file

Fixes #321 

For example
```Shell
asadmin> set-log-attributes foo=bar
remote failure: Invalid logging attribute name or value foo=bar.
Command set-log-attributes failed

asadmin> set-log-attributes --validate=false foo=bar
foo logging attribute value set to bar.
The logging attributes are saved successfully for server.
Command set-log-attributes executed successfully.
```

### DIFF
--- a/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/commands/SetLogAttributes.java
+++ b/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/commands/SetLogAttributes.java
@@ -37,16 +37,22 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2015] [C2B2 Consulting Limited]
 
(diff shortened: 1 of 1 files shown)


smillidge:
payara test please

PR #366, opened 2015-07-05
PAYARA-220 fixes #204  still submits executor task during deployment.

### LINKED ISSUE #204: JBatch fails to execute job on @Startup
Hi,

During my testing with jbatch I observed that the batch runtime fails to execute the job if the job gets started via @PostConstruct of a singleton ejb with @Startup-Annotation.
The jobs runs fine, but (in my case) the actual batchlet gets not executed.

I think/guess the probleme is related to the following issue, which I posted last year on the glassfish issue tracker: https://java.net/jira/browse/GLASSFISH-21216

To reproduce:

1). create a simple web-app
2). create a singleton ejb:
```java

@Singleton
@Startup
public class JobStarter {
    private static final Logger LOG = Logger.getLogger(JobStarter.class.getName());
    @PostConstruct
    protected void start() {
        LOG.log(Level.INFO,
                "simple-job with id {0} started...",
                BatchRuntime.getJobOperator().start("simple-job", null));
    }
}

 ```

3). create a simple Batchlet implementation:

```java
@Named
public class Batchlet extends AbstractBatchlet {
    private static final Logger LOG = Logger.getLogger(Batchlet.class.getName());
    @Inject
    private JobContext jobContext;
    @Inject
    private StepContext stepContext;
    @Override
    public String process() throws Exception {
        LOG.log(Level.INFO, "Batchlet {0} of job {1} finished!",
                new Object[]{
                    stepContext.getStepName(),
                    jobContext.getJobName()
                });

        return BatchStatus.COMPLETED.name();
    }
}
   ```

4). and the simple-job.xml file:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<job xmlns="http://xmlns.jcp.org/xml/ns/javaee" 
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/jobXML_1_0.xsd"
     version="1.0"
     id="simple-job">           
    <step id="step-one"> 
        <batchlet ref="batchlet"/>       
    </step>   
</job>
```
On the console you will see the output of the ejb but not the output of the batchlet.

But if you use a @Schedule annotation instead of @PostConstruct, the batchlet output gets printed.

smillidge:
During deployment an @Startup EJB can submit a Batch Job or ManagedExecutorService task. However the Executor Service was checking the domain.xml to see if the referenced application was enabled although its state had not been written to the domain.xml.

This race condition caused the task/job submission to fail. This pull request fixes this.

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2015] [C2B2 Consulting Limited]
 
 package org.glassfish.concurrent.runtime;
 
@@ -206,12 +207,21 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
     }
 
     private boolean isApplicationEnabled(String appId) {
+        boolean result = false;
         if (appId != null) {
             Application app = applications.getApplication(appId);
-            if (app != null)
-                return deployment.isAppEnabled(app);
+            if (app != null) {
+                result = deployment.isAppEnabled(app);
+            } else { 
+                // if app is null then it is likely that appId is still deploying
+                // and its enabled status has not been written to the domain.xml yet
+                // this can happen for example with a Startup EJB submitting something
+                // it its startup method. Reference Payara GitHub issue 204              
+                logger.info("Job submitted for " + appId + " likely during deployment. Continuing...");
+                result = true;
+            }
         }
-        return false;
+        return result;
     }
 
     private ComponentInvocation createComponentInvocation(ComponentInvocation currInv) {


smillidge:
payara test please

Pandrex247:
payara test please

Pandrex247:
payara test please

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -206,12 +207,21 @@ public void reset(ContextHandle contextHandle) {
...
-                return deployment.isAppEnabled(app);
+            if (app != null) {
+                result = deployment.isAppEnabled(app);
+            } else { 
+                // if app is null then it is likely that appId is still deploying
+                // and its enabled status has not been written to the domain.xml yet
+                // this can happen for example with a Startup EJB submitting something
+                // it its startup method. Reference Payara GitHub issue 204              
+                logger.info("Job submitted for " + appId + " likely during deployment. Continuing...");
+                result = true;
```
jerrinot: so you are making a bet that every unknown `appId` is just a "deployment-in-progress" and it'll be enabled eventually? is it a good bet?  what if the `appId` is genuinely unknown? what if it had been just undeployed?

smillidge: Will dig some more to see if there's an alternative mechanism to determine app deployment status. It isn't too fatal making the assumption as the task will just fail if there's some classloader or somesuch problem.

PR #372, opened 2015-07-11
PAYARA-270 enable hazelcast-configuration synchronisation

smillidge:
Also fix length of hazelcast configuration path in admin console so that custom hazelcast configuration files can be placed into the instance configuration directory and replicated to the instance


smillidge:
payara test please

PR #381, opened 2015-07-28
Fix bug whereby could not set context root 

smillidge:
When deploying from an Input Stream in the new PayaraRuntime api a temp context root was created. This is now by default the application name and can be configured.

Also added an undeploy api


smillidge:
payara test please

Pandrex247:
payara test please

PR #389, opened 2015-08-09
PAYARA-331 fixes #371 

### LINKED ISSUE #371: Creating an HTTP listener with security enabled using web admin console doesn't work
This was tested on GlassFish 4.1, JDK 8u45 and Windows 7 (also Windows Server 2012 R2).

Steps to reproduce:

1. Log into web admin console and navigate to Configurations -> server-config -> Network Config -> Network Listeners then click the New button.

2. Name it "test-listener" then press tab, enter port 8443, enter address 0.0.0.0, check the security enabled checkbox, select the http-thread-pool and then save.

3. Open the domain.xml file and search for test-listener until you find this block:

          <protocol name="test-listener-protocol" security-enabled="true">
            <http default-virtual-server="server">
              <file-cache></file-cache>
            </http>
          </protocol>

It is missing the SSL tag.  It is supposed to look like this:

          <protocol name="test-listener-protocol" security-enabled="true">
            <http default-virtual-server="server">
              <file-cache></file-cache>
            </http>
            <ssl classname="com.sun.enterprise.security.ssl.GlassfishSSLImpl" cert-nickname="s1as"></ssl>
          </protocol>

If you restart GlassFish and try to access https://localhost:8443 you'll get some sort of SSL connection related error.   If you then add the ssl tag to the test-listener in domain.xml and restart, you will be able to access https://localhost:8443 in the browser after being warned about self signed certificates.

I think this is a regression.  I don't remember having this problem with GlassFish 3.1.2.2.

smillidge:
Fixes #371 create-protocol CLI command correctly creates ssl subtag if security is enabled. Also disables SSLv3 automatically in the new protocol.

### DIFF
--- a/appserver/web/admin/src/main/java/org/glassfish/web/admin/cli/CreateProtocol.java
+++ b/appserver/web/admin/src/main/java/org/glassfish/web/admin/cli/CreateProtocol.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-
+// Portions Copyright [2015] [C2B2 Consulting Limited]
 package org.glassfish.web.admin.cli;
 
 import com.sun.enterprise.config.serverbeans.Config;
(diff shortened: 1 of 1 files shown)


smillidge:
payara test please

smillidge:
payara test please

PR #390, opened 2015-08-09
PAYARA-370 Fixes confusing typo in the admin console

smillidge:
Admin console help text says the path to files for cluster nodes is payara/nodes this is incorrect.
Should be glassfish/nodes


smillidge:
payara test please

PR #392, opened 2015-08-11
new quicklook configuration

aq47:
New configuration in quicklook files (as tests were not building properly - some lines of code had to be commented out). This is also to incorporate into Jenkins build


Pandrex247:
payara test please

PR #418, opened 2015-08-30
Fixes PAYARA-415

smillidge:
Move deployAll to after Runtime creation  so that injection of @PayaraMicroRuntime works


smillidge:
payara test please

PR #420, opened 2015-09-05
Changed the way to load payara-bootstrap.properties.

mike-neck:
- In order for `PayaraMicro` class to load `payara-bootstrap.properties` in Gradle task execution context, chage the way to load it.
  - Project : payara-modules/payara-micro
  - Class : `fish.payara.micro.PayaraMicro`
  - Line : 1008
  - Before : `ClassLoader.getSystemResourceAsStream(String)`
  - After : `getClass().getClassLoader().getResourceAsStream(String)`
- Reason : In Gradle task execution context, `ClassLoader.getSystemResourceAsStream("payara-bootstrap.properties")` returns `null` so that payara-micro server cannot get started.


smillidge:
payara test please

PR #421, opened 2015-09-05
Payara 447 Add some simple asadmin commands

smillidge:
Adds some simple asadmin commands to list caches and clear a cache.


smillidge:
payara test please

PR #424, opened 2015-09-05
Fixes #423 ensures instance name is set on the Instance Descritor

### LINKED ISSUE #423: Payara Micro name not passed through to the instance descriptor
When you start Payara Micro with --name this is not being passed through the the Instance Descriptor

smillidge:


smillidge:
payara test please

PR #425, opened 2015-09-05
Make Instance Descriptor null safe when zero applications deployed

smillidge:


smillidge:
payara test please

smillidge:
payara test please

PR #436, opened 2015-09-10
PAYARA-452 Fixes #433 versioned application can not be enabled on cluster instance

### LINKED ISSUE #270: Support rolling updates
There are traces here and there about the support of concurrent versions (a.k.a rolling updates) of a single application inside Glassfish (3.1 and later) and thus achieving zero-downtime deployment.
More info:
* https://wikis.oracle.com/display/GlassFish/ProductionRedeploymentOnePager
* http://stackoverflow.com/questions/19292654/rolling-upgrade-application-glassfish-3-1-2-2

### LINKED ISSUE #433: Clustered / Versioned applications cannot be enabled on individual cluster instances
Regression from GlassFish 3.1.2.2
This is an upstream bug: https://java.net/jira/browse/GLASSFISH-21198

smillidge:
Modifies disable command to check the target is equal to the server executing the command. Also fixes #270

### DIFF
--- a/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DisableCommand.java
+++ b/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DisableCommand.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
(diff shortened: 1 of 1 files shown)


smillidge:
payara test please

PR #439, opened 2015-09-14
PAYARA-427 Allow DISABLED as a JMS Broker type. 

smillidge:
This disables the OpenMQ Broker preventing bootstrap and fails fast the jmsra.rar from attempting to create connections to the broker

To fully disable the default connection factory and connection pools should also be deleted.


smillidge:
payara test please

Issue #445, opened 2015-09-17
Not usable for development on Windows (7, 8.1, 10)


marm5r:
Redeploys fail because of:
https://java.net/jira/browse/GLASSFISH-21389

Automated test (e.g. Arquillian) are to slow, hang while tear down (undeploy), maybe related to:
https://java.net/jira/browse/GLASSFISH-21283

smillidge:
Hi, 
o you have a test application we can look at. I would suspect a classloader leak.

smillidge:
Can't reproduce this on windows 7 using CargoTracker https://cargotracker.java.net/ which is a pretty complex app that uses many JavaEE apis. 

Please try one of the latest unstable builds as a few ClassLoader leaks have been squashed for our customers.http://www.payara.fish/upstream_builds

smillidge:
Also test app referred to in GLASSFISH-21283 redeploys in less than 2 seconds on latest builds

marm5r:
https://java.net/jira/browse/GLASSFISH-21389 is still present in the latest unstable build.

I have published a modified version of my test application on GitHub:
https://github.com/marm5r/verify-gf-21389

Please call the ant script with *ant -Dfish=payara-pre* to run the test with Payara. The script will download Payara from http://payara.co.s3-website-eu-west-1.amazonaws.com/payara-prerelease.zip. You can also put payara-prerelease.zip into the root folder of the test application.

When I run the test application with *ant -Dfish=payara-pre* on my Windows machines the following error is logged on every run:
```
     [exec] Command deploy failed.
     [exec] remote failure: Error occurred during deployment: Exception while loading the app : WELD-001201: Error loadi
ng beans.xml jar:file:/D:/scratch/verify-gf-21389/webapp/target/webapp-1.0-SNAPSHOT/WEB-INF/lib/ejb-dep-1.0-SNAPSHOT.jar
!/META-INF/beans.xml. Please see server.log for more details.
```

Here is the complete stacktrace from server.log:
```
[2015-09-24T15:19:56.181+0200] [Payara 4.1] [SEVERE] [NCLS-CORE-00026] [javax.enterprise.system.core] [tid: _ThreadID=54 _ThreadName=admin-listener(2)] [timeMillis: 1443100796181] [levelValue: 1000] [[
  Exception during lifecycle processing
[...]

Issue #434, opened 2015-09-09
Possible WebappClassLoader leak (permgen issue)


dorpsidioot:
While developing and regularly deploying instances on my local server, I am encountering OutOfMemory errors on the PermGen side of things. After examining the jvm with jvisualvm I can confirm that each redeploy causes an increment in the permgen size and forcing GC does not alleviate the problem.

Even with JDK 1.7 options -XX:+CMSPermGenSweepingEnabled and -XX:+CMSClassUnloadingEnabled enabled, the permgen size does not go down.

Close inspection via heapdump shows that for every deploy a new WebappClassloader is created(which is normal) but existing ones do not disappear(started flag is at false). and thus the heap continues to grow.

![permgen_issue](https://cloud.githubusercontent.com/assets/13536281/9757971/ffec165e-56e3-11e5-987b-8aa22e96a312.png)

Heapdump can be downloaded from https://www.dropbox.com/s/o10i4394g2aq5l9/heapdump-payara.zip?dl=0

smillidge:
These classloaders seem to be held in Wicket-PageSavingThread through AmChartsSerializeHandler through a jackson databind ObjectMapper
![screenshot from 2015-09-10 13 47 30](https://cloud.githubusercontent.com/assets/939073/9788553/95f27cc2-57c2-11e5-9403-1f92b7e9f11b.png)

dorpsidioot:
Strangely that should be auto-cleaned after undeployment, I will check where it gets stuck and why, since the ObjectMapper is completely standalone for the whole thing.

dorpsidioot:
Disabled the Objectmapper and removed references for testing and it still retained heap, used classloader-leak-prevention(https://github.com/mjiderhamn/classloader-leak-prevention) and then I do not experience the leaks anymore.
I will be investigating this further to see if I find where the issue lies, if it's on my side then I can fix it here, if not I will come back to you guys with more information.

dorpsidioot:
The classloader-leak-prevention does its job and forces the cleanup, now I am not sure where the problem lies, but I will investigate further.

dorpsidioot:
Update: the problem no longer occurs when deploying, but when using auto-deploy or redeploy it persists, which is quite annoying while developing(encountering the permgen issue once every half an hour now).

dorpsidioot:
Preliminary testing with pre-release build 154 seems to indicate fewer issues with leaks when using auto-deploy/redeploys, testing further in the next few days.

[...]

PR #467, opened 2015-10-03
Payara 457 classloader work

smillidge:
A load of work to fix a number of classloader leaks and filehandles kept open to make Payara Server more robust on redeployment of applications. Should help #434 #445


smillidge:
payara test please

dorpsidioot:
Is this included in the 154 release, if so I will download and test, since we're experiencing the 'inconsistent module state' issue with a clustered 153 setup.

Pandrex247:
Yes, this will be included in the 154 release.

dorpsidioot:
Can confirm improvements with the latest build, redeploys using autodeploy now work seamlessly and under windows development files are no longer in use when undeploying.

PR #471, opened 2015-10-09
PAYARA-405 fixes #415

### LINKED ISSUE #415: PAYARA-405 It is possible to set invalid ejb container thread pool settings
For the EJB container if you set max-threads less than core-threads then the EJB container thread pool will fail to be created. As the defaults are not shown in the console this can easily be achieved by setting max-threads below the default core-threads (16)

Pandrex247:
Fixes #415

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbContainerUtilImpl.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbContainerUtilImpl.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-
+// Portions Copyright [2015] [C2B2 Consulting Limited]
 package com.sun.ejb.containers;
 
 import com.sun.ejb.base.io.EJBObjectInputStreamHandler;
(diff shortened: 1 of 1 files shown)


appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbContainerUtilImpl.java
```diff
@@ -488,6 +488,27 @@ private ThreadPoolExecutor createThreadPoolExecutor(String poolName) {
...
                 : new SynchronousQueue(true);
 
+        // PAYARA-405 validates attributes of the thread pool to ensure no problems      
+        if (corePoolSize < 0) {
+            _logger.log(Level.WARNING, "Core Pool Size configured to be less than 0. Resetting to 0");
+            corePoolSize = 0;
+        }
+        
+        if (maxPoolSize < 1) {
+            _logger.log(Level.WARNING, "Max Pool Size configured to be less than 1. Resetting to 1");
```
jerrinot: isn't it better to fail-fast here?

smillidge: Tricky to fail fast as you would have to force a server abort somehow and this is just creating the Executor pool. Which can happen after a deployment.

Pandrex247:
payara test please

PR #479, opened 2015-10-16
Windows service wrapper works with config

danPowell52:


danPowell52:
payara test please

danPowell52:
payara test please

PR #483, opened 2015-10-21
PAYARA-448: Update CORBA version to 4.0.1-p2

### LINKED ISSUE #465: Could not load class javax.rmi.CORBA.EnumDesc

davidweaver:
Ignore classNotFoundException when attempting to load class of type, if a loaded class is already held
Fixes #465


Pandrex247:
payara test please

PR #484, opened 2015-10-21
PAYARA-389 fixes #414

### LINKED ISSUE #414: PAYARA-389 Disable size based log rotation when set to zero
Setting size based log roation to zero bytes will set size based rotation off in a similar fashion to the time based log rotation.

Pandrex247:
Fixes #414 
Rebase of #470


Pandrex247:
payara test please

PR #485, opened 2015-10-22
PAYARA-388 - Server log displays webappClassLoader severe warnings on shutdown

Pandrex247:
upgraded hk2.version and hk2.plugin.version to 2.4.0.b32
also used hk2-core instead of -core dependency since -core looked obsolete.


Pandrex247:
payara test please

PR #491, opened 2015-10-24
used @WithAnnotations to narrow down the generic type used.

mulderbaba:
We have 2 extensions on board within our codebase, ServerSentEventCdiExtension and JMSCDIExtension. With Weld 2.x these warning occur when we handle generic types in ProcessAnnotatedType.
They need to be narrowed down by using the @WithAnnotations annotation, as advised in the warning message. I first did some tricks with <weld:exclude> but didn't turned out good so moved 
over by modifying the extensions' code.


smillidge:
Payara test please

Issue #511, opened 2015-11-05
Add support for Informix for persisted ejb timers


dmatej:
There are sql scripts for several databases, but Informix is missing.
I have already created the file and table, but still have some problems with SQL syntax error. I will create push request when it will work.

Issue #509, opened 2015-11-03
EjbDeployer.event logic is incorrect on creation of EJB Timers


dmatej:
Code is readable, but there is too many variants. Our case is that we have only nonpersistent timers in clustered application. But application does not start in clustered instances due to missing tables in timer database.
I think I can fix this problem (check schedules and if all are not persistent, we don't need the database), but there is another one - if we would use the persistent timers, problem would be back again.
This will take a little bit more time to check all conditions around that ...

dmatej:
There is also some "voodoo" in EjbDeployer line 583 (method getOwnerId) - zero is not allowed as upper bound: int useInstance = random.nextInt(instances.size()); It causes IllegalArgumentException if there are no instances in cluster (but this is valid!), what should be an owner? Do we need an owner?
What if we start another instance as the first one?

dmatej:
In method clean is also some manipulation with the policy ... it may be connected somehow with the problem with redeploys ... Another our applications without timers have no such problem ...

smillidge:
We are doing some work in the depths of the EJBTimer code as well as with deployment of timers at the moment. There are a number of issues with this code and we are considering a complete refactoring.

smillidge:
Clustered Persistent EJB Timers only fire on a single node in the cluster therefore there is code in there to work out the owner (via the database). I would need to check the semantics of clustered non-persistent EJB Timers. What semantics are you expecting?

We are also in the middle of integrating the embedded Hazelcast Data Grid as a "persistent" EJB Timer store.

dmatej:
Okay, this is nice to hear. I don't know if I can help somehow ... I got only this server log from one instance:
```
[2015-11-03T17:53:50.116+0100] [Payara 4.1] [INFO] [] [javax.enterprise.system.container.ejb.com.sun.ejb.containers] [tid: _ThreadID=15 _ThreadName=RunLevelControllerThread-1446569598689] [timeMillis: 1446569630116] [levelValue: 800] [[
  Loading EJBTimerService. Please wait.]]

[2015-11-03T17:53:50.750+0100] [Payara 4.1] [INFO] [AS-EJB-00054] [javax.enterprise.ejb.container] [tid: _ThreadID=15 _ThreadName=RunLevelControllerThread-1446569598689] [timeMillis: 1446569630750] [levelValue: 800] [[
[...]

PR #513, opened 2015-11-05
Payara #511 Create table SQL for EJB timer persistence on Informix

dmatej:
- since in orm.xml is set delimited-identifiers property, you have to set
  the property IfxDELIMIDENT=true in your JDBC pool configuration. You will
  get syntax error otherwise. Be careful, Informix documentation speaks about y or 1, but in fact, driver's API takes boolean and Glassfish's parser uses Boolean.valueOf.
- also don't forget for IfxCLIENT_LOCALE and other well known settings.


smillidge:
payara test please

danPowell52:
Flow Test Please

dmatej:
Related to #509 - EJB Timer persistence now can be used also with Informix database.

PR #533, opened 2015-11-21
PAYARA-511 fixes #492

### LINKED ISSUE #492: Redeploying war/ear on cluster fails
When deploying an application(war or ear) for the second time on a cluster, the deployment fails with messages that certain parts already exist (virtual server already has a xxx module loaded at...) or are greeted with the 'Error linking in security policy - Inconsistent module state' message in the cluster's logs.

It happens with both simple projects and big projects. Checking the force redeploy checkbox does nothing. We have to restart the cluster instances to alleviate the issue, which kind of beats the purpose of having a cluster. We encounter the issue both at the instances -> applications -> undeploy and also when undeploying via the applications of the DAS.

What's strange is that we don't even get undeployment messages in the nodes when undeploying an application and timers seem to continue running.

smillidge:
Disable command was not executing correctly on a cluster during undeployment due to not matching the cluster target name with the server name. This is a regression introduced by fixing #433 and had the effect of JNDI names not being removed even though the application was correctly unloaded.


danPowell52:
Flow Test Please

danPowell52:
Flow Test Please

PR #548, opened 2015-12-02
Payara 171 excessive gc time

mulderbaba:
first cut of the health-check service with 3 checkers: CPU, Heap, GC. they're configurable via domain.xml


smillidge:
jenkins test please

PR #551, opened 2015-12-05
Speed up Payara Micro Boot Times PAYARA-527

smillidge:
Tweaks to the config files and creation of a new JBatch lazy persistence manager which cuts Payara Micro boot times by 2 thirds


smillidge:
jenkins test please

PR #556, opened 2015-12-08
PAYARA-496 ensure the truststore is only written to when we have read…

smillidge:
…/write permissions


smillidge:
jenkins test please

smillidge:
Something hasn't actually gone wrong the problem reported by customers was was more that exceptions were being thrown during SSL negotiation when it was trying to write to the truststore which in very secure configurations is set as readonly.

PR #563, opened 2015-12-11
PAYARA-406 - Wrong thread name in server log

Pandrex247:
Add check to GFFileHandler to make sure passed GFLogRecord doesn't get overwritten.


Pandrex247:
Closing, spotted something and want to do more testing

PR #566, opened 2015-12-13
Removed cause of spurious warning RAR8705

smillidge:
DataSourceDeployer was not returning null for non-existent properties.


smillidge:
jenkins test please

PR #575, opened 2015-12-19
PAYARA-547 fix #510

### LINKED ISSUE #510: @DataSourceDefinition defined data source can't be used in persistence.xml
Defining a data source from within the application using either @DataSourceDefinition on a class or the data-source element in web.xml, and then using this in persistence.xml will cause a deployment failure.

See https://java.net/jira/browse/GLASSFISH-20944

smillidge:
Use final application classloader to bootstrap the Persistence Unit as it loads driver classes

This fixes #510


smillidge:
jenkins test please

PR #577, opened 2015-12-19
PAYARA-549 Removed check that file for deployment ends in .war 

### LINKED ISSUE #576: Remove Payara Micro deployment restrictions
Payara Micro currently checks to see if a file ends in .war before deploying. It should allow .ear and .jar

smillidge:
now .ear and .jar will also be deployed

Fixes #576


smillidge:
jenkins test please

johnmanko:
Are ear modules still acceptable deployables?  I've read more recent comment here and elsewhere what only war modules are valid for micro deployments.

PR #580, opened 2015-12-22
PAYARA-551 Fixes #536

### LINKED ISSUE #536: Websocket @OnOpen sometimes isn't called
Hi there,
I deployed my project into payara 4.1.1.115 and 4.1.1.116 (coming from glassfish 4.1 open source ed) and everything works fine except for an issue: the chat, sometimes won't work.
It appears that the client can ALWAYS connect and do the handshake with the server but sometimes the @OnOpen won't be called.
Thinking to a bad implementation I created a new project from scratch, consisting in two components:
client: index.htm
serverside: websocket.java

it is the simplest implementation of a websocket I could made, copying the example at https://blog.idrsolutions.com/2013/12/websockets-an-introduction/

The problem persist:
if you open and close the websocket connection, at the 4th time it will "crash". it's able to do the handshake but the @OnOpen method won't be called and you won't be able to exchange data in that channel.
from this time, if you continue opening and closing the connection, it will works randomly.

I tried to upgrade / downgrade java, tried to upgrade to the new tyrus 1.12 and downgrade to 1.8.1 (the same as glassfish 4.1, which actually works flawlessly) and also tried to downgrade to the gf4.1 grizzly version. none of this thing solved the problem.

I was also able to reproduce this behaviour in glassfish 4.1.1 open source edition.
Tried also with different computers with different os and different jdk version but the results are the same.

I think this is a major issue but my competences stops here.

Wasn't able to reproduce the issue in 4.1.153

smillidge:
sets handler initialised variable to false when the Request is recycled.

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
@@ -734,6 +734,7 @@ public class Request
             return;
         }
 
+        handlerInitialised = false;
         context = null;
         servletContext = null;
         contextPath = null;


MeroRai:
jenkins test please

smillidge:
jenkins test please

smillidge:
jenkins test please

PR #597, opened 2016-01-13
PAYARA-519 Update EclipseLink to patched 2.6.2

### LINKED ISSUE #512: Update usages of EclipseLink
1) GF4 still uses JPA1 persistence.xml (minor thing, but ...)
2) NPE in EclipseLink 2.6.1 - only logged, but it seems finally does not make problems and I hope it should be fixed in 2.7.0. I could not find any solution how to get rid of that stacktrace ...
http://stackoverflow.com/questions/32142789/nullpointer-in-log-during-first-connection-to-database
3) Target server in persistence.xml does not change anything, because it is overriden in class PersistenceUnitLoader (this is a corrent solution - platform is then declared even for deployed applications).
So the only update is in that class, because EclipseLink now supports Glassfish and SunAS9 is deprecated now:
```
         final String ECLIPSELINK_SERVER_PLATFORM_CLASS_NAME_PROPERTY = "eclipselink.target-server"; // NOI18N
         props.put(ECLIPSELINK_SERVER_PLATFORM_CLASS_NAME_PROPERTY,
-                System.getProperty(ECLIPSELINK_SERVER_PLATFORM_CLASS_NAME_PROPERTY, "SunAS9")); // NOI18N
+                System.getProperty(ECLIPSELINK_SERVER_PLATFORM_CLASS_NAME_PROPERTY, "Glassfish")); // NOI18N
```
Result with temporarily set more verbose logging:
```
+ <property name="eclipselink.logging.level" value="FINEST"/>
```
```
[2015-11-05T09:05:57.786+0100] [Payara 4.1] [FINEST] [] [org.eclipse.persistence.session./file:/app/appservers/payara4.1.1.161-ICZ151105/glassfish/nodes/karma-cisas1/cis-web/applications/ejb-timer-service-app/WEB-INF/classes/___EJB__Timer__App.properties] [tid: _ThreadID=17 _ThreadName=RunLevelControllerThread-1446710731051] [timeMillis: 1446710757786] [levelValue: 300] [[
  property=eclipselink.target-server; value=Glassfish; translated value=org.eclipse.persistence.platform.server.glassfish.GlassfishPlatform]]
```
But the NPE still remained because it seems the platform is detected before parsing of persistence.xml ... I will report it to EclipseLink ...

Pandrex247:
Updates EclipseLink to a patched 2.6.2.

Fixes #512 
Depends upon https://github.com/payara/Payara_PatchedProjects/pull/13

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -111,7 +111,7 @@
         <jaxb-api.version>2.2.13-b141020.1521</jaxb-api.version>
         <jaxb.version>2.2.12-b141219.1637</jaxb.version>
         <stax2-api.version>3.1.1</stax2-api.version>
-        <eclipselink.version>2.6.1.payara-p1</eclipselink.version>
+        <eclipselink.version>2.6.2.payara-p1</eclipselink.version>
         <javax.xml.registry-api.version>1.0.7</javax.xml.registry-api.version>
         <javax-persistence-api.version>2.1.0</javax-persistence-api.version> 
         <dbschema.version>3.1.1</dbschema.version>


smillidge:
jenkins test please

PR #600, opened 2016-01-13
PAYARA-568 fixes #596 #pr 

### LINKED ISSUE #596: Payara Blue Secure JMX Connector uses SunX509 algorithm
When creating a secure JMX connector on Payara Blue with no key algorithm specified Payara Server tries to initialise the KeyFactory with the SunX509 Key algorithm rather than the IbmX509 algorithm. The key algorithm can be specified explicitly in the domain.xml via;

```xml
      <jmx-connector address="0.0.0.0" port="8686" name="system" auth-realm-name="admin-realm">
          <ssl classname="com.sun.enterprise.security.ssl.GlassfishSSLImpl" cert-nickname="s1as" key-algorithm="IbmX509"></ssl>
        </jmx-connector>
```

The code needs modifying so that the above step does not need to be done.

smillidge:
Changed SunX509 to KeyManagerFactory.getDefaultAlgorithm()


smillidge:
jenkins test please

PR #617, opened 2016-01-25
PAYARA-657 GLASSFISH-21486: EntityManager should not be detached in extended Persistence context 

pdudits:
Fix for [GLASSFISH-21486](https://java.net/jira/browse/GLASSFISH-21486) - extended persistence context violates all contracts of JPA 2.1 by making fetched entities detached.

### DIFF
--- a/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/EntityManagerWrapper.java
+++ b/appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/EntityManagerWrapper.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2015] [C2B2 Consulting Limited]
+// Portions Copyright [2015-2016] [C2B2 Consulting Limited]
 
 package com.sun.enterprise.container.common.impl;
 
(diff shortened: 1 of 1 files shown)


smillidge:
jenkins test please

smillidge:
Just to update we are about to release Payara 4.1.161 so this will be held back until after that as we are into code freeze and testing. 

Could you let us know what we should put into our test suite to demonstrate the problem?

pdudits:
The test for the issue is https://github.com/pdudits/GLASSFISH-21486 - any query invocation in extended persistence context will detach all entities of it, so they are not updated when the ejb starts or join a transaction.

For me this is absolute showstopper for migration to Payara from GF3, because our project relies heavily on [gateway EJBs](http://www.adam-bien.com/roller/abien/entry/building_perfect_anti_facades_almost). It's a pity this wouldn't fit into 161 time frame, as this will postpone our project for three months.

pdudits:
There was an attempt to fix this in [GLASSFISH-20968](https://java.net/jira/browse/GLASSFISH-20968), but only `em.find` was fixed with that issue, not the rest of the query methods.

smillidge:
It's more about testing time on our side to ensure there are no side effects. Once it is merged though it will be available on the pre-releases downloads page as soon as Jenkins finishes the build so you should have a build to start migration against.

pdudits:
Now that 161.1 is out, how would you proceed with this PR? (and the rest of my PRs, I'll comment on them as well :))

I consider this one the most important of all, because IMO the behaviour without the fix is in violation of JPA spec.

smillidge:
We will be pulling them soon. We'll also liaise about testing them to ensure no side effects.

smillidge:
tagged as PAYARA-657 internally

Issue #618, opened 2016-01-27
Deployment of EJB_Timer_App fails with reference to an application class


pdudits:
When deploying an ejb jar into embedded container on 4.1.161-SNAPSHOT following exception is thrown:

````
Exception Description: Predeployment of PersistenceUnit [__EJB__Timer__App] failed.
Internal Exception: Exception [EclipseLink-7250] (Eclipse Persistence Services - 2.6.2.qualifier): org.eclipse.persistence.exceptions.ValidationException
Exception Description: [<an application class>] uses a non-entity [<another application class in same persistence unit>] as target entity in the relationship attribute [field <field>].
	at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl.createPredeployFailedPersistenceException(EntityManagerSetupImpl.java:2035)
	at org.eclipse.persistence.internal.jpa.EntityManagerSetupImpl.predeploy(EntityManagerSetupImpl.java:2026)
	at org.eclipse.persistence.jpa.PersistenceProvider.createContainerEntityManagerFactoryImpl(PersistenceProvider.java:347)
	at org.eclipse.persistence.jpa.PersistenceProvider.createContainerEntityManagerFactory(PersistenceProvider.java:313)
	at org.glassfish.persistence.jpa.PersistenceUnitLoader.loadPU(PersistenceUnitLoader.java:207)
	at org.glassfish.persistence.jpa.PersistenceUnitLoader.<init>(PersistenceUnitLoader.java:114)
	at org.glassfish.persistence.jpa.JPADeployer$1.visitPUD(JPADeployer.java:223)
	at org.glassfish.persistence.jpa.JPADeployer$PersistenceUnitDescriptorIterator.iteratePUDs(JPADeployer.java:510)
	at org.glassfish.persistence.jpa.JPADeployer.createEMFs(JPADeployer.java:230)
	at org.glassfish.persistence.jpa.JPADeployer.prepare(JPADeployer.java:168)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.prepareModule(ApplicationLifecycle.java:925)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:434)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:219)
	at org.glassfish.ejb.persistent.timer.PersistentEJBTimerService.deployEJBTimerService(PersistentEJBTimerService.java:1463)
[...]

PR #621, opened 2016-01-27
Fix for Payara  #618 - Deployment of EJB_Timer_App fails when app has orm.xml

pdudits:
When application contains META-INF/orm.xml, then this file is also processed
when EJB Timer Service is deployed. To prevent this ejb-timer-service-app
was using non-standard name. This reverts parts of e14ece3, that switched back
to standard


smillidge:
jenkins test please

PR #622, opened 2016-01-27
PAYARA-417 

davidweaver:
Change ejb_remoteview quicktest to access the IIOP Listener on port 5037 instead of defaulting to port 3700 and re-enable tests


Pandrex247:
jenkins test please

PR #624, opened 2016-01-28
Edit PAYARA-496 fix

Pandrex247:
The original fix prevented the asadmin truststore being created if it doesn't exist, stopping you from saving certificates. This remedies that by making sure to create it first.

Also adds some logging for if a certificate can't be added.


smillidge:
jenkins test please

PR #626, opened 2016-01-28
PAYARA-556 Update to use CORBA version 4.0.1-p3

### LINKED ISSUE #539: ACC needs three seconds for InitialContext.lookup("myHome");
I have a performance issue and hope that you can help me.

Inside Payara 4.1.1.154 ACC I'm trying to get remote EJBHome of a stateless session bean. ACC needs three seconds (!) for the following line:

initialContext.lookup("myHome");

(where "myHome" is either java:comp/env/FooBar or myPackage.FooBarHome -- makes no difference in speed)

Do you have an idea how I can speed up this, or how I could check for the actual cause? My customer wants getting the initial session to perform within less than one second, and I have no clue how to speed it up...

davidweaver:
Option to include -Dfish.payara.CORBA.SkipGmbalInit=true to runtime environment to bypass Gmbal MBean initialisation. Fixes #539

### DIFF
--- a/nucleus/pom.xml
+++ b/nucleus/pom.xml
@@ -40,7 +40,7 @@
     holder.
 
 -->
-<!-- Portions Copyright [2014] [C2B2 Consulting Limited] -->
+<!-- Portions Copyright [2014-2016] [C2B2 Consulting Limited] -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
@@ -188,7 +188,7 @@
         <findbugs.version>2.5.2</findbugs.version>
         <findbugs.glassfish.logging.validLoggerPrefixes>javax.enterprise</findbugs.glassfish.logging.validLoggerPrefixes>
 
-        <glassfish-corba.version>4.0.1-p2</glassfish-corba.version>
+        <glassfish-corba.version>4.0.1-p3</glassfish-corba.version>
         <stax-api.version>1.0-2</stax-api.version>
         <management-api.version>1.1-rev-1</management-api.version>
         <servlet-api.version>3.1.0</servlet-api.version>


davidweaver:
jenkins test please

chrjohn:
Just a short question: what is the impact of disabling the Gmbal initialisation? Does this mean I cannot query certain attributes of the affected process via JMX? Can it be safely turned off in case we are not using Gmbal in our code?

smillidge:
This is only for CORBA clients i.e. remote EJB clients. It will speed up the boot of the ORB on the client by preventing scanning of classes looking for MBeans but doesn't have any other effects. You may get some missing MBeans but as far as I can remember they weren't added to JMX anyway.

chrjohn:
We have some external processes and their startup can be sped up substantially by setting that option.
Just wanted to know if it has any side effects.
Thanks for the information. :+1:

kitlei-robert:
Was such a `-Dfish.payara.CORBA.SkipGmbalInit=true` option (mentioned above) added to Payara after all? I don't see it either in the 2016 commit or anywhere in the current master.

It would be potentially very useful to have: in code that's just like #539, the first remote EJB context lookup takes more than 5 seconds on my computer, subsequent ones about 7 millis which is an absurd difference of almost three orders of magnitude. VisualVM clearly shows that the bulk of that time is spent doing Gmbal init.

chrjohn:
@kitlei-robert 
As you can see from my last comment it was definitely included. But to be honest I haven't checked with recent versions. Where did you look for the commits? AFAIR the changes were in corba code and not in Payara code.

kitlei-robert:
@chrjohn 
I ran a text search for `SkipGmbalInit` on all sources of `https://github.com/payara/Payara` and now on `https://github.com/eclipse-ee4j/orb` as well and it didn't show up.

Indeed, I added `<jvm-options>-Dfish.payara.CORBA.SkipGmbalInit=true</jvm-options>` to `domain.xml` in the default domain of a fresh Payara server (under both `java-config`s) and it didn't affect the first context lookup time: Gmbal initialisation was still on.

chrjohn:
You need to set it on the external process that does the lookup.

pzygielo:
> @chrjohn I ran a text search for `SkipGmbalInit` on all sources of `https://github.com/payara/Payara` and now on `https://github.com/eclipse-ee4j/orb` as well and it didn't show up.

But it's there - in [payara/patched-src-glassfish-corba](https://github.com/payara/patched-src-glassfish-corba) repo:

https://github.com/payara/Payara/blob/26f91509956145122b87e6d3dd28bcfd0d6803ce/orbmain/src/main/java/com/sun/corba/ee/impl/orb/ORBImpl.java#L575

kitlei-robert:
@pzygielo 
I do see the option in commit `26f9150`, but it's not there in the current master anymore. It should be right around this line:

https://github.com/payara/patched-src-glassfish-corba/blob/12025d15719bc71a132615976f7f754e7882c0ed/orbmain/src/main/java/com/sun/corba/ee/impl/orb/ORBImpl.java#L573C1-L575C1

pzygielo:
> I do see the option in commit `26f9150`, but it's not there in the current master anymore. It should be right around this line:
> 
> https://github.com/payara/patched-src-glassfish-corba/blob/12025d15719bc71a132615976f7f754e7882c0ed/orbmain/src/main/java/com/sun/corba/ee/impl/orb/ORBImpl.java#L573C1-L575C1

I provide no authority for Payara's doing, but I suppose the master in patched-src-glassfish-corba reflects the master in the [source repo: javaee/glassfish-corba](https://github.com/javaee/glassfish-corba) at some historical point.

And the flag is still used in glassfish-corba-4.2.4.payara-maintenance branch in Payara's fork.

pzygielo:
> (As an aside: I wonder why your permalink gets a nice snippet box with a preview while mine doesn't. Not even when I select just one line.)

This is a combination of:
- GitHub renders snippets for links to sources in issue/pull comments within the same repository only (and you link to different one)
- if GitHub fails to find referenced commit in given repository it checks in forks.

PR #629, opened 2016-01-28
Fix for Payara #627 - Invocation Webservice deployed from directory f…

pdudits:
…ails

Fix for GLASSFISH-17039 contained a change in ModuleContextValidator where a
slash in fullFileUri was replaced with File.separator. However URIs are not
platform dependant and cannot contain backslashes.


smillidge:
jenkins test please

Pandrex247:
fixes #627

PR #640, opened 2016-02-02
GLASSFISH-20818: Pass passwords from passwordfile to commands (PAYARA-658)

pdudits:
…dded glassfish

When CLIUtil.readPasswordFileOptions is invoked with boolean parameter of true, then Map
`passwords` contains keys prefixed with `AS_ADMIN_`. This is never the parameter name and
therefore the password parameters didn't get filled.

I couldn't locate the origin of bug, because the code is same in Glassfish 3.1.2, I suppose it
either used some side channel, or command model transformation.


smillidge:
jenkins test please

pdudits:
Just like [original issue](https://java.net/jira/browse/GLASSFISH-20818) describes, the error happens when trying to invoke create-file-user with embedded Payara. Should I provide a test case?

smillidge:
Any test case is welcome.

smillidge:
Tagged as PAYARA-658 internally

PR #669, opened 2016-02-19
PAYARA 626 added support for Hazelcast Lite Members 

smillidge:
Lite members are full cluster members but do not store any JCache or web session data. This provides more flexibility for cluster configurations.

Also corrects instance names in list-cluster-members to be the instance name and the name set in payara micro using --name


danPowell52:
jenkins test please

PR #670, opened 2016-02-19
PAYARA-640 fixes #667

### LINKED ISSUE #667: #505 is not fixed when TxType is REQUIRES_NEW
I'm very happy with Payara 4.1.1.161 released.
I wrote issue https://github.com/payara/Payara/issues/505 .
I tried Payara 4.1.1.161, and I saw this bug is fixed when TxType is `REQUIRED`.

But, it seems this bug is NOT fixed when TxType is `REQUIRES_NEW`.
Could you try my test program again?
https://github.com/MasatoshiTada/TransactionalSample-Doma/blob/master/src/main/java/com/example/jdbc/JdbcCdiRequiresNewTestDao.java

smillidge:
Fix rollback on exception for Transactional Annotation fixes #667


danPowell52:
jenkins test please

PR #676, opened 2016-02-24
Fixes #673 PAYARA-642 do explicit check for OSGi classloader 

### LINKED ISSUE #673: Problems on deployment of OSGi hybrid bundles (ClassNotFoundException - IIOPSSLSocketFactory)
Please see attached minimal MDB. Works on payara 4.1.152.1 but fails on 4.1.1.161
It works when deployed as normal JAR, but deployed as OSGi bundle there are exceptions thrown (same as in #516 and #535)

Deployment with:
`asadmin deploy --type osgi payarabug-0.0.1-SNAPSHOT.jar`

NB: it also works when first deploying as normal JAR, then undeploy, then deploy as OSGi bundle. So please deploy to a freshly started instance.

Prerequisites: you need to have a connection factory and test queue:
`asadmin create-jms-resource --restype=javax.jms.QueueConnectionFactory --enabled=true ConnectionFactory`
`asadmin create-jms-resource --restype=javax.jms.Queue --enabled=true queue/test`

The bean is just this:
```
package payarabug;

import javax.annotation.Resource;
import javax.ejb.MessageDriven;
import javax.jms.Message;
import javax.jms.MessageListener;
import javax.jms.QueueConnectionFactory;

@MessageDriven( mappedName = "queue/test" )
public class PayaraBugMDB
    implements MessageListener {

    @Resource( name = "ConnectionFactory" )
    private QueueConnectionFactory queueConnectionFactory;

    

    public void onMessage( final Message msg ) {
        System.out.println( "XXX onMessage()" );
    }

}

```

Attached is the JAR file (renamed as ZIP).
[payarabug-0.0.1-SNAPSHOT.zip](https://github.com/payara/Payara/files/140753/payarabug-0.0.1-SNAPSHOT.zip)

smillidge:
and don't switch to parent in that case.

Bit of a hack that needs fixing by the Janitor Service in the future to clear up threads context classloader as the ORB spawns a bunch of threads.

### DIFF
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
@@ -585,11 +585,10 @@ public final class GlassFishORBManager {
             orb = ORBFactory.create() ;
             prevCL = Utility.getClassLoader();
             try {
-               if (processType != processType.Other) {
+              if (processType != processType.Other && !prevCL.getClass().getName().contains("OSGi")) {
                 Utility.setContextClassLoader(prevCL.getParent());
                }
-            
-                ORBFactory.initialize( orb, args, orbInitProperties, useOSGI);
+               ORBFactory.initialize( orb, args, orbInitProperties, useOSGI);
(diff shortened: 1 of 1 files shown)


smillidge:
jenkins test please

Issue #680, opened 2016-02-26
File handles on EAR libs held on DAS during undeployment on Windows


pdudits:
When performing undeployment from a standalone instance, the undeployment takes too long to finish or fails on the DAS after it completed sucessfully on the instance.

This is caused by undeploy command being stuck when deleting `applications/<ear>/lib/*.jar`:

````
at com.sun.enterprise.util.io.FileUtils.doWithRetry(Lcom/sun/enterprise/util/io/FileUtils$RetriableWork;)I (FileUtils.java:770)
at com.sun.enterprise.util.io.FileUtils.internalDeleteFile(Ljava/io/File;Z)Z (FileUtils.java:615)
at com.sun.enterprise.util.io.FileUtils.deleteFileWithWaitLoop(Ljava/io/File;)Z (FileUtils.java:577)
at com.sun.enterprise.deploy.shared.FileArchive.deleteDir(Ljava/io/File;)Z (FileArchive.java:598)
at com.sun.enterprise.deploy.shared.FileArchive.deleteDir(Ljava/io/File;)Z (FileArchive.java:595)
at com.sun.enterprise.deploy.shared.FileArchive.delete()Z (FileArchive.java:229)
at org.glassfish.deployment.admin.UndeployCommand.execute(Lorg/glassfish/api/admin/AdminCommandContext;)V (UndeployCommand.java:442)
````

The DAS keeps open handles to jar files, as the corresponding `EarLibClassloader` have not been closed, even if the app is not loaded on the DAS only on the instance.

The solution is to always clean applications on the DAS during undeployment. PR follows.

OndrejM:
@pdudits thank you for reporting and the PR.
It looks like @smillidge already did some improvements in this commit: https://github.com/payara/Payara/commit/4378070744ab7752abb720bcfdd942e1d69a2b44, but it did not cover this specific scenario.

smillidge:
@pdudits is this on Windows platform? Also do you have an EAR file that demonstrates this repeatedly? I can not reproduce this on either Linux or Windows.

pdudits:
@smillidge Yes, it was on windows; I'm afraid I cannot provide the EAR that causes this, but in my case it was remote undeployment from a standalone instance. So `localhost != DAS host` and `DAS process != target process`.

PR #682, opened 2016-02-26
PAYARA-671 Always perform clean up on application on DAS upon undeployment.

pdudits:
`EarLibClassloader` of an EAR app deployed on non-DAS instance was initialized but not destroyed on DAS upon deployment.
That lead to JarFile handles to its lib directory being held, and the DAS was unable to clear its application directory.
Calling unload unconditionally gurantees ApplicationInfo.clean to be called, which in turns calls preDestroy() / done() on all application
class loaders.

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [C2B2 Consulting Limited]
 
 package com.sun.enterprise.v3.server;
 
@@ -1093,11 +1094,9 @@ public class ApplicationLifecycle implements Deployment, PostConstruct {
 
         events.send(new Event(Deployment.UNDEPLOYMENT_START, info));
 
-        // for DAS target, the undeploy should unload the application
-        // as well
-        if (DeploymentUtils.isDASTarget(params.target) || DeploymentUtils.isDomainTarget(params.target)) {
-            unload(info, context);
-        }
+        // we unconditionally unload the application, even if it is not loaded, because we must clean the
+        // application, especially the classloaders need to be closed to release file handles
+        unload(info, context);
 
         if (report.getActionExitCode().equals(ActionReport.ExitCode.SUCCESS)) {
             events.send(new Event(Deployment.UNDEPLOYMENT_SUCCESS, context));


smillidge:
Thanks for this I've been trying to track down these for a while. My only concern is to ensure you test the scenario of retargeting an application i.e. target to two stand-alone instances and then untarget one of them thereby undeploying if from one while leaving it deployed to the second.  

I added in the additional DeploymentUtils.isDomainTarget(params.target) into the OR (but didn't add the copyright comment!) 

What is the target in the scenario you describe?

pdudits:
See the referenced issue for the scenario - undeployment from standalone instance, which is the only target. We either deploy applications to entire domain or to single instance.

When retargetting an application I'd use delete-application-ref / create-application-ref anyway, so I'm not sure if undeploy should be concerned about that scenario.

pdudits:
Of course, cleaner fix for this would be, if the classloader would get unloaded during startup after the application is verified, and will not be loaded. But I was unable to find correct spot for this logic.

pdudits:
I just verified, that like I wrote above, one cannot use `undeploy` to remove one of the targets of an application.

````
Application test is referenced by more than one targets. Please remove other references or 
specify all targets (or domain target if using asadmin command line) before attempting undeploy 
operation.
```

`DeleteApplicationRefCommand` is not affected by this change, as it undeploys only if target matches the executing instance.

Using application-refs to move app between instances works correct in my setup.

pdudits:
I am still encountering file lock when doing local .war redeployment on DAS. I'll yet check, that code path is also sufficiently clean.

smillidge:
I suspect this one will get swept up and incorporated into a larger piece of work as we are also working on stale NFS handles related to file locking. I'm still grateful for the pull request and any other things you find out relating to file locking on the DAS while you investigate.

smillidge:
OK I've debugged this and maybe I'm doing something wrong. I have a domain with a standalone instance and the DAS. I deploy an EAR file targeting only the standalone instance. I then undeploy the application. In my case unload is called on the DAS as the target is "domain" therefore the condition is true and unload is called.

Do I need to do something different to reproduce the problem?

pdudits:
Yes, I'm targeting the instance specifically when calling undeploy. I never realized, that using domain target would be a workaround for this :).

smillidge:
Does using --target domain in your undeploy fix it? BTW when undeploying using the console domain is the target. 

I've also just made a pull request for a related fix #699 which closes some Files that were being left open prior to deletion especially if the jars contained classes referenced in a servlet initialiser.

pdudits:
Yes, using `--target domain` does fix the issue.

However I would still argue that undeploy command is  broken without this change, as using valid argument -- instance name -- results in incorrect behaviour -- classloader not being released on DAS and thus the .jar file handles not being cleared.

Especially if this is the only difference in handling of the command, because `source.delete()` always follows `deploymentContext.clean()`, except for using (undocumented) `keepreposdir`, or directory deployment, which usually not the case for instance deployment.

pdudits:
I again went into glassfish SVN to find the probable cause for this behaviour. I found [rev 46606](https://java.net/projects/glassfish/sources/svn/revision/46606)

> 2011-05-03 14:05:48 UTC (4 years)
> 
> Incremental changes to support cloud deployment. Have the clean phase 
> invoked for unload so both disable and undeploy commands will trigger 
> this phase (the phase will be triggered with different origins so the 
> container code could do things accordingly). This change makes the 
> deploy/enable and undeploy/disable paths more symmertric and allows the 
> orchestrator do necessary service clean up for both disable and undeploy 
> operations.
> Ran QL, PE/EE deployment dev tests, ejb dev tests.

The revision moved `ApplicationInfo.clean()` call from method `undeploy` into method `clean`, but failed to implement part _"so both disable and undeploy commands will trigger this phase"_, as it is apparently not invoked on DAS when targetting an instance.

smillidge:
OK thanks I'll review again.

smillidge:
jenkins test please

smillidge:
I've had a play around and I think you are right it is clean to just unload on the DAS whatever. I can't see a valid case whereby you can deploy and app on a set of servers and then undeploy on a subset of those servers and thereby doing this would be dangerous.

smillidge:
Fixes #680

PR #683, opened 2016-02-26
Resolve com.sun.aas.instanceName in JVM arguments (PAYARA-660)

pdudits:
`com.sun.aas.instanceName` is the only glassfish system property that will not resolve when passed as JVM argument, which is strange omission.

In my use case I redirect all logging to logback, and logback gets initialized earlier than the system property is set during appserver startup. Passing it in JVM arguments guarantees it is available right from the process start.

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -38,7 +38,7 @@
  * holder.
  */
 
-// Portions Copyright [2014-2015] [C2B2 Consulting Limited]
+// Portions Copyright [2014-2016] [C2B2 Consulting Limited]
 
 package com.sun.enterprise.admin.launcher;
 
@@ -726,6 +726,7 @@ public abstract class GFLauncher {
         all.putAll(envProps);
         all.putAll(asenvProps);
         all.putAll(sysProps);
+        all.put(SystemPropertyConstants.SERVER_NAME, getInfo().getInstanceName());
         all.putAll(sysPropsFromXml);
         all.putAll(jvmOptions.getCombinedMap());
         all.putAll(profiler.getConfig());


smillidge:
Tageed as PAYARA-660 internally

danPowell52:
jenkins test please

danPowell52:
I have reviewed the changes and everything seems in order. I will merge once the test suite confirms.

danPowell52:
We have reverted this issue for now because on second review we are unsure what this addition is achieving. Could you please provide clarification as to what your problem is and what your intended fix is?

We've debugged this area of code, and we're not sure what your "resolving" issue is. For the DAS, it is already resolved as _server_ at this point, which your fix does not change. In addition, if you specify the instance name in the domain.xml, it will always resolve as this once it hits the `resolve` methods a few lines later on (regardless of your fix).

For an instance, the instance name is _null_ by default, but it is resolved as what you pass in with `-Dcom.sun.aas.instanceName` from the asadmin script, or as what you specify in the instance's domain.xml file. This is the only point we see your fix making a difference: if you don't pass in anything from the asadmin script and don't specify the name in the config, your fix will grab the default name of the instance (the name it was created with) and resolve it to this. Again though, if you specify an instance name in the instance's domain.xml, it will override the instance name your fix resolves.

pdudits:
In my case I've got `<jvm-options>-Dlogback.serviceName=${com.sun.aas.instanceName}</jvm-options>` in my DAS as well as instance configs, that is used to determine the log folder location and additional property of the log events when they're passed to ELK stack. 

Value for instanceName is not present in the map at point of process start, therefore the process is started with value of the property `logback.serviceName` equal to `${com.sun.aas.instanceName}`.

The system property is eventually set, but it is too late for logging subsystem, that writes out log events  directly after the process is started, and only resolves the property once, therefore I've put it in `jvm-options` directly.

I'll check once again if the resolution did not work for other reason, but it definetely did not ever happen after this line was added.

pdudits:
No, in my debugging session, without the patch, the token `${com.sun.aas.instanceName}` will not be replaced with respective instance name when used in jvm options, whereas any of other standard java properties, environment variables and `SystemPropertyConstants.*_PROPERTY` properties would.

danPowell52:
Ok, thanks for the information and prompt reply. I am afraid I misunderstood what you were doing.

danPowell52:
I have merged the fix back in, thanks for your contribution.

PR #684, opened 2016-02-26
Do not wait for all applications to be loaded when doing TX recovery

pdudits:
This improves previous fixes of [GLASSFISH-21175](https://java.net/jira/browse/GLASSFISH-21175).

I was unsatisfied with timeout approach that prolonged server starts or cause startup of application to fail. To fix this in Glassfish 3.1.2 I made `ApplicationLoaderService` spawn a new thread after all connector modules were started. With that, its `postConstruct` would finish before transaction recovery started.

Glassfish 4 has an explicit application ordering built in, so it already guarantees that all connector modules are started when applications start, and need a transaction. Therefore this wait point in `ConnectorResourceRecoveryHandler` is not relevant anymore.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/recovery/ConnectorsRecoveryResourceHandler.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/recovery/ConnectorsRecoveryResourceHandler.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [C2B2 Consulting Limited]
 
 package com.sun.enterprise.resource.recovery;
 
@@ -103,11 +104,7 @@ public class ConnectorsRecoveryResourceHandler implements RecoveryResourceHandle
 
     @Inject
     private Provider<ConnectorResourceDeployer> connectorResourceDeployerProvider;
-
-    @Inject
-    @Named("ApplicationLoaderService")
-    private Provider<ApplicationLoaderService> startupProvider;
-    
+   
     @Inject
     private ConfigBeansUtilities configBeansUtilities;
 
@@ -216,8 +213,11 @@ public class ConnectorsRecoveryResourceHandler implements RecoveryResourceHandle
         //Done so as to initialize connectors-runtime before loading connector-resources. need a better way ?
         ConnectorRuntime crt = connectorRuntimeProvider.get();
 
-        //Done so as to load all connector-modules. need to load only connector-modules instead of all apps
-        startupProvider.get();
+        // ApplicationLoaderService already guarantees that connectors are loaded before any other applications.
+        // Recovery will not start sooner than transaction is first needed on EE application startup, therefore
+        // it is safe to continue here without waiting for startupProvider, as that ultimately creates a deadlock
+        // between ApplicationLoaderService needed a transaction manager, and recovery wainting for applications
+        // to finish loading.
 
         Collection<ConnectorResource> connectorResources = getAllConnectorResources();
 


pdudits:
Any thoughts on this? Especially is my reasoning correct? Do you see any possibility that something would trigger transaction enlis transaction recovery before all connectors are loaded?

I had a look on JCA 1.7 spec, and as far as I understand, this could happen if the resource adapter implemented Transaction inflow (chapter 15) and wanted to submit a work without having any endpoints registered yet. But I have never seen such beast so I don't know how to verify this scenario.

smillidge:
This area of code needs a fair amount of work from us in a future release. The main problem is with EJB Timers which are loaded before TX recovery and then proceed to use JPA to access the database. Boot needs some reworking to prevent this.

smillidge:
When you did this fix in GlassFIsh 3.1.2 did you encounter problems with the EJB Timer initialisation? If you look at https://java.net/jira/browse/GLASSFISH-21175 all the problems seem to stem for EJB Timer subsystem bootstraps -> starts a transaction -> waits for TX recovery to complete. However TX recovery has't even started and never will as the boot is now waiting for TX recovery!! All we did was put in place a timeout on the wait to mitigate the problem.

pdudits:
Yes, I was fighting GLASSFISH-21175 in that case. I tried your patch first, but it wasn't resolving the problem for me, as the wait for TX eventually timed out, `@Startup` EJBs would fail to initialize, and the applications would not start anyway.

The deadlock as I understand it is, that in order to start transaction recovery, all resource adapter modules need to be loaded, so they can participate in the process. Therefore the recovery handler waits for all applications to be loaded, at it has no better way of doing it just for connectors. All applications are loaded after ApplicationLoaderService's `@PostConstruct` is finished. However, other applications may require transaction at startup, therefore `@PostConstruct` would not finish, because it waits for TX recovery, which in turn waits on it.

We need to ensure that all connector modules are loaded. What I did in 3.1.2 was, that I had ApplicationLoaderService spawn a new thread after connector applications started. So the @PostConstruct finished after satisfying the condition for TX recovery, and would not block it. Applications would get their transactions correctly as soon as recovery finished. That approach make contract of ApplicationLoaderService bit unclear, but it worked well for me, it is 6 months in production without any TX recovery trouble.

Now that we know that connector applications are started before any EJB can request a transaction one has to wonder why is that line there in `ConnectorsRecoveryResourceHandler` in the first place.

* Either it was there sooner than application ordering was implemented in `ApplicationLoaderService`. I think it is worth going on archeology expedition to early days of V3, but ever since svnsearch.org is gone there's no useful tool at hand for that.
* There is a use case where TX recovery would be initiated sooner, by something else than EJB startup. That was my point about Transaction inflow. 

I'd vote for first option, and therefore removing it is the easiest thing to make things work. If second option would be true, waiting doesn't solve the issue anyway, because we know that whenever TX is needed at startup, the server would fail.

pdudits:
One has to love temporary fixes:

> https://java.net/projects/glassfish/sources/svn/revision/25156
>
> 2009-03-06
> Fixing a bug to get ConnectorRuntime (initialize connector-runtime before loading resource-> adapter, resources) Making sure that applications are loaded before recovery (temporary, later only connector modules should be loaded)

pdudits:
> https://java.net/projects/glassfish/sources/svn/revision/31400
> 
> 2009-09-16
> Fixed issue 5382. Load standalone rars before other applications in 
server start up. Also some minor clean up in ApplicationLifecycle.

Option 1 proved. Time to remove temporary fix, I presume.

Pandrex247:
jenkins test please

smillidge:
I've tested the pull request and it does fix the problem. Now with an EJB Timer application and transaction recovery enabled we no longer hit the timeout. Looks like you are right in your analysis. I'm not sure I'd have worked out that was the problem so thanks.

smillidge:
jenkins test please

smillidge:
jenkins test please

PR #698, opened 2016-03-11
PAYARA-661 don't fail validation for wrap-jdbc-objects=false when sql…

smillidge:
…-trace-listeners is empty string


smillidge:
jenkins test please

Issue #647, opened 2016-02-06
NCLS-DEPLOYMENT-00009 error with embedded v4.1.x and lambda


zappee:
Hi,

I have a problem with embedded glassfish server v4.1.x and lambda expression.
Could you please read this topic on [stackoverflow.com](http://stackoverflow.com/questions/35230667/glassfish-embedded-v4-1-1-and-lambda
)?

I tried to use Payara embedded full, but I got same error.
I used this maven dependency:

```
<dependency>
    <groupId>fish.payara.extras</groupId>
    <artifactId>payara-embedded-all</artifactId>
    <version>4.1.152.1</version>
    <type>jar</type>
    <scope>test</scope>
</dependency>
```

smillidge:
Any chance you can create a simple project on GitHub as a test case? Does the same EJB jar fail on deploying to Payara Server or just on embedded

The latest Payara Embedded should be 4.1.1.161.

zappee:
Hi,

I have created a simple test application: https://github.com/zappee/lambda-web-demo.git

zappee:
any good news?

smillidge:
Sorry with the 161 release We've not had time to try your test case.

smillidge:
Hi thanks this is reproducible. The jar is failing to deploy on embedded with the Lambda expression. The hidden error is;
`
[...]

PR #700, opened 2016-03-13
refactored asm-all dependency to asm-all-repackaged

mulderbaba:


smillidge:
Does it fix the lambda bug?

smillidge:
jenkins test please

mulderbaba:
yes Steve, I also tried with lambda-web-demo provided by zappee. 

I replaced all org.glassfish.external:asm-all:3.3 dependencies with org.glassfish.hk2.external:asm-all-repackaged:hk2_version, which is a repackaging of asm-all 5.0.3 in the background. I refactored all org.objectweb.asm.* imports to org.glassfish.hk2.external.org.objectweb.asm.*. The interfaces that were turned into abstract classes are now being extended. Since those abstract classes don't have any default constructors I added the call super(Opcodes.ASM5); to the constructor of the classes that extend those abstract classes. I removed the class EmptyVisitor, the open source projects that were upgraded from ASM 3.x to ASM 5.x had the same approach. Instead of the EmptyVisitor I introduced the regarding visitor, such as MethodVisitor. I removed the dependency auto-depends, since it was introducing asm-all transitively. Instead of it I used the dependency hk2-junitrunner where it provides a newer version for HK2Runner. It's a not JUnit Runner extension anymore so I introduced it in the class hierarchy as a super class to ConfigApiTest.

smillidge:
Fixes #647

smillidge:
jenkins test please

PR #709, opened 2016-03-16
PAYARA-606 explicitly set TCCL before attempting Batch repository purge on undeployment

smillidge:


smillidge:
jenkins test please

PR #716, opened 2016-03-22
PAYARA-714 Fixed logging dependencies of several modules

dmatej:
- separated LogStrings.properties - no dependencies
- some properties moved to place where they are used, some unused deleted
- connectors-inbound, connectors-internal-api, connectors-runtime, jdbc-core,
  jdbc40, jdbc-runtime, jms-core, orb-iiop, javamail-runtime, resources-runtime,
  nucleus


smillidge:
jenkins test please

Issue #717, opened 2016-03-22
isSupportClientInfo in ConnectionHolder40 is not implemented correctly


dmatej:
It throws exceptions when it should only log some info and return false.
If the ClientInfo is not supported, there should be no problem for the ConnectionHolder - then there is nothing to cache and return.
This issue depends on #714 / #716 - I have fixed logging before I fixed this.

smillidge:
Fixed by #718

PR #718, opened 2016-03-22
#717

dmatej:
Fixes isSupportClientInfo in ConnectionHolder40 for old dumb drivers (latest ifxjdbc).
It should be pulled after #716.

#716 fixes only logging, both issues together changes behavior so that 
- if connection.getClientInfo throws an exception, it is catched in isSupportClientInfo not in the caller. The caller retrieves false and does not try to call the method. The result is always cached (former version cached only true).
- an exception is logged with FINE level, because it does not cause any problems.
- an exception from rs.close is logged with SEVERE level, because if we cannot close the resultset, it is always an error (but rare).


smillidge:
jenkins test please

PR #720, opened 2016-03-22
PAYARA-676 Restart cluster asadmin command

mikecroft:
Added an asadmin command to restart a cluster and added an admin console button. No l10n on the strings, though.

### DIFF
--- a/appserver/admingui/cluster/src/main/resources/cluster/clusterGeneral.jsf
+++ b/appserver/admingui/cluster/src/main/resources/cluster/clusterGeneral.jsf
@@ -127,6 +127,17 @@
         />
     </sun:button>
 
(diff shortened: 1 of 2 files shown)


appserver/admingui/cluster/src/main/resources/cluster/clusterGeneral.jsf
```diff
@@ -127,6 +127,17 @@
         />
     </sun:button>
 
+    <sun:button id="restartButton" text="Restart Cluster" disabled="#{disableStop}" primary="#{false}"
```
mikecroft: Set the `disabled` boolean to the value of `disableStop` since we want the restart button enabled at the same times as the stop button.

smillidge:
jenkins test please

PR #726, opened 2016-03-26
changed gosh command in XML files

mikecroft:
Use the `--nointeractive` mode rather than the current `--noshutdown -c noop=true`


smillidge:
jenkins test please

PR #728, opened 2016-03-28
Payara 679 Uber Jar command line option

smillidge:
A command line option for payara micro to output an uber jar which includes all command line deployments and options so that when run again all deployments and command line options are preserved and packaged into the jar.


smillidge:
jenkins test please

PR #733, opened 2016-03-31
Prevent duplicate server respawn when run under supervision. Fixes #732.

### LINKED ISSUE #732: Two instances start on restart-domain / restart-instance
Our Payara instances run as windows service and are remotely restarted during deployment.

However after restart we see two new ASMain processes start up, and fight for network ports. Due to some other bug, the instance which does not acquire the ports does not always terminate, although it says so in its log.

The issue is caused by two respawn strategies being active at same time - one of watchdog parameter, and another one of restart-<instance|domain> commands. More info in upcoming pull request.

pdudits:
Service wrapper runs instances / domains with `--watchdog true`. This causes server to be relaunched once it exits with predefined exit code. See `StartDomainCommand.executeCommand` and `StartLocalInstanceCommand.executeCommand`.

On the other hand, RestartServer will also relaunch the server by starting separate process just before the process exits. See `RestartServer.doReincarnation()`.

RestartServer doesn't take` --watchdog` argument into account in its `doExecute` method, and therefore two server instances are started, from each of the code paths.

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherInfo.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherInfo.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [C2B2 Consulting Limited]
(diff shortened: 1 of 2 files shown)


smillidge:
jenkins test please

PR #738, opened 2016-04-03
Remove dependency on Hazelcast 

smillidge:
PAYARA-692 hazelcast dependency should be removed from JBatch until the JBatch persistence store for Hazelcast is available


danPowell52:
jenkins test please

PR #739, opened 2016-04-03
Payara 687 Implement Hazelcast Cluster Group

smillidge:
Hazelcast Cluster Group Functionality enables users to set the cluster group name and cluster group password. these have to be identical for nodes to cluster.


danPowell52:
jenkins test please

PR #743, opened 2016-04-06
setting values for the checker services dynamically is now working correctly

mulderbaba:
…rrectly.


smillidge:
jenkins test please

Issue #702, opened 2016-03-14
Payara not showing modules for versioned application PAYARA-670


ratcashdev:
Mostly related to #455 and slightly #445.
Glassfish is not able to list the deployed modules when using versioned applications.
See the attached screenshots:
![versioneddeployment](https://cloud.githubusercontent.com/assets/2928922/13747256/b24e480c-e9f7-11e5-85f2-1bd3b6fb0168.PNG)
expected:
![expected](https://cloud.githubusercontent.com/assets/2928922/13747258/b35186a6-e9f7-11e5-8e6a-3dc8c50933ac.PNG)

I was wondering, if this is not a simple path mistake? Note the different between the colon in the application's name, and the ~ (tilde) that is used in the app's path on the filesystem.

devyoustice:
#461 seems to be duplicate (and sorry, i forgot about that one). I do think we've got more relevant information here, than there. As #461 also deals with lost requests, I suggest to focus on the lost request as part of #461 and with components/monitoring as part of this issue.

ratcashdev:
#461 seems to be duplicate (and sorry, i forgot about that one). I do think we've got more relevant information here, than there. As #461 also deals with lost requests, I suggest to focus on the lost request as part of #461 and with components/monitoring as part of this issue.

ratcashdev:
Couple of findings:
1. The component listing seems to be just a GUI problem, because the REST management services work find. E.g. 
`GET http://localhost:4848/management/domain/applications/application/versionedSampleApp:1/list-sub-components` works just fine and I get the necessary components

2. The json response from the REST resource includes the following:
```
"commandLog": [
            "list-sub-components --modulename versionedSampleApp:1"
        ]
```
whereas the correct command-line is:
`asadmin list-sub-components --appname="versionedSampleApp:1" versionedSampleApp` i.e. without the ":1" version suffix, and without --modulename (the latter being not even understood by the cli)

This hints to the fact that the GUI is not propely stripping the version from the app-name when used as module-name.

However, it also seems that for different REST endpoints, the expected style of the module-name is varying. 
Examples:
* AppUtil.getSnifferListOfModule(), that is `/applications/application/{appName}/module/{moduleName}`  expects the appName and the moduleName both with the version included.
* ApplicationHandlers.getSubComponents(), that is `/applications/application/{appName}/get-context-root.xml?modulename={moduleName}` expects the appname with version and the moduleName w/o version
[...]

Issue #461, opened 2015-09-29
Monitoring on versioned applications


ratcashdev:
Using Payara 4.1.153
Monitoring is not able to work with versioned apps, it seems. See the attachment.
![monitoring_with_versions](https://cloud.githubusercontent.com/assets/2928922/10157381/ecda20b4-6689-11e5-97d1-39a097241d1c.JPG)

In addition while doing the "asadmin enable myapp:1.2" command, the server returns 404 for a number of requests, which is not what I would expect.

devyoustice:
Let's deal with the missing components/monitoring as part of #702 and this issue should focus solely on the lost requests.

ratcashdev:
Let's deal with the missing components/monitoring as part of #702 and this issue should focus solely on the lost requests.

smillidge:
Fixed by #746

PR #746, opened 2016-04-11
Fix component list for versioned apps

ratcashdev:
Bugfix: always strip off the app's version suffix (apps with rolling updates) when looking up the app's modules.

This is probably a dirty hack to solve issue #702, and partly also #461 (note: only the GUI issue, when components were listed), because:
1.  it leaves the GUI in a kind of unpolished state: it will still show the app's main moduleName with the version suffix, even though it does not seem right (before this hack the  the REST resource did not accept module names with version suffix).
2. it does not enforce the correct naming of the module in various calls, rather makes the REST resource more forgiving
3. certain REST resources may return URLs with incorrect moduleName (with the suffix)

However! 
1. It works and is just a 6 line change. Simplicity matters.
2. The potential of regressions is practically zero, because this change affects only versioned apps, which did not work before at all.
3. The alternative is to review/change the admin_gui Handlers (ApplicationHandler.java and MonitoringHandler.java) and the Management REST Resource generator(s) and maybe a couple of more files, where unwanted regressions are basically guaranteed.

### DIFF
--- a/appserver/deployment/javaee-core/src/main/java/org/glassfish/javaee/core/deployment/GetContextRootCommand.java
+++ b/appserver/deployment/javaee-core/src/main/java/org/glassfish/javaee/core/deployment/GetContextRootCommand.java
@@ -98,7 +98,10 @@ public class GetContextRootCommand implements AdminCommand {
         if (appInfo != null) {
             Application app = appInfo.getMetaData(Application.class);
             if (app != null) {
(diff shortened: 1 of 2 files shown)


smillidge:
jenkins test please

ratcashdev:
So will this get into the next release?

smillidge:
Should do it just needs reviewing and merging

PR #755, opened 2016-04-15
PAYARA-726 ensure we use a file in the payara-micro.jar to get a hand…

smillidge:
…le into payara micro


smillidge:
jenkins test please

smillidge:
jenkins test please

Issue #730, opened 2016-03-30
Hibernate 5.x with Payara and Oracle driver throws reflection exception


ubench-tdv:
When using hibernate 5.x with Payara and a connectionpool to an Oracle database (ojdbc7.jar) you"ll see an exception:

```java
javax.resource.ResourceException: java.lang.IllegalAccessException: Class 
com.sun.gjc.util.MethodExecutor can not access a member of class 
    oracle.jdbc.driver.PhysicalConnection with modifiers "public"
    	at com.sun.gjc.util.MethodExecutor.invokeMethod(MethodExecutor.java:240)
	at com.sun.gjc.spi.jdbc40.ConnectionHolder40.getSchema(ConnectionHolder40.java:609)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.hibernate.engine.jdbc.env.internal.DefaultSchemaNameResolver$SchemaNameResolverJava17Delegate.resolveSchemaName(DefaultSchemaNameResolver.java:72)
	at org.hibernate.engine.jdbc.env.internal.DefaultSchemaNameResolver.resolveSchemaName(DefaultSchemaNameResolver.java:59)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentImpl.determineCurrentSchemaName(JdbcEnvironmentImpl.java:294)`
```

Digging a little deeper this is a bug in com.sun.gjc.spi.jdbc40.ConnectionHolder40

```java
public String getSchema() throws SQLException {
    if(DataSourceObjectBuilder.isJDBC41()) {
            checkValidity();
            try {
                return (String) getMethodExecutor().invokeMethod(con, "getSchema", null);
            } catch (ResourceException ex) {
                _logger.log(Level.SEVERE, "jdbc.ex_connection_holder", ex);
                throw new SQLException(ex);
            }
        }
        throw new UnsupportedOperationException("Operation not supported in this runtime.");
}
```

Which passes the connection (internal Oracle T4CConnection class) straight to com.sun.gjc.util.MethodExcutor:

```java
public Object invokeMethod(Object object, String methodName, 
[...]

PR #762, opened 2016-04-20
invokeMethod now depends on getDeclaredMethod as recursive instead of getMethod

mulderbaba:


smillidge:
jenkins test please

smillidge:
fixes #730

PR #783, opened 2016-04-28
Payara 738 references to weld osgi bundle.jar should be fixed

mulderbaba:
weld-osgi-bundle.jar text references replaced with weld-osgi-bundle-glassfish4.jar.
Also staged the weld-environment-common.jar to lib/appclient folder by first putting it into modules folder like weld-se-core and that propagating it to lib/appclient during the build process.

These actions fix the quicklook tests that were failing and the MDBTests to be specific.


danPowell52:
jenkins test please

PR #798, opened 2016-05-09
Payara 775

### LINKED ISSUE #751: Unwrapping principal interferes with JASPIC's register session
I unfortunately found a problem with the solution as committed for https://github.com/payara/Payara/issues/290. In case a custom principal is used AND the JASPIC 1.1 "javax.servlet.http.registerSession" directive is used, the groups (and possibly any credentials) are not restored as they should be.

This happens because the protocol for "javax.servlet.http.registerSession" requires [getting the Principal from the `HttpServletRequest` and then feeding it back to the handler in a SAM](http://arjan-tijms.omnifaces.org/2013/04/whats-new-in-java-ee-7s-authentication.html#RegisterSession). Subsequently [BaseContainerCallbackHandler#processCallerPrincipal](https://github.com/payara/Payara/blob/master/appserver/security/core-ee/src/main/java/com/sun/enterprise/security/jmac/callback/BaseContainerCallbackHandler.java#L374) checks if the type is `WebPrincipal`. 

If this is not the case, no "reuse" will take place.

Now I'm the first to admit this "getting principal from request and feeding it back" is a little brittle, but that's how it works in JASPIC 1.1.

What I guess what needs to happen is that when in a SAM's validateRequest method the unwrapping should *not* take place. There are a few characteristics one can check for, e.g. if the principal has reference equality to `org.apache.catalina.Session#getPrincipal`, but perhaps the more robust way is setting a request attribute to indicate we're in a SAM:

    //JSR 196 is enabled for this application
    try {
         context.fireContainerEvent(ContainerEvent.BEFORE_AUTHENTICATION, null);
         ((HttpServletRequest)request.getRequest()).setAttribute("...", true);
         result = validate(request, response, config, authenticator, calledFromAuthenticate);
     } finally {
         ((HttpServletRequest)request.getRequest()).removeAttribute("...");
         context.fireContainerEvent(ContainerEvent.AFTER_AUTHENTICATION, null);
     }

Any other mechanism like the internal *notes* may work as well, as long as the `getUserPrincipal()` can internally check for it.

What do you think? I've asked Ron Monzillo if he has any opinion on this matter and let you know as soon as he replies.

cc @Circkumflex 

p.s. I originally commented on https://github.com/payara/Payara/issues/290, but since that issue is already closed it's perhaps better to open a new issue.

smillidge:
Fixes #751 enable access to both the custom principal and the WebPrincipal in the Security Context when doing JASPIC

### DIFF
--- a/appserver/security/core-ee/src/main/java/com/sun/enterprise/security/jmac/callback/BaseContainerCallbackHandler.java
+++ b/appserver/security/core-ee/src/main/java/com/sun/enterprise/security/jmac/callback/BaseContainerCallbackHandler.java
@@ -375,9 +375,10 @@ abstract class BaseContainerCallbackHandler
         final Subject fs = cpCallback.getSubject();
         Principal principal = cpCallback.getPrincipal();
         
+        // PAYARA-755 If the SAM has set a custom principal then we check that the original WebPrincipal has the same custom principal within it
         if (principal != null && !(principal instanceof WebPrincipal)) {
             Principal additional = SecurityContext.getCurrent().getAdditionalPrincipal();
-            if (additional != null) {
+            if ((additional != null) && (additional instanceof WebPrincipal) && ((WebPrincipal)additional).getCustomPrincipal() == principal) {
                 principal = additional;
             }
         }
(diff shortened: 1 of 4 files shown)


danPowell52:
jenkins test please

PR #825, opened 2016-05-20
PAYARA-800 Payara Micro Maven Deployer gives a FileNotFoundException

Pandrex247:
Add break clause to loop to stop it overwriting the correct artefactURL with an incorrect one if there are still more repositories to search


smillidge:
jenkins test please

Issue #662, opened 2016-02-15
Weld NPE when invalidating sessions


Pandrex247:
[PAYARA-638](https://jira.c2b2.co.uk/browse/PAYARA-638)
[GLASSFISH-21166](https://java.net/jira/browse/GLASSFISH-21166)

smillidge:
Fixed by #837

PR #837, opened 2016-05-25
Fixes PAYARA-638 - NPE on logout in EAR files

lprimak:
added WeldTerminationListenerProxy to check if BeanManager is available since all WARs in an EAR are always assumed to be CDI

removes NPE that was caused by this
Fixes PAYARA-638 and Issue #662


lprimak:
jenkins test please

lprimak:
jenkins test please

PR #844, opened 2016-06-02
Proposal for a new Payara github page

OndrejM:
I created a proposal for a new Payara github page, as we discussed. Have a look at this url: http://ondrejm.github.io/Payara/. Internal issue PYSVC-66.

After PR accepted, the page should be available at http://payara.github.io/Payara/


smillidge:
Looks good.

PR #872, opened 2016-06-23
Added error logging when unable to bind to port.

OndrejM:
A very often error that prevents server startup is that a port cannot be bound because it is already occupied. Payara server just logs a stacktrace and shuts down.

This update adds error logging of address and port, which could not be bound.


smillidge:
Jenkins test please

PR #877, opened 2016-06-24
Response writer -> Grizzly is now just one buffer - byte buffer

lprimak:
Fixes PAYARA-797


lprimak:
jenkins test please

lprimak:
This is now tested & Fixed in Grizzly 2.3.26: https://java.net/jira/browse/GRIZZLY-1839
For now, this workaround can be used:
in ```<domain_dir>/config/default-web.xml```, add the following:
```
<init-param>
  <param-name>genStrAsByteArray</param-name>
  <param-value>false</param-value>
</init-param>
```

PR #878, opened 2016-06-27
Fix payara minimal

Pandrex247:
Our packages weren't being included in payara minimal, which was breaking the rest service


Pandrex247:
jenkins test please

Pandrex247:
Only a band-aid, rejecting for now as it doesn't fully solve the problem

PR #883, opened 2016-06-29
Moved Payara Micro out of nucleus and into appserver/extras

Pandrex247:
Payara Micro living in nucleus is quite restrictive when trying to avoid having dependencies from appserver. 

Since Payara Micro is its own thing, I've moved it into appserver/extras and made the necessary refactors. I also removed the payara-micro package, as I'm not entirely sure why Payara Micro was being included in the full release.

This needs to be debated and tested intensively!


appserver/extras/payara-micro/payara-micro-distribution/pom.xml
```diff
@@ -81,26 +79,39 @@
...
+                    </configuration>
+                </execution>
+            </executions>
         </plugin>
+        <plugin>
+            <artifactId>maven-jar-plugin</artifactId>
+            <executions>
+                <execution>
+                    <id>default-jar</id>
+                    <phase>never</phase>
```
Pandrex247: This was done to avoid maven creating an empty JAR in addition to the JAR built by ant. If someone knows of a better way please inform me!

OndrejM:
I think having payara-micro in payara-appserver-modules together with the CDI bus suits better.
It solves the problem a) too, while it seems a better place than extras/embedded, as there is very little Java code in subfolders of  embedded folder - these are mostly glue maven projects putting the pieces together.

Pandrex247:
Hmm, I still prefer having them in _extras/payara-micro_.
I prefer the idea of keeping _appserver/payara-appserver-modules_ for modules that actually go into appserver.
So for example, _JCache_ and the _EJB timer_ can live in there, as _JCache_ is actively used in Payara Server, and last I checked @smillidge had intentions to integrate the _hazelcast-ejb-timer_ into Payara Server (though atm it's only integrated into Payara Micro).
But the CDI module in there is specific to Payara Micro (afaik), so I don't think it's right to store it with stuff that it has no relation to.

OndrejM:
Actually, the CDI module has no dependency on Micro, and since Micro and Server can now join the same hazelcast cluster, it would make sense to integrate the CDI module into the server too.
But I agree that PayaraMicro API will always be related only to Payara Micro.

Pandrex247:
Are we looking at the same CDI module? I'm looking at the _payara-micro-cdi_ module, which is __only__ used and depended on by Payara Micro.

OndrejM:
Yes, but you are right. CDI module depends heavily on Payara Micro. I thought it depends only on Hazelcast, which is not the case.
But it would be nice to make it depend on plain hazelcast and integrate it into the full server to enable simple message passing between server and micro.

smillidge:
Jenkins test please

Pandrex247:
Jenkins test please

smillidge:
Ultimately I think we should be pulling Payara Micro into its own repo. Especially if we moved to some level of modularity.

Pandrex247:
The problem I see with that is that it currently depends on unpublished build artefacts, though it would be a lot cleaner if we did restructure it so that it could live on its own.

PR #915, opened 2016-07-12
PAYARA-832 add blue info to version

danPowell52:


smillidge:
jenkins test please

danPowell52:
Isn't this going to fail anyway because its going into payara blue?

smillidge:
Yes it will :-)

PR #920, opened 2016-07-13
PAYARA-900 Fixed NPE in Cluster Store if HZ is not enabled

smillidge:
get on the cluster store did not check whether the store was enabled resulting in a NPE


smillidge:
jenkins test please

PR #926, opened 2016-07-14
Fixes PAYARA-810

lprimak:
Code is now fetching correct (current) SecurityContext instead creating a new, invalid one


lprimak:
jenkins test please

PR #935, opened 2016-07-20
PAYARA-904 asadmin fails on add-resources when creating a jdbc pool w…

MeroRai:
…ith connection validation using a file


MeroRai:
jenkins test please

PR #945, opened 2016-07-25
Added links to JCache API

### LINKED ISSUE #942: JSR107 Javadocs should be linked from the root GitHub page
There is a link to the Payara Micro API javadoc, but no link to the jsr107 javadoc here:
http://payara.github.io/Payara/appserver_API/New%20folder/payara-appserver-modules_api/payara-jsr107_api/site/apidocs/

OndrejM:
Preview of the changes available at https://ondrejm.github.io/Payara/
Fixes #942.

### DIFF
--- a/index.html
+++ b/index.html
@@ -42,20 +42,32 @@
                 </p>
 
                 <h2>Payara Server Resources</h2>
-                <div class="resources buttons">
-                    <a class="btn" href="https://payara.gitbooks.io/payara-server/content/" target="_blank">Payara Server Documentation</a>
-                    <a class="btn" href="http://payara.github.io/Payara/nucleus_API/payara-modules/payara-micro/target/apidocs/" target="_blank">Payara Micro API Reference</a>
-                    <a class="btn" href="https://github.com/payara/Payara/issues" target="_blank">Issue Tracking</a>
-                    <a class="btn" href="https://github.com/payara/Payara" target="_blank">Sources on GitHub</a>
-                    <a class="btn" href="https://github.com/payara/Payara-Server-Documentation" target="_blank">GitHub repository for documentation pages</a>
-                    <a class="btn" href="https://jenkins.payara.fish/" target="_blank">Payara Jenkins CI</a>                    
-                </div>
+                    <div class="resources buttons">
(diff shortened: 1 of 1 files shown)


MeroRai:
jenkins test please

OndrejM:
The branch gh-pages does not contain any code therefore any tests. It is only static html page available here: https://payara.github.io/Payara/

smillidge:
jenkins test please

OndrejM:
@smillidge, @MeroRai - this branch does not contain any test, only static html page. Therefore Jenkins fails saying no tests found.

jerrinot:
Hi, some feedback. As a user I find this confusing:  
image

What's the difference between "Payara JSR 107 API" and  "JSR 107 (JCache) API" ? 
(I actually know what the difference is, but I still find it non-intuitive)

OndrejM:
Thanks for the fedback, @jerrinot. I also find it confusing, but I could find a better way to name the links. What would you suggest?

smillidge:
Payara Cache API?

OndrejM:
How is the Payara API related to JCache API? Does it depend on JCache, extend JCache? "Payara Cache API" suggests it is an alternative to standard JCache API, which I think is not the case...

jerrinot:
Payara API is just a glue, isn't it? Integration / CDI extension, something along these lines.

smillidge:
Yep just glue and a CDI extension for @NamedCache and some producers for injecting CacheManager and CachingProvider.

OndrejM:
What do you think about "Payara JCache extensions" and "JCache API"? (I would get rid of JSR 107, as it is very obscure for beginners)

jerrinot:
I agree "JSR-107" sounds obscure. Analogy: I use "Servlet API", not "JSR 340"

"JCache API" sounds good. 
"Payara JCache Extensions" less so. It does not give away what is it good for. What about "Payara JCache Integration(s)" ? + perhaps some intro at directly at the [site](http://payara.github.io/Payara/appserver_API/New%20folder/payara-appserver-modules_api/payara-jsr107_api/site/apidocs/) ?

PR #956, opened 2016-07-27
PAYARA-930 upgrade mimepull to expected version for jersey

### LINKED ISSUE #954: Wrong version of Mimepull included in the modules folder
I'm using Payara 162 and have a REST service running on the server. It uses the FormDataParam and FormDataContentDisposition classes provided by jersey-media-multipart. Now when trying the REST service, I get an exception that the method org.jvnet.mimepull.MIMEMessage.close() does not exist.

After some digging around I think I have found the cause. In the modules directory is jersey-media-multipart.jar included, which is version 2.22.1. This uses mimepull version 1.9.6. But when I look at the version of mimepull that is included in the modules directory, I see that that version is 1.9.4. When I look at that code I can confirm that that version does not include the "close" method.

Fix it by updating the mimepull version included in the modules folder to 1.9.6.

smillidge:
Fixes #954

### DIFF
--- a/nucleus/pom.xml
+++ b/nucleus/pom.xml
@@ -210,7 +210,7 @@
         <jettison.version>1.3.7</jettison.version>
         <jax-rs-api.spec.version>2.0</jax-rs-api.spec.version>
         <jax-rs-api.impl.version>2.0.1</jax-rs-api.impl.version>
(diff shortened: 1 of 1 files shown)


smillidge:
jenkins test please

PR #957, opened 2016-07-27
PAYARA-926 trace EJB timers

smillidge:
This adds tracing of EJB timers by adding a start and end trace to the timer task


smillidge:
jenkins test please

PR #968, opened 2016-07-29
PAYARA-851 Access logging command line option for Payara Micro

MeroRai:


smillidge:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -265,6 +275,30 @@ public void setUserLogFile(String fileName) {
     }
     
     /**
+     * Set user defined file directory for the access log
+     *
+     * @param filePath
+     */
+    public void setAccessLogDir(String filePath) {
+        if (!filePath.endsWith("/")) {
+
```
OndrejM: If value does not end in "/", the value is ignored and access log file is not created. We could treat access log argument in the same way as logFile argument - if "/" is not at the end, it specifies path to a file instead of directory. Even better, I would fix both access log and userLogFile to add a check whether the passed argument is a file or directory, in case "/" is not at the end. Otherwise people might create a directory beforehand and pass it as argument without a slash, and this would cause an error - Payara Micro would expect it to be a file, but it is a directory.

OndrejM: In any case, if there is an unexpected value in the argument, Payara Micro should log this as an error. I could not see anything in the logs to give me a hint on what I was doing wrong (that I did not add "/" at the end of the path) and I had to look into the source code.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -1368,13 +1427,19 @@ private void scanArgs(String[] args) {
                                 + "  --disablePhoneHome Disables sending of usage tracking information\n"
                                 + "  --version Displays the version information\n"
                                 + "  --logToFile <file-path> outputs all the Log entries to a user defined file\n"
+                                + "  --accessLog <directory-path> Sets user defined directory path for the access log\n"
+                                + "  --accessLogFormat Sets user defined a log format for access log\n"
```
OndrejM: I think the article "a" is redundant. I would change it to: "Sets user defined log format ..."

smillidge:
I've added a new pull request to add some better error checking and feedback to the user

OndrejM:
Just to link them together, smillidge's PR is #981

Issue #967, opened 2016-07-29
Race condition on restart command


pdudits:
I am seeing race condition where during restart the main thread exits sooner than restart command executes `System.exit(ret)` in `RestartServer`.

This causes JVM to exit with status 0, so the watchdog process (`asadmin start-domain` or `start-local-instance`) would not relaunch the server.

So there's CommandRunner Thread with following stack trace:

````
"Thread-22"
	at com.sun.enterprise.v3.admin.RestartServer.doExecute(RestartServer.java:127)
	at com.sun.enterprise.v3.admin.RestartDomainCommand.execute(RestartDomainCommand.java:109)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:539)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:535)
	at java.security.AccessController.doPrivileged(AccessController.java)
	at javax.security.auth.Subject.doAs(Subject.java:360)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2.execute(CommandRunnerImpl.java:534)
	at org.glassfish.api.AsyncImpl$1$1.run(AsyncImpl.java:76)
````

Racing with 
````
"DestroyJavaVM"
	at java.lang.Shutdown.shutdown(Shutdown.java:234)
````

I think the solution is that RestartServer should set it's thread as non-daemon, so the JVM would wait until it exits.

PR #969, opened 2016-07-29
Issue #967: Payara-945 Perform System.exit in non-daemon thread during server restart.

pdudits:
It was bit more involved than I originally thought, but one cannot make a running daemon thread a non-daemon thread. Starting an another one was the only way to keep it close to the actual restart code.


smillidge:
jenkins test please

PR #980, opened 2016-07-30
Production Payara domain enhancements - PAYARA-846

lprimak:


ggam:
What about enabling Hazelcast by default?

lprimak:
jenkins test please

lprimak:
@ggam I am not sure, I will pose the question to the rest of the team

ggam:
@lprimak it is one of Payara killer features. What are the objections there?

lprimak:
@ggam It's just a big step and I want other team members to chime in first, that's all.
I personally think it's a good idea.

lprimak:
Also, it drags in a few more questions, like,
- do we switch default session replication to hazelcast?
- there is also new feature about EJB timers stored in hazelcast, do we enable that?

It's not quite as trivial as it seems at first glance

smillidge:
I think a new config within the payaradomain is a good idea. hz-cluster-config or something like that?

pdudits:
I think enabling hazelcast with its default multicast discovery would be quite dangerous production setting. Hazelcast themselves recommend TCP-IP joiners for production.

lprimak:
I think given all the comments, this deserves some discussion, and probably a separate issue,
instead of shoehorning it into this issue.

ggam:
@lprimak I'll create another issue for that.

Another possible production improvement: enable precompilation of jsp by default.

Btw a development configuration would also ve nice.

ggam:
Another one I just found: when creating a new data source, enable ping by default

lprimak:
@ggam Unfortunately, your above comments would not be in scope for this enhancement.  This deals with application deployment and are not part of the domain creation

lprimak:
jenkins test please

lprimak:
jenkins test please

PR #982, opened 2016-08-01
Renamed links

OndrejM:
According to discussion in PR #945


OndrejM:
Only changes in gh-pages branch - static HTML pages. No tests in this branch.

PR #996, opened 2016-08-04
Add support for default pincipal->group mapping #906 and PAYARA-936

### LINKED ISSUE #906: Provide default group to role mapping
Most of the time a 1:1 group to role with same name mapping is done and so the glassfish-web.xml could be avoided. 

This is already done by WildFly and is proposed for JSR 375 (Security API).

An option should be added to revert to the original behaviour where no automatic mapping is done.

lprimak:
added property default-role-mapping in glassfish-web.xml
Fixes #906 and PAYARA-936

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/BundleDescriptor.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/BundleDescriptor.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [C2B2 Consulting Limited and/or its affiliates]
 
 package com.sun.enterprise.deployment;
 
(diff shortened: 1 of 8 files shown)


lprimak:
jenkins test please

lprimak:
PR #996

OndrejM:
Hi @lprimak,

For glassfish-application.xml and glassfish-ejb-jar.xml, the new element <property> should be also added to their DTD, is that correct? I don't see the element in the [DTD for application.xml](https://github.com/payara/Payara/blob/master/appserver/deployment/dtds/src/main/resources/glassfish/lib/dtds/glassfish-application_6_0-1.dtd).

lprimak:
@OndrejM you are right.  I assumed that if it was in glassfish-web.xml it would be in the other two, I guess i was wrong

lprimak:
jenkins test please

PR #1000, opened 2016-08-07
Fixes PAYARA-813 and PAYARA-690 spurious "resource not found" messages

lprimak:
added log strings file to EJB container to quiet the "resource bundle not found" messages

Also fixed logging in JavaEE kernel (MEJBService and it's proxy) to quiet similar warning messages


lprimak:
jenkins test please

PR #1012, opened 2016-08-09
PAYARA-986 Add -configuration suffix to the set commands of both requ…

MeroRai:
…est tracing and notification service.


MeroRai:
jenkins test please

Issue #668, opened 2016-02-19
CDI EJB Field producers' validation fail with incompatible types


pdudits:
I'm deploying an EAR, within which there is a war with producer:

````
public class ServiceProducer {
    @EJB(lookup = "java:/app/ejb-module/EjbInterfaceImpl")
    @Produces
    @AlternativeStereotype // stereotype enabled in beans.xml
    EjbInterface contractService;
}
````

Deployment fails with
````
[2016-02-19 17:14:17,673] [admin-listener(1)] [ERROR] [javax.enterprise.system.core]: Exception while loading the app : CDI definition failure:The type of the injection point fieldName is EjbInterface.  The type of the physical resource is com.pletely.DifferentEjb They are incompatible. 
javax.enterprise.inject.spi.DefinitionException: The type of the injection point fieldName is EjbInterface.  The type of the physical resource is com.pletely.DifferentEjb They are incompatible. 
	at org.glassfish.weld.services.InjectionServicesImpl.throwProducerDefinitionExeption(InjectionServicesImpl.java:319) ~[weld-integration.jar:na]
	at org.glassfish.weld.services.InjectionServicesImpl.validateResourceClass(InjectionServicesImpl.java:310) ~[weld-integration.jar:na]
	at org.glassfish.weld.services.InjectionServicesImpl.validateEjbProducer(InjectionServicesImpl.java:234) ~[weld-integration.jar:na]
	at org.glassfish.weld.services.InjectionServicesImpl.registerInjectionTarget(InjectionServicesImpl.java:197) ~[weld-integration.jar:na]
	at org.jboss.weld.manager.InjectionTargetFactoryImpl.postProcess(InjectionTargetFactoryImpl.java:156) ~[weld-osgi-bundle.jar:2015-09-08 14:57]
	at org.jboss.weld.manager.InjectionTargetFactoryImpl.createInjectionTarget(InjectionTargetFactoryImpl.java:93) ~[weld-osgi-bundle.jar:2015-09-08 14:57]
	at org.jboss.weld.bean.ManagedBean.<init>(ManagedBean.java:100) ~[weld-osgi-bundle.jar:2015-09-08 14:57]
	at org.jboss.weld.bean.ManagedBean.of(ManagedBean.java:80) ~[weld-osgi-bundle.jar:2015-09-08 14:57]
[...]

PR #1017, opened 2016-08-12
Fixes PAYARA-947 and github issue #668

lprimak:
fixed a bug when injection value was checked against first available EJB, not the correct one


lprimak:
jenkins test please

PR #1024, opened 2016-08-12
Fixes PAYARA-989

lprimak:


lprimak:
jenkins test please

hobrom:
This commit has a severe side-effect: 
`Paths.get( xxx )` will throw an `InvalidPathException` if the originating request URL contains characters not allowed in your operating system's file-system path (e.g. semicolon ':' on Windows).

how to reproduce:
I'm using apache isis, which heavily depends on URLs like `http://localhost:8080/myapp/w/entity/domainapp.dom.refund.WorkContract:1`. Note the semicolon at the end! Those URLs provoke an exception!

PR #1052, opened 2016-08-20
PAYARA-1010 update ConstantPool info to match JDK7+

### LINKED ISSUE #959: BufferUnderflowException in case of using java 8 with lambdas
As described in 
http://stackoverflow.com/questions/28301584/javaee-glassfish-bufferunderflowexception - in case an EJB Module is using Lambdas, ConstantPoolInfo first fails to recognize 18 (INVOKEDYNAMIC, see http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.4) and then goes all wrong.
We have verified the fix proposed by Koen Serneels in the above SO thread.
We're  currently using Payara 4.1.1.162.
Is the fix for the above issue already included in some later branch/release, or are there an plans to include it?
In case you would like to include it by didn't yet have time for it - please let me know. I'd be happy to create a pull request.

smillidge:
Fixes #959


smillidge:
jenkins test please

PR #1071, opened 2016-08-30
Payara 866 Microprofile.io distribution

smillidge:
Initial cut of the Payara Microprofile.io distribution


smillidge:
jenkins test please

PR #1090, opened 2016-09-13
Correct JMS Port property - fixes PAYARA-892

lprimak:
re-add JMS port property to domainProperties because its a "substitutable property"


lprimak:
jenkins test please

PR #1107, opened 2016-09-23
PAYARA-931 Payara Micro Request Tracing

Pandrex247:
Adds in command line options to enable and configure request tracing in Payara Micro.
Also turns on the Notification service by default.

The new options are:

* `--enableRequestTracing` - Just a flag; you **don't** need to add _true_ or _false_ afterwards. 
* `--requestTracingThresholdUnit` - Accepts a String value determining the time unit, defaulting to seconds if no value is given. The accepted strings are (case insensitive): _NANOSECONDS_, _MICROSECONDS_, _MILLISECONDS_, _SECONDS_, _MINUTES_, _HOURS_, and _DAYS_.
* `--requestTracingThresholdValue` - Accepts a _long_ to determine the threshold value before a request is traced, defaulting to 30 if no value is entered.

As an example of usage:
`java -jar payaramicro.jar --enableRequestTracing --requestTracingThresholdUnit NANOSECONDS --requestTracingThresholdValue 7`

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
@@ -1472,7 +1472,9 @@ public class PayaraMicro {
                                 + "  --logProperties <file-path> Allows user to set their own logging properties file\n"
                                 + "  --accessLog <directory-path> Sets user defined directory path for the access log\n"
                                 + "  --accessLogFormat Sets user defined log format for the access log\n"
-                                + "  --enableRequestTracing Enables the Request Tracing Service"
+                                + "  --enableRequestTracing Enables the Request Tracing Service\n"
+                                + "  --requestTracingThresholdUnit Sets the time unit for the requestTracingThresholdValue option\n"
+                                + "  --requestTracingThresholdValue Sets the threshold time before a request is traced\n"
                                 + "  --help Shows this message and exits\n");
                         System.exit(1);
                         break;
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
@@ -1572,6 +1572,7 @@ public class PayaraMicro {
                                 + "  --logProperties <file-path> Allows user to set their own logging properties file\n"
                                 + "  --accessLog <directory-path> Sets user defined directory path for the access log\n"
                                 + "  --accessLogFormat Sets user defined log format for the access log\n"
+                                + "  --requestTracing Enables the Request Tracing Service and optionally sets the threshold unit and/or value\n"
                                 + "  --enableRequestTracing Enables the Request Tracing Service\n"
                                 + "  --requestTracingThresholdUnit Sets the time unit for the requestTracingThresholdValue option\n"
                                 + "  --requestTracingThresholdValue Sets the threshold time before a request is traced\n"
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
@@ -1483,9 +1483,7 @@ public class PayaraMicro {
                                 } 
                                 // If the first entry is a String
                                 else if (requestTracing[0].matches("\\D+")) {
-                                    
-                                    String parsedUnit = 
-                                                parseRequestTracingUnit(
+                                    String parsedUnit = parseRequestTracingUnit(
                                                         requestTracing[0]);
                                     try {
                                         TimeUnit.valueOf(parsedUnit.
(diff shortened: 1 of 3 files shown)


appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -1410,6 +1428,17 @@ private void scanArgs(String[] args) {
                     case "--disablePhoneHome":
                         disablePhoneHome = true;
                         break;
+                    case "--enableRequestTracing":
+                        enableRequestTracing = true;
+                        break;
+                    case "--requestTracingThresholdUnit":
```
smillidge: There should be some validation on the the units passed in on the command line to ensure they are valid units.

appserver/extras/payara-micro/payara-micro-core/src/main/resources/microdomain.xml
```diff
@@ -155,6 +155,12 @@ Portions Copyright [2016] [Payara Foundation and/or its affiliates]
             </transaction-service>
```
smillidge: microdomain nocluster also need configuring correctly

Pandrex247:
Updated with required changes.
I've also made it so it accepts lowercase characters for the _requestTracingThresholdUnit_.

smillidge:
jenkins test please

smillidge:
jenkins test please

michaelranaldo:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -1410,6 +1429,34 @@ private void scanArgs(String[] args) {
...
+                                    + " for --requestTracingThresholdUnit", 
+                                    requestTracingThresholdUnit);
+                            throw e;
+                        }
+                        i++;
+                        break;
+                    case "--requestTracingThresholdValue":
+                        try {
+                            requestTracingThresholdValue = 
+                                    Long.parseLong(args[i + 1]);
```
jerrinot: this smells like `ArrayIndexOutOfBoundsException` when a user forgets to include the actual units.

jerrinot:
Two parameters for a single thing look a bit awkward. 

Why not 
` --requestTracingThreshold 2` <-- defaults to seconds = common case

+parsing units from the value itself when needed?
` --requestTracingThreshold 500ms`

From the other hand It could become somewhat clumsy with microseconds (`us` or `μs` ?) of with days (1 day**s**?)

Pandrex247:
I like the idea of a single option, we could even go full hog and combine all 3 into one (enable, time, and units) - though in either situation I'm fond of keeping the discrete options for flexibility. I'll have a play and see what I can cobble together.

I'll check for any obvious _OutOfBounds_ cases, but I believe several of the configuration options for Payara Micro can fall afoul of this, so is probably worth its own PR to fix it rather than tacking it onto this one.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/PayaraMicro.java
```diff
@@ -1410,6 +1431,112 @@ private void scanArgs(String[] args) {
                     case "--disablePhoneHome":
                         disablePhoneHome = true;
                         break;
+                    case "--requestTracing":
```
jerrinot: this is screaming for refactoring.

Pandrex247:
Added in a new option: `--requestTracing`

This option enables the request tracing service, and optionally configures the threshold unit and value. I've also added extra unit validation (to both this option, and the original one), allowing you to now enter:

- Nanoseconds (case-insensitive), Nanosecond (case-insensitive), ns
- Microseconds (case-insensitive), Microsecond (case-insensitive), us, μs
- Milliseconds (case-insensitive), Millisecond (case-insensitive), ms
- Seconds (case-insensitive), Second (case-insensitive), s
- Minutes (case-insensitive), Minute (case-insensitive), Mins (case-insensitive), Min (case-insensitive), m
- Hours (case-insensitive), Hour (case-insensitive), h
- Days (case-insensitive), Day (case-insensitive), d

I decided against enforcing proper grammar and refusing options such as "3day", as not everyone cares as much as I do.

Example usages: 

- Enable request tracing - `java -jar payara-micro.jar --requestTracing`
- Enable request tracing and set threshold unit to nanoseconds (value defaults to 30) - `java -jar payara-micro.jar --requestTracing ns`
- Enable request tracing and set threshold value to 2 (unit defaults to seconds) - `java -jar payara-micro.jar --requestTracing 2`
- Enable request tracing and set threshold to 4 microseconds - `java -jar payara-micro.jar --requestTracing 4ms`

In its current state, entering something like `--requestTracing 4 minutes` will have the _minutes_ ignored. If this is particularly wanted though, it can be done.

I've elected to not fix the potential Array index out of bounds issues in this PR, as we're looking at spinning the _scanArgs_ method out into a proper parsing class.

Note, these options override each other without warning (like most others) - the last one entered will override the previous. For example:

- `java -jar payara-micro.jar --requestTracingThresholdUnit DAYS --requestTracing 1ns` will enable request tracing, and set the threshold to 1 _Nanosecond_.
- `java -jar payara-micro.jar --requestTracing 1ns --requestTracingThresholdUnit DAYS` will enable request tracing, and set the threshold to 1 _Day_.

smillidge:
jenkins test please

smillidge:
jenkins test please

PR #1121, opened 2016-10-04
PAYARA-1056 Global thread pool statistics incorrect

Pandrex247:
This pull request makes the global thread pool statistics much more complete, with it now counting the total maximum and minimum threads in all of the thread pools of a config.

It also edits the way the monitoring works so that it is more accurate by querying the JVM for the actual current number of threads, rather than just relying upon events to alter internally stored values (which may have been missed or reset).


Pandrex247:
After some extra thought, don't think this is ready yet.

PR #1123, opened 2016-10-06
PAYARA-1056 global thread pool statistics incorrect

Pandrex247:
This pull request makes the global thread pool statistics much more complete, with it now counting the total maximum and minimum threads in all of the thread pools of a config.

It also edits the way the monitoring works so that it is more accurate by querying the JVM for the actual current number of threads, rather than relying upon events to alter internally stored values (which may have been missed or reset depending on when the monitor listener was registered).

Also renames the global thread pool stats from _thread-pool_ to _global-thread-pool-stats_ to make it more clear.


smillidge:
jenkins test please

smillidge:
jenkins test please

michaelranaldo:
jenkins test please

PR #1125, opened 2016-10-06
JCache beans created by JSR107Producer use generics (PAYARA-1107)

jerrinot:
This allows to use JCache beans to be applied into injection sites with
type parameters.

For example this works now:
```java
@Inject
private Cache<LongKey, Schedule> scheduleCache;
```


jerrinot:
I am not a CDI expert, but this appears to do the trick. 

Injection into raw `JCache` (without type parameters) still works. I am not quite sure how to write an automated test for this.

smillidge:
jenkins test please

michaelranaldo:
jenkins test please

smillidge:
We'll add tests to our test suite

PR #1126, opened 2016-10-06
Use name from CacheDefaults when injecting JCache (PAYARA-1106)

### LINKED ISSUE #1124: Proposal: Use javax.cache.annotation.CacheDefaults when injecting JCache
Payara uses `CacheDefaults` for cache resolution in the CDI interceptor as the spec mandates. 

I believe it would make a lot of sense to use `CacheDefaults` as a cache resolution strategy when injecting JCache via `@Inject`. 

Currently it does use name of a declaring class as a cache name. There is also the `@NamedCache` but that creates a dependency on Payara-specific code.

jerrinot:
Fixes #1124

Disclaimer: I don't really understand CDI lifecycle. I assume the producer method is called for each injection point. 

[Example usage](https://github.com/jerrinot/microprofile-conference/commit/53a08594d6ce36896becc4e7d6a0a07aee57dd71) (depends also on https://github.com/payara/Payara/pull/1125)


smillidge:
jenkins test please

michaelranaldo:
jenkins test please

smillidge:
jenkins test please

PR #1139, opened 2016-10-17
Don't cast cached Throwable to Exception as it's too narrow. (PAYARA-1126)

jerrinot:
Reasoning:
JCache spec and Payara implementation allows caching of exceptions.
In theory everything extending Throwable can be cached -> casting to
Exception is  not safe as it can throw ClassCastException


smillidge:
jenkins test please

PR #1146, opened 2016-10-24
PAYARA-1114 distributed deadlock in JTA

lprimak:
added java.util.concurrent locks in order to resolve deadlock when callbacks are called from within synchornized methods

### DIFF
--- a/appserver/transaction/jts/src/main/java/com/sun/jts/CosTransactions/RegisteredSyncs.java
+++ b/appserver/transaction/jts/src/main/java/com/sun/jts/CosTransactions/RegisteredSyncs.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [Payara Foundation and/or its affiliates]
 
 //----------------------------------------------------------------------------
 //
@@ -67,11 +68,9 @@ import java.util.*;
 import org.omg.CORBA.*;
 import org.omg.CosTransactions.*;
 
-import com.sun.jts.trace.*;
 import java.util.logging.Logger;
 import java.util.logging.Level;
 import com.sun.logging.LogDomains;
-import com.sun.jts.utils.LogFormatter;
 
 /**
  * The RegisteredSyncs class provides operations that manage a set of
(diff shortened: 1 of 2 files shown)


lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

smillidge:
Is there a lot of white space changes or did you do a Major refactor of TopCoordinator as it won't display on GH?

lprimak:
jenkins test please

lprimak:
It's a major refactor.  This is only used for XA transaction.  Not sure if remote only, but looks like only remove (CORBA-based) XA transaction use this code

lprimak:
jenkins test please

lprimak:
I've double-checked and the classes are not getting used outside of remote (CORBA) XA transactions, so it's much lower rick than previously thought

pdudits:
I've definetely met TopCoordinator when debugging my previous XA issues, and those were usually non-CORBA transactions, rather ActiveMQ MDBs doing some JPA.

appserver/transaction/jts/src/main/java/com/sun/jts/CosTransactions/RegisteredSyncs.java
```diff
@@ -113,6 +112,15 @@
      */
     RegisteredSyncs() {}
 
+    private RegisteredSyncs(List<Synchronization> rhs) {
+        registered.addAll(rhs);
+    }
+    
+    @Override
+    protected RegisteredSyncs clone() throws CloneNotSupportedException {
```
pdudits: If you'd dropped the throws clause here, you wouldn't need to handle it in TopCoordinator.

lprimak: Didn't test the MDB/MQ scenario. Thanks @pdudits !
As far as the CloneNotSupported exception, I get a warning otherwise, and it's good form to support it since it overrides the interface...

lprimak:
jenkins test please

lprimak:
jenkins test please

Issue #1062, opened 2016-08-25
Fail-Fast when attempting to build on JDK 8


jerrinot:
I'm on https://github.com/payara/Payara/commit/4660a87c296768a77b9f562f3ad3fa08a750adc4 and when I tried to build Payara by invoking the `gfbuild.sh` script I've got this error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:2.4:compile (default-compile) on project rest-service: Compilation failure
[ERROR] error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider org.glassfish.admin.rest.annotation.processor.RestModelExtensionProcessor not found
[ERROR] -> [Help 1]
[ERROR] 
```

This is with `Oracle JDK 8u101`

When I set my `JAVA_HOME` to `Oracle JDK 7u80` then the build script finishes just fine. 

The build script should check the current Java version and fail-fast - with a descriptive error message - when an unsupported Java version is found. The current error is rather cryptic and does not give away what's wrong.

smillidge:
actually the build script needs deleting as you should just be able to build Payara using

mvn install from the root.

mikecroft:
We've got issue #338 to purge those old bits and pieces.

mikecroft:
Fixed by #1151

jerrinot:
so is Payara buildable on JDK 8?

smillidge:
yes it's buildable on JDK8 with maven

mikecroft:
Also, if you run `mvn clean install -PQuickBuild -DskipTests` you'll build almost everything you need very quickly. It's really handy for initial builds.

More detail here:
https://github.com/payara/Payara/pull/1073

Issue #338, opened 2015-06-14
Remove Oracle specific build scripts.


smillidge:
Root directory contains Oracle specific build scripts. They should be removed.

mikecroft:
Internal issue no `PAYARA-1059`

OndrejM:
When resolved, also close the issue #1062, as it's a duplicate

mikecroft:
Fixed by #1151

PR #1151, opened 2016-10-26
PAYARA-1059 removed old Oracle build files we don't need

smillidge:


mikecroft:
Fixes #338, fixes #1062

PR #1152, opened 2016-10-27
PAYARA-618 Enable the DAS to discover Payara Micro instances

Pandrex247:
Allows the DAS to view information about the Payara Server and Payara Micro instances in the Hazelcast cluster.

Introduces a new Asadmin command, `list-hazelcast-cluster-members`, that lists information about the instances in the domain's Hazelcast cluster:

Usage: `list-hazelcast-cluster-members [--type=type]`
The `--type` option accepts a String of _micro_ or _server_, and filters the result to only list information about Payara Micro or Payara Server instances respectively.

This PR also modifies the _Hazelcast -> Cluster Members_ page under _Domain_ (formally could be found under _server_) in the Admin Console to display the new information.

Also moves configuration of Hazelcast for the DAS from _server_ into _Domain_, as we're looking to move to a model with a single Hazelcast domain-wide cluster utilising deployment groups over having a separate hazelcast configuration for each instance.


Pandrex247:
Might be worth refactoring the package and module names, as it's no longer just for Payara Micro.

Pandrex247:
Jenkins test please

Pandrex247:
This PR also changes the behaviour of the `BuildMicro` profile slightly - this profile now activates building of just the Micro distribution - the Core and CDI modules will still be compiled if this profile is omitted.

A new profile is also introduced to split out the MicroProfile build from that of Micro: `BuildMicroProfile`

Pandrex247:
Jenkins test please

Issue #585, opened 2015-12-29
PWC6117: File "null" not found


andreasernst:
Sometimes after a restart and admin console login this error occurs:

[2015-12-29T20:33:35.823+0100] [Payara 4.1] [SEVERE] [] [org.apache.jasper.servlet.JspServlet] [tid: _ThreadID=40 _ThreadName=admin-listener(2)] [timeMillis: 1451417615823] [levelValue: 1000] [[
  PWC6117: File "null" not found]]

A second restart helps. Payara Server 4.1.1.154 #badassfish (build 270)

mikecroft:
Does this happen with a completely fresh Payara or one that you have used for a while and changed? It looks like this is coming from Jasper, so I would assume you have deployed an application with a servlet to Payara?

Do you have a reliable way that we could reproduce the problem here?

andreasernst:
This is for me a known issue, also with older Glassfish versions. This happens sometimes after restarting Payara, i guess Payara does not find the index/home file for the admin console.

mikecroft:
How does this affect the admin console? Does it not load correctly?

Can you also attach the server.log which contains the issue, please?

andreasernst:
The console crashes, you cannot login anymore, until you restart the server.

[2015-12-31T21:37:04.242+0100] [Payara 4.1] [SEVERE] [] [org.apache.jasper.servlet.JspServlet] [tid: _ThreadID=49 _ThreadName=jk-connector
-8009(4)] [timeMillis: 1451594224242] [levelValue: 1000] [[
  PWC6117: File "null" not found]]

Hhmm, strange is that the jk-connector-8009 is shown, because the admin console runs on admin port with 4848. Jk-connector-8009 is for the Apache HTTP server.

mikecroft:
That's strange - I grepped my own `/opt/payara` directory which contains lots of versions of Payara and found no occurrence of that message. I searched for the error code and found a stackoverflow answer about Jasper used in Jetty which created a file in `/tmp` which was deleted and caused an error the next time Jetty tried to use it. Payara also creates temporary files in /tmp when new applications are deployed, so I wonder if there is a required file for a separate deployment which is causing Jasper to fail?

How is your Payara instance configured? Does it run under a permissions-limited user?

jerrinot:
I am not sure if it's related but some Linux distributions occasionally clean `/tmp`. 

We used to have a similar issue with Hazelcast Management Centre which indeed uses an embedded Jetty.

[...]

PR #1157, opened 2016-10-29
PAYARA-1008 remove nuisance PWC6117 errors

lprimak:
suppress Apache Jasper logging severe when a JSP file is not found, w…hich is a common occurence in production web sites

Also may apply, or at least make #585 less confusing


lprimak:
jenkins test please

PR #1163, opened 2016-11-03
PAYARA-1065 re-introduce blocking stateless bean pool

### LINKED ISSUE #646: Bean-pool unexpected behavior
Hi all.

Looks like impossible to control the number of concurrently executed threads on a stateless bean.

Lets consider following case - out application communicate with billing platform, but billing have some limitations - number of concurrent requests (5 for instance).
So, in this case we can to achieve this ("we hope to achieve :)") with bean pool for BillingDelegateBean for instance.

Lets consider following vendor specific "glassfish-ejb-jar.xml" deployment descriptor:

```xml
<?xml version="1.0"`encoding="UTF-8"?>
<!DOCTYPE glassfish-ejb-jar PUBLIC "-//GlassFish.org//DTD GlassFish Application Server 3.1 EJB 3.1//EN"`"http://glassfish.org/dtds/glassfish-ejb-jar_3_1-1.dtd">
<glassfish-ejb-jar>
    <enterprise-beans>
        <ejb>
            <ejb-name>BillingDelegateBean</ejb-name>
            <bean-pool>
                <steady-pool-size>2</steady-pool-size>
                <resize-quantity>1</resize-quantity>
                <max-pool-size>5</max-pool-size>
            </bean-pool>
        </ejb>
    </enterprise-beans>
</glassfish-ejb-jar>
```

But it doesn't works at all, no guarantee that will be instantiated only 5 BillingDelegateBean classes by EJB container.

Futhermore, max-wait-time-in-millis was depricated

https://docs.oracle.com/cd/E19798-01/821-1750/beauq/index.html

So it looks like long time ago application server was able to provide this behavior, but it was depricated :(

After googling i found many posts about "How to control the number of concurrently executed threads"
- http://www.adam-bien.com/roller/abien/entry/do_we_need_stateless_session

>  With pools you can throttle the concurrency, because most (all I know) application servers allow you the configuration of max number of instances. You can control the concurrency of every bean type (Boundary) in a fine grained way. This is important for the avoidance of DoS attacks and dealing with legacy resources.
- https://java.net/nonav/projects/glassfish/lists/users/archive/2011-03/message/585

It looks like misunderstanding, because many people expect this behavior, but it dosn't work at all, this is terrible. And i don't know alternative how to solve this problem in EJB - Java EE way.

From glassfish documentation - "GlassFish Application development guide - Pooling and Caching"

> Note that the steady-pool-size and max-pool-size parameters only govern the
> number of instances that are pooled over a long period of time. They do not
> necessarily guarantee that the number of instances that may exist in the JVM at a given
> time will not exceed the value specified by max-pool-size. For example, suppose an
> idle stateless session container has a fully-populated pool with a steady-pool-size of
> 10. If 20 concurrent requests arrive for the EJB component, the container creates 10
> additional instances to satisfy the burst of requests. The advantage of this is that it
> prevents the container from blocking any of the incoming requests. However, if the
> activity dies down to 10 or fewer concurrent requests, the additional 10 instances are
> discarded.

"The advantage of this is that it prevents the container from blocking any of the incoming requests." 

In my opinion this is not advantage, at now this is unexpected behavior, i think container should be able to provide some blocking mechanisms in some cases.

lprimak:
- removed duplication of code from AbstractPool
- added BlockingPool for limiting resource consumption of @Stateless beans
- refactored and upgraded code base for Java 7 syntax

Fixes #646

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016] [Payara Foundation and/or its affiliates]
 
 package com.sun.ejb.containers;
 
@@ -63,6 +64,7 @@ import org.glassfish.ejb.deployment.descriptor.runtime.IASEjbExtraDescriptors;
 import com.sun.ejb.ComponentContext;
 import com.sun.ejb.EjbInvocation;
 import com.sun.ejb.containers.util.pool.AbstractPool;
+import com.sun.ejb.containers.util.pool.BlockingPool;
 import com.sun.ejb.containers.util.pool.NonBlockingPool;
 import com.sun.ejb.containers.util.pool.ObjectFactory;
 import com.sun.ejb.monitoring.stats.EjbMonitoringStatsProvider;
@@ -255,10 +257,19 @@ public class StatelessSessionContainer
 
         poolProp = new PoolProperties(ejbContainer, beanPoolDes);
         String val = ejbDescriptor.getEjbBundleDescriptor().getEnterpriseBeansProperty(SINGLETON_BEAN_POOL_PROP);
-        pool= new NonBlockingPool(getContainerId(), ejbDescriptor.getName(),
-           sessionCtxFactory, poolProp.steadyPoolSize,
-           poolProp.poolResizeQuantity, poolProp.maxPoolSize,
-           poolProp.poolIdleTimeoutInSeconds, loader, Boolean.parseBoolean(val));
+        if(beanPoolDes.getMaxWaitTimeInMillis() != -1) {
+            pool= new BlockingPool(getContainerId(), ejbDescriptor.getName(),
+                sessionCtxFactory, poolProp.steadyPoolSize,
+                poolProp.poolResizeQuantity, poolProp.maxPoolSize,
+                poolProp.poolIdleTimeoutInSeconds, loader, Boolean.parseBoolean(val),
+                beanPoolDes.getMaxWaitTimeInMillis());
+        }
+        else {
+            pool= new NonBlockingPool(getContainerId(), ejbDescriptor.getName(),
+                sessionCtxFactory, poolProp.steadyPoolSize,
+                poolProp.poolResizeQuantity, poolProp.maxPoolSize,
+                poolProp.poolIdleTimeoutInSeconds, loader, Boolean.parseBoolean(val));
+        }
     }
 
     protected void registerMonitorableComponents() {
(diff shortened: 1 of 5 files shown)


lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
```diff
@@ -248,17 +252,26 @@ protected void initializeHome()
...
             beanPoolDes = iased.getBeanPool();
         }
 
         poolProp = new PoolProperties(ejbContainer, beanPoolDes);
         String val = ejbDescriptor.getEjbBundleDescriptor().getEnterpriseBeansProperty(SINGLETON_BEAN_POOL_PROP);
-        pool= new NonBlockingPool(getContainerId(), ejbDescriptor.getName(),
-           sessionCtxFactory, poolProp.steadyPoolSize,
-           poolProp.poolResizeQuantity, poolProp.maxPoolSize,
-           poolProp.poolIdleTimeoutInSeconds, loader, Boolean.parseBoolean(val));
+        if(poolProp.maxWaitTimeInMillis != -1) {
```
smillidge: What is the default here? Will people that have just created an EJB without any special descriptor or annotations get the original NonBlockingPool?

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
```diff
@@ -759,6 +772,8 @@ public void destroy(Object obj) {
 
     private static class PoolProperties {
         int maxPoolSize;
+        int maxWaitTimeInMillis = Integer.getInteger(String.format("fish.payara.ejb-container.%s", RuntimeTagNames.MAX_WAIT_TIME_IN_MILLIS),
```
smillidge: Have you created a new deployment descriptor field here? Does this need documenting by Mike's team.

lprimak: No. this field already exists but marked deprecated. I think wE do need to document it as it works a bit differently now.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
lprimak: Yes. Default is the unchanged behavior. I tried to document it in the github issue and in the Payara issue

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
mikecroft: documentation issue ID: `PYSVC-136`

PR #1164, opened 2016-11-04
PAYARA-703 Create HipChat Notifier

mulderbaba:


mulderbaba:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/RequestTracingNotificationEventFactory.java
```diff
@@ -44,6 +45,12 @@ NotificationEvent build(long elapsedTime, NotifierType notifierType) {
             event.setMessage(getRequestEventsAsStr());
             return event;
         }
+        if (NotifierType.HIPCHAT.equals(notifierType)) {
```
smillidge: I don't like this if type equals code in this class as this ties Request Tracing to HipChat. Shouldn't request tracing just send a notification not caring what types of notifiers exist or are configured ?

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotifierService.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    @Subscribe
+    public void handleNotification(HipchatNotificationEvent event) {
+
+        HipchatNotifierConfigurationExecutionOptions executionOptions =
+                (HipchatNotifierConfigurationExecutionOptions) getNotifierConfigurationExecutionOptions();
+        String urlStr = MessageFormat.format(RESOURCE, executionOptions.getRoomName(), executionOptions.getToken());
+        try {
+            URL url = new URL(concatenateEndpoint(urlStr));
```
smillidge: Can you put the sending of the request to HipChat on a seperate thread. I wouldn't want request tracing to be notifying HipChat in the middle of a user's request. 
Also can you make this a try with resources to ensure the output stream is always closed if there is an exception

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/configuration/HipchatNotifier.java
```diff
@@ -0,0 +1,23 @@
+/*
```
smillidge: Can you use the new license boilerplate
```java
/*
 * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
 *
 * Copyright (c) 2016 Payara Foundation and/or its affiliates. All rights reserved.
 *
 * The contents of this file are subject to the terms of either the GNU
 * General Public License Version 2 only ("GPL") or the Common Development
 * and Distribution License("CDDL") (collectively, the "License").  You
 * may not use this file except in compliance with the License.  You can
 * obtain a copy of the License at
 * https://github.com/payara/Payara/blob/master/LICENSE.txt
 * See the License for the specific
 * language governing permissions and limitations under the License.
 *
 * When distributing the software, include this License Header Notice in each
 * file and include the License file at glassfish/legal/LICENSE.txt.
 *
 * GPL Classpath Exception:
 * The Payara Foundation designates this particular file as subject to the "Classpath"
 * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
 * file that accompanied this code.
 *
 * Modifications:
 * If applicable, add the following below the License Header, with the fields
 * enclosed by brackets [] replaced by your own identifying information:
 * "Portions Copyright [year] [name of copyright owner]"
 *
 * Contributor(s):
 * If you wish your version of this file to be governed by only the CDDL or
 * only the GPL Version 2, indicate your decision by adding "[Contributor]
 * elects to include this software in this distribution under the [CDDL or GPL
 * Version 2] license."  If you don't indicate a single choice of license, a
 * recipient has the option to distribute your version of this file under
 * either the CDDL, the GPL Version 2 or to extend the choice of license to
 * its licensees as provided above.  However, if you add GPL Version 2 code
 * and therefore, elected the GPL Version 2 license, then the option applies
 * holder.
 */
```

nucleus/payara-modules/notification-core/pom.xml
```diff
@@ -71,7 +71,14 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>18.0</version>
+            <version>19.0</version>
```
smillidge: Shouldn't this use the property defined in the higher level pom?

nucleus/payara-modules/notification-core/pom.xml
```diff
@@ -71,7 +71,14 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>2.2.6</version>
```
smillidge: If we are going to start using mokito can the version be moved up to a top level pom and defined as a property?

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/domain/execoptions/NotifierConfigurationExecutionOptionsFactory.java
```diff
@@ -28,11 +29,19 @@ and Distribution License("CDDL") (collectively, the "License").  You
...
 
-    public NotifierConfigurationExecutionOptions build(NotifierConfiguration notifierConfiguration) {
+    public <T extends NotifierConfigurationExecutionOptions> T build(NotifierConfiguration notifierConfiguration) {
         if (notifierConfiguration instanceof LogNotifierConfiguration) {
             LogNotifierConfigurationExecutionOptions executionOptions = new LogNotifierConfigurationExecutionOptions();
             executionOptions.setEnabled(Boolean.parseBoolean(notifierConfiguration.getEnabled()));
-            return executionOptions;
+            return (T) executionOptions;
+        }
+        if (notifierConfiguration instanceof HipchatNotifierConfiguration) {
```
smillidge: I hate this instanceof stuff it messes with my OO trained brain and mashes up the dependencies. Can't we use some sort of factory pattern or builder pattern here?

mulderbaba:
jenkins test please

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotifierService.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    @Subscribe
+    public void handleNotification(HipchatNotificationEvent event) {
+        HipchatNotifierConfigurationExecutionOptions executionOptions =
+                (HipchatNotifierConfigurationExecutionOptions) getNotifierConfigurationExecutionOptions();
+        new Thread(new HipchatNotificationRunnable(executionOptions,
+                        event.getUserMessage(),
+                        event.getMessage()), PREFIX + threadNumber.getAndIncrement())
+                .start();
```
jerrinot: does this start a new thread for each message? would it better to have a single thread and a message queue? or use (already existing?) executor service? 

imagine there is a connectivity issue, hip-chat API is unavailable or whatever - then you rely on timeout and if your producing rate is high then you can have quite a few threads hanging.

mulderbaba: correct! I should revise with an executor service.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotificationRunnable.java
```diff
@@ -0,0 +1,118 @@
...
+                                String message) {
+
+        this.executionOptions = executionOptions;
+        this.userMessage = userMessage;
+        this.message = message;
+    }
+
+    @Override
+    public void run() {
+        String urlStr = MessageFormat.format(RESOURCE, executionOptions.getRoomName(), executionOptions.getToken());
```
jerrinot: shouldn't the roomname be encoded? I don't know what chars are allowed in the roomname though.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotifierService.java
jerrinot: Why not a thread consuming messages from a queue? `HipchatNotifierService` would only place a message to the queue.

This way you could even re-use a connection.  Now each message triggers TLS negotiation, etc...

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/configuration/HipchatNotifier.java
mulderbaba: done on the new ones. should I replace for all the payara-nucleus-modules that I have as well?

nucleus/payara-modules/notification-core/pom.xml
mulderbaba: that's the same case as guava version. we need to decide on where that should land in project-wise.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotificationRunnable.java
mulderbaba: I checked and it supports url encoded parameters, will revise it thanks.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/HipchatNotifierService.java
mulderbaba: as for `HttpURLConnection` I cannot cache it as its Javadoc states and don't need also since underlying connection is seamlessly reused by other instances but queue is a good idea indeed.

jerrinot: I didn't know the underlying connection is re-used behind the scene, good one!

mulderbaba:
jenkins test please

mulderbaba:
jenkins test please

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/service/hipchat/HipchatMessageQueue.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Queue;
+
+/**
+ * Created by mertcaliskan
+ */
+@Service
+@RunLevel(StartupRunLevel.VAL)
+public class HipchatMessageQueue {
+
+    private Queue<HipchatMessage> messageQueue = new LinkedList<>();
```
jerrinot: you better use `ConcurrentLinkedQueue` or something. `LinkedList` is not thread-safe.

mulderbaba: good choice, I didn't care for it since I only have 1 consumer/producer pondering on the queue

mulderbaba:
jenkins test please

PR #1192, opened 2016-11-19
Payara 1143 Integrated CDI Event Bus into Payara Full

smillidge:
Pull request integrates the CDI Event bus into Payara Full.
Also moves event processing to a separate thread using the default ManagedExecutor
Also adds additional targeting of Outbound events by a specific event name or by instance name. Also allows loopback whereby the event will also be received inbound on the instance the fired the outbound event.


smillidge:
jenkins test please

smillidge:
jenkins test please

smillidge:
I also have concerns about it but can't think of a better way at the moment to get off the non-JavaEE Hazelcast Thread onto a "JavaEE" thread.

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/ClusteredCDIEventBus.java
```diff
@@ -0,0 +1,203 @@
...
+    private final static String INSTANCE_PROPERTY = "InstanceName";
+    
+    private final static String EVENT_PROPERTY = "EventName";
+    
+    @PostConstruct
+    void postConstruct() {
+        runtime.addCDIListener(this);
+        capturedClassLoader = Thread.currentThread().getContextClassLoader();
+        im = Globals.getDefaultHabitat().getService(InvocationManager.class);
+        capturedInvocation = im.getCurrentInvocation();
```
OndrejM: I tried running a simple app with the nightly build from 13 Jan 2017, and `im.getCurrentInvocation();` returns null, leading to 
```
	at fish.payara.micro.cdi.ClusteredCDIEventBus.eventReceived(ClusteredCDIEventBus.java:140)
	at fish.payara.appserver.micro.services.PayaraInstance.receiveMessage(PayaraInstance.java:162)
```

smillidge:
micro is in flux on nightly at the moment.

OndrejM:
OK, let me know when ready for testing :)

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/ClusteredCDIEventBus.java
OndrejM: Can't we use the Java EE ContextService to avoid using the hack with "captured" invocation manager? What is the issue here - is it because you are trying to access the executor from a hazelcast Java SE thread? I think that ContextService should enrich the current thread to enable using the executor.

smillidge: Yeah it's because we need to get the invocation off the Hazelcast Thread onto a "JavaEE" thread with the correct application specific classloader.

lprimak: I may be able to help with that along with some of my other work, so stay tuned

PR #1194, opened 2016-11-20
PAYARA-822 JASPIC utilities

smillidge:
Add utility classes to enable simple registration of a custom Server Auth Module


smillidge:
@lprimak most of the api is defined by JASPIC so can't be changed to generics. Some of the internal constructors may be changeable. I will try and update the PR.

smillidge:
Done for the constructors. The other methods are overrides and can't be touched.

lprimak:
jenkins test please

PR #1196, opened 2016-11-21
PAYARA-1207 Reduce RequestEvent test testFlushMultipleThreads 

davidweaver:
threshold from 100 to 95 to handle vm time variations


davidweaver:
jenkins test please

PR #1215, opened 2016-11-30
Change of maven project name form GF to Payara

OndrejM:
Change of maven project name form GF to Payara


smillidge:
jenkins test please

PR #1216, opened 2016-11-30
PAYARA-1241 Extend RequestTrace elapsedtime test upper threshold from…

davidweaver:
… 1100 to 1500 to handle potential system delay


davidweaver:
jenkins test please

PR #1217, opened 2016-12-02
PAYARA-1224 Concurrency bug in session stats provider

Pandrex247:
Replaces the circular FIFO bounded queue with a ThreadLocal variable, which is used to check if a session was just created (within the same thread).

Also corrects the weird formatting introduced in the previous PR.


Pandrex247:
This PR does not fix the weird monitoring figures we're observing in this area, which is being handled in a separate issue: PAYARA-1246

davidweaver:
jenkins test please

PR #1224, opened 2016-12-08
PAYARA-1260 upgraded jline to 2.14.2 which is the latest stable version

mikecroft:
http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22jline%22%20AND%20a%3A%22jline%22


smillidge:
jenkins test please

Issue #936, opened 2016-07-20
Web fragment definitions cannot be overriden in web.xml


pdudits:
A servlet or filter mapping for a servlet that is defined in a fragment cannot be changed in `web.xml`. It appears to be validated sooner than fragments are processed, resulting in error:

```
java.lang.RuntimeException: There is no web component by the name of s4 here.
    at org.glassfish.web.deployment.node.WebCommonNode.endElement(WebCommonNode.java:248)
    at org.glassfish.web.deployment.node.WebBundleNode.endElement(WebBundleNode.java:158)
    at com.sun.enterprise.deployment.node.SaxParserHandler.endElement(SaxParserHandler.java:622)
    at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.endElement(AbstractSAXParser.java:609)
    at com.sun.org.apache.xerces.internal.impl.xs.XMLSchemaValidator.endElement(XMLSchemaValidator.java:857)
    at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanEndElement(XMLDocumentFragmentScannerImpl.java:1783)
    at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:2970)
    at com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:606)
    at com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.next(XMLNSDocumentScannerImpl.java:118)
    at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanDocument(XMLDocumentFragmentScannerImpl.java:510)
    at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:848)
    at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:777)
    at com.sun.org.apache.xerces.internal.parsers.XMLParser.parse(XMLParser.java:141)
    at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.parse(AbstractSAXParser.java:1213)
    at com.sun.org.apache.xerces.internal.jaxp.SAXParserImpl$JAXPSAXParser.parse(SAXParserImpl.java:643)
[...]

PR #1228, opened 2016-12-12
PAYARA-1007 fixes overrides in web.xml

lprimak:
correct behavior when overriding servlet-mapping from web-fragments in web.xml

fixes issue #936


lprimak:
jenkins test please

PR #1230, opened 2016-12-13
PAYARA-1259 fixed access logging admin console interface

lprimak:
fixed bug where it was impossible to turn off access logging to console via admin interface


lprimak:
jenkins test please

Issue #1122, opened 2016-10-04
Bug in parametrized log messages


rds76:
For example:
Source: [LogUtils.java](https://github.com/payara/Payara/blob/master/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/LogUtils.java#L289`)
correct: `message = "Invalid request scheme for Endpoint {0}. Expected ''{1}'', received ''{2}''.",`

Regarding [MessageFormat](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html), params must be escaped with double single quotes.

**There are many such a wrong escaped log messages in source tree!**

mikecroft:
Created internal issue `PAYARA-1115`

PR #1255, opened 2016-12-25
PAYARA-1115 log messages single quote escaped

lprimak:
fixed single quote escapes
fixes issue #1122


lprimak:
jenkins test please

PR #1263, opened 2017-01-03
PAYARA-1185 Application Client JMS connectivity

lprimak:
Now using ORB/EJB host for default JMS connection address


lprimak:
jenkins test please

PR #1268, opened 2017-01-06
PAYARA-1263 Change name of thread pool internally to differentiate po…

Pandrex247:
…ols and listeners


smillidge:
jenkins test please

PR #1282, opened 2017-01-12
Replicate original set command key instead of processed one. Fixes #1280

### LINKED ISSUE #1280: asadmin set fails to replicate properties that contain dot in name
# Description #
----------

When launching command like

    asadmin set "config-name.security-service.message-security-config.SOAP.provider-config.ClientProvider.property.signature\.key\.alias=s1as"

then the value of the property is not replicated to respective instances. The value is also not replicated when changing the value via GUI. This affects not only security service, but any property nodes (connection pools, custom resources, ...)

## Expected Outcome

The new value of the property is applied and can be inspected over instance's REST interface or AMX.

## Current Outcome

Asadmin reports that command failed to replicate.

> Failure: Command set failed on server instance test-instance: remote failure: No
 configuration found for test-instance.security-service.message-security-config.SOAP.provider-config.ClientProvider.property.signature.key

The cause for this behavior is, that replicated set command argument does not contain the escape characters (`\`) from the input, and the instance therefore misinterprets the command, failing with error.

1. Start the domain

       asadmin start-domain payaradomain

2. Create a local instance

       asadmin create-local-instance test-instance

3. Start it

       asadmin start-instance test-instance

4. Change the property containing dot.    

       asadmin set test-instance.security-service.message-security-config.SOAP.provider-config.ClientProvider.property.signature\.key\.alias=other

5. Compare the values on DAS vs the instance

       asadmin get test-instance.security-service.message-security-config.SOAP.provider-config.ClientProvider.property.signature.key.alias
       asadmin --port 24848 get test-instance.security-service.message-security-config.SOAP.provider-config.ClientProvider.property.signature.key.alias

## Workaround

Workaround is to restart the instance after changes are applied to DAS.

    asadmin restart-instance test-instance

## Context

In my case I was creating a custom resource of type java.util.Properties where property key contained dots (as they usually do). However this affects any dotted property name.

## Fix

Fix appears to be changing both [invocation ](https://github.com/payara/Payara/blob/32f9e683cf01ed0b96de79648dafefa0a6087aa2/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java#L373) [sites](https://github.com/payara/Payara/blob/32f9e683cf01ed0b96de79648dafefa0a6087aa2/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java#L498) of `replicateSetCommand` to pass `target` instead of `targetName`.

However there is quite [complex logic](https://github.com/payara/Payara/blob/32f9e683cf01ed0b96de79648dafefa0a6087aa2/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java#L279) around obtaining `targetName` in method `getAliasedParent` and branches following the call. However in both of my test cases - security service config and custom resource config, they do not change the name except for dropping the escape characters. I also don't see under which circumstances the target instances would get different result to `getAliasedPattern` provided the other input is `Domain`.

## Environment ##

- **Payara Version**: 4.1.1.163
- **Edition**: Full
- **JDK Version**: 8u45

pdudits:
As discussed in the issue,  the patch is correct unless there exists a situation, where instance would give different result to ` getAliasedParent(domain, pattern)` than the DAS.

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2017] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.v3.admin;
 
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/SetCommand.java
@@ -370,7 +370,7 @@ public class SetCommand extends V2DottedNameSupport implements AdminCommand, Pos
                     ConfigSupport.createAndSet((ConfigBean) parentNode, Property.class, attributes);
                     success(context, targetName, value);
                     runLegacyChecks(context);
-                    if (targetService.isThisDAS() && !replicateSetCommand(context, targetName, value))
+                    if (targetService.isThisDAS() && !replicateSetCommand(context, target, value))
                         return false;
                     return true;
                 } catch (TransactionFailure transactionFailure) {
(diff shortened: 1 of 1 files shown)


lprimak:
jenkins test please

lprimak:
jenkins test please

PR #1290, opened 2017-01-18
PAYARA-1324 session replication resilience and improvements

lprimak:
more resilience of session serialization in presence of errors, such as:
- ignore non-serializable fields with a warning, even if they are inside a serializable object

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/session/ManagerBase.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/session/ManagerBase.java
@@ -55,6 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// Portions Copyright [2016-2017] [Payara Foundation and/or its affiliates]
 
 package org.apache.catalina.session;
 
@@ -73,10 +74,15 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.*;
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
+import java.io.BufferedOutputStream;
 import java.io.DataInputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
+import java.io.OutputStream;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.text.MessageFormat;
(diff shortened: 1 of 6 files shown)


lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

jerrinot:
Is skipping non-serializable fields a good idea? This can lead to some hard-to-find errors. Imagine you put a non-serilizable field into a collection and it starts to disappear. the warning can help, but I still struggle to find a real-world use-case for this. Also it can have security implications.

lprimak:
jenkins test please

lprimak:
@jerrinot it won't skip individual fields / entries in a collection, but will not replicate the whole session attribute if it won't serialize, thus it won't affect any security or introduce bugs.
This is for replication only, the current process state of the session will remain unaffected, thus there will be no regressions of any kind.

Currently, if there is some kind of serialization issue in the session, the moment you turn on ```<distributable/>``` things will start to fail in weird ways, and good luck with session failover, rolling upgrades, etc. 
Basically, if there is any issue, the session would have been lost and not replicated, and thus failover will not work for most applications.

The cool thing about these fixes is that all of the sudden most applications will be magically able to run in distributable mode and have at least a good shot at working failover.
Local processing will remain the same, so no issues with the current deployments will crop up

lprimak:
These are basically bug fixes to make session replication work as intended by it's designers.

lprimak:
jenkins test please

lprimak:
@jerrinot as far as use cases, take a look at the referenced issues.
There are other similar issues that have happened here as well

jerrinot:
@lprimak: thanks for explanation. I was concerned about dropping _inconvenient attributes_ for the sake of serialization. 

If there is a non-serialized object in a session then in my opinion a container should _fail-fast_, at least by default. So an application developer has a feedback and can act on it. 

As a developer I don't like when a library/container/server/whatever is trying to fix my errors. I want to know about them! 

From the other hand I _think_ I saw a non-serialized attribute dropping in some other container. I might be mistaken, but perhaps it was in WebSphere? So maybe it makes sense in some cases. Should it be enabled by default though? Doesn't it compromise Payara as a drop-in replacement of Glassfish?

lprimak:
This only affects session replication, local session is unaffected, thus it preserves "drop-in replacement" for GlassFish.
In the past, session replication won't happen at all in that case, with weird errors,
so this is only an improvement from zero to replication, not a "change" in replication.

There is a big warning presented when serialization doesn't happen properly.
The big issue here is that non-serializable parts are mostly due to dependencies which users can't control, and don't really matter in the grand scheme of things.

PR #1292, opened 2017-01-18
PAYARA-1283 Edit logging of set-monitoring-configuration error to mak…

Pandrex247:
…e it less verbose and easier to read.

The command will now log something along the lines of this:

`Exception during command set-monitoring-configuration: A Property with the same key "HeapMemoryUsage" already exists in MonitoringServiceConfiguration null|#]`


mulderbaba:
jenkins test please

PR #1294, opened 2017-01-18
PAYARA-1282 jmx monitoring service is not dynamic until first restart

Pandrex247:
This fix is implemented by adding the default configuration tags for this service to the default domain.xml files, so will not take affect for those migrating from previous versions (if they have not already made a configuration change to the _monitoring-service-configuration_ service)


mulderbaba:
jenkins test please

lprimak:
jenkins test please

PR #1297, opened 2017-01-19
PAYARA-1150 Provide greater clarity in error log for when setting an invalid value for the log file rotation size limit

Pandrex247:
Also adds in an extra message for when setting the file rotation time to an invalid value.


mulderbaba:
jenkins test please

Issue #1072, opened 2016-08-31
JDBC connection attempt does not time out


chrjohn:
We are using Payara 4.1.152.1 and recently had the following problem: the connection to our data base went down (network problem). When we tried to get a connection from the pool the components of our application were hanging until Payara was restarted about eight minutes later. The stack trace on shutdown shows that several components were waiting for a connection from the pool. Example:

```
"p: thread-pool-1; w: 9" #373 daemon prio=5 os_prio=0 tid=0x00007f12f6a73000 nid=0x4233 waiting on condition [0x00007f12e09f0000]
   java.lang.Thread.State: WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000000853c6460> (a java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireShared(AbstractQueuedSynchronizer.java:967)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(AbstractQueuedSynchronizer.java:1283)
        at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lock(ReentrantReadWriteLock.java:727)
        at com.sun.enterprise.resource.pool.datastructure.RWLockDataStructure.getResource(RWLockDataStructure.java:116)
        at com.sun.enterprise.resource.pool.ConnectionPool.getResourceFromPool(ConnectionPool.java:716)
        at com.sun.enterprise.resource.pool.ConnectionPool.getUnenlistedResource(ConnectionPool.java:632)
        at com.sun.enterprise.resource.pool.ConnectionPool.internalGetResource(ConnectionPool.java:526)
        at com.sun.enterprise.resource.pool.ConnectionPool.getResource(ConnectionPool.java:381)
        at com.sun.enterprise.resource.pool.PoolManagerImpl.getResourceFromPool(PoolManagerImpl.java:245)
        at com.sun.enterprise.resource.pool.PoolManagerImpl.getResource(PoolManagerImpl.java:171)
        at com.sun.enterprise.connectors.ConnectionManagerImpl.getResource(ConnectionManagerImpl.java:360)
[...]

PR #1148, opened 2016-10-26
PAYARA-769 Make resetAssociation() more robust

retoo:
We've ported the patch from pull request https://github.com/payara/Payara/pull/770 to a 4.1.1.163 

We have been running this patched version in multiple environments for 3 months. 

This fixes #769

PR #1301, opened 2017-01-20
PAYARA-769 Make resetAssociation() more robust (#1148)

lprimak:
Committed by @retoo
* Payara-769: Make resetAssociation() more robust by avoiding begin()ing tx in error handling/finally

Otherwise we will crash right again in our finally/resetAssociation block which will
cause the connection pool entry to have avery inconsistent state from which the
server does not recover.

* Payara-769: When errors occur during TXStart we should evict the connection from the pool

* Payara copyright

* Payara Copyright

* Payara Copyright


lprimak:
jenkins test please

lprimak:
I submitted the original PR #1148 into the wrong branch
It has already been reviewed & tested

lprimak:
fixes #1072

PR #1302, opened 2017-01-22
PAYARA-1343 allow host aware partitioning to be enabled on Hazelcast

### LINKED ISSUE #870: Add support to configure hazelcast partition groups
Hazelcast provides easy-to-configure mechanism of grouping members into bigger partitions to prevent replicas being stored in a cluster member on the same system.

Hazelcast partitioning defaults to having a partition per member, and replicas can be stored in any other member. Payara could support configuration of partition groups, as decribed in [hazelcast documentation](http://docs.hazelcast.org/docs/3.3/manual/html-single/hazelcast-documentation.html#partition-grouping). 

It is possible to define adavanced partitioning using hazelcast XML configuration file, but this cannot be combined with the configuration on domain.xml. Therefore at least support for `HOST_AWARE` partitioning would be very nice, as it is just a single configuration option.

smillidge:
Fixes #870


smillidge:
jenkins test please

smillidge:
jenkins test please

PR #1308, opened 2017-01-24
PAYARA-1314-Update-Request-Tracing-asadmin-commands-to-configure-historical-storing

mulderbaba:


mulderbaba:
jenkins test please

PR #1313, opened 2017-01-24
PAYARA-1044 Refresh logging configuration directly in the log manager…

smillidge:
… if Payara Micro


smillidge:
jenkins test please

PR #1318, opened 2017-01-25
PAYARA-1358-Notification-HealthCheck-RequestTracing-services-should-be-notified-on-configuration-change

mulderbaba:


mulderbaba:
jenkins test please

mulderbaba:
jenkins test please

smillidge:
@mulderbaba I'm actually not sure this code is right. Can you take a look at https://github.com/payara/Payara/blob/master/nucleus/payara-modules/service-exemplar/src/main/java/fish/payara/service/example/ExampleService.java and how it does Config change listener

PR #1321, opened 2017-01-26
PAYARA-1330 Send-asadmin-command does not return anything for commands that do not inherently log anything themselves

Pandrex247:
Adds in two new options: --verbose and --logOutput

The first option prints out the command output to the terminal form which you run the command.
The second option, enabled by default for the admin console, prints the output to the server log.


Pandrex247:
This seems to include a "joint commit" from myself and Steve due to rebasing and squelching.
@smillidge Do you want me to make a new branch and cherry-pick my commits to get rid of it? Leaving as is may make our commit history weird.

mulderbaba:
jenkins test please

PR #1326, opened 2017-01-27
PAYARA-895 Console log level should be FINE to allow SQL Trace to be …

smillidge:
…seen if enabled


smillidge:
jenkins test please

PR #1329, opened 2017-01-30
PAYARA-1324 rolling upgrade enhancements

lprimak:
Fixes numerous deployment issues connected to rolling upgrades,
including infamous "inconsistent module state" problems (maybe not all, but some)
Also automates many tasks that were previously manual for rolling upgrades


lprimak:
jenkins test please

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExternalConfigurationImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public Map<String, Object> getConfigurationProperties() {
+        return propsMap;
+    }
+
+    @Override
+    public void cleanup() {
+        // intentionally left blank
```
Pandrex247: As long as it doesn't take a multi-page essay to explain, can you expand on this comment a little bit as to why it's being left blank please? It helps if we ever have to come back to it :)

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExternalConfigurationImpl.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+
+    private final Map<String, Object> propsMap;
```
Pandrex247: I'm a pedant for convention - can this be declared at the top of the class please? :)

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
```diff
@@ -474,11 +475,14 @@ public WeldApplicationContainer load(WeldContainer container, DeploymentContext
         String archiveName = !isSubArchive? appInfo.getName() : archive.getName();
         int suffixIdx = archiveName.lastIndexOf('-');
         if(isSubArchive && suffixIdx > 0) {
-            archiveName = archiveName.substring(0, suffixIdx);
-            if(!context.getArchiveHandler().getArchiveType().isEmpty()) {
-                archiveName = String.format("%s.%s", archiveName, context.getArchiveHandler().getArchiveType());
+            String versionStr = archiveName.substring(suffixIdx + 1, archiveName.length());
+            if(versionStr.matches("^[0-9]+\\..*")) {
```
Pandrex247: This regex probably needs testing in a Windows environment before we can merge. At a bare minimum we need to check that this String will only ever contain "\\"s and not "/"s.

Also is the extra "." in your regex necessary? It seems redundant. If I'm reading it right it's currently matching one or more numbers at the beginning of the string, followed by a single reverse stroke, followed by any character, and then followed by any number of any characters. Why not just ".+"?

lprimak:
This isn't ready yet, I will put a ready label here when done

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExternalConfigurationImpl.java
lprimak: I usually leave a comment like this when writing an empty function so you know not to remove it. It seems less pertinent with the @override annotation now but still is a good habit I think.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExternalConfigurationImpl.java
lprimak: I like to put the functionality on top, always done so and it makes more sense to me to put API on top, but I realize it breaks with current GF conventions so if you still want me to put it on top, I will do so.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
lprimak: This checks for digit followed by dot followed by anything else. Windows shouldn't be an issue because the backslash is used as an escape sequence for the dot only.

lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
jenkins test please

lprimak:
@Pandrex247 can you review please? Thanks!

PR #1338, opened 2017-02-04
PAYARA-1386 allow environment variable substitution in the domain.xml

smillidge:
This PR provides the capability to add ${ENV=envvariable} into the domain.xml to enable environment variable substitution. This also works for some deployments including application deployed datasources.


smillidge:
jenkins test please

smillidge:
jenkins test please

icereed:
@smillidge Thank you for merging this feature.
This makes Payara Docker ready and configurable via environment variables.
I have two questions:

1. Can the placeholder be integrated into the domain.xml via asadmin, too?
2. Is there any possibility to provide default values for environment variable substitution, in case there is no environment variable given?

Thank you in advance!

smillidge:
for (1) it should work in most cases but you may have to escape it in the shell.

for (2) no it doesn't do that but it is a good idea and I will expand the feature to allow default values.

icereed:
Thank you for the quick answer. This is very good news.

PR #1351, opened 2017-02-07
PAYARA-1413-Healthcheck-threshold-configuration-fails-if-all-the-values-are-not-provided

mulderbaba:
sets default values for thresholds


mulderbaba:
jenkins test please

PR #1354, opened 2017-02-08
PAYARA-1383 Admin console integration for Healthcheck service

Pandrex247:
Adds an extra command for the admin console to use, combining the `healthcheck-service-configure` and `healthcheck-service-configure-thresholds` commands: `healthcheck-service-configure-checker-with-thresholds`

Can be used from the command line as a convenience method, taking the same parameters as both commands.

Also moves the REST target endpoints for the healthcheck commands to the correct location (under configs/config/health-check-service-configuration).


Pandrex247:
jenkins test please

Pandrex247:
jenkins test please

Pandrex247:
A remote target bug has appeared after my rebase - DON'T MERGE

PR #1359, opened 2017-02-08
PAYARA-1423 notification svc preview

OndrejM:
I'm not sure, but I think that the Notification service shouldn't be in technical preview since 164, or at least since 171. In that case, there should be no warning message in admin console.

If it is still in technical preview, **don't merge this PR yet**.


mulderbaba:
jenkins test please

PR #1364, opened 2017-02-09
PAYARA-1430 Asadmin recorder migration improvement

Pandrex247:
Have the asadmin recorder service create its domain.xml entry on startup to help negate errors upon migration.


smillidge:
jenkins test please

PR #1365, opened 2017-02-09
PAYARA-1422 Extend environment support to enable embedding in paths etc.

smillidge:


smillidge:
jenkins test please

nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java
```diff
@@ -124,20 +117,43 @@ public static Object getTranslatedValue(Object value) {
...
+                } 
                 i++;     
             }
             if (i >= MAX_SUBSTITUTION_DEPTH) {
                 Logger.getAnonymousLogger().severe(
                         Strings.get("TranslatedConfigView.badprop", 
                         i, origValue));
             }
+            if (!stringValue.equals(origValue))
+                return stringValue;
```
OndrejM: Returning the value here does not allow to mix ENV and system property references in the same value. If it makes the life of a developer easier, we can have the same syntax for both (with the ENV prefix), therefore the same logic. System properties would take preference if there exist both env var and sys prop with the same name).

smillidge: I'll test dropping down through to the env substitution code to see if that allows them to contain both otherwise it is an edge case for next release.

PR #1368, opened 2017-02-10
PAYARA-1383 Deprecate name parameter and fix pages

Pandrex247:
The name parameter is a reserved word, and so can't be used from the admin console or rest interface.


Pandrex247:
jenkins test please

PR #1374, opened 2017-02-10
PAYARA-1436-Healthcheck-UI-should-have-a-checkbox-for-log-notifier

mulderbaba:


mulderbaba:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/HealthCheckConfigurer.java
```diff
@@ -85,9 +85,12 @@ and Distribution License("CDDL") (collectively, the "License").  You
     @Param(name = "target", optional = true, defaultValue = "server-config")
     protected String target;
 
-    @Param(name = "enabled", optional = false)
+    @Param(name = "enabled")
     private Boolean enabled;
 
+    @Param(name = "notifierEnabled", optional = true, defaultValue = "false")
```
Pandrex247: We've just created a few PRs to remove default values from parameters when they're not necessary, as it essentially makes the parameter required.

If a parameter is optional, but also has a default value, it means that you could configure the healthcheck to have the notifier enabled in one command, but if you then run the command again without this parameter specified it will overwrite the current configuration and disable it again.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/HealthCheckConfigurer.java
```diff
@@ -144,8 +147,30 @@ public Object run(final HealthCheckServiceConfiguration healthCheckServiceConfig
...
+    private void enableLogNotifier(ActionReport actionReport, AdminCommandContext context) {
+        CommandRunner runner = serviceLocator.getService(CommandRunner.class);
+        ActionReport subReport = context.getActionReport().addSubActionsReport();
+
+        CommandRunner.CommandInvocation inv = runner.getCommandInvocation("healthcheck-log-notifier-configure", subReport, context.getSubject());
+
+        ParameterMap params = new ParameterMap();
+        params.add("dynamic", dynamic.toString());
+        params.add("target", target);
+        params.add("enabled", notifierEnabled.toString());
```
Pandrex247: In line with my comment about removing the default value from this parameter, this will need a null check:

```
if (notifierEnabled != null) {
    params.add("enabled", notifierEnabled.toString());
}
```

mulderbaba:
jenkins test please

mulderbaba:
jenkins test please

PR #1378, opened 2017-02-10
PAYARA-1422 

smillidge:
support mixed system property and environment property multiple times in value


smillidge:
jenkins test please

PR #1383, opened 2017-02-12
Payara 1434 cdi bus outbount instance names as array

OndrejM:
Changed instanceName from type String to array.
A single instance name can still be defined in the same way (due to how arrays work in annotations):

```
@Inject
@Outbound(instanceName = "Instance-1")
Event<String> event;
```

However, instead of using comma to separate more instances, multiple instance names are specified as an array:

```
    @Inject
    @Outbound(instanceName = {"inst1", "inst2"})
    Event<String> event;
```

Internaly, three commas (instead of a single comma) are used as a separator in the flattened string value, so that it doesn't clash if comma is used also inside an instance name.


smillidge:
jenkins test please

PR #1395, opened 2017-02-17
Increment version number to 172-SNAPSHOT

Pandrex247:


mulderbaba:
I think this is duplicated @Pandrex247 ?

smillidge:
this is into Payara Blue branch so it is not duplicate

PR #1399, opened 2017-02-20
PAYARA-1471-clarify-default-unpack-behaviour

mikecroft:
verified that setting -Djava.io.tmpdir does still change the unpack directory and that this is the default behaviour. Added this fact to the description.


mikecroft:
This is pretty low priority - its debatable whether this is really necessary here, but I think it's helpful.

mikecroft:
Also noticed that the unpackdir description wasn't being printed. It looks like it just wasn't found due to a capital letter in the key in the command options. Seems odd, though, because there are other capitals which don't cause an issue. Ideally should be case insensitive, but that's a little more than a 2 minute patch.

smillidge:
jenkins test please

PR #1402, opened 2017-02-20
Improve micro command help

mikecroft:
A few changes to help with readability of the `--help` option in Payara Micro. Added padding to the option strings and alphabetised them.

Also split away the error strings from the commandoptions.properties file into their own file, since they're all camelcase and the actual options aren't. I looked at just ignoring case but there wasn't a native way of doing that with ResourceBundle and this was easier.

I haven't created any internal JIRA issue for this yet because it's low priority and other people might prefer some other behaviour.

Here's what it looks like:

```shell
➜  /store/dev/payara-projects/payara git:(improve-micro-command-help) java -jar  appserver/extras/payara-micro/payara-micro-distribution/target/payara-micro.jar --help
--accesslog                       <directory-path> Sets user defined directory path for the access log
--accesslogformat                 Sets user defined log format for the access log
--additionalrepository            <repo-url> specifies an additional repository to search for deployable artefacts in
--autobindhttp                    sets autobinding of the http port to a non-bound port
--autobindrange                   <number-of-ports> sets the maximum number of ports to look at for port autobinding
--autobindssl                     sets autobinding of the https port to a non-bound port
--clustername                     <cluster-name> sets the Cluster Group Name
--clusterpassword                 <cluster-password> sets the Cluster Group Password
--deploy                          <file-path> specifies a war file to deploy
--deploydir                       <directory-path> if set to a valid directory all ear/war files in this directory will be deployed
--deployfromgav                   <list-of-artefacts> specifies a comma separated groupId,artifactId,versionNumber of an artefact to deploy from a repository
--deploymentdir                   <directory-path> if set to a valid directory all ear/war files in this directory will be deployed
--disablephonehome                Disables sending of usage tracking information
[...etc...]
```


michaelranaldo:
jenkins test please

PR #1404, opened 2017-02-21
Payara 1039 we need a public api jars on maven central so people can use public payara apis

mulderbaba:


lprimak:
jenkins test please

PR #1405, opened 2017-02-21
PAYARA-539 change self signed

Cousjava:
The self-signed certificate that is generated by default in Payara no longer says it's generated by Oracle


Cousjava:
jenkins test please

PR #1446, opened 2017-03-09
PAYARA-1395 Allow full server to deploy applications from a maven repository and URLs

Pandrex247:
Introduces a new asadmin command, `deploy-remote-archive`, that allows you to deploy an artefact from a URL, or by providing comma separated GAV (GroupId, ArtefactId, VersionNumber) coordinates and Maven repositories to check and download from (similar to how the feature works in Payara Micro).

The new command takes the same parameters as the standard deploy command, with the addition of a new parameter: `--additionalRepositories`.
This parameter accepts a comma separated list of URLs that Payara will search for Maven artefacts in (Payara will only check in Maven Central by default).

Command usage for deploying directly from a URL (as in, without GAV conversion): `asadmin deploy-remote-archive url`

For example:
`asadmin deploy-remote-archive https://raw.github.com/Pandrex247/Payara_PatchedProjects/Payara-Maven-Deployer/fish/payara/testing/clusterjsp/1.1/clusterjsp-1.1.war`

If you don't specify a name or context root for the artefact when deploying direct from a URL, it will simply be the file name with the archive type stripped off (in the example above, it would be _clusterjsp-1.1_).

Command usage for deploying using GAV coordinates: `asadmin deploy-remote-archive --additionalRepositories repo1,repo2,...,repoN groupId,artefactId,VersionNumber`

For example:
`appserver/distributions/payara/target/stage/payara41/bin/asadmin deploy-remote-archive --additionalRepositories https://google.co.uk,https://raw.github.com/Pandrex247/Payara_PatchedProjects/Payara-Maven-Deployer fish.payara.testing,clusterjsp,1.1`

If you don't specify a name or context root for the artefact when deploying using GAV coordinates, it will just be the artefact name (in the example above, it would be _clusterjsp_).

As of this PR, the command will only deploy from HTTP or HTTPS URLs.
Please note that although the command accepts Maven GAV coordinates, it does not actually make use of Maven, and so artefacts will not be saved into your local maven repository.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -39,6 +39,7 @@
  */
 package fish.payara.micro.impl;
 
+import fish.payara.deployment.util.GAVConvertor;
 import fish.payara.micro.BootstrapException;
 import fish.payara.micro.boot.runtime.BootCommand;
 import fish.payara.micro.cmd.options.RuntimeOptions;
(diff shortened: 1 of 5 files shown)


mulderbaba:
jenkins test please

PR #1458, opened 2017-03-14
PAYARA-1199 Remove dependency on IIOP from MDB container. Fixes #164

### LINKED ISSUE #164: Allow disabling of iiop-service when MDBs present
Essentially the MessageBeanContainer fails to start unless iiop-service is configured. This seems like an undesirable dependency. We typically disable all uneeded elements when going to production but have found we can not disable iiop-service if we use MDBs. 

This was initially reported in

https://java.net/jira/browse/GLASSFISH-21156

smillidge:
Fixes #164


smillidge:
jenkins test please

PR #1461, opened 2017-03-16
PAYARA-1473-When-historicaltraceenabled-is-true-on-healthcheck-configure-notifierenabled-is-required

mulderbaba:


mulderbaba:
jenkins test please

PR #1467, opened 2017-03-20
PAYARA-1404 Deselect tickboxes on failure

Pandrex247:
The tickboxes in the Hazelcast micro instances table will now be properly deselected on failure.


smillidge:
jenkins test please

PR #1473, opened 2017-03-22
PAYARA-1509 condition for access logger may be read in from domain.xml

### LINKED ISSUE #1450: Support conditional Access Logs
# Description #
----------

Support condition on the Access Log in Payara. See https://tomcat.apache.org/tomcat-7.0-doc/config/valve.html#Access_Log_Valve/Attributes for the equivalent in Tomcat. The source code in PEAccessLog supports the conditrion attribute but there is no way to configure it in the domain.xml
See https://github.com/payara/Payara/blob/master/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java#L593 

## Expected Outcome

Conditional access logging is supported.

## Current Outcome

This is a pretty simple fix and will extend the functionality of access logging to bring it into line with the Tomcat equivalent.

Cousjava:
fixes #1450

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016] [Payara Foundation]
+// Portions Copyright [2016-2017] [Payara Foundation]
 
 package com.sun.enterprise.web;
 
@@ -907,6 +907,14 @@ public final class PEAccessLogValve
         } else if (maxHistoryFiles > 0) {
             historyFiles = new LinkedList<File>();
         }
+        
+        //Conditional access log
+        if (accessLogConfig != null) {
+            setCondition(accessLogConfig.getCondition());
+        } else {
+            setCondition(ConfigBeansUtilities.getDefaultCondition());
+        }
+        
         
         // log to console
         accessLogToConsole = Boolean.parseBoolean(accessLogConfig.getLogToConsoleEnabled());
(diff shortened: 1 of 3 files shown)


Cousjava:
Jenkins, would you kindly?

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/ConfigBeansUtilities.java
```diff
@@ -74,6 +75,13 @@ public ConfigBeansUtilities(Applications paramApps, Domain paramDomain) {
         domain = paramDomain;
     }
 
+    /*
```
smillidge: Why is this change needed if the default if not available will be null anyway?

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
```diff
@@ -908,6 +908,14 @@ void updateAccessLogAttributes(HttpService httpService,
             historyFiles = new LinkedList<File>();
         }
         
+        //Conditional access log
+        if (accessLogConfig != null) {
```
smillidge: Why check if not null and then do an else that sets it to null. Why not just set it regardless and not bother with the getDefaultCondition change?

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/ConfigBeansUtilities.java
Cousjava: I wrote it this way to match how others are set, and I'm not sure if the default condition is meant to be null - the other defaults mention that they come from a dtd which I have not been able to find.

PR #1478, opened 2017-03-23
PAYARA-1402 OSGi classloading failure improvements

lprimak:
log a warning when OSGi exception results from an API load, instead of swallowing the error


lprimak:
jenkins test please

PR #1479, opened 2017-03-23
PAYARA-992 fixed dynamic cluster config for JMX logging

lprimak:
fixed bug when starting and stopping JMX monitor on a cluster, some cleanup


lprimak:
jenkins test please

PR #1491, opened 2017-03-28
PAYARA-1495 catch null pointer in HTTPServiceStatsProvider

Cousjava:


Cousjava:
jenkins test please

smillidge:
Please do a null check rather than catch NPE

Cousjava:
replaced by #1495

PR #1493, opened 2017-03-28
PAYARA-1542 Embedded Micro doesn't shut down

Pandrex247:
Allows Embedded Payara Micro to actually shut down, as the refactor brought in from 171 broke this.


smillidge:
jenkins test please

PR #1514, opened 2017-04-08
PAYARA-1565 ensure Micro Instance uses the correct classloader when b…

### LINKED ISSUE #1508: Can't get a handle on running Payara Micro
# Description #
----------

I followed the tutorial under http://blog.payara.fish/the-health-check-service-in-depth-payara-micro

## Expected Outcome
That Payara would bring my application up and then start health checks. My application does run without trying to enable health checks, when my application starts.

## Current Outcome
I get this in the server.log:

java.lang.IllegalStateException: Payara Micro is not running
        at fish.payara.micro.impl.PayaraMicroImpl.getRuntime(PayaraMicroImpl.java:1027)
        at fish.payara.micro.impl.PayaraMicroImpl.getRuntime(PayaraMicroImpl.java:101)
        at fish.payara.micro.PayaraMicro.getRuntime(PayaraMicro.java:179)

I created a standard REST service and put the Payara initialisation code (see tutorial) in a @PostConstruct method in my ApplicationConfig class. But I also tried to execute it much later, when my service is first called, same result. Payara thinks it is not running, but it is, because my app works.

I can send the entire app, if this cannot be reproduced.

## Environment ##

- **Payara Version**: 4.1.1.171.1
- **Edition**: Micro
- **JDK Version**: 1.8.0_121-b13
- **Operating System**: Linux

smillidge:
…oot via java -jar

Fixes #1508

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroLauncher.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroLauncher.java
@@ -46,7 +46,6 @@ import java.net.URI;
 import java.security.CodeSource;
 import java.security.ProtectionDomain;
 import java.util.List;
-import sun.tools.tree.ThisExpression;
 
 /**
  * This class boots a Payara Micro Executable jar It establishes the Payara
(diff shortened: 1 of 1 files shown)


smillidge:
jenkins test please

PR #1515, opened 2017-04-09
PAYARA-1255 ensure ActivationConfigProperty values of MDBs are transl…

smillidge:
…ated.


lprimak:
Jenkins test please

lprimak:
@smillidge I tested this scenario. EnvironmentProperty has its own mechanism for system property substitution. 
Although I don't see any harm in this change, what prompted you to make it?
Just curious.

smillidge:
It didn't work with the ${ENV=} syntax for environment variables.

lprimak:
Ah!  I see.  This is the best way to make the change then, thanks!

PR #1516, opened 2017-04-10
Thread-safe initialization of InstanceResolverImpl. Fixes #1511. PAYARA-1577

### LINKED ISSUE #1511: Race condition when initializing JAX-WS endpoints
# Description #
----------

When a JAX-WS endpoint is hit with concurrent traffic right after application startup, the system might end up with endpoint instance, on which PostConstruct methods have not been invoked.

## Expected Outcome

All managed beans should be properly initialized before application code is invoked.

## Current Outcome

JAX-WS runtime delegates to instance that is held by [InstanceResolverImpl](https://github.com/payara/Payara/blob/32f9e683cf01ed0b96de79648dafefa0a6087aa2/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/InstanceResolverImpl.java). The class has field `instance`, with class obtained with injection manager, and field `resolver`, that should contain another singleton wrapper of the same instance.

During debugging I observed situation, where `instance` and `reslover.singleton` pointed to two different instances. Moreover the one held by resolver had state as if `@PostConstruct` method haven't been invoked on it.

`@PostConstruct` is invoked during [call to inject](https://github.com/payara/Payara/blob/32f9e683cf01ed0b96de79648dafefa0a6087aa2/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/InstanceResolverImpl.java#L94), so it could happen that field `instance` have changed between previous line and this line due to other thread also executing this branch. Since these are non-volatile reads we may get any constellation of field values from both threads.

Hit a SOAP endpoint that has `@PostConstruct` method with two concurrent requests right after application start.

## Environment ##

- **Payara Version**: 4.1.1.171
- **Edition**: Full
- **JDK Version**: 8u102

pdudits:
Resolver guarded with proper double-check lock. Field instance only changes when resolver is
initialized, therefore I believe it doesn't need any further synchronization.

### DIFF
--- a/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/InstanceResolverImpl.java
+++ b/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/InstanceResolverImpl.java
@@ -66,8 +66,8 @@ import javax.xml.ws.WebServiceException;
 public final class InstanceResolverImpl<T> extends InstanceResolver<T> {
    
     //delegate to this InstanceResolver
-    private  InstanceResolver<T> resolver;
-    private  T instance;
+    private volatile InstanceResolver<T> resolver;
+    private T instance;
     private final Class<T> classtobeResolved;
 
     private WSWebServiceContext wsc;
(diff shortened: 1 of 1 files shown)


Cousjava:
jenkins test please

Cousjava:
jenkins test please

PR #1519, opened 2017-04-10
Changed config name in admin console for hazelcast configuration

MattGill98:
Hazelcast configuration showed incorrect configuration name


MattGill98:
jenkins test please

PR #1522, opened 2017-04-11
Fixed application reload links

MattGill98:
The page was referencing an absolute path instead of a relative path.


MattGill98:
jenkins test please

PR #1530, opened 2017-04-13
PAYARA-1255 

smillidge:
do environment variable scanning for ConnectionFactoryDefinitions set via annotations


lprimak:
Jenkins test please

PR #1538, opened 2017-04-18
PAYARA-1573 Support for spec data-source element without user element

arjantijms:


arjantijms:
Jenkins, would you kindly?

PR #1544, opened 2017-04-20
PAYARA-1006 ensure there is no Null Pointer Exception if a supplied c…

smillidge:
ensure there is no Null Pointer Exception if a supplied HZ configuration file does not hace a Global Serializer Config


lprimak:
Jenkins test please

PR #1560, opened 2017-04-27
PAYARA-1614 Fix NPE on list-historic-healthchecks when healthcheck er…

Cousjava:
…ivce isn't enabled and historic healthchecks is


Cousjava:
jenkins test please

PR #1558, opened 2017-04-26
Increment version number to reflect GlassFish change

Pandrex247:
The fixes included in GlassFish 4.1.2 have already been implemented in Payara, so this is just a version change.

PR #1561, opened 2017-04-27
Move API back into Payara Project

Pandrex247:
Moves the Payara API back into the Payara Project at the top level.

### DIFF
--- a/api/payara-api/pom.xml
+++ b/api/payara-api/pom.xml
@@ -43,7 +43,7 @@
     <parent>
         <groupId>fish.payara.api</groupId>
         <artifactId>api-parent</artifactId>
-        <version>4.1.1.172-SNAPSHOT</version>
+        <version>4.1.2.172-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
 
--- a/api/payara-api/pom.xml
+++ b/api/payara-api/pom.xml
@@ -1,0 +1,169 @@
+<!--
+  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ 
+  Copyright (c) [2016-2017] Payara Foundation and/or its affiliates. All rights reserved.
+ 
+  The contents of this file are subject to the terms of either the GNU
+  General Public License Version 2 only ("GPL") or the Common Development
+  and Distribution License("CDDL") (collectively, the "License").  You
+  may not use this file except in compliance with the License.  You can
+  obtain a copy of the License at
+  https://github.com/payara/Payara/blob/master/LICENSE.txt
+  See the License for the specific
+  language governing permissions and limitations under the License.
+ 
+  When distributing the software, include this License Header Notice in each
+  file and include the License file at glassfish/legal/LICENSE.txt.
+ 
+  GPL Classpath Exception:
+  The Payara Foundation designates this particular file as subject to the "Classpath"
+  exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+  file that accompanied this code.
+ 
+  Modifications:
+  If applicable, add the following below the License Header, with the fields
+  enclosed by brackets [] replaced by your own identifying information:
+  "Portions Copyright [year] [name of copyright owner]"
+ 
+  Contributor(s):
+  If you wish your version of this file to be governed by only the CDDL or
+  only the GPL Version 2, indicate your decision by adding "[Contributor]
+  elects to include this software in this distribution under the [CDDL or GPL
+  Version 2] license."  If you don't indicate a single choice of license, a
+  recipient has the option to distribute your version of this file under
+  either the CDDL, the GPL Version 2 or to extend the choice of license to
+  its licensees as provided above.  However, if you add GPL Version 2 code
+  and therefore, elected the GPL Version 2 license, then the option applies
+  only if the new code is made subject to such option by the copyright
+  holder.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>fish.payara.api</groupId>
+        <artifactId>api-parent</artifactId>
+        <version>4.1.1.172-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <groupId>fish.payara.api</groupId>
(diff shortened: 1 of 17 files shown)


Pandrex247:
This will probably get into conflict with https://github.com/payara/Payara/pull/1558

smillidge:
jenkins test please

Pandrex247:
Jenkins, would you kindly?

Pandrex247:
Jenkins test please

OndrejM:
@lprimak - with what you suggest, it wouldn't be possible to access any new additions in the API added in a quarterly release. I think it's pretty normal that the API tracks server versions. But we need to maintain backwards compatibility, so that anything compiled with e.g. API version 4.1.1.171 would run on Payara Server 172 and later.

lprimak:
@OndrejM Think about it this way:
Let's say Java EE APIs tracked every version of every app server version ever released.  This would be impossible right?
I think API should remain as stable as possible, definitely not quarterly release with incompatible changes.

OndrejM:
@lprimak That's easy to do with Java EE, because it's never updated until the next major version. But e.g. WildFly provides a BOM for every version, and the BOM specifies exactly the versions provided by the corresponding server version. And that's exactly what's happening with Payara Server too - we're adding new features in every version and some of them are exposed in the API (e.g. some notifiers added new interfaces in 171, CDI event bus was updated for 171 and may be updated again in future versions)

OndrejM:
What I wanted to say that Payara API changes with every release, while Java EE changes only with a major EE release. It's not possible to fix the version of Payara API and add new API for developers unless we provide each API addition as a separate dependency.

OndrejM:
But we should keep the API backwards compatible, so when you compile against API v4.1.1.171, you can easily run it on Payara Server v4.1.1.172 and future versions. I don't see how this is confusing.

lprimak:
As someone who's dealt with APIs all my life, this breaks number of conventions which makes it very confusing:

- Semanic versioning for APIs is broken here, i.e APIs don't change with every minor release, only major releases
- Users are forced to track quarterly changes for APIs

In reality, API changes are very rare, and usually are prompted by the developer rather than operation need, such as upgrading the server itself.
API changes are rather rare, compare with server releases,
thus realistically it makes no sense to use the same version for the server as for APIs

OndrejM:
> @lprimak In reality, API changes are very rare

This is just no the case in Payara Server - API changes happen almost with every 3-monthly release. How could we ship these changes to the developers if the API version doesn't change?

We could have a "stable" API artifact with a simple version like API-4.1, but what would it include? Only the features available in the first 4.1 version Payara Server? Those would be only JCache integration and possibly CDI bus, but even those were probably added later than in the first version.

We could release a new version of the API only if it changes, having versions like 4.0 for the initial API (JCache, CDI bus) matching versions 144+, 4.1 for later additions (RequestTracing interceptor introduced in 163) matching versions 163+, 4.2 for other additions in v171, etc.

But wouldn't this be even more confusing, as people would need to check which Payara Server version is compatible with which API artifact? In real world projects, applications are being developed against a particular version of a server, not an API. With each server upgrade, there's a migration step for the application. There's no point in having too much abstraction, because people general know what they are doing with the API artifact. If they don't need any specific APi feature from Payara Server, they can use the stable Java EE API, which would change only with a major Payara Server version.

OndrejM:
Even if we start versioning the API with the 171 release, having API-4.0 matching Payara Server 171, I still expect that the API will be updated regularly with smaller or bigger updates, so there would be API-4.1, 4.2 ... in the future.

Payara v5 might change things a bit, but even once we release Payara v5 and API-5.0, I expect that the API might be changed pretty frequently - at least every second or third public release. And then I repeat my question, would it be less confusing to map API versions to server versions, or to keep the versions in synch and make the mapping evident? We would ensure that apps built against older API versions would run with new versions of Payara Server and devs don't have to update the API version if they don't  need the new features.

lprimak:
This will be different from every other API versioning scheme out there, and that alone makes it a non-starter for me.  I don't care if it's by date, 1,2,3,4,5 any combination of the two, or something else, it just *can't* be tied to the server version.  It's too much of a breaking tradition with every other naming convention that's out there

OndrejM:
Lenny, I think we have a different view on the API. 

I see the API as a gateway to the features exposed by Payara Server, and every new version can add new features. Without an API artifact there's no way to access these features, because Payara Server isn't a library that you can simply add to your classpath. For me the API is very close to a library that you drop into your app, except it's provided by the server.

I think what you expect from the API is something stable that doesn't change a lot. It's natural with Java EE, because that's solid API, and usually the implementations add non-standard features on top of it. E.g. hibernate provides additional API, which is AFAIK versioned with the impl, because with hibernate IMPL = API. With most libraries IMPL = API in fact. But we can't just expose Payara features as a library that devs can build against by importing it via a BOM into a maven project. Payara features are not compiled as a standalone library and the only way to build against these features is to build against the API.

I'd like to see an example of the tradition you mentioned, because so far the only example is Java EE, On the other hand, there are many examples whre API is versioned together with the IMPL, besides all libraries that don't provide a separate API (API = IMPL in that case), WildFly also couples API versions with IMPL, e.g. [wildfly-security-api](http://search.maven.org/#search%7Cga%7C1%7Cwildfly-security-api), [OpenEJB-API](http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22openejb-api%22) is also couple with impl, even [SLF4J](http://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.slf4j%22) keeps the same version for the API and connectors to the underlying impl - in this case the connectors are in fact the impl of the common API.

I think it's very hard to find an example where API and IMPL are maitained by the same project and have different versions. And I believe that if there are such cases, the main focus is on the API, while with Payara Server's API artifact, in my view, the main focus is on the server and exposing the new functionality to the developers.

smillidge:
The glassfish public api jar just tracks the GlassFish version. I see this as similar to the glassfish api jar.
https://github.com/payara/Payara/blob/master/nucleus/common/glassfish-api/pom.xml 
https://mvnrepository.com/artifact/org.glassfish.main.common/glassfish-api

lprimak:
I'm not sure this applies. GF APIs are not publicized as I know of and nobody uses it that I know of. 

I think our API should project stability and ease of use

OndrejM:
Glassfish-API wasn't known even to me until now :) Maybe it's not a question of versioning but of how we can make the API more accessible to the developers (blogs, guides, tutorials, examples, etc.)

I would continue with versions of glassfish-api and payara-api in line with server versions, but maybe we can improve versioning and distribution in the future. I just now don't see a better versioning scheme, which would be more stable yet still understandable and easy to use for developers.

smillidge:
Both these apis are really just apis to internal server functionality so I think they should be aligned with the server version. More generic for features layered on top of the server I would envisage having a different version lifecycle. 

GlassFish api defines thins like start up classes, SQL Trace Listeners etc.

Issue #1401, opened 2017-02-20
HTTP access logging can break asynchronous servlets (and maybe asynchronous JAX-RS services too)


jo-ka:
# Description #

----------

There's a recent open bug report for GlassFish [here](https://java.net/jira/browse/GLASSFISH-21651) that points out that the log entries created for asynchronous servlets are wrong. However, I have strong reason to believe that under certain circumstances, this broken logging can even lead to the failure of the execution of asynchronous servlets itself, which manifests itself in seemingly random, race-condition-like exceptions.

## Expected Outcome

With regards to the log entries: correct HTTP status code and message size. With regards to the execution of asynchronous servlets: deterministically successful execution.

## Current Outcome

(Note: I'm quoting exception stack traces from GF 4.1 here, but I could reproduce the bug on Payara 4.1.1.164 in principle, so apart from potentially wrong line numbers, they should still apply)

As noted in the linked java.net thread, access log records for asynchronous servlets (usually) have their status code set to 200 and their message size to 0, e.g.

`"10.0.30.83" "08/Feb/2017:08:58:42 +0100" "GET [...] HTTP/1.1" 200 0`

The much bigger problem are exceptions like shown below though, which were observed with a frequency of roughly 1 in 10000 requests.

    [2016-12-15T14:32:56.550+0100] [glassfish 4.1] [SEVERE] [] [javax.enterprise.web.core] [jk-connector(11)] [timeMillis: 1481808776550] [] [[
    An exception or error occurred in the container during the request processing
    java.lang.IllegalStateException: Current state = RESET, new state = FLUSHED
        at java.nio.charset.CharsetDecoder.throwIllegalStateException(CharsetDecoder.java:968)
        at java.nio.charset.CharsetDecoder.flush(CharsetDecoder.java:657)
        at java.nio.charset.CharsetDecoder.decode(CharsetDecoder.java:786)
        at java.nio.charset.Charset.decode(Charset.java:806)
        at org.glassfish.grizzly.memory.HeapBuffer.toStringContent(HeapBuffer.java:859)
        at org.glassfish.grizzly.http.util.BufferChunk.toString(BufferChunk.java:497)
        at org.glassfish.grizzly.http.util.DataChunk.toString(DataChunk.java:291)
        at org.glassfish.grizzly.http.util.DataChunk.toString(DataChunk.java:283)
        at org.glassfish.grizzly.http.HttpRequestPacket.getQueryString(HttpRequestPacket.java:251)
[...]

PR #1571, opened 2017-05-02
PAYARA-1585 Async servlet access is not correctly logged

jGauravGupta:
Issue : Response information is not correctly logged in the HTTP Access log when the web container deals with Asynchronous servlet requests.
Fix : Logging invocation is now done at the completion notification of Asynchronous Context using AsyncListener.
Issue  Ref : https://github.com/payara/Payara/issues/1401


jGauravGupta:
Jenkins test please

PR #1575, opened 2017-05-03
PAYARA-1625 fixed data-source from application.xml in JPA

arjantijms:
DataSourceDefinitionDeployer now also takes data sources that are defined in application.xml into account when doing the special early deployment for JPA.


lprimak:
jenkins test please

Issue #1532, opened 2017-04-16
@Priority Interceptor from external JAR not picked up


Pscheidl:
# Description #
----------

here is a simple WAR file with one external dependency. The external dependency is a JAR with one CDI extension (working correctly) and one [interceptor annotated with @Priority annotation](https://github.com/Pscheidl/FortEE/blob/master/src/main/java/cz/pscheidl/fortee/FailsafeInterceptor.java). This interceptor is not activated in Glassfish 4.1.2 (and also in derived Payara 171). 

An attempt was made to enable the interceptor manually via beans.xml.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns="http://xmlns.jcp.org/xml/ns/javaee"
       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
       bean-discovery-mode="all">

    <interceptors>
        <class>cz.pscheidl.fortee.FailsafeInterceptor</class>
    </interceptors>

</beans>
```

This leads to:
```text
java.io.IOException: com.sun.enterprise.admin.remote.RemoteFailureException: Error occurred during deployment: Exception while loading the app : CDI deployment failure:WELD-001417: Enabled interceptor class <class>cz.pscheidl.fortee.FailsafeInterceptor</class> in file:/home/pavel/Software/application_servers/glassfish4/glassfish/domains/domain1/applications/payara-interceptor-bug-1.0/WEB-INF/beans.xml@8 does not match an interceptor bean: the class is not found, or not annotated with @Interceptor and still not registered through a portable extension, or not annotated with @Dependent inside an implicit bean archive. Please see server.log for more details.
```

## Expected Outcome

When a JAR library with an interceptor annotated with `@Priority` annotation is included in a web archive, it should be enabled by default.

## Current Outcome

Interceptor is not enabled. Manual interceptor registration and enbaling does result in an exception.

### Reproducing from scratch
1. Create a JAR with a Java EE interceptor class.
2. Annotate the interceptor with @Priority annotation with arbitrary priority.
3. Include the resulting JAR in a WAR.
4. Use the interceptor in the WAR application.
5. Deploy the WAR into Payara.
6. Observe the interceptor not being active.

### Reproducing with sample application
1. Start Payara 171/164
[...]

PR #1576, opened 2017-05-04
PAYARA-1584 @Priority Interceptor from external JAR not picked up

jGauravGupta:
Issue Ref :  https://github.com/payara/Payara/issues/1532


jGauravGupta:
Jenkins test please

Quix0r:
Has this, by chance being integrated into Payara Server  4.1.2.173 ? Because I'm having same issue here and I cannot find the true reason of it.

PR #1585, opened 2017-05-10
Payara 1322 application deployment time

MattGill98:
Double data type gets funny in hk2, so I've put a double validator instead

### DIFF
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/config/ApplicationDeploymentTime.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/config/ApplicationDeploymentTime.java
@@ -1,0 +1,70 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2009-2010 Oracle and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://glassfish.dev.java.net/public/CDDL+GPL_1_1.html
+ * or packager/legal/LICENSE.txt.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at packager/legal/LICENSE.txt.
+ *
+ * GPL Classpath Exception:
+ * Oracle designates this particular file as subject to the "Classpath"
+ * exception as provided by Oracle in the GPL Version 2 section of the License
+ * file that accompanied this code.
+ *
+ * Modifications:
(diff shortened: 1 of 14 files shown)


MattGill98:
jenkins test please

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/config/ApplicationDeploymentTime.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+@Configured
+public interface ApplicationDeploymentTime extends ConfigBeanProxy, Payload {
+
+    /**
+     * Time taken to deploy the application.
+     *
+     * @return deployment time
+     */
+    @Attribute(defaultValue = "0.0")
```
mulderbaba: why do we have a default value as ```0.0``` on time:string ?

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/config/ApplicationDeploymentTime.java
```diff
@@ -0,0 +1,68 @@
+/*
```
mulderbaba: license plate is wrong, it should be the one from Payara

MattGill98: Ah sorry forgot those, I'll update them now.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/config/ApplicationDeploymentTime.java
MattGill98: It was just to give it a sensible default value, should I remove it?

mulderbaba: I haven't seen its usage anywhere, so if that's the case we can remove it

nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeploymentCommandUtils.java
```diff
@@ -89,6 +89,31 @@
...
+     * get the time that has passed since the timer started.
+     */
+    public static void startTimer() {
+        startTime = System.nanoTime();
+    }
+    
+    /**
+     * Gets the time (in seconds) that has passed since the timer was started. 
+     * If the timer hasn't been started yet you may get strange results.
+     * The time returned is to millisecond accuracy.
```
MattGill98: Because I thought that System.currentTimeMillis() is more a measure of wall time than time passed. It's not always guaranteed to be positive. Having said that it might not be necessary for low accuracy timings like this. What do you think?

michaelranaldo:
Jenkins test please

MattGill98:
Jenkins test please

mulderbaba:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

mulderbaba:
jenkins test please

PR #1592, opened 2017-05-11
PAYARA-1638 Rest Endpoint Null Pointer Exception Fix

MattGill98:
Changed list rest endpoints command to still return success for empty component so that link appears in admin console, and fixed null pointer exception bug when deploying an application with no web components.


MattGill98:
jenkins test please

PR #1621, opened 2017-05-24
PAYARA-1693 Persistent Timer Info Bug when Using a custom class

lprimak:
using Class.forName() for arrays, as ClassLoader.loadClass() does not…work for arrays


lprimak:
jenkins test please

PR #1623, opened 2017-05-25
PAYARA-1486 Changed room name column to room id for xmpp notifier

MattGill98:
To complement https://github.com/payara/Payara-Server-Documentation/pull/202


MattGill98:
jenkins test please

PR #1626, opened 2017-05-26
PAYARA-1694-Healthcheck-still-prints-out-error-in-log

mulderbaba:


mulderbaba:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/HistoricHealthCheckEventRetriever.java
```diff
@@ -110,11 +105,6 @@ public void execute(AdminCommandContext context) {
             actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
             return;
         }      
-        if (!service.isHistoricalTraceEnabled()) {
-            actionReport.setMessage("Health Check Historical Trace is not enabled!");
-            actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
```
MattGill98: Correct me if I'm being dumb, but would it have worked to just change this line to a success code? Only because I like it returning a 'disabled' message when it's disabled. It's more of a question than a review really, so feel free to disregard it!

mulderbaba: @MattGill98 this is being consumed from admin console and if we return failure, it throws out an error while opening the historic configuration page.

PR #1627, opened 2017-05-26
PAYARA-1691-useSeparateLogFile-is-not-optional-in-notification-log-configure-command

mulderbaba:


mulderbaba:
jenkins test please

PR #1635, opened 2017-05-29
PAYARA-1715 Implemented processing of transactional attribute

arjantijms:
This attribute was already being collected in the original code with
even a check if it was set of not, but then eventually the last part was
not implemented.


arjantijms:
Jenkins, would you kindly?

PR #1640, opened 2017-05-31
PAYARA-1696 stop previous thread when notifier is configured

Cousjava:
Only a single thread will now exist for each notifer.


Cousjava:
jenkins test please

PR #1642, opened 2017-06-01
PAYARA-1681 Add default JVM options check

Pandrex247:
If the JVM option _org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER_ is not present in the server/instance config, we will override the Grizzly default and use the _HeapMemoryManager_.
To override this behaviour (so to use a different memory manager), you must specify this JVM option in the server/instance config with your desired setting.

This is to help with people upgrading from an older version of Payara where this JVM option has not been added to the default domain.xml.

When this override happens, a message is logged at a level of FINE.

When this override happens, the JVM option is **NOT** added to the domain.xml.

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherLogger.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherLogger.java
@@ -37,6 +37,9 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+
+// Portions Copyright [2017] [Payara Foundation and/or its affiliates.]
+
 package com.sun.enterprise.admin.launcher;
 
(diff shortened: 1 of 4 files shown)


Cousjava:
jenkins test please

nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncherLogger.java
```diff
@@ -185,4 +188,13 @@ private GFLauncherLogger() {
     action = "NA",
     level = "INFO")
     public static final String COMMAND_LINE = "NCLS-GFLAUNCHER-00005";
+    
+    @LogMessageInfo(
+            message =
+    "Using Payara default value \"{1}\" for JVM option: {0}",
+    comment = "JVM Info",
+    cause = "JVM Option not present in domain.xml",
```
michaelranaldo: option vs Option

PR #1645, opened 2017-06-02
PAYARA-1133 Slow SQL logger should keep the top and slowest SQL traces

Pandrex247:
Adds the capability to monitor the slowest SQL traces executed against a connection pool.

To enable this, turn on monitoring of the JDBC Connection pools in a config, for example:
`asadmin set server.monitoring-service.module-monitoring-levels.jdbc-connection-pool=HIGH`

It is configured in the same way as the frequent SQL tracing, using the _number-of-top-queries-to-report_ and _time-to-keep-queries-in-minutes_ properties:

```
asadmin set server.resources.jdbc-connection-pool.<POOL_NAME>.property.number-of-top-queries-to-report=15                    
asadmin set server.resources.jdbc-connection-pool.<POOL_NAME>.property.time-to-keep-queries-in-minutes=10  
```

The monitoring figures can be viewed via the admin console (under Resources > _Resource_ > JDBC Connection Pool Statistics), or using `asadmin get -m server.resources.<POOL_NAME>.slowSqlQueries-current`

For ease of use, I've also created a default SQL Trace listener that simply does nothing, as the monitoring of the slowest SQL traces requires an SQL Trace listener. This can be replaced with any other valid SQL Trace listener without affecting the monitoring (as far as I've tested anyway). The new default SQL Trace listener **is now the default value of the _sqlTraceListeners_ option for the _create-jdbc-connection-pool_ command**.

This PR also fixes monitoring of the most Frequent SQL Traces, though this still retains all of its previous shortcomings.

### DIFF
--- a/appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/util/FrequentSQLTraceCache.java
+++ b/appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/util/FrequentSQLTraceCache.java
@@ -1,0 +1,109 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2010 Oracle and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://glassfish.dev.java.net/public/CDDL+GPL_1_1.html
+ * or packager/legal/LICENSE.txt.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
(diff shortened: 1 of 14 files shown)


Pandrex247:
Jenkins, would you kindly?

Cousjava:
jenkins test please

mulderbaba:
jenkins test please

smillidge:
jenkins test please

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/util/FrequentSQLTraceCache.java
```diff
@@ -0,0 +1,106 @@
...
+     * ordered by the number most frequently used, followed by the usage
+     * timestamp. Only the top 'n' queries represented by the numTopQueriesToReport are
+     * chosen for display.
+     *
+     * @return string representation of the list of sql queries sorted
+     */
+    public String getTopQueries() {
+        purgeEntries();
+        
+        List<SQLTrace> sqlTraceList = new ArrayList(cache.values());
```
lprimak: Should be ArrayList<> instead of raw type

lprimak:
Jenkins, would you kindly?

lprimak:
Jenkins, would you kindly?

arjantijms:
Jenkins, would you kindly?

PR #1651, opened 2017-06-05
PAYARA-1694 Healthcheck still prints out error in log

Pandrex247:
Don't mark command as failed when not enabled


Pandrex247:
Jenkins, would you kindly?

Cousjava:
should this replace #1626 ?

Pandrex247:
Nah it's in addition to it.
Mert's PR stops the command from returning as failed later on in its execution when it's retrieving the stuff, mine stops it from failing at the command's entry point.

Pandrex247:
jenkins test please

PR #1660, opened 2017-06-06
PAYARA-1733-Package-Payara-Public-API-dependency-into-Embedded-editions

mulderbaba:


mulderbaba:
jenkins test please

appserver/extras/embedded/web/pom.xml
```diff
@@ -289,6 +290,14 @@
            <type>zip</type>
            <optional>true</optional>
        </dependency>
+
+        <dependency>
+          <groupId>fish.payara.packager</groupId>
+          <artifactId>payara-api-package</artifactId>
+          <version>${project.version}</version>
+          <type>zip</type>
+          <optional>true</optional>
```
Cousjava: Why is this optional? I would have thought this is required.

mulderbaba: @Cousjava it excludes the transitive dependencies of payara-api-package (note that cascading applies in transitive dependencies), so it only fetches one zip file and extract it.

PR #1668, opened 2017-06-09
PAYARA-1738 Some minor refactoring

arjantijms:


arjantijms:
Jenkins, would you kindly?

arjantijms:
This was intended as preparation to accept reflective calls via a Method instance obtained from a MDB proxy (as JBoss supports), but now I think we should just adjust the client code to obtain the Method instance from the Bean Class. Not 100% sure what the spec requires, will try to ask this on the JMS and EJB lists, but those EGs are not active currently.

Did some refactoring before that, and didn't want to waste them.

mulderbaba:
jenkins test please

smillidge:
Should this be merged?

PR #1690, opened 2017-06-19
PAYARA-1561 Payara Micro fails fast if there is an error in boot

Cousjava:


Cousjava:
jenkins test please

Cousjava:
jenkins test please

PR #1695, opened 2017-06-20
PAYARA-1770 Throw an unprocessed change when editing a HTTP listener with monitoring enabled

Pandrex247:
Also edits the similar message for Thread Pools to be clearer, and makes it so that the HTTP listener is only recreated if something has actually changed.


Pandrex247:
Jenkins, would you kindly?

PR #1699, opened 2017-06-21
PAYARA-1560 Third party library classloading on micro

Cousjava:


Cousjava:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1309,10 +1331,33 @@ private void setSystemProperties(File propertiesFile) throws IllegalArgumentExce
...
 
     private void deployAll() throws GlassFishException {
-      
+
         // Deploy from within the jar first.
         int deploymentCount = 0;
-        Deployer deployer = gf.getDeployer();                
+        Deployer deployer = gf.getDeployer();
+
+        //add libraries to deployment
```
smillidge: I think this is too late for adding the jars to the classloader. I would add the jars before the server is booted. At the beginning of the bootstrap method. That way if jars are required for custom Hazelcast configurations or settings in domain.xml they will be available during boot.

Cousjava:
jenkins test please

Cousjava:
jenkins test please

Cousjava:
jenkins test please

PR #1703, opened 2017-06-22
Payara 5 173 issue 1702

khasunuma:
for the naming strategy, the version number should be '5.0.173-SNAPSHOT' instead of '5.1.2.173-SNAPSHOT'.


khasunuma:
Sorry. the destination is missed.

Issue #1702, opened 2017-06-22
Payara 5's version number should adapt to the naming strategy


khasunuma:
# Description #
----------
**This issue is only related to 'Payara-5-173' branch**

Now on 'Payara-5-173' branch, the version number is set '5.1.2.173-SNAPSHOT'. But the current GlassFish 5's version number is '5.0-SNAPSHOT' (see https://github.com/javaee/glassfish/blob/master/pom.xml, _et al._), thus the number should be '5.0.173-SNAPSHOT' to adapt the naming strategy described on https://payara.gitbooks.io/payara-server/content/general-info/general-info.html .

Please see also the pull request I'll sent soon after.

mikecroft:
Thanks for sending #1704 - I agree with your points about the numbering. It certainly shouldn't begin `5.1.2....`.

khasunuma:
Even if 'Payara-5-173' is working branch, I can't understand set it the version number not 5.0.0.173-SNAPSHOT but 5.1.2.173-SNAPSHOT against the naming strategy of Payara. Because it's unprecedented. For example, 'Payara-5' branch, previous working branch for Payara 5 used '5.0.0.163-SNAPSHOT'.

When I wrote this issue first, I thought which version number should be chosen in working branch. Then I judged 'Payara-5-173' because I seemed it aims at Payara 5.

If this problem is not solve, perhaps we have to a new versioning rule for working branches such as 'Payara-5-173'.

mikecroft:
Hi @khasunuma - the version will change, but it's not a priority yet. We're still a little way off providing snapshot builds and the final version will not be released for a few months. The main priorities for Payara 5 are bringing in support for the latest APIs and there is still more work to do on that.

PR #1704, opened 2017-06-22
Modify Payara version number to adapt to the naming strategy

khasunuma:
Request again, because I had mistake of pull request #1703 

Please ignore #1703


smillidge:
@arjantijms can you review please

Pandrex247:
Jumping in quickly before this gets merged.
The version should be 5.0.0.173. Note the extra 0.

If we're sticking with our normal versioning of GlassFishVersion.PayaraVersion, we need it there in case Oracle do a GlassFish 5.x.x.
If it's not there, we break "latest" versions, as we could go from 5.0.173, to 5.0.1.174 - the former will always show as the "latest" because it technically has a higher version number

arjantijms:
Another thing worth mentioning, the current number in the Payara-5-173 branch is just a working number. I made a snap decision there, since it's based on Payara 4.1.2.173 and not on GlassFish 5.0. Basically it's our own independent upgrade of Payara 4 to incorporate all new components.

The idea was basically to set the version number right when it gets merged to master, and Payara 4.x gets branched to a rolling new branch, but that was just an idea, no decision has been made there

smillidge:
I agree with @Pandrex247 this should be 5.0.0.173

khasunuma:
I see. I agree with @Pandrex247 comments.

smillidge:
Subject to raging debate internally on version numbers. This is now a good working 5 version number. @arjantijms Can you review?

khasunuma:
@arjantijms , I see your opinion. But I still have following problems;
* I don't think a good version number without this request based on issue #1702 .
* I can't explain why the version number on Payara-5 branch was '5.0.0.163-SNAPSHOT'.

Pandrex247:
This was unintentionally closed, new PR at #1777

PR #1718, opened 2017-06-25
PAYARA-1792 Environment variable replacement does not work in glassfish-resources.xml

smillidge:
ensure environment variable, password alias and system property replacement is possible when using glassfish-resources.xml


smillidge:
redone to resolve on deployment

lprimak:
Maybe I don't understand something... why is this necessary at all?
All configurations should retain variables, and then get substituted during runtime, not even deployment.
Am I missing something?

lprimak:
jenkins test please

PR #1735, opened 2017-06-30
PAYARA-1767

Pandrex247:
Makes it so you can get a local node to use the node host for its host name, rather than the localhost value.

### DIFF
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
@@ -109,8 +109,10 @@ public class IiopFolbGmsClient implements CallBack {
 
     private GroupInfoService gis;
     
-    private static final String USE_NODE_HOST_FOR_LOCAL_NODE_PROPERTY = 
-            "fish.payara.iiop.gmsClient.useNodeHostForLocalNode";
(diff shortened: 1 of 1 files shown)


Pandrex247:
Extra work has been requested - closing whilst we wait for confirmation that the additional changes work.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
```diff
@@ -352,12 +357,17 @@ private ClusterInstanceInfo getClusterInstanceInfo( Server server,
...
+        fineLog( "getClusterInstanceInfo: iservice {0}", iservice ) ;
+        
         final String nodeName = server.getNodeRef() ;
         String hostName = nodeName ;
         if (nodes != null) {
             Node node = nodes.getNode( nodeName ) ;
             if (node != null) {
-                if (node.isLocal()) {
+                if (node.isLocal() && (!Boolean.getBoolean(USE_NODE_HOST_FOR_LOCAL_NODE_SYSTEM_PROPERTY) 
+                        || !Boolean.parseBoolean(iservice.getOrb().getPropertyValue(
```
OndrejM: A condition of type "NOT condition1 OR NOT condition2" always smells - it should be "NOT condition1 AND NOT condition2", or changed to "NOT (condition1 OR condition2)"

PR #1701, opened 2017-06-22
PAYARA-1740 Deployment and initialization order for Payara 5

lprimak:

PR #1748, opened 2017-07-05
PAYARA-1740 Deployment and initialization order for Payara 5

lprimak:
This is created for the new Payara-5 branch
The old PR is #1701


arjantijms:
Will merge it in now and do some further testing later.

PR #1750, opened 2017-07-06
PAYARA-1723 JMX Connector bind on all network interfaces

jGauravGupta:
JMX Connector allowing connections from remote hosts on all network interfaces by specifying  "0.0.0.0" or "*" .

### DIFF
--- a/nucleus/common/mbeanserver/src/main/java/org/glassfish/admin/mbeanserver/RMIConnectorStarter.java
+++ b/nucleus/common/mbeanserver/src/main/java/org/glassfish/admin/mbeanserver/RMIConnectorStarter.java
@@ -37,6 +37,8 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016-2017] [Payara Foundation and/or its affiliates]
+
 package org.glassfish.admin.mbeanserver;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


jGauravGupta:
jenkins test please

nucleus/common/mbeanserver/src/main/java/org/glassfish/admin/mbeanserver/RMIConnectorStarter.java
```diff
@@ -137,7 +139,7 @@ public RMIConnectorStarter(
         }
 
         final boolean ENABLED = true;
-        mBindToSingleIP = ENABLED && !(address.equals("0.0.0.0") || address.equals(""));
+        mBindToSingleIP = ENABLED && !(address.equals("0.0.0.0") || address.equals("*")|| address.equals(""));
```
MattGill98: This line seems like it should work, but for whatever reason when it should be binding multiple IPs it still doesn't. Possibly due to the logic on line 160.

jGauravGupta: Is host does not listens on 0.0.0.0 ? I tested with CurrPort tool.  Is host does not listens on 0.0.0.0 ? I tested with CurrPort tool.

MattGill98: The port is correctly opened when scanning over nmap and port scanning tools, but actually connecting over JMX will fail

smillidge:
Does this PR fix the bug or not? If not can you close the PR and repopen when working? If it is Matt can you either dismiss your review or approve?

jGauravGupta:
This PR works but looks like @MattGill98  getting some environment specific issue.

PR #1751, opened 2017-07-06
Payara 1739 Pre and post boot command files a little too thorough

MattGill98:
Previous fix also removed the end of commands with hashtags in them as comments. 
E.g. 
`set configs.config.server-config.http-service.virtual-server.test.default-web-module=ear-test#ear-test-web-1.0-SNAPSHOT.war`
Now only removes comments if they have whitespace before them.


MattGill98:
jenkins test please

MattGill98:
https://regex101.com/r/nCGjJp/1
Those are the new test cases for comments

PR #1773, opened 2017-07-18
PAYARA-1842 race condition in admin gui when modifying logging attributes

lprimak:
trigger server restart here because the GUI cannot wait for the log file to be re-read and
thus triggering restart message too late, as it won't show up in the GUI when logging attribtues are changed


lprimak:
jenkins test please

lprimak:
jenkins test please

PR #1790, opened 2017-07-24
PAYARA-1840 RAR detector now checks for all JCA annotations

Cousjava:
All archives are now checked for annotations, regardless of type

### DIFF
--- a/appserver/connectors/connectors-connector/src/main/java/com/sun/enterprise/connectors/connector/module/RarDetector.java
+++ b/appserver/connectors/connectors-connector/src/main/java/com/sun/enterprise/connectors/connector/module/RarDetector.java
@@ -36,17 +36,17 @@
  * and therefore, elected the GPL Version 2 license, then the option applies
  * only if the new code is made subject to such option by the copyright
  * holder.
+ *
+ * Portions Copyright [2017] Payara Foundation and/or affiliates
  */
 
 
 package com.sun.enterprise.connectors.connector.module;
 
-import com.sun.enterprise.deploy.shared.FileArchive;
 import org.glassfish.api.deployment.archive.ArchiveDetector;
 import org.glassfish.api.deployment.archive.ArchiveHandler;
 import org.glassfish.api.deployment.archive.ArchiveType;
 import org.glassfish.api.deployment.archive.ReadableArchive;
-import org.glassfish.deployment.common.DeploymentUtils;
 import com.sun.enterprise.deployment.deploy.shared.Util;
 import org.glassfish.deployment.common.GenericAnnotationDetector;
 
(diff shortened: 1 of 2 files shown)


Cousjava:
jenkins test please

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/GenericAnnotationDetector.java
```diff
@@ -151,8 +152,7 @@ public void scanArchive(ReadableArchive archive) {
                     } finally {
                         is.close();
                     }
-                } else if (entryName.endsWith(".jar") && 
-                    entryName.indexOf('/') == -1) {
+                } else if (!entryName.contains("/")) {
```
smillidge: why has this changed? It looks like it may produce side effects.

appserver/connectors/connectors-connector/src/main/java/com/sun/enterprise/connectors/connector/module/RarDetector.java
```diff
@@ -69,8 +69,24 @@
...
 public class RarDetector implements ArchiveDetector {
+    
+    /**
+     * Connector annotations for JCA 1.7
+     */
     private static final Class[] connectorAnnotations = new Class[]{
-            javax.resource.spi.Connector.class};
+            javax.resource.spi.Connector.class,
+            javax.resource.spi.ConfigProperty.class,
+            javax.resource.spi.ConnectionDefinition.class,
```
smillidge: I don't think this is correct as other applications can have these annotations and aren't rar files.

You need to see which ones are rar only and which ones can be used by clients of a rar

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/GenericAnnotationDetector.java
Cousjava: because otherwise it will not scan non-jars for annotations. The code below it is in a try statement anyway so any problems will be caught.

Cousjava:
jenkins test please

appserver/connectors/connectors-connector/src/main/java/com/sun/enterprise/connectors/connector/module/RarDetector.java
```diff
@@ -69,8 +69,16 @@
...
 @Singleton
 public class RarDetector implements ArchiveDetector {
+    
+    /**
+     * Connector annotations for JCA 1.7
+     */
     private static final Class[] connectorAnnotations = new Class[]{
-            javax.resource.spi.Connector.class};
+            javax.resource.spi.Connector.class,
+            javax.resource.spi.ConfigProperty.class,
```
MattGill98: `@ConfigProperty` is used in WARs which use Resource Adapters, I'd swap it for `@Activation`

Cousjava:
jenkins test please

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/GenericAnnotationDetector.java
maxencelaurent: Side effect when calling jarSubArchive.entries().

com.sun.enterprise.deployment.deploy.shared.InputJarArchive#entries(boolean) does not throw an IOException but a RuntimeException. The effect is the exception is not catch by the inner try-catch bloc but the outer one, skipping all other entries in the archive.

PR #1794, opened 2017-07-25
Use logging instead of System.out.println PAYARA-1878

jcosmo:
Whilst the scenario causing to this println is not supposed to happen, it's currently happening a very large number of times a second on one of my servers, and being unable to turn off the message to get at more pertinent logging is making it hard to fix.


lprimak:
jenkins test please

jcosmo:
CLA submitted

smillidge:
Can you also give us a stack trace to see how you get into this code?

smillidge:
Thanks for CLA this can now be merged.

jcosmo:
Hi,

I don't have a stack trace, unfortunately it just keep printing this message to the console and not logging how it was getting into this part of the code.  With this patch at least we will get a stack trace in the future.

The "fix" was eventually to restart payara, so it must have been in some way related a (possibly failed) redeployment.

smillidge:
OK thanks

PR #1795, opened 2017-07-25
PAYARA-1874:Changed value to send to remove email from '' to null

DanielContrerasAladro:


DanielContrerasAladro:
jenkins test please

PR #1821, opened 2017-08-01
PAYARA-1898 Disabled findbugs plugin

Cousjava:
The plugin will no longer run as part of the site lifecycle by default, it already has its own profile.


Cousjava:
jenkins test please

Issue #1586, opened 2017-05-10
"java.lang.IllegalStateException: Use getGMSAdapterByName method when there are multiple clusters" when an EJB is deployed and GMS is enabled


angtes:
# Description #
I have a domain with several clusters and GMS enabled.
Whenever I try to deploy an EJB I get the following error:
java.lang.IllegalStateException: Use getGMSAdapterByName method when there are multiple clusters

If GMS is disabled all runs OK.

Is this a bug? Can I use EJB and GMS simultaneously? Do I need to configure something? 

## Environment ##
- **Payara Version**: 4.1.1.171.1
- **Edition**: Full
- **JDK Version**: 8 u121 - Oracle
- **Operating System**: Linux (CentOS 6.8)

MattGill98:
Hi Angtes,
Unfortunately I'm having a bit of trouble reproducing your bug. I tried by setting up a clean Payara 171.1 and deploying the [SimpleRemoteEJB Payara Example](https://github.com/payara/Payara-Examples/tree/master/Java-EE/SimpleRemoteEJB). GMS was enabled for both clusters, and each cluster had two instances running in it. Is there maybe anything else specific about your configuration that might be causing this?

angtes:
Thanks MattGill98 for your answer.
I've deployed the SimpleRemoteEjb on my system without any problem.
The differences I've seen regarding my ejb are that mine is stateful and is EJBv2 instead EJBv3
I'm going to do some tests and I'll comment the conclusions.
Thank you very much

angtes:
I've tried to deploy a simple Stateful EJBv3 and I've got the same error:
java.lang.IllegalStateException: Use getGMSAdapterByName method when there are multiple clusters

I've compiled the sources from [https://www.go4expert.com/articles/ejb-30-stateful-session-bean-example-t4212/](url).
When I deploy the jar I got the error.

I've attached the jar file with the example.

[ejb_stateful.zip](https://github.com/payara/Payara/files/1001449/ejb_stateful.zip)

MattGill98:
Hi @angtes,
Thank you for attaching the jar file. I've managed to reproduce the bug consistently now. I found the bug only happens when you try to deploy to the DAS as well as the clusters. If you just deploy the application to the clusters then you should find it deploys successfully. Is there any reason that you would need to deploy to the DAS?

Let me know if it works for you.
Kind regards,
Matt

angtes:
I've tried it and it works ok.

[...]

PR #1823, opened 2017-08-01
PAYARA-1679 Deploying to DAS no longer fails when there are multiple …

Cousjava:
…clusters avaliable


Cousjava:
jenkins test please

Cousjava:
fixes #1586

PR #1834, opened 2017-08-03
PAYARA-1322 has a default value for deployment time if none set

Cousjava:
No longer throws an exception if domain upgraded from previous version


Cousjava:
jenkins test please

PR #1861, opened 2017-08-14
PAYARA-1949 moved the jmx-monitoring-plugin package

Cousjava:
The plugin is for the admingui, and should not be built with Micro


Cousjava:
jenkins test please

PR #1863, opened 2017-08-14
PAYARA-1950 fix NPE in CDI 

smillidge:
Fixes a NPE coming from microprofile while deploying a versioned application. Doesn't hit on all versioned applications.


Pandrex247:
Jenkins test please

PR #1885, opened 2017-08-21
PAYARA-1964

Pandrex247:
Changes the groupId and artifactId of the parent poms (and edits their children appropriately) so that we can deploy them to Maven snapshots to fix the "parent pom not found" errors when downloading the embedded/micro snapshots.

Also adds in a profile so that we can test uploading snapshots before pushing them to Sonatype.


Pandrex247:
--- DO NOT MERGE ---
This PR will break the Jenkins nightly builds as the profile name has changed, so this cannot be merged without changing Jenkins at the same time!
--- DO NOT MERGE ---

Pandrex247:
Jenkins test please

Issue #1872, opened 2017-08-17
Payara overwriting Hazelcast serialization XML configuration


scottg88:
# Description #
----------

Payara appears to overwrite Serialization configuration in built-in Hazelcast.

## Expected Outcome

DataSerialization factories specified in hazelcast.xml configuration should flow through to the configuration used by built-in Hazelcast.

## Current Outcome

There are no configured factories in the running configuration of Hazelcast after startup.

## Steps to reproduce (Only for bug reports) 

Configure hazelcast.xml with:

```xml
<serialization>
  <data-serializable-factories>
    <data-serializable-factory factory-id="111">
      com.example.ExampleDataSerializableFactory
    </data-serializable-factory>
  </data-serializable-factories>
</serialization>
```

Bundle `com.example.ExampleDataSerializableFactory` with Payara Micro/place in Payara lib directory to ensure it's available during Hazelcast startup.

Attempt to pull a custom POJO requiring deserialization from a distributed object (e.g. queue). Receive error message indicating factory was not registered:

```
[2017-08-17T15:52:58.993+1000] [] [SEVERE] [] [com.hazelcast.spi.impl.eventservice.impl.EventServiceImpl] [tid: _ThreadID=88 _ThreadName=hz._hzInstance_1_dev.event-4] [timeMillis: 1502949178993] [levelValue: 1000] [[
  [192.168.56.1]:5701 [dev] [3.8] hz._hzInstance_1_dev.event-4 caught an exception while processing task:com.hazelcast.spi.impl.eventservice.impl.LocalEventDispatcher@1c02c7aa
com.hazelcast.nio.serialization.HazelcastSerializationException: No DataSerializerFactory registered for namespace: 111
        at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.readInternal(DataSerializableSerializer.java:134)
        at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:103)
        at com.hazelcast.internal.serialization.impl.DataSerializableSerializer.read(DataSerializableSerializer.java:49)
        at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:48)
        at com.hazelcast.internal.serialization.impl.AbstractSerializationService.toObject(AbstractSerializationService.java:184)
        at com.hazelcast.spi.impl.operationservice.impl.InvocationFuture.resolve(InvocationFuture.java:106)
[...]

PR #1887, opened 2017-08-22
Rework global serializer setting, do not override user settings

scottg88:
Fixes issue #1872
This no longer overrides user settings if the global serializer is null. Instead will only set the global serializer if there are user serialization settings. Same behaviour if all serialization settings are null.
Tested with example provided in issue.


scottg88:
I'm not familiar with the purpose of the `PayaraHazelcastSerializer` so I'm not familiar with what tests to perform to ensure it still operates with these changes.

lprimak:
Hi,

Unfortunately this PR will break PayaraHazelcastSerializer.  It needs to be modified to pass through the custom serializer if exists.  Feel free to enhance your PR to provide this or we can close this PR.  I will get to it fairly soon when/if you close it

scottg88:
Thanks for the feedback. I believe this addresses it. This will create a whole serialization config if no user config, otherwise always override the global serialiser with the `PayaraHazelcastSerializer`.

lprimak:
I am not sure this is working as intended (not yet sure).
Give me a few days and I will get back to you,
once again thanks for your contribution

lprimak:
I had to rework some other ways the config is processed.
See PR #1915 
Let me know if that one works for you

PR #1900, opened 2017-08-25
PAYARA-1994 added addlibs alias for addjars

michaelranaldo:


michaelranaldo:
Jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/resources/commandoptions.properties
```diff
@@ -65,7 +65,7 @@ deployfromgav=<list-of-artefacts> specifies a comma separated groupId,artifactId
 additionalrepository=<repo-url> specifies an additional repository to search for deployable artefacts in
 outputuberjar=<file-path> packages up an uber jar at the specified path based on the command line arguments and exits
 copytouberjar=<directory-path> copy a directory into the root of an uberjar
-addjars=<file-or-directory-path> add specified jar files into classloader, or packaged into an uberjar
+addlibs=<file-or-directory-path> add specified jar files into classloader, or packaged into an uberjar. Multiple files can be specified as colon separated values.
```
smillidge: why did you delete addjars? Don't you need all enum values in the properties file?

mikecroft: +1 we need to maintain both addJars and addLibs in help. If we're not supporting anything other than JAR files with this, then I'm not sure the alias has much benefit and might just cause confusion.

MattGill98: I would just change --addJars description to 'Alias for --addLibs' to keep it supported.

michaelranaldo:
jenkins test please

PR #1901, opened 2017-08-25
PAYARA-1954 Added get rest monitoring configuration command

michaelranaldo:


michaelranaldo:
Jenkins test please

appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-service/src/main/java/fish/payara/appserver/monitoring/rest/service/admin/GetRestMonitoringConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+    private Target targetUtil;
+
+    @Param(name = "target", defaultValue = SystemPropertyConstants.DAS_SERVER_NAME, optional = true)
+    private String target;
+
+    @Override
+    public void execute(AdminCommandContext context) {
+        Config config = targetUtil.getConfig(target);
+        
+        RestMonitoringConfiguration restMonitoringConfiguration = config
```
Scoty: Correct me if I am wrong, but it looks NPE is possible here ? I guess this row should be below the null check for the config variable?

Pandrex247: Good spot! I agree this should be below the null check.

PR #1908, opened 2017-08-29
PAYARA-2013 The number of times EJB create method is called, is incorrect in jConsole

jGauravGupta:


jGauravGupta:
jenkins test please

PR #1916, opened 2017-08-30
PAYARA-2010 fixes connection pool health checker does not work on 173

smillidge:
fix OSGI manifest for connection pool checker to import hk2 configuration Property


smillidge:
jenkins test please

PR #1933, opened 2017-09-04
removed duplicated "-Djava.ext.dirs"

t-gergely:
I don't think deleting one line can be copyrighted so I skip the legal bits for now, sorry. Feel free to drop my contribution and delete that line yourself.


smillidge:
yeah deletion doesn't need IP assignment

smillidge:
jenkins test please

PR #1936, opened 2017-09-04
PAYARA-1761 Removed jsr181 and JAX-WS apis, since these are Java SE now

arjantijms:
PAYARA-1761

See: https://github.com/javaee/glassfish/issues/22205


arjantijms:

PR #1947, opened 2017-09-07
Payara 2053 update-connector-security-map cannot change user-groups to principals and vice versa

jGauravGupta:


jGauravGupta:
jenkins test please

PR #1951, opened 2017-09-08
PAYARA-1269-Upgrade-libpam4j-to-v1.8-with-groupid-org-kohsuke

mulderbaba:


mulderbaba:
jenkins test please

smillidge:
Does this class not need to change?

https://github.com/payara/Payara/blob/master/appserver/security/core-ee/src/main/java/com/sun/enterprise/security/auth/realm/pam/PamRealm.java

mulderbaba:
no @smillidge. I used that class for the realm config.

smillidge:
So the classnames are still org.jvnet?

mulderbaba:
exactly, see https://github.com/kohsuke/libpam4j/tree/master/src/main/java/org/jvnet/libpam

smillidge:
Doesn't it need an upgraded JNA?

mulderbaba:
upgraded jna comes with the libpam4j as a transitive dependency. our version of jna is now 4.0.0. I'm planning to update it as well, after forking libpam4j and sending a PR. if that doesn't get accepted in a timely manner we can host it in patched projects repo.

PR #1954, opened 2017-09-10
PAYARA-2063 fixed precedence on loading user defined configuration properties

### LINKED ISSUE #1953: Certain documented system properties seem not to override their command line equivalents
# Description #
----------

Documented system properties referenced [here](https://docs.payara.fish/v/172/documentation/payara-micro/configuring/config-sys-props.html) do not appear to serve as replacements for their command line equivalents.

## Expected Outcome

I was expecting that if I had a `system.properties` file like this:

    payaramicro.name=foo

...supplied like this:

    --systemProperties system.properties

...then output/behavior should be identical to what would happen if all I supplied instead was this:

    --name foo

...in accordance with the documentation.

## Current Outcome

Supplying just `payaramicro.name=foo` in `system.properties` as above (i.e. no `--name` option) does not in fact cause the instance name to be set.

This seems to be true as well for the `payaramicro.disablePhoneHome` property (only the `--disablePhoneHome` option, which turns (under the covers) into an `asadmin set` operation, works).

## Environment ##

- **Payara Version**: 4.1.1.172
- **Edition**: Micro
- **JDK Version**: payara/micro:172 JVM
- **Operating System**: payara/micro:172 OS
- **Database**: N/A

Thanks for a great product and timely help out there in the world!

smillidge:
Fixes #1953
Precedence in highest precedence first order is
1) Command line option
2) Property loaded via --systemproperties command line option
3) Property set in the uber jar MICRO-INF/deploy/payaramicro.properties
4) Property set in the uber jar MICRO-INF/payara-boot.properties
5) Property set on the command line via -D

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/FileSystemItemValidator.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/FileSystemItemValidator.java
@@ -40,8 +40,6 @@
 package fish.payara.micro.cmd.options;
 
 import java.io.File;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.text.MessageFormat;
 
 /**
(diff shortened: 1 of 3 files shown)


smillidge:
jenkins test please

PR #1968, opened 2017-09-12
Payara-2074 Handle log messages in MessageFormat by escaping double single quote

jGauravGupta:


jGauravGupta:
jenkins test please

PR #1972, opened 2017-09-12
Payara 1166 Add JDK8 Metaspace Configuration To Payaradomain

MattGill98:
Added -XX:MaxMetaspaceSize=256m to payaradomain template


MattGill98:

OndrejM:
I think that limiting the metaspace size to just 256M is very risky - in production, you generally want the service to operate at all costs, even if it takes more memory. With 256M for class metadata it could just not be enough and the server could crash without any warning.

With Java7 and permgen space, this scenario ended in server becoming stuck in case of a leak because perm gen space was always limited and took a portion of the heap. This often led to the practice of restarting the server after each redeployment to clean up the memory. This isn't desired in production, where you want to redeploy and get trunning fast.

The only reason to limit the metaspace is to prevent abnormal memory consumption in case of a classloader memory leak which often happens after several redeployments of an app without a server restart. If we want to limit the metaspace, I would use a limit which should be high enough for most applications and many redeployments - maybe 2GB. This is just a virtual limit and is used only if needed.

This answer explains more https://stackoverflow.com/questions/31455644/should-i-set-a-maxmetaspacesize#31463972
Also, worth reading the Oracle docs at https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/considerations.html#sthref66

OndrejM:
On the other hand, it makes sense to set MetaspaceSize to a higher level. It is usually set to around 20MB by default, causing GC of metaspace very early. We could raise it to 256M. This is the same as setting min heap space to a higher value to prevent too many GCs and resizing of the heap.

MattGill98:
Thanks for the review. I hadn't found the Oracle docs on it! I've changed the initial size to 256m and the max to 2g

Cousjava:
Jenkins quick test payara 5 please

MattGill98:
Jenkins quick test payara 5 please

MattGill98:
Jenkins quick test payara 5 please

MattGill98:

arjantijms:

PR #1984, opened 2017-09-14
PAYARA-1615-Healthcheck-machine-memory-usage-inaccurate

mulderbaba:


mulderbaba:
jenkins test please

mulderbaba:
@smillidge An IOException gets thrown but it's catched inside a try block and a healthcheck result entry gets generated with message: "Memory information cannot be read for retrieving physical memory usage values". We can do a OS check but it'll result with the same approach. Do you want me to inform user that this will only work on Linux for now?

smillidge:
Yeah probably a good idea and fall back to the old functionality when not on Linux as the results may be accurate on Windows (I don't know)

Pandrex247:
What's the state of this? It's been hanging around for ages.

mulderbaba:
jenkins test please

PR #1987, opened 2017-09-14
PAYARA-2124 update-connector-security-map can't delete backend principal password

jGauravGupta:


jGauravGupta:
jenkins test please

PR #1994, opened 2017-09-16
PAYARA-2095 Fix EJB timers broken in Payara 5 alpha1

smillidge:


smillidge:
@lprimak I deleted your disabling and reenabling of Hazelcast for a couple of reasons
1) It broke the EJB Timer store
2) Hazelcast in 5 is becoming more fundamental to the core server so more and more needs to be there during boot.

smillidge:

arjantijms:

lprimak:
I'll need to test this with an example.
There is likely an issue with this.  The intent is that Hazelcast is enabled after applications are deployed but not yet started

smillidge:
I want Hazelcast enabled by default. I don't want it booting after application deployment.

lprimak:
Remember that the new deployment sequence is split up between loading and initialization.
1) Applications and classes are loaded
2) Hz is started
3) Applications are initialized.

Hz needs to be started after all application classes are loaded and available
I think the issue is coming up because the timers are started in phase (1) and should be started in phase (3)
I can take a look at that as well

arjantijms:

PR #1998, opened 2017-09-18
PAYARA-2136 Dead lock occurs when both invalidate() and getAttribute() are called at the same time

jGauravGupta:


jGauravGupta:
jenkins test please

PR #1999, opened 2017-09-18
PAYARA-2139 Number of Stateful session beans in Passive state is not updated in Jconsole

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2000, opened 2017-09-19
PAYARA-2145 SessionContext.getCallerPrincipal() returns previous principal on TimerService

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2007, opened 2017-09-19
PAYARA-2126 fix NPE in JDBC-RA in non-server (ACC/embedded) environments

lprimak:
fixes NPE where Domain and it's configuration is not available in the environment,
i.e. embedded or standalone client


lprimak:
jenkins test please

PR #2031, opened 2017-09-24
PAYARA-2150 Rebuild context for CDI Event Bus as Weld initialisation …

smillidge:
…is different in Payara 5


lprimak:
Why did you add appName to the context code serialization?  ComponentID already has appName in it.
In previous iterarions of this code I used to use appName but found it unnecessary and messes up versioned apps 
Perhaps there is another solution that does satisfies your needs and doesn't use appName

smillidge:
Because appname is required by the Concurrency implementation. It uses it to check the application is deployed. Event bus wasn't working because all requests to the managed executor were rejected because appname wasn't present on the current invocation.

lprimak:
Unfortunately, this definitely breaks the basic tenet of JavaEEContextUtil, which should be brain-dead easy to use.  Also, it breaks the Hazelcast unmerged PR in a complicated way.
Also, it definitely breaks versioned applications.

I ran into a similar issue with ManagedExecutorService in the past, and I think it's a bug/missing feature there and should be fixed there not to require appName.

lprimak:
I have a pretty good idea on how to fix all of this without changing JavaEEContextUtil API in a breaking way, so you can assign this issue to me

arjantijms:
Open for too long, closing it now. Let's see if we can revisit this later.

PR #2050, opened 2017-10-05
PAYARA-2172, PAYARA-2191, PAYARA-2192 Configuring the Notifiers for a Standalone/Cluster instance targets DAS

MeroRai:


Pandrex247:
Jenkins test please

PR #2052, opened 2017-10-06
PAYARA-2178 Micro postboot commands should skip empty lines

MeroRai:


MeroRai:
Jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/BootCommands.java
```diff
@@ -80,7 +80,7 @@ public void parseCommandScript(URL scriptURL) throws IOException {
             String commandStr = br.readLine();
             while (commandStr != null) {
                 // # is a comment
-                if (!commandStr.startsWith("#")) {
+                if (!commandStr.startsWith("#") && commandStr.trim().length() > 0) {
```
Cousjava: How about doing commandStr.trim before the if? That way if there is whitespace and then a # it will also be skipped.

MeroRai: Ya cool, will do that. Thanks.

MeroRai:
Jenkins test please

PR #2061, opened 2017-10-10
PAYARA-2199 Can't create Connector Connection Pool for inbound RA deployed on a cluster

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2065, opened 2017-10-11
PAYARA-2209 Warn if SFSB lifecycle callbacks use container-transaction that is not RequiresNew or NotSupported

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2097, opened 2017-10-25
PAYARA-2109 Protect Admin Console Web Pages against Clickjacking/UI Redress attacks

MeroRai:


MeroRai:
Jenkins test please

appserver/admingui/web/src/main/resources/org/glassfish/web/admingui/Strings.properties
```diff
@@ -300,6 +300,8 @@ http.XpoweredBy=XPowered By:
 http.XpoweredByHelp=Include X-Powered-By: Servlet/3.0 in servlet-generated HTTP response headers
 http.ServerHeader=Server Header:
 http.ServerHeaderHelp=Include Server Header: Servlet/3.0 in servlet-generated HTTP response headers
+http.XframeOptions=XFrame Options:
+http.XframeOptionsHelp=Include X-Frame-Options: Servlet/3.0 in servlet-generated HTTP response headers
```
smillidge: Probably should say "Default X-Frame-Options value to be included in all responses"

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/GlassfishNetworkListener.java
```diff
@@ -437,24 +447,35 @@ protected boolean onHttpHeaderParsed(final HttpHeader httpHeader,
...
           
             if (xPoweredBy == null && httpHeader.containsHeader(Header.XPoweredBy)) {
                 httpHeader.getHeaders().removeHeader(Header.XPoweredBy);
             }
+            
+            if (serverVersion == null && httpHeader.containsHeader(Header.Server)){
+                httpHeader.getHeaders().removeHeader(Header.Server);
+            }
+            
+            if (xFrameOptions == null && httpHeader.containsHeader(xFrameOptionsHeader)){
```
smillidge: I don't think you should remove an XFrameOptions Header if it is already present. XPoweredBy and Server are removed when configured to be removed because they leak security information. However XFrame-Options could be set by the application developer for legitimate reasons so should be honoured if present.

MeroRai:
Jenkins test please

MeroRai:
Jenkins test please

PR #2114, opened 2017-10-31
PAYARA-2240 JMX monitoring values with spaces now recognised

michaelranaldo:
JMX Monitoring values which have spaces as part of their name, path, or type can now be entered by escaping the space:
`name=PS\ Scavenge`


michaelranaldo:
jenkins test please

PR #2121, opened 2017-11-02
PAYARA-2247 disable jmx log monitoring service by default.

smillidge:
Disables JMX Log Monitoring by default in Payara Micro


smillidge:
jenkins test please

PR #2122, opened 2017-11-02
PAYARA-2248 Disable notification service at boot of Payara Micro to s…

smillidge:
…ave some boot time


smillidge:
jenkins test please

PR #2136, opened 2017-11-07
PAYARA-2253 Update Hibernate Validation to 6.0.3

arjantijms:
PAYARA-2253 Hibernate Validation updated to 6.0.3, which is also the plain upstream version now.

### DIFF
--- a/nucleus/pom.xml
+++ b/nucleus/pom.xml
@@ -226,8 +226,8 @@
         
         <!-- Bean Validation -->
         <javax.validation.version>2.0.0.Final</javax.validation.version>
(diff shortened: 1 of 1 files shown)


arjantijms:
jenkins quick test payara 5

arjantijms:
jenkins quick test payara 5

arjantijms:
Fails because GitHub seems to return an invalid ID, or Jenkins parses the ID wrongly as inter:

```
ERROR: Build step failed with exception com.fasterxml.jackson.core.JsonParseException: Numeric value (4303347774) out of range of int at [Source: 
```

Trying to re-do the PR and see if we can get another ID.

Issue #2150, opened 2017-11-12
Excessive number of JAR scanner threads leads to high JVM native memory usage


lreimer:
# Excessive number of JAR scanner threads leads to high JVM native memory usage #
----------

When running Payara server full on large OpenShift cluster installations with cluster nodes with a lot of CPU cores (like 128 cores or more) we are seeing a high number of `deployment-jar-scanner` and `dol-jar-scanner` threads being started and waiting. Depending on the configured thread stack size of the JVM (1MB per default) this will lead to a high JVM native memory usage.

## Expected Outcome

Once the server is up and running and all applications are deployed these threads should not be required anymore (?) and could be stopped and released again. Thus reducing the JVM native memory usage of the Docker container.

## Current Outcome

Currently, we are seeing a large number of `deployment-jar-scanner` and `dol-jar-scanner` threads WAITING and thus waisting JVM native memory.

## Steps to reproduce (Only for bug reports) 

Basically, all you need to do is to start the default Payara server on an OpenShift cluster with cluster nodes that have a lot of CPU cores (like 128 and more).

1 -** Use the default `payara/server-full`. Deploy any sample app (or none).

2 -** Deploy the image to an OpenShift cluster. Beware: the cluster nodes need to have a lot of CPU cores to show the behaviour.

3 -** Log into one of the pods of the OpenShift deployment.

   $ oc rsh <pod-name>

4 -** Get the stack trace and count the numbers of `deployment-jar-scanner` and `dol-jar-scanner` threads using `jstack 1 | grep jar-scanner | wc -l

### Samples

Any Payara example will suffice and should show the described behaviour. The problem is, that you will need a really large OpenShift cluster setup with cluster nodes with 128 cores or more. 

## Context (Optional)

Currently, I am seeing excessive native memory usage for all our Payara server pods on OpenShift. Due to the high number of JAR scanner threads this leads, depending on the the configured thread stack size, to a lot of native memory being occupied by the JVM.

## Environment ##

- **Payara Version**: 4.1.2.173
- **Edition**: Full
- **JDK Version**: OpenJDK 8_141
- **Operating System**: Linux
- **PaaS**: OpenShift 3.5

lreimer:
The relevant classes that cause the described behaviour are in `com.sun.enterprise.v3.server.ApplicationLifecycle` line 2258 ff

```java
private ExecutorService createExecutorService() {
        Runtime runtime = Runtime.getRuntime();
[...]

PR #2151, opened 2017-11-12
Improvement for high number of JAR scanning threads.

lreimer:
This fixes and improves #2150. It uses `Executors.newCachedThreadPool(..)` instead of a fixed thread pool with the number of available processors as size instead.

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/annotation/impl/ModuleScanner.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/annotation/impl/ModuleScanner.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2016-2017] [Payara Foundation and/or its affiliates.]
 
 package com.sun.enterprise.deployment.annotation.impl;
 
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/annotation/impl/ModuleScanner.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/annotation/impl/ModuleScanner.java
@@ -348,9 +348,8 @@ public abstract class ModuleScanner<T> extends JavaEEScanner implements Scanner<
         if (executorService != null) {
             return executorService;
         }
-        Runtime runtime = Runtime.getRuntime();
-       int nrOfProcessors = runtime.availableProcessors();
-        executorService = Executors.newFixedThreadPool(nrOfProcessors, new ThreadFactory() {
+
+        executorService = Executors.newCachedThreadPool(new ThreadFactory() {
             @Override
             public Thread newThread(Runnable r) {
                 Thread t = new Thread(r);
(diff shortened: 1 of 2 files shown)


ggam:
Shouldn't you set a maximum number of threads? By default cached thread pool uses Intenger.MAX_VALUE as its limit. I guess the number of cores would be a more apropiate value.

lreimer:
You are right, Integer.MAX_VALUE is the theoretical limit. But I doubt that you will ever reach this. Even better is the following: _Threads that have not been used for sixty seconds are terminated and removed from the cache._

According to the Javadocs (see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newFixedThreadPool(int).

My assumption is that the threads for these pools are very short-lived and are only really required during the startup of the server. If so, the improved behaviour is way better than before having 80+ threads just WAITING around and wasting my JVM native memory.

svendiedrichsen:
Maybe using `ThreadPoolExecutor` could do both. 
* limit the time threads exist in the pool and thus save memory
* limit the max number of threads used to the number of cores

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html

Just use i.e. https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html#ThreadPoolExecutor(int,%20int,%20long,%20java.util.concurrent.TimeUnit,%20java.util.concurrent.BlockingQueue,%20java.util.concurrent.ThreadFactory)

ggam:
@svendiedrichsen I was thinking precisely about that, thanks.

svendiedrichsen:
Using the `ThreadPoolExecutor` you could also bring down the time the idle threads sit waiting to i.e. 10 seconds.

lreimer:
:+1: Sounds fine to me. I will adjust the used executor service accordingly.

lreimer:
Changed everything to ThreadPoolExecutor. Mininum size is 0, maximum size is number of processors. Timeout for threads is 30 SECONDS. Used a `LinkedBlockingQueue` instead of a `SynchronousQueue`.

Cousjava:
jenkins test please

lprimak:
Since I've been banging my head against the wall with ThreadPoolExecutor recently, I will review this code after it's stabilized.

smillidge:
Did you see any deployment time impact with these changes?

lreimer:
Hard to tell. I did some testing. Here is the setup:
1) Unmodified Payara Server Full 174
2) Payara Server Full (Master) with modified Thread Pool Core Size 0
3) Payara Server Full (Master) with modified Thread Pool Core Size 2

Then, for all 3 server variants I did the following iterations and steps:
- Iteration 1
  - Fresh start of domain1
  - Access the admingui so it gets deployed
  - Check deployment thread count using `jstack`
  - Stop domain1
- Iteration 2
  - Start domain1 again
  - Check deployment thread count using `jstack`
  - Stop domain1
- Iteration 3
  - Start domain1 again
  - Check deployment thread count using `jstack`
  - Stop domain1

The results are attached. [timings.txt](https://github.com/payara/Payara/files/1499065/timings.txt)

For 2) iteration 3 I would say it is about 300ms slower, but threads are left after deployment is done.

svendiedrichsen:
How about the memory consumption after start? I thought the main reason for this was the loss of native memory after start.

lprimak:
jenkins test please

PR #2163, opened 2017-11-16
PAYARA-2281 System Property Substitution Fails For Asadmin Recorder

MattGill98:
The asadmin recorder output file location will now always be substituted, regardless of thread-local configuration


MattGill98:
jenkins test please

PR #2166, opened 2017-11-18
PAYARA-1909 Domain Discovery Mode

smillidge:
Major Refactor of the Hazlecast Integration. By default Hazlecast now uses information in the domain.xml to determine where members of the cluster are likely to be. This results in pretty transparent clustering of all instances in the domain.
All instances and the DAS now form a single unified Hazelcast Data Grid for data sharing including web session data and JCache data.
Tested on Azure in a VPC all instances in the domain clustered automatically using SSH nodes
Tested on AWS in a VPC all instances in the domain clustered automatically using SSH nodes
Tested on Google Compute in a VPC all instances in the domain clustered automatically using SSH nodes
Tested on LAN in Linux all instances in the domain clustered automatically using SSH nodes

Payara Micro has new flag --clustermode which can be domain:host:port to cluster with a DAS (or any other instance) or it can be tcpip:host:port:host:port to provide a list of hosts and ports to check. This change make it easy to cluster micros in the cloud where IPs are known e.g. VPC. Default is the original multicast

Host aware partitioning is now the default.


smillidge:
Jenkins quick test Payara 5 please

smillidge:
Jenkins quick test Payara 5 please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1960,7 +2000,7 @@ private void setArgumentsFromSystemProperties() {
         sslPort = getIntegerProperty("payaramicro.sslPort", Integer.MIN_VALUE);
         hzMulticastGroup = getProperty("payaramicro.mcAddress");
         hzPort = getIntegerProperty("payaramicro.mcPort", Integer.MIN_VALUE);
-        hostAware = getBooleanProperty("payaramicro.hostAware");
+        hostAware = getBooleanProperty("payaramicro.hostAware","true");
```
arjantijms: >hostAware = getBooleanProperty("payaramicro.hostAware","true");

Could do with a space after the ,

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -2341,6 +2391,19 @@ private Boolean getBooleanProperty(String value) {
         }
         return "true".equals(property);
     }
+    
+    private Boolean getBooleanProperty(String value, String defaultValue) {
+        String property;
```
arjantijms: `property = System.getProperty(value);`

could become

`String property = System.getProperty(value);`

Why separate those?

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/utils/JsonUtil.java
```diff
@@ -140,7 +141,9 @@ public static JsonValue getJsonValue(Object object, boolean hideConfidentialProp
             result = Json.createValue((BigDecimal) object);
         } else if (object.getClass().isEnum()){
             result = Json.createValue(object.toString());
-        } else {
+        } else if (object instanceof InetAddress) {
+            result = Json.createValue(object.toString());
+        }else {
```
arjantijms: `}else {`

Could use a space between after {

nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/MemberAddressPicker.java
```diff
@@ -0,0 +1,158 @@
...
+ * @author Steve Millidge (Payara Foundation)
+ */
+public class MemberAddressPicker implements MemberAddressProvider {
+
+    private final ServerEnvironment env;
+    private final HazelcastRuntimeConfiguration config;
+    private final HazelcastConfigSpecificConfiguration localConfig;
+    private InetSocketAddress bindAddress;
+    private InetSocketAddress publicAddress;
+    private static final Logger logger = Logger.getLogger(MemberAddressPicker.class.getName());
```
arjantijms: By convention, static variables are put before instance variables, so the logger could be moved up to line 61

nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/MemberAddressPicker.java
```diff
@@ -0,0 +1,158 @@
...
+    @Override
+    public InetSocketAddress getBindAddress() {
+        return bindAddress;
+    }
+
+    @Override
+    public InetSocketAddress getPublicAddress() {
+        if (publicAddress != null) {
+            return publicAddress;
+        } else {
```
arjantijms: It's a matter of style, but the else is not needed here. Can return directly after the if block.

smillidge:
Jenkins quick test Payara 5 please

lprimak:
Agreed with @arjantijms style comments above, although all are minor

lprimak:
GUI UX usability issue:
When on the Data Grid configuration screen, it has a radio box: multicast/tcpip/domain
The issue is that parameters for all three options are present below, which is confusing.
The parameters should only be there for applicable setting of the radio button, and change depending which radio button is selected.
I think this is implemented somewhere else in the admin console as well, I don't remember where though.

lprimak:
Ran a simple test: 1 DAS and 1 Micro.  No changes (default domain1 configuration)
Got the following error from Micro:

```
[2017-12-07T00:08:35.605-0600] [] [SEVERE] [] [com.hazelcast.cluster] [tid: _ThreadID=42 _ThreadName=hz._hzInstance_1_development.priority-generic-operation.thread-0] [timeMillis: 1512626915605] [levelValue: 1000] [10.0.1.15]:5900 [development] [3.9] Node could not join cluster. A Configuration mismatch was detected: Incompatible joiners! expected: tcp-ip, found: multicast Node is going to shutdown now!
```

With default options, Micro should start correctly, otherwise it breaks the "works out-of-the-box" principle

smillidge:
That will require chanting micro default cluster mode to domain. However micro is a standalone product and clusters by multicast by default. As your tested scenario is not the default usage model for micro I disagree about it working out of the box.

smillidge:
Hazelcast 3.9 still needs all the interface checking unless it has changed in a later minor release

smillidge:
Ah I see what you mean. Payara Micro does not start if a DAS is running. I'll try and work out how it is trying to join the DAS in a cluster.

smillidge:
I'll investigate and recreate the PR

Issue #1841, opened 2017-08-06
Misleading outcome when using "stop-domain" without specifying domain


ggam:
# Description #
----------

The `stop-domain` command always defaults to the `domain1` domain. But when you've started another domain (e.g. `payaradomain`), calling `stop-domain` gives you the following output:

```
c:\payarabin>asadmin stop-domain
CLI306: Warning - The server located at C:\servers\payara-4.1.2.173-nightly\glassfish\domains\domain1 is not running.
Command stop-domain executed successfully.
```

It's the "command executed successfully" what's confusing to me. The command has finished correctly, but doing nothing. 

## Expected Outcome

I think a message like the following would be clearer:
```
c:\payarabin>asadmin stop-domain
Trying to stop domain1...
CLI306: Warning - The server located at C:\servers\payara-4.1.2.173-nightly\glassfish\domains\domain1 is not running.
domain1 could not be stopped. You may want to specify a different domain to stop.  Running domains: payaradomain.

Command stop-domain executed successfully.
```

Another possibility would be to both default to `domain1` and then `payaradomain` if `domain1` was not started.

## Environment ##

- **Payara Version**: 4.1.2.173 nightly
- **Edition**: Full
- **JDK Version**: Oracle JDK 8
- **Operating System**:  Windows 10

Cousjava:
Created internal issue PAYARA-2113 to change it

smillidge:
Fixed by #2169

PR #2169, opened 2017-11-20
PAYARA-2113 Made stop-domain a little more descriptive

michaelranaldo:
Addresses #1841

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ListDomainsCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ListDomainsCommand.java
@@ -51,8 +51,8 @@ import com.sun.enterprise.util.HostAndPort;
 import com.sun.enterprise.util.io.DomainDirs;
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
+import java.util.Map;
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.*;
 import org.glassfish.hk2.api.PerLookup;
@@ -65,7 +65,7 @@ import org.jvnet.hk2.annotations.*;
 @PerLookup
 public final class ListDomainsCommand extends LocalDomainCommand {
 
-    private static final LocalStringsImpl strings = new LocalStringsImpl(ListDomainsCommand.class);
+    private static final LocalStringsImpl Strings = new LocalStringsImpl(ListDomainsCommand.class);
     private String domainsRoot = null;
 
     @Param(name = "long", shortName = "l", optional = true)
@@ -112,7 +112,7 @@ public final class ListDomainsCommand extends LocalDomainCommand {
                     }
                 }
             } else {
-                logger.fine(strings.get("NoDomainsToList"));
+                logger.fine(Strings.get("NoDomainsToList"));
             }
         } catch (Exception ex) {
             throw new CommandException(ex.getLocalizedMessage());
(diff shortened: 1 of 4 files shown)


michaelranaldo:
jenkins test please

michaelranaldo:
jenkins test please

michaelranaldo:
jenkins test please

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ListDomainsCommand.java
```diff
@@ -60,99 +64,107 @@
...
             }
-        }
-        catch (Exception ex) {
+        } catch (Exception ex) {
             throw new CommandException(ex.getLocalizedMessage());
         }
         return 0;
     }
 
+    protected List<String> getRunningDomains() throws IOException, DomainException, CommandException {
```
MattGill98: I would probably have this method return something more generic. Maybe a Map<String, String> mapping domains to their status or something similar. IMHO, the list-domains command shouldn't care about whether the domain is running or not. It's just me being pedantic, but I think it's better to invert the control on this one

michaelranaldo: Right, makes sense. I shall bow to the pedantry in this case ;)

michaelranaldo:
jenkins test please

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/ListDomainsCommand.java
```diff
@@ -65,7 +65,7 @@
 @PerLookup
 public final class ListDomainsCommand extends LocalDomainCommand {
 
-    private static final LocalStringsImpl strings = new LocalStringsImpl(ListDomainsCommand.class);
+    private static final LocalStringsImpl Strings = new LocalStringsImpl(ListDomainsCommand.class);
```
MattGill98: This should follow naming conventions of all capitals

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
```diff
@@ -170,11 +171,18 @@ protected int dasNotRunning() throws CommandException {
...
+                for (String domain : domains.keySet()) {
+                    if (domains.get(domain)) {
+                        runningDomains += "\n" + domain;
+                    }
                 }
             } catch(Exception e) {        
             }
+            runningDomains = 
+                    (runningDomains.length() < 1) 
+                    ? "\nNo domains are currently running."
```
MattGill98: I like this change, but could it be implemented using a different message name in the LocalString.properties?

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/StopDomainCommand.java
```diff
@@ -140,34 +145,56 @@ protected int executeCommand()
...
+                    (runningDomains.length() < 1) 
+                    ? "\nNo domains are currently running." 
+                    : "\nPlease specify one of the currently running domains:" + runningDomains;
+            logger.warning(Strings.get("StopDomain.dasNotRunning", getDomainRootDir(), runningDomains));
+        } else {
             logger.warning(Strings.get("StopDomain.dasNotRunningRemotely"));
-        return 0;
+        }
+        // If it has gotten this far, the domain has failed to be stopped so the command has failed
+        return 1;
```
smillidge: Why have you changed the return value for the command?

michaelranaldo: At this point the command has failed to find a valid running instance that matches the input, so it would make more sense to have the command report it as a failure.

michaelranaldo:
jenkins test please

Cousjava:
HK2 cannot be used as it may be run when there is no running server and so it does not have access to HK2.

Pandrex247:
Sounds like something that should be done in a separate PR

MattGill98:
I believe Jonathan's comment is pertinent to the PR. If HK2 can't be used, the current code won't work. I believe it works just fine, but it does need clarifying

michaelranaldo:
Interesting; definitely something worth investigating.

PR #2175, opened 2017-11-22
PAYARA-2294 Switch to Secure Random

smillidge:


smillidge:
jenkins test please

jerrinot:
what's the reason for this? see https://www.2uo.de/myths-about-urandom/

smillidge:
Synch from upstream

appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/AuthenticatorBase.java
```diff
@@ -193,7 +193,7 @@
     /**
      * A random number generator to use when generating session identifiers.
      */
-    protected Random random = null;
+    protected SecureRandom random = null;
```
arjantijms: Assignment to null is technically not needed

PR #2177, opened 2017-11-23
PAYARA-2285 Jmx Monitoring Same Attribute Different Beans Merge

MattGill98:
- Updated command set-monitoring-configuration from the format `--addproperty name=x value=x` to `--addattribute attributeName=x objectName=x`. Old command syntax is maintained for backwards compatibility.
- Multiple properties are added by providing multiple --addproperty commands.
- Better validation and output messages added to set-monitoring-configuration command.
- New MonitoredAttribute bean created, and REST endpoint to access it: `monitoring-service-configuration/monitored-attribute.json`. Functions vastly similar to the old `property.json` page, but with the new format.
- Admin console updated to reflect new changes.
- Updated admin console REST util to allow data in the body of a PUT request.


MattGill98:
jenkins test please

PR #2180, opened 2017-11-24
Update FileDirContext.java

markus4zander:
I've noticed that Payara on Windows delivers files that are located in a junction folder in docroot, although the parameter "allowlinking" is set to false. This is because a path check of the docroot base path with the canonical path of the file is performed in the FileDirContext class. According to Javadoc, the function getCanonicalPath resolves symlinks only on Unix systems. Therefore the junction could not be recognized under Windows. Switching to the function getRealPath solves this problem under Windows. The function has to be tested under the other operating systems.

### DIFF
--- a/appserver/web/web-naming/src/main/java/org/apache/naming/resources/FileDirContext.java
+++ b/appserver/web/web-naming/src/main/java/org/apache/naming/resources/FileDirContext.java
@@ -55,6 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// Portions Copyright [2017] [Payara Foundation and/or its affiliates]
 
 package org.apache.naming.resources;
 
@@ -897,7 +898,7 @@ public class FileDirContext extends BaseDirContext {
             // Check that this file belongs to our root path
             String canPath = null;
             try {
-                canPath = file.getCanonicalPath();
+                canPath = file.toPath().toRealPath().toString();
             } catch (IOException e) {
             }
             if (canPath == null) {


markus4zander:
jenkins test please

smillidge:
jenkins test please

smillidge:
Thanks for the PR
Could you describe how to reproduce the original issue so we can test the change?

markus4zander:
To reproduce the original issue, you need a windows machine. Let's assume, your docroot folder is `C:\Payara\domain\docroot`. Lets further assume, you have a folder `C:\topsecret\recipes` with the file `mothers_cheesecake.txt` in it. Now create a junction with the command `mklink /j C:\Payara\domain\docroot\leak C:\topsecret\recipes`. If you have set `allowLinking` to `false` (which is the default value), it should not be possible to access the receipe for mother's cheescake via payara. But if you go to your browser and open `http://localhost:8080/leak/mothers_cheesecake.txt`, you will see the recipe. With my change, you can't access the file anymore and get a 404 response.

Pandrex247:
Hmm, this hasn't fixed the issue for me - I can still see the file with your branch.
Have I missed something?

* Build your branch using JDK8 - _https://github.com/markus4zander/Payara/tree/issue-mz001_
* Copy the _payara41_ directory to _C:\Payara_
* Make the recipes directory and file (_C:\TopSecret\Recipes\mothers_cheesecake.txt_)
* Make the symlink using _Command Prompt (Admin)_ in _C:\Payara\payara41\glassfish\domains\domain1\docroot_ directory - `mklink /J leak C:\TopSecret\Recipes`
* Start the domain and navigate to _http://localhost:8080/leak/mothers_cheesecake.txt_

markus4zander:
Hi Andrew,

I could reproduce the behavior you described. I will cover two things: First, what is the difference between your test and my test and second, what is the technical reason, why it works in my case.

In the simplified test case, I described, the junction was created in the standard docroot folder. In my real project, it was created in an alternate docroot, which was specified in the `web.xml` file of an .war file.

Why does this make a difference? The reason for this is in the class `org.glassfish.grizzly.http.server.util.Mapper.java`, namely in the method `internalMapWrapper`. There are many nested if-else-cases and in most of them, there is a call of `context.resources.lookup(pathStr)`, but not in all. But the check for `allowLinking` is only done, when this method is called. And in the test case with the standard docroot, the method is not called. Could you please check, whether it is a good idea to add the call to `context.resources.lookup(pathStr)` in all cases?

lprimak:
I don't think ```allowLinking``` actually does anything in Payara.  This was brought in with Tomcat code, but since then diverged.

If you symlink a directory into docroot, it's the proper behavior to have the files in the linked-to folder to be accessible by Payara.
I further think it's a dead end to change anything regarding this, including this PR, because it will have issues with NAS and can introduce other problems in complex network and storage environments.

lprimak:
Bottom line, I think the current behavior in Payara is appropriate and no changes are necessary

markus4zander:
@lprimak The actual behavior is at least inconsistent, because it differs between linux and windows and it differs between the standard docroot and several other places (e.g. an alternate docroot defined in the web.xml of an war file). We can discuss, what the right solution would be, but something has to be fixed.

The parameter is still documented in the actual glassfish reference manual, which is linked on the payara homepage. [https://javaee.github.io/glassfish/doc/4.0/reference-manual.pdf](url)
If you search for "allowLinking" in this document, you will find 3 matches.

More, this is a security issue. The glassfish documentation itself notes, that it is not a good idea to set allowLinking to true. See old Glassfish 3.1-documentation: [https://docs.oracle.com/cd/E18930_01/html/821-2417/beayb.html#indexterm-368](url) . But currently, payara follows symlinks sometimes under linux and always under windows.

If you have compatibility concerns, it would be possible to set the default value for allowLinking to true, but IMHO this is not a good idea.

lprimak:
Can you post an http://sscce.org for your real use case please?
I.e. the one where your altdocroot in the war file which you mention before?
Clearly, linking in standard docroot yields expected result.
Perhaps I am not understanding your "real" use case so please provide it

markus4zander:
Ok, I can Do that, but first we have to be clear, what we expect as correct behavior.

I would expect that:
1) payara respects the "allowLinking" as documented in the Glassfish 4.0 reference documentation
2) The default value for "allowLinking" is "false" as documented in the Glassfish 4.0 reference documentation
3) there is no difference between operating systems
4) there is no difference between the different docroot locations
5) As a result of 1) to 4), a document in a junction folder should not be served by payara in the default configuration.

Can you confirm those expectations or do you have other ones?

lprimak:
It sounds reasonable, but "there is no difference in the operating systems" is just way too broad a standard, especially if there is a risk of breaking existing customer installations.

markus4zander:
Ok, I'll create an SSCCE as soon as I'm back from my Christmas vacation (December 23,2017 - January 7,2018).

markus4zander:
I'll still have a lot to do at work. Please let this PR open.

arjantijms:
@markus4zander Do you perhaps have a broad indication of when you may have time for this?

markus4zander:
I should have time for this in the next two weeks.

AdamBien:
One option could be to deactivate the symbolic links following all together. In Germany we have a list of rules defined by a government institution called BSI( https://www.bsi.bund.de/DE/Themen/ITGrundschutz/ITGrundschutzKataloge/itgrundschutzkataloge_node.html). 

One of the rule prescribes the deactivation of symbolic link following for the Apache webserver. Some clients just re-apply the rule to application servers. 

I had a chat with @markus4zander he will prepare a VM a replicate the rule for Windows OS.

arjantijms:
@AdamBien Cool, looking forward to it, thanks!

markus4zander:
My colleagues and I are currently running dozens of tests. We expect results this week.

markus4zander:
We have completed our tests. To demonstrate the most important results we have set up two VirtualBox-VMs. One with Windows and one with Linux.

We have performed the following steps on both VMs:

- Installed JDK V8u162
- Installed Maven V3.5.2
- Downloaded the sources of the original Payara from https://github.com/payara/Payara/archive/master.zip 
- Downloaded the sources of the Payara modified by me from https://github.com/markus4zander/Payara/archive/master.zip 
- Compiled both versions with `mvn clean install -DskipTests`
- Unpacked the resulting payara. zip. The original payara to `C:\payaraStd` (Windows) or `/home/user/payaraStd` (Linux) and the modified Payara to `C:\payaraMZ` (Windows) or `/home/user/payaraMZ` (Linux). 
- Created the file `cheesecakes. txt` in the folder `C:\TopSecret` (Windows) or `/home/user/TopSecret` (Linux)
- Created a folder symlink named `Leak` to the `TopSecret` directory in the folder `c:\payaraStd\payara41\glassfish\domains\domain1\docroot` and `c:\payaraMZ\payara41\glassfish\domains\domain1\docroot` (Windows) or `/home/user/payaraStd/glassfish/domains/domains1/docroot` and `/home/user/payaraMZ/glassfish/domains/domains1/docroot` (Linux)
- Afterwards we tested the access to this file via the URL http://localhost:8080/leak/cheesecakes.txt with different settings for the property `allowLinking`. The results can be found in the table below.
 
Payara | AllowLinking undefined (default) | AllowLinking set to true | AllowLinking set to false
------ | -------------------------------- | ------------------------ | -------------------------
according to reference manual | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404
original on Windows | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200 (wrong!)
original on Linux | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404
modified on Windows | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404
modified on Linux | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404

The test results clearly show that the PR leads to the desired behavior under Windows and does not cause any behavioral changes in standard installation as well as under Linux.

Currently we upload both VMs to our server. As soon as this is finalized, I'll post it here.

Pandrex247:
Hmm, I'm still not getting it in the same way you are. What I'm seeing is this:

Payara | AllowLinking undefined (default) | AllowLinking set to true | AllowLinking set to false
------ | -------------------------------- | ------------------------ | -------------------------
original on Windows | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200
original on Linux | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404
modified on Windows | ---> **HTTP-Status 404** <--- | HTTP-Status 200 | HTTP-Status 404
modified on Linux | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404

Can you double check the highlighted one please?
I'll get someone else here to try as well, it might just be my machine is being weird.

Pandrex247:
Wait, I thought the desired behaviour was for `allowLinking` to default to _false_? From your results it defaults to true both before and after your changes. From my results it seems to be working as you want it to?

markus4zander:
I have to double check this with the sources and the reference manual. But anyway, my change should not affect Linux. Perhaps, there is an other issue, that comes into play. Give me a little bit more time, I will check this.

Pandrex247:
Sure, looking at the docs for GlassFish 4 and 5, it seems the default should be _false_.
The default was _true_ in [GlassFish 3](https://docs.oracle.com/cd/E26576_01/doc.312/e24938/create-virtual-server.htm#GSRFM00062), but it looks like it was changed to _false_ in [GlassFish 4](https://javaee.github.io/glassfish/doc/4.0/application-deployment-guide.pdf#G11.1030730).

So it looks like your change is bringing the behaviour back in line with the documentation.

Pandrex247:
Sorry, got confused by the docs.
The default value for _allowLinking_ is true on both 3 and 4.
It defaults to false in the sun-web-app / glassfish-web-app xml file though.

So it **should** behave as your results show (allowed when no property is present), but I'm getting a different result (not allowed when no property is present).

markus4zander:
Hi @Pandrex247. Your hint that the default for allowLinking in `domain. xml` and `glassfish-web-app.xml` are different was crucial. In a fresh Payara installation the result is "200"as shown in my table, because the default `allowLinking=true` from the `domain.xml` is active. However, as soon as you call the admin interface only once or deploy an application, the default `allowLinking=false` from the web. xml will kick in and you will get "404" as a result. This also applies after a restart of the server. This is a very unexpected behavior and I think it's a good idea to align the two default values to prevent such an unexpected behavior.

Pandrex247:
Aha! Yes, that seems to be where the discrepancy in our results is coming from - I was using the admin console to configure the `allowLinking` property.

So if I redo my tests with this in mind, my actual results are this:

Payara | AllowLinking undefined (Admin Console not loaded) | AllowLinking undefined (Admin Console loaded) | AllowLinking set to true (Admin Console not loaded) | AllowLinking set to true (Admin Console loaded) | AllowLinking set to false (Admin Console not loaded) | AllowLinking set to false (Admin Console loaded) 
------ | -------------------------------------------------- | ---------------------------------------------- | --------------------------------------------------- | ----------------------------------------------- | ---------------------------------------------------- | -------------------------------------------------
original on Windows | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 200
original on Linux | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 404
modified on Windows | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 404
modified on Linux | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 200 | HTTP-Status 200 | HTTP-Status 404 | HTTP-Status 404

This looks to be about right in line with the docs now.
I agree it's a little jarring to have the same property have a different default value in the two places it can be used, but I'm counting it as a separate issue so I'm going to move forward and merge this PR.
I'll make an issue on our internal JIRA to align the default values - my immediate gut reaction is to change the default value in the _domain.xml_ to **not** allow the linking, as the documentation for GlassFish 4 states that allowing the linking potentially opens up some JSP vulnerabilities on Windows (which I would guess are still present):

> Caution: Setting this property to true on Windows systems exposes JSP source code.

Feel free to beat us to the punch and make your own PR though :P

Thanks for the PR and all your effort! :)

PR #2185, opened 2017-11-28
PAYARA-2300 Added logging of servlet mappings for each module at boot

arjantijms:
PAYARA-2300

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/data/ModuleDescriptor.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/data/ModuleDescriptor.java
@@ -40,9 +40,12 @@
 package fish.payara.micro.data;
 
 import java.io.Serializable;
+import java.util.List;
+import java.util.Map.Entry;
 
 /**
  * Describes an application module
+ * 
  * @author Steve Millidge
  */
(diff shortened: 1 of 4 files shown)


arjantijms:
Jenkins quick test Payara 5 please

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/data/ModuleDescriptor.java
```diff
@@ -51,18 +54,27 @@
      * Returns the context root of the application if any
      * @return 
      */
-    public String getContextRoot();
+    String getContextRoot();
```
smillidge: Why has the visibility changed as this is a breaking change? These classes are exposed via the public Payara Micro api and now these methods aren't visible.

arjantijms: ModuleDescriptor is an interface. Methods in an interface are public by default, not package or anything else. The public access modifier is redundant to use for interface methods. I.e. removing or adding them makes no difference for the visibility.

arjantijms:
Jenkins quick test Payara 5 please

arjantijms:
Jenkins quick test Payara 5 please

PR #2193, opened 2017-12-01
PAYARA-2303 ConcurrentModificationException in JTA

lprimak:
using working list to prevent concurrent modification exceptions in JTA


lprimak:
jenkins test please

PR #2201, opened 2017-12-08
PAYARA-1909 Domain Discovery Mode

smillidge:
Domain 

Major  Refactor of the Hazlecast Integration. By default Hazlecast now uses  information in the domain.xml to determine where members of the cluster  are likely to be. This results in pretty transparent clustering of all  instances in the domain. All instances and the DAS now form a single unified Hazelcast Data Grid  for data sharing including web session data and JCache data. Tested on Azure in a VPC all instances in the domain clustered automatically using SSH nodes Tested on AWS in a VPC all instances in the domain clustered automatically using SSH nodes Tested on Google Compute in a VPC all instances in the domain clustered automatically using SSH nodes Tested on LAN in Linux all instances in the domain clustered automatically using SSH nodes Payara Micro has new flag --clustermode which can be domain:host:port  to cluster with a DAS (or any other instance) or it can be  tcpip:host:port:host:port to provide a list of hosts and ports to check.  This change make it easy to cluster micros in the cloud where IPs are  known e.g. VPC. Default is the original multicast Host aware partitioning is now the default.

Also now fixes issue with multicast micros trying to join the domain mode cluster and failing


smillidge:

smillidge:
@lprimak same code just fixed the issue with the Micro not booting if the DAS is running

smillidge:

lprimak:
Although minor, I think the configuration screen is still confusing

smillidge:
I'll raise a separate JIRA for the config screen as it isn't my area of expertise.

PR #2204, opened 2017-12-09
PAYARA-2304 Support raw deployments on Payara Micro

smillidge:
deploy applications that are missing --deploy flag to improve user experience on Payara Micro with this PR
```java -jar payara-micro.jar test.war ``` will just work


smillidge:
jenkins test please

PR #2224, opened 2017-12-21
PAYARA-2350 Fixed configuration of hostaware and lite and fixed typo …

smillidge:
…in asadmin command parameter set-hazelcast-configuration


lprimak:

PR #2246, opened 2018-01-08
PAYARA 2386 Fix Payara Micro Arquillian Container

MattGill98:
- Container now stops in undeploy phase, to prevent multiple running instances.
- Tests now wait completely for startup by following log output.
- No longer confuses output from different tests.


MattGill98:

PR #2250, opened 2018-01-09
PAYARA-2327 History log files are not retained however their count is less than the maximum

MeroRai:


MeroRai:
Jenkins test please

PR #2279, opened 2018-01-16
PAYARA-2267 JPA Conversion now unwraps connection properly

Cousjava:


Cousjava:
jenkins test please

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/base/ConnectionHolder.java
```diff
@@ -130,9 +130,12 @@ public ConnectionHolder(Connection con, ManagedConnectionImpl mc,
      * Returns the actual connection in this holder object.
      *
      * @return Connection object.
+     * @throws java.sql.SQLException
      */
-    public Connection getConnection() {
-        return con;
+    public Connection getConnection() throws SQLException {
+        //To ensure that the actual connection is always returned and not a proxy
+        Connection result = con.unwrap(java.sql.Connection.class);
```
smillidge: I don't believe blindly unwrapping on getConnection is the correct fix. I think the problem is why unwrap is failing. However I could be wrong.

smillidge: Is also missing //Portions Copyright comment

Cousjava:
jenkins test please

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/base/ConnectionHolder.java
Cousjava: unwrap was never called, and based on the comment above the method then the actual connection should be returned and not a wrapped version of it.

arjantijms:
@smillidge Is this now okay to be merged?

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/base/ConnectionHolder.java
Cousjava: unwrap connection is not not used anywhere else in Payara or Eclipselink, unless its done via reflection.

dmatej: I'm not sure if this would not introduce some connection leaks etc. I used own wrappers in 15 years old application to monitor behavior and connection leaks in very bad implementation (I left that company so . 
The problem is that when you unwrap connection outside the proxy, proxy methods will not be called. So you allow to obey mechanisms of the proxies.

I think the getConnection should return wrapped object and it should not do any "voodoo" with it. But I understand another view - how to get original connection? (and why someone needs that?)

But unwrap method (not implemented here) allows more sofisticated (recursive) method: "return first wrapped object implementing given interface". In our case we generated proxy wrapper for the IfmxConnection and returned that - this mechanism did not introduce any leaks and it was completely transparent.
And ... later I implemented wrappers for the Informix JDBC driver to allow usage of java.time API with the EclipseLink and Informix JDBC driver (which still implements only JDBC 3.50). The fact is that there was no access to the original wrapped connection (except reflection), but I think it was alright (https://github.com/dmatej/jdbc-adapter/blob/master/src/main/java/org/jdbc/adapter/ifx/Ifx42Connection.java).

So ... be careful ...

dmatej: Short version: it would be better to implement unwrap and isWrapperFor and let getConnection to return directly what was given in constructor.

lprimak: I also do not believe this is the correct fix.
unwrap() is called by the user to get access to underlying proprietary classes.
Proxy should be returned, not the raw connection here

PR #2303, opened 2018-01-21
PAYARA-2377 singleton resources

lprimak:
trying to fix race conditions with singleton resource handling


lprimak:
jenkins test please

PR #2325, opened 2018-01-25
Revert PAYARA-2354 and PAYARA-2303

lprimak:
The bug is actually in JMSContext, the fixes would have hidden this bug further,
so reverting these fixes


lprimak:
jenkins test please

lprimak:
Fix no longer applicable

hmeerlo:
Ok, this has cost me some hours :-) @lprimak can you explain why this was reverted, because now the ConcurrentModificationException has returned for me (only under load).

hmeerlo:
Never mind, must be caused elsewhere because the iterator construction is perfectly fine.

PR #2330, opened 2018-01-28
PAYARA-2377 JMS 2.0 context race condition workaround on by default

lprimak:
make skip resource registration is on by default
There are no bad sideffects due to this change


lprimak:
jenkins test please

lprimak:
jenkins test please

PR #2399, opened 2018-02-14
PAYARA-2406 Add Clustering Switch To Micro Arquillian Container

MattGill98:
Also added environment variable for already present `outputToConsole` option.


arjantijms:

PR #2420, opened 2018-02-18
PAYARA-2511 Ensure Optional property is returned as empty and not as unconfigured value

smillidge:
Fix to Microprofile Config api to ensure the optional is injected as empty if no value is specified for the property
```java
@Inject
@ConfigProperty(name="test")
Optional<String> value;
```


smillidge:
jenkins test please

PR #2444, opened 2018-02-22
PAYARA-2515 Micro Arquillian Tests Fail In Parallel

MattGill98:
- Payara Micro now reserves the ports it's configured once it's found them, so that two Micro instances starting up simultaneously don't choose the same port. Allows many Micro instances to startup at the same time.
- Merged `Started Grizzly Framework` messages into one.


arjantijms:

PR #2448, opened 2018-02-23
PAYARA-2464 Asadmin Ignores Create Jacc Provider Properties

MattGill98:
- Added code for handling properties


MattGill98:
jenkins test please

PR #2452, opened 2018-02-24
PAYARA-2523 Handle some simple mistakes when specifying domain cluster mode

smillidge:
PAYARA-2523 Handle some simple mistakes when specifying domain cluster mode in micro

localhost or 127.0.0.1 specified most likely IP address chosen
4848 specified as port a warning is printed that the default is 4900


smillidge:

PR #2455, opened 2018-02-25
PAYARA-2492 added options to unpack jars when Payara Micro is used programmatically

smillidge:
This option is required to compile JSPs but otherwise is not required.


smillidge:
jenkins test please

PR #2470, opened 2018-02-26
PAYARA-2542 Weld BDA context is the same in Server as in Micro

lprimak:
otherwise Weld clustering deserialization fails


lprimak:
jenkins quick test payara 5

arjantijms:
jenkins quick test payara 5

PR #2472, opened 2018-02-27
PAYARA-2542 @Stateful HA fixes

lprimak:
Never worked in HA to begin with, so fails on micro as wel


lprimak:
jenkins quick test payara 5

PR #2475, opened 2018-02-27
PAYARA-2553 Timeout option for Arquillian Micro connector

arjantijms:


appserver/extras/arquillian-containers/payara-micro/src/main/java/fish/payara/arquillian/container/payara/managed/PayaraMicroDeployableContainer.java
```diff
@@ -237,8 +235,18 @@ public ProtocolMetaData deploy(Archive<?> archive) throws DeploymentException {
                 }
             }, 1500, 200, MILLISECONDS);
 
+            int startupTimeoutInSeconds = configuration.isDebug()? -1 : configuration.getStartupTimeoutInSeconds();
```
Pandrex247: Missed a space before your '?'

arjantijms:
jenkins quick test payara 5

arjantijms:
jenkins quick test payara 5

appserver/extras/arquillian-containers/payara-micro/src/main/java/fish/payara/arquillian/container/payara/managed/PayaraMicroDeployableContainer.java
```diff
@@ -237,8 +235,18 @@ public ProtocolMetaData deploy(Archive<?> archive) throws DeploymentException {
                 }
             }, 1500, 200, MILLISECONDS);
 
+            int startupTimeoutInSeconds = configuration.isDebug() ? -1 : configuration.getStartupTimeoutInSeconds();
```
MattGill98: Why are the debug and timeout options mutually exclusive?

arjantijms: The idea is that if you are debugging you probably don't want a timeout as you're very directly interacting with the code. Alternatively this could be changed into; if you are in debug mode, the default is -1, otherwise the default is 180.

On the other hand, if you're not happy with the defaults you can simply not use the debug mode and enter the debug options manually as extra payara options instead. It's documented in the config btw that timeout is infinite (-1) when debug is used.

arjantijms:
jenkins quick test payara 5

appserver/extras/arquillian-containers/payara-micro/src/main/java/fish/payara/arquillian/container/payara/managed/PayaraMicroDeployableContainer.java
MattGill98: That makes sense, I was just checking it was deliberate!

PR #2481, opened 2018-02-28
PAYARA-2555 Micro Arquillian Container Doesn't Allow Spaces In Arguments

MattGill98:
This should&trade; also work on windows.


MattGill98:
Should work with multiple spaces now: https://regex101.com/r/Kyt77o/1/

PR #2491, opened 2018-03-02
PAYARA-2563 Fix tck failure in config api

smillidge:


smillidge:
jenkins quick test payara 5

PR #2494, opened 2018-03-03
PAYARA-2564 multi-threaded InitialContext() fix

lprimak:
added LinkageError to the catch statement so the error actually gets caught


lprimak:
jenkins test

PR #2497, opened 2018-03-05
PAYARA-2528-Payara-Micro-ignores-invalid-arguments-if-they-dont-startwith-double-dash

mulderbaba:


mulderbaba:
jenkins test please

PR #2501, opened 2018-03-05
PAYARA-2522 Deploying an application to an Instance in a Deployment Group and the Deployment Group causes the application to only target the instance

MeroRai:


MeroRai:
Jenkins Test Please

MeroRai:
Jenkins Test Please

PR #2534, opened 2018-03-16
PAYARA-1688-Redesign-the-public-notifier-API

mulderbaba:


mulderbaba:
jenkins test please

api/payara-api/src/main/java/fish/payara/notification/NotificationData.java
```diff
@@ -0,0 +1,49 @@
...
+ * holder.
+ */
+package fish.payara.notification;
+
+import java.io.Serializable;
+
+/**
+ * @author mertcaliskan
+ */
+public abstract class NotificationData implements Serializable {
```
OndrejM: It would be nice if NotificationData abstract class provided a method to convert itself to a subclass, such as:
```
<T> T as(Class<T extends NotificationData> cls)
```
so that the conversion can be fluent:

```
HealthCheckNotificationData hcData = notification.getData().as(HealthCheckNotificationData.class);
```

Similar to the Shrinkwrap API.

mulderbaba: make sense @OndrejM, will handle it in a separate PR

PR #2539, opened 2018-03-17
Implement SNI support for both Micro and Server PAYARA-1528

### LINKED ISSUE #1431: Add SNI SSL support
# Description #
----------

It would be nice to be able to configure/use SSL SNI (Server Name Identification).
## Expected Outcome
 

 Currently it's possible to have only one SSL certificate per IP address. With SNI, multiple SSL certificates (for different domains) can be used with the same IP address. 
This feature is already supported in Grizzly but I coudn't find a way to enable/configure it in Payara. 
## Environment ##

- **Payara Version**: 4.1.1.x
- **Edition**: 
- **JDK Version**: 
- **Operating System**:

smillidge:
Fixes #1431


smillidge:
jenkins test

PR #2546, opened 2018-03-18
PAYARA-2615 Add support for Payara Micro command line option --sslCert

smillidge:
to specify the certificate alias to use in the keystore

Also adds some extra logging is SSL is misconfigured
Also correctly packages up alternate keystores into the Uber Jar if specified via system properties


Pandrex247:

PR #2552, opened 2018-03-20
PAYARA-2584 Don't Let Payara 5 Start if JDK Version is Less than 8u161

MeroRai:


MeroRai:
Jenkins test please

Pandrex247:
Jenkins test please

ggam:
Could this also apply to JDK 9?

smillidge:
I think the minimum version is 161 not 162

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/MainHelper.java
```diff
@@ -107,6 +118,16 @@ private static int getMinorJdkVersion() {
         }
     }
 
+    private static int getSecurityJdkVersion() {
+        String jv = System.getProperty("java.version");
```
arjantijms: Avoid using abbreviations like "jv", just use "javaVersion" or "jdkVersion" etc here.

Pandrex247:
We actually need to udpate the JDK version on Jenkins before this will pass, as it doesn't use the new required version!

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/MainHelper.java
lprimak: There is also already JDK.java class in nucleus/common to take care of things like this

arjantijms:
Jenkins test

smillidge:
I have a concern with this in that it is actually possible to run on less than JDK 8u161 by switching the Grizzly NPN version.

lprimak:
That's probably true, so maybe we should morph this fix into new version-dependent options in domain templates

MeroRai:
@lprimak Ya, I think it will better to change this fix to new version-dependent options in domain templates

PR #2556, opened 2018-03-21
PAYARA-2620 Fix Application Edit Post Request

MattGill98:


MattGill98:
jenkins test please

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/PropertiesBagResource.java
```diff
@@ -176,11 +176,17 @@ protected ActionReportResult clearThenSaveProperties(List<Map<String, String>> p
...
+                String escapedName = getEscapedPropertyName(unescapedName);
+
                 String value = ((Object)property.get("value")).toString();
-                String description = ((Object)property.get("description")).toString();
-                final String unescapedValue = value.replaceAll("\\\\", "");
+                String unescapedValue = value.replaceAll("\\\\", "");
+                
+                String description = null;
+                if (property.get(description) != null) {
+                    description = Object.class.cast(property.get("description")).toString();
```
lprimak: That cast is super weird.  Can't you just get rid of it?

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/PropertiesBagResource.java
```diff
@@ -176,11 +176,17 @@ protected ActionReportResult clearThenSaveProperties(List<Map<String, String>> p
 
             for (Map<String, String> property : properties) {
                 Property existingProp = existing.get(property.get("name"));
-                String unescapredpropname = ((Object)property.get("name")).toString();
-                String escapedName = getEscapedPropertyName(unescapredpropname);
+
+                String unescapedName = ((Object)property.get("name")).toString();
```
lprimak: Why does this need to be cast to Object?  Makes zero sense

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/PropertiesBagResource.java
MattGill98: I agree it is super weird. Annoyingly we ran into issues before where the toString() method returned a JsonString so would cause a ClassCastException. So it's irritating, but necessary

Pandrex247:
jenkins test please

PR #2532, opened 2018-03-15
PAYARA-2579 Added keyfile to Payara Micro distribution

Cousjava:
fixes #2466

PR #2557, opened 2018-03-22
PAYARA-2579 Added keyfile to Payara Micro distribution

Cousjava:
Payara 4 version of #2532


Cousjava:

PR #2565, opened 2018-03-25
PAYARA-2630 Fix exception during boot when Payara started via java

arjantijms:
PAYARA-2630 when Payara is started via the java command, instead of via asadmin, it throws a bunch of exceptions and hangs. This fixes this.


arjantijms:
jenkins test please

PR #2520, opened 2018-03-11
PAYARA-2582 …

smillidge:
change application to be a local variable to prevent classloader leak when validating resources

PR #2577, opened 2018-03-27
PAYARA-2582 … (#2520)

arjantijms:
* PAYARA-2582 change application to be a local variable to prevent classloader leak when validating resources

* payara-2582 Fixed messed up infinite loop


arjantijms:

PR #2599, opened 2018-04-04
asadmin to import plain PKCS#8 RSA keypairs into the java keystore

ratcashdev:
ASADMIN command to import unencrypted PKCS8 RSA keypairs into JAVA keystores using the api directly (not via keytool). Such keys and certificates are produced by LETSEncrypt.

Don't allow empty or short key and/or keystore passwords.
Added convenience methods to enforce glassfish convention of keys using the same passwords as the keystore.

A couple of considerations:

1. Please have a look on the API in KeystoreManager. In many places it expects the `keyStoreType`, but as of https://bugs.openjdk.java.net/browse/JDK-8062552 it could be omitted and presume "JKS" as it works both for PKCS12 and JKS keystores - there's a relevant test for it.
2. Currently oly RSA keys are supported/tested
3. Only non-encrypted, plain key files are supported - but this is what LetsEncrypt delivers.
4. There are methods that may violate the glassfish convention of same passwords for the key and keystore, but those are clearly documented.


smillidge:
thanks I will try and review soon

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java
```diff
@@ -79,6 +79,22 @@
...
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.security.Key;
+import java.security.KeyFactory;
+import java.security.PrivateKey;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateFactory;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Base64;
```
smillidge: This is a JDK8 only class. This is OK for Payara 5 branch but not for Payara 4

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java
```diff
@@ -471,6 +487,210 @@ protected void changeKeystorePassword(String oldPassword, String newPassword,
...
+	 * @param privateKeyLines
+	 * @return the decoded binary content
+	 */
+	byte[] extractPrivateKeyBytes(Stream<String> privateKeyLines) {
+		String base64KeyData = privateKeyLines.filter((line) -> line.charAt(0) != '-')
+				.collect(Collectors.joining());
+		return Base64.getDecoder().decode(base64KeyData);
+	}
+
+	public Collection<? extends Certificate> readPemCertificateChain ( File pemFile ) throws KeyStoreException {
```
smillidge: javadoc needed

smillidge:
jenkins test please

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java
ratcashdev: Good point. So should we make this backportable to Payara 4 too?

ratcashdev: BTW: com.sun.enterprise.v3.admin.AdminAdapter is using this class too, so I thought it's ok.

ratcashdev:
jenkins test please

smillidge:
jenkins test please

Issue #2332, opened 2018-01-28
Error creating managed object for class: class org.jboss.weld.module.web.servlet.WeldListener


ahamana:
# Description #
----------

When server start up withe Payara 5, the error that "Error creating managed object for class: class org.jboss.weld.module.web.servlet.WeldListener" occurs.

## Expected Outcome

Server start up without fail.

## Current Outcome

\* I think the cause of this error relates with [this stackoverflow](https://stackoverflow.com/questions/48235803/error-creating-managed-object-for-class-class-org-jboss-weld-module-web-servle).
\* Both of the delegate options(true/false) of glassfish-web.xml, this error occurs.
\* If there are any ideas to resolve this error, please let me know.

[2018-01-28T19:54:06.756+0900] [Payara 5..Beta2] [警告] [NCLS-CORE-00069] [javax.enterprise.system.core] [tid: _ThreadID=58 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1517136846756] [levelValue: 900] [[
  Exception while dispatching an event
org.glassfish.api.invocation.InvocationException
	at org.glassfish.api.invocation.InvocationManagerImpl.postInvoke(InvocationManagerImpl.java:178)
	at org.glassfish.weld.WeldDeployer.event(WeldDeployer.java:293)
	at org.glassfish.kernel.event.EventsImpl.send(EventsImpl.java:131)
	at org.glassfish.internal.data.ApplicationInfo.load(ApplicationInfo.java:333)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.prepare(ApplicationLifecycle.java:494)
	at org.glassfish.deployment.admin.DeployCommand.execute(DeployCommand.java:512)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:549)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:545)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:360)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2.execute(CommandRunnerImpl.java:544)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$3.run(CommandRunnerImpl.java:575)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$3.run(CommandRunnerImpl.java:567)
[...]

PR #2606, opened 2018-04-07
PAYARA-2666 - show DepoymentException in logs in case of deployment error

OndrejM:
This logs InvocationException first and then DeploymentException with information about the original error. It may be worth fixing the InvocationException too because it has no real value for the user.

The log message for the InvocationException should be probably externalized similar to WELD_BOOTSTRAP_SHUTDOWN_EXCEPTION but I don't know how to do it correctly.


OndrejM:
Fixes https://github.com/payara/Payara/issues/2332

lprimak:
jenkins test

PR #2633, opened 2018-04-13
PAYARA-2687 ensure Config built with ConfigBuilder has the default co…

smillidge:
…nverters


mulderbaba:
jenkins test

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfigBuilder.java
```diff
@@ -106,6 +106,7 @@ public ConfigBuilder withConverters(Converter<?>... converters) {
 
     @Override
     public Config build() {
+        this.converters.addAll(resolver.getDefaultConverters());
```
arjantijms: "this." would not be needed here

PR #2634, opened 2018-04-13
PAYARA-2687 ensure Config built with ConfigBuilder has the default co…

smillidge:
…nverters


mulderbaba:
jenkins quick test payara 4

Issue #2613, opened 2018-04-10
Listener specific keystore not used


ratcashdev:
# Description #
----------

**Keystore explicitely assigned to be used by a listener is ignored.**

## Expected Outcome

the SSL implementation used by the listener (by default `com.sun.enterprise.security.ssl.GlassfishSSLImpl`) would read the keystore explicitely assigned to the listener.

## Current Outcome
The listener is using the default keystore specified as a `-Djavax.net.ssl.keyStore=` JVM parameter.

## Steps to reproduce (Only for bug reports) 

1. copy the default keystore `keystore.jks` in the config folder. Name it `keystore3.jks`
2. remove the s1as alias from keystore3
3. add a new keypair to keystore-http2, name it `ssl12`. 
4. configure http-listener-2 to use keystore3 and use alias ssl12, i.e. the `domain/config.xml` should look like:
`<protocol name="http-listener-2" security-enabled="true">
            <http max-connections="250" default-virtual-server="server">
              <file-cache></file-cache>
            </http>
            <ssl classname="fish.payara.letsencrypt.ssl.PayaraSSLImpl" cert-nickname="ssl12" key-store="${com.sun.aas.instanceRoot}/config/keystore3.jks"></ssl>
          </protocol>`
![http-listener-2-config](https://user-images.githubusercontent.com/2928922/38544716-849409be-3ca8-11e8-99e9-a6a8890da1b5.PNG)
Note: disregard the classname: it is a direct copy of the `com.sun.enterprise.security.ssl.GlassfishSSLImpl` with some logging enhancement.
5. start the domain. It will go all right
6. try to access `https://localhost:8181`. The browser will show **Site can't be reached**. (Chrome will show `This site can’t be reached. localhost unexpectedly closed the connection.`). This should however, work, since keystore3.jks contains the necessary alias and keypair. 
7. If you add alias ssl12 to the default keystore (keystore.jks) it will work correctly.

## Context (Optional)

This was discovered while trying to develop a keystore implementation that would facilitate hot-reload of the keystore and certificates for the purposes of https://github.com/payara/Payara/issues/1047
However, as it can be seen, it is not being called for the keystore in question,only for the default keystore (visible from the logs).

The actual `GlassfishServerSocketFactory` uses the following code snippet to retrieve the keystore inside methods `getKeyManagers(...)`:
```
String keystoreFile = (String) attributes.get("keystore");
        if (logger.isLoggable(Level.FINE)) {
[...]

Issue #1047, opened 2016-08-19
Let's Encrypt Integration


jerrinot:
An integration with [Let's Encrypt](https://letsencrypt.org/) for easy to use & secure HTTPS could be a killer feature. Configuring TLS is PITA in most application containers. 

For inspiration see [Caddy](https://github.com/mholt/caddy) 

Also: https://github.com/shred/acme4j

matthiasblaesing:
Not sure whether this is appropriate to hijack this - but it is related, as it covers the TLS integration. At work I have a gateway system for which access is protected by a custom CA which is configured to regularly create a CRL. Big problem: I have to restart the gateway system to reload the CRL.

If I'm not mistaken this is also true for changes to the trust- and keystores. So to sum it up it would be nice if the TLS implementation would dynamicly adapt to changes to the trust-/keystores and CRLs.

smillidge:
How would you see this working from a user perspective. What steps would you see them doing?

@matthiasblaesing dynamically reacting to keystore changes is doable but is an enhancement request can you raise an issue for that?

jerrinot:
@smillidge: See https://caddyserver.com/docs/automatic-https

Also:
![image](https://cloud.githubusercontent.com/assets/158619/17853275/132828a0-6874-11e6-8af1-c3ccab99fa2d.png)
:-)

For Payara Micro one could use whatever Payara Micro is using for configuration.

smillidge:
Cool picture. As long as people realise it probably won't work seamlessly behind a proxy, it would be a great feature to have.

mikecroft:
Internal issue id `PAYARA-1061`

Davio:
If you would use and enable this, you could maybe also enable secure admin by default?
And perhaps this would require the hostname to be set to some real value, i.e. not localhost.

ratcashdev:
"The simplest way to run your Java application in production."
I like the idea, although IMHO it would take more just an integration with letsencrypt. The wireframe should also include options to:
- adding the header: Strict-Transport-Security:max-age with a sane default value
- force Perfect Forward secrecy by pre-selecting (or disabling) the necessary cipher suites

While we're at it, if on-the-fly certificate reload and change would be possible, I would actually prefer doing it through asadmin CLI in a scriptable fashion (sample script being part of the payara package).

a-genius:
Hi all,

[...]

Issue #499, opened 2015-10-29
invalid cert-nickname, no error logged.


tbugfinder:
There is no error message logged in case an invalid cert-nickname is configured or the alias within keystore.jks file does not match.
In that case SSL just fails.

``` xml
<protocol security-enabled="true" name="https-listener-443">
            <http default-virtual-server="server">
              <file-cache></file-cache>
            </http>
            <ssl classname="com.sun.enterprise.security.ssl.GlassfishSSLImpl" cert-nickname="HEELOs1as"></ssl>
</protocol>
```

smillidge:
Thanks for reporting this we will try to reproduce the issue

mikecroft:
Internal defect ID `PAYARA-833`

PR #2635, opened 2018-04-14
Bugfix: listener not using the correct keystore

ratcashdev:
Based on the analysis in https://github.com/payara/Payara/issues/2613 i'd like to present a one-liner fix. This fix has the added benefit of reloading the relevant keystore(s) whenever a listener is enabled. 
Thus this change makes the procedure outlined in https://github.com/payara/Payara/issues/1047, specifically,
```
asadmin set server.network-config.network-listeners.network-listener.http-listener-2.enabled=false
asadmin set server.network-config.network-listeners.network-listener.http-listener-2.enabled=true
```
to activate the new keypair from letsencrypt (and alias) working.

I have tested this with secure-admin enabled admin port (4848), as well as two separate listeners on different ports and using different keystores/aliases. All was working.

According to my research, the only references to GlassfishSSLImpl in the project are inside domain.xml configurations as SSL implementations for the listener(s).

The simplicity of this fix raises the question, why is/was the existing solution (using GlassfishServerSocketFactory employing SecuritySupportImpl) necessary in the first place? It seems an overly complicated code and lots of lines to be able to load the default keystore(s), and ignore any others? Especially, because all the code providing keystore and truststore unification is useless, because the `SecuritySupport` contract does not allow specifying the necessary keystore(s) for the given listener or adding more keystores. From this perspective, this "Pluggable" is broken as well, which would deserve its own bug.

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/ssl/GlassfishSSLImpl.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/ssl/GlassfishSSLImpl.java
@@ -43,6 +43,7 @@ package com.sun.enterprise.security.ssl;
 import java.net.Socket;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLSocket;
+import org.glassfish.grizzly.config.ssl.JSSE14SocketFactory;
 import org.glassfish.grizzly.config.ssl.SSLImplementation;
 import org.glassfish.grizzly.config.ssl.ServerSocketFactory;
 import org.glassfish.grizzly.ssl.SSLSupport;
@@ -64,7 +65,7 @@ public class GlassfishSSLImpl extends SSLImplementation {
     }
 
     public ServerSocketFactory getServerSocketFactory() {
-        return new GlassfishServerSocketFactory();
+        return new JSSE14SocketFactory();
     }
 
     public SSLSupport getSSLSupport(Socket socket) {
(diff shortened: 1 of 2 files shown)


lprimak:
jenkins test

ratcashdev:
Tested a MICRO instance (5.181) with an updated ssl-impl.jar from the latest master. With, or without the  change in this PR, the System.getProperty("javax.net.ssl.keyStorePassword") is always returning 'changeit' even though I use the `-Djavax.net.ssl.keyStorePassword=asdasd` as a cmdline parameter for JAVA. It seems it gets overriden somewhere.
For this reason it seems #65cdd01 is having little effect.

ratcashdev:
Mystery clarified: the culprit is IdmService. I forgot to update nucles/core/kernel.jar and thus it did not include commit 0a4a08a. Back to testing.

ratcashdev:
Also solves https://github.com/payara/Payara/issues/499

When providing a non-existent alias, the following appears in the server.log:
```
[2018-04-17T15:08:46.218+0200] [Payara 5.181] [WARNING] [] [javax.enterprise.network.config] [tid: _ThreadID=43 _ThreadName=http-thread-pool::http-listener-2(1)] [timeMillis: 1523970526218] [levelValue: 900] [[
  GRIZZLY0050: SSL support could not be configured!
java.io.IOException: 5330: Alias name s1as2 does not identify a key entry
```

smillidge:
Hi thanks for the PR. Just for double safety could you add a system property that toggles the old behaviour. e.g. `-Dfish.payara.ssl.GlassfishServerSocketFactory=true` that way if there are any issues in the field it can easily be switched back while we investigate.

ratcashdev:
My pleasure. See the updated PR.

lprimak:
jenkins test

PR #2643, opened 2018-04-17
PAYARA-2633 set domain.xml file encoding - Payara 4

lprimak:
added fish.payara.configFileEncoding JVM option / System property to set config file encoding (i.e. domain.xml)


lprimak:
jenkins quick test payara 4

Issue #2430, opened 2018-02-20
Resource Validator fails on java:/app/ paths


pdudits:
# Description #
----------

When deploying a module that references an EJB via JNDI path `java:/app/moduleName/beanName`, the resource validator fails to deploy the app with exception `JNDI Lookup failed / Got null ComponentInvocation`.

Workaround is to disable resource validation with `asadmin create-jvm-options -Ddeployment.resource.validation=false`

## Expected Outcome

Deployment should pass as the JNDI names are valid and worked in versions prior to 174.

## Current Outcome

````
org.glassfish.deployment.common.DeploymentException: JNDI lookup failed for the resource: Name: [java:module/env/net.dudits.payara.issue688.EjbProducer/firstEjb], Lookup: [java:app/ejbmodule/FirstEjb], Type: [Session]
	at com.sun.enterprise.deployment.util.ResourceValidator.validateJNDIRefs(ResourceValidator.java:874)
	at com.sun.enterprise.deployment.util.ResourceValidator.validateResources(ResourceValidator.java:736)
	at com.sun.enterprise.deployment.util.ResourceValidator.event(ResourceValidator.java:141)
	at org.glassfish.kernel.event.EventsImpl.send(EventsImpl.java:131)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:424)
	at com.sun.enterprise.v3.server.ApplicationLifecycle.deploy(ApplicationLifecycle.java:220)
	at org.glassfish.deployment.admin.DeployCommand.execute(DeployCommand.java:508)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:544)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:540)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:360)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$2.execute(CommandRunnerImpl.java:539)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$3.run(CommandRunnerImpl.java:570)
	at com.sun.enterprise.v3.admin.CommandRunnerImpl$3.run(CommandRunnerImpl.java:562)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:360)
[...]

PR #2644, opened 2018-04-17
PAYARA-2691 fix ResourceValidator Class Loaders

### LINKED ISSUE #2581: @Remote bean can't be accessed from a different app, when using Payara5
# Description #
----------

With the latest version of Payara 4 I am able to deploy the following two apps:

1. App1 has the following code(the below interface has single implementation):
```
import javax.ejb.Remote;

@Remote
public interface EmailNotificationManagerRemote {

	public void sendMail(String[] recipientsTo, String[] recipientsCC, String[] recipientsBCC, String subject,
            String body, String senderName, String senderMail, String mimeType, String mailSession, Boolean isSMS)
            throws InvalidInputException;
	
	 public void sendSupportMail(String errorCode, String subject, String body, String logLevel)  throws InvalidInputException;
}
```

App2 has the following code:

```
@EJB(mappedName="EmailNotificationManager")
private EmailNotificationManagerRemote emailNotificationManagerRemote;
```

When deploying App2(after App1 has already been successfully deployed), I get the following exception:

```
Exception during lifecycle processing
org.glassfish.deployment.common.DeploymentException: JNDI lookup failed for the resource: Name: [java:module/env/com.***.KeyLicenseNotifierBean/emailNotificationManagerRemote], Lookup: [EmailNotificationManager#com.***.EmailNotificationManagerRemote], Type: [Session]
...
Caused by: javax.naming.CommunicationException: Communication exception for SerialContext[myEnv={java.naming.factory.state=com.sun.corba.ee.impl.presentation.rmi.JNDIStateFactoryImpl, java.naming.factory.url.pkgs=com.sun.enterprise.naming} [Root exception is java.lang.RuntimeException: Cant copy Serializable object:]
...
Caused by: java.lang.RuntimeException: Cant copy Serializable object:
...
Caused by: java.lang.IllegalArgumentException: Illegal null argument to ObjectInputStreamWithLoader
```

## Expected Outcome

If App1 is successfully deployed, App2 should be deployed without exceptions.

## Current Outcome

App2 deployment fails.

Create two wars/jars one with a @Remote EJB definition and one which uses this remote bean. Check that when deploying the app, which uses the bean, the bean can't be found...

## Context (Optional)

The same two wars are deployable and are working on Payara4.

## Environment ##

- **Payara Version**: 5
- **Edition**:  Full
- **JDK Version**:  1.8.0_151
- **Operating System**:  Windows 10.0.15063 Build 15063

lprimak:
unset capturedInvocation (reset to null) if JNDI environment is not present,

signifying invalid invocation.  fixes #2581

### DIFF
--- a/appserver/web/gf-web-connector/src/main/java/fish/payara/appserver/context/JavaEEContextUtilImpl.java
+++ b/appserver/web/gf-web-connector/src/main/java/fish/payara/appserver/context/JavaEEContextUtilImpl.java
@@ -89,7 +89,6 @@ public class JavaEEContextUtilImpl implements JavaEEContextUtil, Serializable {
      */
     @Override
     public Context pushContext() {
-        ClassLoader oldClassLoader = Utility.getClassLoader();
         InvocationManager invMgr = serverContext.getInvocationManager();
         boolean invocationCreated = false;
         if (invMgr.getCurrentInvocation() == null && capturedInvocation != null) {
@@ -98,10 +97,11 @@ public class JavaEEContextUtilImpl implements JavaEEContextUtil, Serializable {
             invMgr.preInvoke(newInvocation);
             invocationCreated = true;
         }
+        ClassLoader oldClassLoader = null;
         if (invocationCreated) {
-            Utility.setContextClassLoader(getInvocationClassLoader());
+            oldClassLoader = Utility.setContextClassLoader(getInvocationClassLoader());
         }
-        return new ContextImpl.Context(oldClassLoader, invocationCreated? invMgr.getCurrentInvocation() : null, invMgr);
+        return new ContextImpl.Context(invocationCreated? invMgr.getCurrentInvocation() : null, invMgr, oldClassLoader);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


smillidge:
Can you check it in reference to https://github.com/payara/Payara/commit/b16f673dd19fc74924f088df5f4a2e23a743529d#diff-935446c41541b08168ed81591ca15470

lprimak:
Looks good to me

lprimak:
On second thought, the event AFTER_CLASSLOADER_CREATION the class loader is already set and JNDI environment isn't yet set up completely yet, so pushContext() is a no-op at this point.  Is there a particular reason that was needed?  Perhaps this is a vestage from Payara 4 where the application initialization was different

lprimak:
@smillidge

smillidge:
There was no  invocation which was causing an issue. This was the original issue I fixed. https://github.com/payara/Payara/issues/2430

lprimak:
Thanks!  I'll re-test that one

lprimak:
ok, added new setEmptyContext() to JavaEEContextUtil for this purpose, now works in both tickets

lprimak:
jenkins test

appserver/web/gf-web-connector/src/main/java/fish/payara/appserver/context/JavaEEContextUtilImpl.java
```diff
@@ -69,6 +69,14 @@
 @Service
 @PerLookup
 public class JavaEEContextUtilImpl implements JavaEEContextUtil, Serializable {
+    private transient @Getter(AccessLevel.PACKAGE) ServerContext serverContext;
+    private transient ComponentEnvManager compEnvMgr;
+    private transient ComponentInvocation capturedInvocation;
+    private @Getter(onMethod = @__(@Override)) String instanceComponentId;
```
Pandrex247: Unless I missed something we're still not allowing lombok.

lprimak: This has been there for a long time now.  I just changed the location as per coding standards.
I will get to making a massive PR in the future to remove lombok from everything so it can be correctly evaluated.
For now, please don't hold up a regression fix for this, I am just keeping the status quo for now, not adding new lombok code

Issue #2646, opened 2018-04-17
Measure coverage


ratcashdev:
Sonarqube is up & running (https://github.com/payara/Payara/issues/150), Jenkins as well, but coverage is not yet measured/published (AFAIK).
Putting coverage against complexity helps targetting test efforts effectively.

Cousjava:
@ratcashdev The reason for coverage being 0% is that we have very little in the way of unit tests in the Payara project, our tests are in their own repository at https://github.com/payara/PayaraTests

rdebusscher:
There are various test repositories, also internal ones, so a global coverage is not published.

PR #2648, opened 2018-04-18
Code coverage analysis using Jacoco

ratcashdev:
Fix for https://github.com/payara/Payara/issues/2646


lprimak:
jenkins test

lprimak:
@ratcashdev is there any particular reason this is in nucleus/pom.xml rather than root or appserver/pom.xml?

ratcashdev:
@lprimak putting it into root.pom was ineffective. Later I realized, that root.pom is not the parent of nucleus.pom, and that they don't share common ancestors. The current parent of nucleus is:
```
<parent>
        <groupId>net.java</groupId>
        <artifactId>jvnet-parent</artifactId>
        <version>5</version>
        <relativePath />
    </parent>
```
On the other side, appserver/pom.xml uses nuclues/pom.xml as its parent.

Issue #2619, opened 2018-04-11
Context root defined in glassfish-web.xml always gets overridden when deployed via browser UI


scrmtrey91:
# Description #

We decided to change names for our war files (add version). But we still need the old context root of the application (because of internal calls inside IS).

We created test app with glassfish-web-xml in webapp/WEB-INF folder with structure:
 
```
<!DOCTYPE glassfish-web-app PUBLIC "-//GlassFish.org//DTD
GlassFish Application Server 3.1 Servlet 3.0//EN"
        "http://glassfish.org/dtds/glassfish-web-app_3_0-1.dtd"> 
<glassfish-web-app>
    <context-root>ApplicationAccess</context-root>
</glassfish-web-app>
```

## Expected Outcome

When deployed on server, the context root should set correcly to defined  value in xml deployment descriptor if exist (via browser ui or via asadmin).

## Current Outcome

When deployed on server via browser UI, it should automaticaly set context root value from xml deployment descriptor. But it does not.
`Loading application [TestApp-1.0.0] at [/TestApp-1.0.0]]]`

When deployed via intellij (maven asadming) it does work as excpected (context root is set correcly as defined in xml file). 
`Loading application [TestApp-1.0.0] at [/ApplicationAccess]]]`

dmatej:
I would not be so rude, but I agree this descriptor is not too "usage friendly". Documentation is unavailable (this is not enough: https://docs.oracle.com/cd/E18930_01/html/821-2417/beaql.html#scrolltoc ), some elements are duplicated, and their meaning is not clear. It should be revised and perfectly documented - and maybe new xsd/dtd should be released.

I had a problem with arquillian deployments too because the context root changed with each version (version is a part of the name of the artifact). Finally I had to change the filename of the deployed archive  `ShrinkWrap.create(WebArchive.class, "ec.war")` so the context is `/ec` now.

matthiasblaesing:
@lprimak is right - the GUI guesses the context root from the war name. However having an option in the GUI to let it take the values from the war deployment descriptor instead of the GUI fields would be helpful. I routinely see badly deployed applications at work because the developer does not notice the broken name. Of course this could be mitigated by using the `asadmin` command (that uses the deployment descriptor unless explicitly asked to override the values), but then the web gui is convenient.

phillipross:
[...]

PR #2659, opened 2018-04-23
PAYARA-2697 default context root admin console improvement

lprimak:
default context root value for deployment is now taken internally,
the default GUI valus is now blank
Issue #2619


lprimak:
jenkins test

PR #2679, opened 2018-04-26
PAYARA-2425 Fix for using multiple envrionment variable in a line

Cousjava:


Cousjava:
Jenkins test please

smillidge:
I assume this requires a 4.x PR as well

mikecroft:
Jenkins test please

Cousjava:
Jenkins test please

mikecroft:
Note that the above error isn't resolved by the new commit. The problem is that you **can't add a JVM option which doesn't start with a hyphen**, not that it is parsed incorrectly. This is validation happening in the `create-jvm-options` command which needs to be tweaked slightly. (perhaps a check that, if the option does not start with a hyphen, then it checks to see if it is an environment variable, so only environment variables are permitted to not have hyphens).

Cousjava:
Jenkins test please

nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/JvmOptions.java
```diff
@@ -61,6 +61,12 @@
     private final static Pattern envP = Pattern.compile("([^\\$]*)\\$\\{ENV=([^\\}]*)\\}([^\\$]*)");
```
lprimak: the envP pattern needs a comment explaining what it does, with examples

Cousjava:
Jenkins test please

PR #2682, opened 2018-04-26
PAYARA-2710 do not use JDK 8-only Soteria CdiUtils class

lprimak:


lprimak:
jenkins quick test payara 4

lprimak:
jenkins quick test payara 4

PR #2693, opened 2018-04-30
PAYARA-2711 Compress on Rotation doesn't work for Payara Notification Logger

MeroRai:
… Logger (#2680)


Pandrex247:

PR #2702, opened 2018-05-02
PAYARA-2575 Yubikey Integration

MarkWareham:
PAYARA-2575 Yubikey Integration


appserver/featuresets/payara/pom.xml
```diff
@@ -138,5 +138,12 @@
             <version>${project.version}</version>
             <type>zip</type>
         </dependency>
+        <!-- Yubikey Authentication -->
```
Cousjava: Copyright needs updating in this file

appserver/packager/yubikey-authentication/src/main/assembly/yubikey-authentication.xml
```diff
@@ -0,0 +1,39 @@
...
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://glassfish.dev.java.net/public/CDDL+GPL_1_1.html
+ * or packager/legal/LICENSE.txt.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at packager/legal/LICENSE.txt.
+ -->
```
Cousjava: Use longer copyright text, as you did in the pom.xml

appserver/packager/yubikey-authentication/src/main/resources/pkg_proto.py
```diff
@@ -0,0 +1,36 @@
...
+# The contents of this file are subject to the terms of the Common Development
+# and Distribution License("CDDL") (collectively, the "License").  You
+# may not use this file except in compliance with the License.  You can
+# obtain a copy of the License at
+# https://glassfish.dev.java.net/public/CDDL+GPL_1_1.html
+# or packager/legal/LICENSE.txt.  See the License for the specific
+# language governing permissions and limitations under the License.
+# 
+# When distributing the software, include this License Header Notice in each
+# file and include the License file at packager/legal/LICENSE.txt.
```
Cousjava: Again, use longer copyright

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/authentication/twoFactor/TwoFactorAuthenticationMechanismDefinitionAnnotationLiteral.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.enterprise.util.AnnotationLiteral;
+import javax.security.enterprise.authentication.mechanism.http.LoginToContinue;
+
+/**
+ * A literal representation of the {@link fish.payara.security.annotations.TwoFactorAuthenticationMechanismDefinition}
+ * 
+ * @author Mark Wareham
+ */
+@SuppressWarnings("AnnotationAsSuperInterface")
+public class TwoFactorAuthenticationMechanismDefinitionAnnotationLiteral
```
Cousjava: Does this class actually do anything?

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/identitystores/YubicoClientFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+import fish.payara.security.annotations.YubikeyIdentityStoreDefinition;
+import com.yubico.client.v2.YubicoClient;
+
+/**
+ * Factory class for obtaining an instance of Yubico Client.
+ * 
+ * @author Mark Wareham
+ */
+public class YubicoClientFactory {
```
Cousjava: Could this be done as an application-scoped CDI bean instead?

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/identitystores/YubikeyCredential.java
```diff
@@ -0,0 +1,81 @@
...
+    
+    public String getPublicID(){
+        String fullKey = getOneTimePasswordString();
+        if(fullKey==null){
+            return null;
+        }
+        if (fullKey.isEmpty()){
+            return "";
+        }
+        if(fullKey.length()<12){
```
Cousjava: fullKey.length<12 makes fulKey.isEmpty redundant as the same result would be returned by this one

Cousjava:
Also Yubikey needs adding to embedded and web build

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/annotations/TwoFactorAuthenticationMechanismDefinition.java
```diff
@@ -0,0 +1,61 @@
...
+import javax.enterprise.util.Nonbinding;
+import javax.security.enterprise.authentication.mechanism.http.LoginToContinue;
+
+/**
+ * Used to define a Two Factor Authentication Mechanism with an annotation and provide configuration options
+ * @author Mark Wareham
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface TwoFactorAuthenticationMechanismDefinition {
```
Cousjava: I think this would be better named something more like YubikeyMechanismDefiniton, as a 2FA definition won't necessarily use Yubikey as one of its methods. If you do stick with TwoFactorAuthenticationMechanismDefinition, then using yubikey should be another option inside it.

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/identitystores/YubikeyCredential.java
MarkWareham: I was thinking of StringUtils.isEmpty() which checks whitespace insensitively

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/identitystores/YubicoClientFactory.java
MarkWareham: Good idea. This way the YubicoClient is per-app

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/annotations/TwoFactorAuthenticationMechanismDefinition.java
MarkWareham: The 2FA Mechanism is completely independent of yubikey. You could use the 2FA mech with an LDAPIdentityStore and a DatabaseIdentityStore for example. It selects the identity stores based on what's defined in the app and their priority

smillidge:
Also are the lambdas strictly necessary as this will push the feature to be Payara 5 only.

Cousjava:
@MarkWareham My review comments are still relevant

arjantijms:
For this issue the same comment holds as https://github.com/payara/Payara/pull/2704#issuecomment-391014755

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/annotations/TwoFactorAuthenticationMechanismDefinition.java
Cousjava: This is still not giving the option of multiple authentication mechanism, which is what 2FA is, instead this is making multiple identity stores used in an AND system rather than the default OR.

MarkWareham: I agree, it does not allow two authentication mechanisms, but two identity stores. I will rename to avoid confusion

MarkWareham:
jenkins test please

appserver/payara-appserver-modules/yubikey-authentication/src/main/java/fish/payara/security/annotations/YubikeyIdentityStoreDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+    
+    int yubikeyAPIClientID() default 0; //default is needed to allow yubikeyAPIClientIDExpression alone.
+
+    String yubikeyAPIKey();
+    
+    int priority() default 100;
+    
+    String priorityExpression() default "";
+    
+    String yubikeyAPIClientIDExpression() default "";
```
Cousjava: A Javadoc comment here would be useful

PR #2704, opened 2018-05-03
PAYARA-2718 Added Microprofile Config support to OAuth definition ann…

Cousjava:
…otation

### DIFF
--- a/appserver/payara-appserver-modules/security-oauth2/src/main/java/fish/payara/security/oauth2/OAuth2AuthenticationMechanism.java
+++ b/appserver/payara-appserver-modules/security-oauth2/src/main/java/fish/payara/security/oauth2/OAuth2AuthenticationMechanism.java
@@ -269,7 +269,7 @@ public class OAuth2AuthenticationMechanism implements HttpAuthenticationMechanis
         String result = value;
         Optional<String> configResult = provider.getOptionalValue(mpConfigKey, String.class);
         if (configResult.isPresent()) {
-            result = configResult.get();
+            return configResult.get();
         }
         result = (String) TranslatedConfigView.getTranslatedValue(result);
         if (isELExpression(value)){


appserver/payara-appserver-modules/security-oauth2/src/main/java/fish/payara/security/oauth2/OAuth2AuthenticationMechanism.java
```diff
@@ -108,44 +114,77 @@ public OAuth2AuthenticationMechanism() {
...
-     * Sets the properties of the OAuth2AuthenticationMechanism as defined in an
-     * {@link OAuth2AuthenticationDefinition} annotation.
+     * Sets the properties of the OAuth2AuthenticationMechanism as defined in an {@link OAuth2AuthenticationDefinition} annotation.
+     *
      * @param definition
-     * @return 
+     * @return
      */
-    public OAuth2AuthenticationMechanism setDefinition(OAuth2AuthenticationDefinition definition) {;
+    public OAuth2AuthenticationMechanism setDefinition(OAuth2AuthenticationDefinition definition) {
```
MarkWareham: I think you could reduce the complexity of this method with a utility method that does the check of which config to use if present. Especially as they are all Strings.
something like 
authConfig = getConfig(definition.authEndpoint())

appserver/payara-appserver-modules/security-oauth2/pom.xml
```diff
@@ -108,6 +108,18 @@
             <type>jar</type>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>${microprofile-config.version}</version>
+            <type>jar</type>
+        </dependency>
+        <dependency>
```
MarkWareham: Add <scope>provided</scope> here to prevent problems

Cousjava:
Jenkins test please

appserver/extras/embedded/all/pom.xml
```diff
@@ -341,6 +341,78 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.main.packager</groupId>
+            <artifactId>asadmin-recorder</artifactId>
+            <version>${project.version}</version>
+            <type>zip</type>
+            <optional>true</optional>
+        </dependency>
+        <!-- phomehome package -->
+        <dependency>
```
smillidge: What does any of this have to do with the name on the PR

arjantijms:
The general approach needs some changes. The code should not directly treat the annotation attribute as a potential MP Config key. The convention/`pattern that has emerged` is to use global keys that override. For instance in MP JWT 1.1 (draft):

> 9.3.1. mp.jwt.verify.publickey
> The **mp.jwt.verify.publickey** config property allows the Public Key text itself to be supplied as a string. The Public Key will be parsed from the supplied string in the order defined in section Supported Public Key Formats. 

Where **mp.jwt.verify.publickey** is a global well known key.

MP Fault tolerance uses something similar, but based on composite keys:

https://github.com/eclipse/microprofile-fault-tolerance/blob/master/spec/src/main/asciidoc/configuration.asciidoc

Finally, the issue originally asked for expression language (EL) support in the annotation attributes, which is consistent with EE Security. See https://javaee.github.io/javaee-spec/javadocs/javax/security/enterprise/package-summary.html

Since it's also a Payara annotation, which already uses the "${}" syntax, it might be best to only support the "#{}" syntax and treat that as resolve once / immediate-like.

Cousjava:
Jenkins test please

Cousjava:
Jenkins test please

arjantijms:
Jenkins test

appserver/payara-appserver-modules/security-oauth2/src/main/java/fish/payara/security/oauth2/OAuth2AuthenticationMechanism.java
```diff
@@ -248,4 +264,29 @@ private AuthenticationStatus redirectForAuth(HttpMessageContext context) {
         return context.redirect(authTokenRequest.toString());
     }
 
+    
+    private String getConfiguredValue(String value, Config provider, String mpConfigKey){
+        String result = value;
+        Optional<String> configResult = provider.getOptionalValue(mpConfigKey, String.class);
+        if (configResult.isPresent()) {
+            result = configResult.get();
```
arjantijms: If config result is present, just return it. Don't let that value go through the processors. 

(a concern was raised that this might introduce security loop holes, as the MP Config values can be set externally)

Cousjava:
Jenkins test please

PR #2717, opened 2018-05-11
PAYARA-2707 Dynamic registration of MBean as MP metric

Cousjava:
This allows for placeholder replacement in MP custom metrics


Cousjava:
Jenkins test please

Cousjava:
Jenkins test please

smillidge:
@Cousjava this now has conflicts can you resolve them

Cousjava:
Jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -192,7 +204,7 @@ public void resetMetricsSecureProperty() {
         MetricRegistry registry = getRegistry(registryName);
         Map<String, Metric> metricMap = registry.getMetrics();
         if (metricMap.containsKey(metricName)) {
-            return singletonMap(metricName, metricMap.get(metricName));
+            return Collections.singletonMap(metricName, metricMap.get(metricName));
```
arjantijms: Does "Collections.singletonMap" really make the code more readable than "singletonMap"?

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -202,7 +214,7 @@ public void resetMetricsSecureProperty() {
         MetricRegistry registry = getRegistry(registryName);
         Map<String, Metadata> metricMetadataMap = registry.getMetadata();
         if (metricMetadataMap.containsKey(metricName)) {
-            return singletonMap(metricName, metricMetadataMap.get(metricName));
+            return Collections.singletonMap(metricName, metricMetadataMap.get(metricName));
```
arjantijms: Does "Collections.singletonMap" really make the code more readable than "singletonMap"?

Cousjava: Yes.

arjantijms: What does it add for semantics knowing that the method comes from the `Collections ` class?

What you need to know reading the code is that a `singletonMap` is returned. Is there also another type of `singletonMap` that is not a `Collection`?

Cousjava: By just calling `singletonMap()` I would presume that is was directly present in the class or a superclass, it is non-obvious that it comes from the `Collections` class.

arjantijms: If it's only about "where it comes from" and not "the meaning itself isn't clear", then it doesn't matter ;)

arjantijms:
Jenkins test please

Cousjava:
Jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
Cousjava: I would still say it isn't clear is a local method could well do data processing, which the one in the Collections framework does not.

PR #2727, opened 2018-05-16
PAYARA-1061 LetsEncrypt integration script

### LINKED ISSUE #1047: Let's Encrypt Integration

ratcashdev:
Requires certbot with the webroot plugin.

The idea is to deploy an empty war (as a directory, using in-place deployment) and provide that directory's path to certbot's webroot plugin to pass the challenge.
Script is written in python, because certbot itself requires python. As certbot works only under linux, so does this script.
Proper functionality (reloading the keystore and restarting the listener) requires https://github.com/payara/Payara/commit/b3bd6b073dca50de17b5161d37b59b12827f3573 and https://github.com/payara/Payara/commit/83fb3c15e445575384d7f9966e4a9209d9571639

fixes #1047

### DIFF
--- a/appserver/packager/glassfish-common-web/src/main/resources/bin/letsencrypt.py
+++ b/appserver/packager/glassfish-common-web/src/main/resources/bin/letsencrypt.py
@@ -91,11 +91,11 @@ def invoke_certbot(domain_names):
 def check_http_port():
 	proc = Popen([ASADMIN_PATH, "get", "server.network-config.network-listeners.network-listener.*.port"], stdout=PIPE)
 	ports = proc.stdout.read()
-	running_on_port_80 = '.port=80\n'.encode() in ports
-	print(WARNING + "WARNING: " + ENDC + "None of the listeners of Payara are running on the standard HTTP port (80).\nUnless there is a port mapping, "
-		"a reverse-proxy exposing port 80 or other solution making the deployed web-app visible through port 80, the following invocation of certbot will "
-		"likely fail (due to a failing 'HTTP Challenge' of the ACME protocol; see Chapter 8.3 on https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html). "
-		"Proceeding nevertheless.")
+	if not '.port=80\n'.encode() in ports:
+		print(WARNING + "WARNING: " + ENDC + "None of the listeners of Payara are running on the standard HTTP port (80).\nUnless there is a port mapping, "
+			"a reverse-proxy exposing port 80 or other solution making the deployed web-app visible through port 80, the following invocation of certbot will "
+			"likely fail (due to a failing 'HTTP Challenge' of the ACME protocol; see Chapter 8.3 on https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html). "
+			"Proceeding nevertheless.")
 
 
 if __name__=="__main__":
--- a/appserver/packager/glassfish-common-web/src/main/resources/bin/letsencrypt.py
+++ b/appserver/packager/glassfish-common-web/src/main/resources/bin/letsencrypt.py
@@ -6,9 +6,10 @@ import os
 import sys
 from subprocess import PIPE,call, check_call, CalledProcessError
 import argparse
+import shutil
 
 ASADMIN_PATH="./asadmin"
-WEB_ROOT_PATH=Path(path.expanduser("~")) / "payara_le_war"
+WEB_ROOT_PATH=Path("/tmp/payara_le_war")
 LE_LIVE_PATH=Path("/etc/letsencrypt/live/")
 APP_NAME="le"
 FNULL = open(os.devnull, 'w')
@@ -54,6 +55,7 @@ def deploy_war():
 		print('[' + OKGREEN + " OK " + ENDC + ']')
 	except CalledProcessError: 
 		print('[' + FAIL + "FAIL" + ENDC + ']' + " Is the server up and running?\n", sys.exc_info()[1])
+		shutil.rmtree(WEB_ROOT_PATH)
 		return 1
 
 	return 0
@@ -63,8 +65,10 @@ def undeploy_war():
 	try:
 		check_call([ASADMIN_PATH, "undeploy", APP_NAME], stdout=FNULL, stderr=FNULL)
 		print('[' + OKGREEN + " OK " + ENDC + ']')
+		shutil.rmtree(WEB_ROOT_PATH)
 	except CalledProcessError: 
 		print('[' + FAIL + "FAIL" + ENDC + ']' + " Is the server up and running?\n", sys.exc_info()[1])
+		shutil.rmtree(WEB_ROOT_PATH)
 		return 1
 
 	return 0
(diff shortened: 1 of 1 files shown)


ratcashdev:
@lprimak  hi, can you please ask someone to look on this? thank you.

lprimak:
We will get to this.  We are busy with a quarterly release right now.
Thanks for your patience

mulderbaba:
Jenkins test please

lprimak:
Im not sure that I am good person to review this, as I don't really know Python
Also, how do you run the script?  Do you just go to the Payara install location and run it?
Any documentation / examples would be good so I or someone can at least test the functionality

MattGill98:
Hi @ratcashdev,

Firstly, thank you for your contribution. I see that the certbot script is in Python, of which I'm not an expert. Even so, I don't personally see any functionality that couldn't be reproduced with Java in a way more native to our codebase. Am I incorrect in this assessment?

Kind regards,

Matt

lprimak:
I am going to review this code this weekend/next week... sorry for the delay I was busy with other tasks

ratcashdev:
@lprimak no worries, take your time.

MattGill98:
Hi @ratcashdev,

Thank you for your response. I disagree with the precedence argument, since I think every case should be reviewed on it's own merit. However, your first argument is quite compelling. This in mind, I'm quite happy adding this python script if others seem to agree (which they seem to)!

Kind regards,

Matt

lprimak:
Also, it's assuming domain1 why not ```payara``` domain.  Also, I am using separate domain directory outside Payara distribution, using --domaindir options to asadmin.  This needs to be supported

ratcashdev:
@lprimak Thank you for the review and testing. Will have a look on this. So my understanding is that, the following changes are required:
1. add the --domain-dir option. 
2. add the --domain-name option
3. check for certbot's availability
4.(Optional) echo back the comands run. This could be enabled with a --debug param, so that users will see which command went wrong.

lprimak:
@ratcashdev that's what I came up with so far.
Thank you!

ratcashdev:
You're right. As a matter of fact you can specify multiple domain names. I've written it too long ago to remember.

ratcashdev:
Updated code. Start by `./letsencrypt.py -h`, it will print the usage and arguments. Speaking about which, let me point out that it is necessary to distinguish between fully-qualified-domain-names (needed by certbot for the subject-alternative-names in the certificate) and payara domains (needed by asadmin).
except from the help message:
```
  -n DOMAIN_NAME, --domain-name DOMAIN_NAME
                        The FQDN of the domain(s) the certificate will be
                        bound too. You may use this arg multiple times.
  -p PAYARA_DOMAIN, --payara-domain PAYARA_DOMAIN
                        The name of the payara-domain where the certificate
                        will be uploaded.
  -d DOMAIN_DIR, --domain-dir DOMAIN_DIR
                        The directory where payara domains are defined.
                        Necessary to provide only when the domains are in non-
                        standard locations.
```

lprimak:
jenkins test

lprimak:
Can you please rename --domain-dir option to --domaindir to be consistent with Payara? Thanks

lprimak:
Also, it's really easy to get confused about what the difference is between -p and -n arguments.
I know what your intent is, but it needs to be renamed.
```-n -> -c | --cert-domain```
```-p -> -n | --name```

**Comment by lprimak** [comment]
example:

```
letsencrypt.py -d /Users/lprimak/domains -n payara-domain -c www.payara.fish
```

**Comment by ratcashdev** [comment]
1. `payara_le_war`: OK, will be put under /tmp and cleaned up afterwards.
2. `-c` and `-p` also fine. I am actually happy for your idea of cert-domain. I was searching for a better name for quite some time. On the other hand, I would like to be consistent. If c stands for Cert-domain, p should stand for Payara-domain, and not n (for `name` wich is not self-descriptory).
3. finally, about domaindir: Rather than introducing a new piece of code with non-standardized argument naming (see GNU coding standards for long argument naming here: http://www.gnu.org/prep/standards/) I would prefer payara cleaning up the historical mess of glassfish (e.g. by introducing the `domain-dir` into payara and keep domaindir in parallel)

In any case, please let me know the results of your reconsideration, and I will do as proposed.

**Comment by lprimak** [comment]
I hear you on the --domain-dir but the letsencrypt script is not "worthy enough" to change Payara because all the customer's scripts will break

**Comment by ratcashdev** [comment]
You're right. I am discussing a single 'dash' here, and it's really not worth it (even though customer scripts would not break, because, as I wrote, asadmin would keep --domaindir also in parallel for backward compatibility). So let's skip No3 and agree on `--domaindir`.

What about No2 from above?

**Comment by lprimak** [comment]
I like —name to be consistent with payara asadmin commands still

**Comment by lprimak** [comment]
I just want it to be consistent with the existing payara asadmin command to relieve users' cognitive load when learning a new command such as letsencrypt.

**Comment by ratcashdev** [comment]
@lprimak  would this be ok? Allowing both domaind-dir and domaindir.
```
  -c CERT_DOMAIN, --cert-domain CERT_DOMAIN
  -n NAME, --name NAME
  -d DOMAIN_DIR, --domain-dir DOMAIN_DIR, --domaindir DOMAIN_DIR
```

lprimak:
sure thing

ratcashdev:
@lprimak done with the changes. Please have a look.

ratcashdev:
It's important to note, that, by default, production domain also runs on port 8080. Thus, without some port mapping and configuration change in domain.xml the validation according to ACME (https://ietf-wg-acme.github.io/acme/draft-ietf-acme-acme.html see chapter 8.3, HTTP Challenge) will likely fail.

smillidge:
Please remove Do Not Merge label when this is agreed

lprimak:
> It's important to note, that, by default, production domain also runs on port 8080. 
Yes, I did see this.  Alternatively there can be some port mapping / assignment done by the router.
It would be pretty easy / self-explanatory for the user to get done.  I will write some docs / blog accompanying this script as well and there can be something there explaining this

lprimak:
jenkins test

lprimak:
Looks great!

lprimak:
Jenkins test

ratcashdev:
Hi @lprimak,
I made one last fix. Sorry about the  confusion. I was considering the idea of temporarily changing listener-1's port to 80 (and then back to 8080 or whatever it was), but decided against it. That is to say, you're free to merge. There are no changes planned for the foreseeable future. And thanks for you patience, cooperation and help.

lprimak:
Oh, no problem.  Happy for the contribution thanks!

lprimak:
jenkins test

johnmanko:
It looks like the use case for this script is only for new non-Docker production installations.  Perhaps it can be extended for use with Docker where the certs would be generated outside the container.

ratcashdev:
@johnmanko importing existing certs is already possible - this script already relies on that possibility.
See the `asadmin add-pkcs8` command and its parameters. Also see the `restart_listener` method inside the script for an example how to restart a listener (and activating the newly imported keypair) in a running app-server.

PR #2731, opened 2018-05-17
PAYARA-2742 Errors adding a metric don't cause metrics service to cra…

Cousjava:
…sh


Cousjava:

PR #2740, opened 2018-05-18
PAYARA-2670 JMX monitoring logging no longer logs when disabled

Cousjava:
* PAYARA-2670 stop jmx-monitoring sevice sending notifications when not enabled

* PAYARA-2670 JMX Logging no longer still logs when disabled


Cousjava:

PR #2744, opened 2018-05-18
PAYARA-2729 When listing batch job executions using "--long" options,  it causes creation of excessive JDBC connections

MeroRai:


Cousjava:

PR #2748, opened 2018-05-21
Changed printing to stdout to logging with the JDK logger.

OndrejM:
This partially fixes PAYARA-2730 by allowing disabling the annoying log message


OndrejM:
There are some more `System.out.print` in:
* EjbContainerListener
* BtraceClientGenerator

OndrejM:
I realized that this isn't related to PAYARA-2730 so this is of low priority but it would be nice to reduce logging with `sys.out` anyway.

arjantijms:
Jenkins test

PR #2796, opened 2018-05-30
PAYARA-2779 fix NPE regression in concurrent

### LINKED ISSUE #2794: Payara 5 fix for PAYARA-2346 leads to NPEs when starting managed threads
## Description

Payara 5 fix for PAYARA-2346 leads to NPEs

Suspecting commit for #2218 

## Expected Outcome

When running a managed thread-factory I expect managed threads to execute Runnnables.

I defined a 
@Resource(lookup = "myThreadFactory")
ManagedThreadFactory myThreadFactory;

and MyThreadPoolExecutor extends ThreadPoolExecutor

When calling myThreadPoolExecutor.execute(myRunnable) the run method should be performed.

## Current Outcome

myThreadPoolExecutor.execute(myRunnable) is called and a NullPointerException is seen at ContextSetupProviderImpl line 174, before my debugger enters the Runnable run method.

I cannot easily provide a test application, but I do see the following:

See commit https://github.com/payara/Payara/commit/e11ebf3af0cc323587631b5b790ab7b99a8a5e9f

Line 170 of ContextSetupProviderImpl states

`if (handle.getInvocation() != null) {`

Line 174 adds:

`&& handle.getInvocation().getJNDIEnvironment() != null`

without the same null check.

`handle.getInvocation() `

can still be null, the result is a NPE. NPE occurs on line 174 in my case when a managed threadfactory starts a new thread. The run method in my Runnable is not called, I see this NPE occur.

Even if this is fixed the new code at line 179 will fail:

`if(handle.getInvocation().getComponentId() != null && compEnvMgr != null) {`

because `handle.getInvocation() ` is still null

## Stacktrace

This is the most of an exception I could reproduce:

```
	at org.glassfish.concurrent.runtime.ContextSetupProviderImpl.setup(ContextSetupProviderImpl.java:174)
	at org.glassfish.enterprise.concurrent.internal.ContextProxyInvocationHandler.invoke(ContextProxyInvocationHandler.java:94)
	at com.sun.proxy.$Proxy718.run(Unknown Source)
	at com.myRunnable....
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
	at org.glassfish.enterprise.concurrent.ManagedThreadFactoryImpl$ManagedThread.run(ManagedThreadFactoryImpl.java:244)
```

## Environment ##

- **Payara Version**: 5.181
- **Edition**: Full
- **JDK Version**:8 u102 - Oracle

lprimak:
moved code that uses handle.getInvocation() into the if() block that _does_ a null check for the former
fixes #2794

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -168,24 +168,24 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         InvocationContext handle = (InvocationContext) contextHandle;
         String appName = null;
 
+        ClassLoader backupClassLoader = null;
         if (handle.getInvocation() != null) {
             appName = handle.getInvocation().getAppName();
-        }
-        if(appName == null && handle.getInvocation().getJNDIEnvironment() != null) {
-            appName = DOLUtils.getApplicationFromEnv((JndiNameEnvironment)handle.getInvocation().getJNDIEnvironment()).getName();
-        }
-        ClassLoader backupClassLoader = null;
-        if (appName == null) {
-            // try to get environment from component ID
-            if(handle.getInvocation().getComponentId() != null && compEnvMgr != null) {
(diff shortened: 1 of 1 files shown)


lprimak:
jenkins test please

smillidge:
Fixes #2794

PR #2800, opened 2018-05-30
PAYARA-2770 Rest resource and Rest Client on same classpath cause ambiguous rest endpoint error

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2813, opened 2018-06-01
PAYARA-2788 Changing port now works for an non-admin http listener fo…

Cousjava:
…r the set-network-listener command


Cousjava:
Jenkins test please

PR #2823, opened 2018-06-07
PAYARA-2826 Exception is no longer thrown in AMXJ2EEStartupService wh…

Cousjava:
…en creating a new instance


Cousjava:
Jenkins test please

PR #2830, opened 2018-06-11
PAYARA-2775 MP Metrics support for @Stateless JAX-RS resource

jGauravGupta:
**Root cause :**
- Bean<?> injection not supported in JCDIService
- @Inject and @Intercepted annotation not added to constructor of EJB serializable interceptor created by ASM class writer


jGauravGupta:
jenkins test please

PR #2839, opened 2018-06-13
#2838 Calculate const values using doubles

### LINKED ISSUE #2838: PrometheusExporter for MP metrics conversion error
# Description #
----------

When exporting metrics of a rest method annotated with `@Timed`, some of the values are zero because of a conversion bug.

These conversion values:
```
    private static final double NANOSECOND_CONVERSION = 1 / 1_000_000_000;
    private static final double MICROSECOND_CONVERSION = 1 / 1_000_000;
    private static final double MILLISECOND_CONVERSION = 1 / 1_000;
```
all convert to zero and therefore `@Timed` quantiles are 0. 
The conversion to double is done after the integer division so all 3 conversion values are `0.0d`.

## Expected Outcome

Correct conversion is done for the timer metrics samples exported in prometheus format.

## Current Outcome

All reported quantiles are 0.

Create a rest resource with a `@Timed` annotated method and call that method a few times.
Then observe /metrics in prometheus format.

### Samples

```
application:de_llorenzen_resource_accept_seconds{quantile="0.5"} 0.0
application:de_llorenzen_resource_accept_seconds{quantile="0.75"} 0.0
application:de_llorenzen_resource_accept_seconds{quantile="0.95"} 0.0
application:de_llorenzen_resource_accept_seconds{quantile="0.98"} 0.0
application:de_llorenzen_resource_accept_seconds{quantile="0.99"} 0.0
application:de_llorenzen_resource_accept_seconds{quantile="0.999"} 0.0
```

## Context (Optional)

`fish.payara.microprofile.metrics.writer.PrometheusExporter`

## Environment ##

- **Payara Version**: 5.181, 5.182
- **Edition**: All
- **JDK Version**: 8 Oracle
- **Operating System**: Linux / Mac

svendiedrichsen:
Fixes #2838

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/writer/PrometheusExporter.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/writer/PrometheusExporter.java
@@ -1,8 +1,8 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
- * 
+ *
  *    Copyright (c) [2018] Payara Foundation and/or its affiliates. All rights reserved.
- * 
+ *
  *     The contents of this file are subject to the terms of either the GNU
  *     General Public License Version 2 only ("GPL") or the Common Development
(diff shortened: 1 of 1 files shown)


smillidge:
Jenkins test please

PR #2847, opened 2018-06-15
PAYARA-2827 Add empty java config to prevent null pointer

Pandrex247:


Pandrex247:
Embedded uses a different domain.xml, and that already has some entries in.

Pandrex247:
Jenkins test please

PR #2869, opened 2018-06-20
PAYARA-2835 local SSL remote EJB call fallback property

lprimak:
added system property fish.payara.orb.disable-ssl-check to make remote EJB calls accessible in development


lprimak:
jenkins test

PR #2879, opened 2018-06-21
PAYARA-2828 - Add rotation on file fize options on Access logging Payara4

MeroRai:
PAYARA-2828 Add rotation on file fize options on Access logging


MeroRai:

PR #2883, opened 2018-06-21
PAYARA-2874 Enabling Security Manager Causes Log Errors

MattGill98:
This fix stops the NullPointerExceptions, and I don't see any obvious problem with the contextID being null in the first place. Needs a review to make sure this is the case.


MattGill98:
jenkins test please

PR #2910, opened 2018-06-28
PAYARA-2881 Upgrade EclipseLink to 2.7.2-p1

### LINKED ISSUE #2608: Recursive annotations cannot be deployed with JPA on Payara 5
# Description #
----------

When an annotation class annotated with itself gets scanned by EclipseLink, it enters into a recursive loop which triggers a StackOverflowError and fails the deployment. The same was working correctly on version 4.

## Expected Outcome

Successful deployment

## Current Outcome

A failed deployment with a log like in
https://javaee.groups.io/g/glassfish/topic/6519037
https://pastebin.com/61trs1kj

## Steps to reproduce

Compile only these 3 files into a WAR and deploy in payara micro:

The culprit:
```
@Self
public @interface Self {}
```

The enabler:
```
<?xml version="1.0" encoding="UTF-8"?>
<persistence version="2.1" xmlns="http://xmlns.jcp.org/xml/ns/persistence">
        <persistence-unit name="persistence" transaction-type="JTA">
                <exclude-unlisted-classes>false</exclude-unlisted-classes>
        </persistence-unit>
</persistence>
```

The trigger:
```
@Singleton
public class Bean {
    @PersistenceContext
    private EntityManager em;
}
```

### Samples

[scce.zip](https://github.com/payara/Payara/files/1889163/scce.zip)

## Context

I encountered the issue in a Kotlin project where `kotlin.annotation.Target` is annotated with itself.

## Environment ##

- **Payara Version**: 5.181 and 5.182-SNAPSHOT
- **Edition**: Micro
- **JDK Version**: openjdk version "1.8.0_162"
- **Operating System**: Fedora release 26
- **Database**: default embedded H2 datasource

lprimak:
fixes #2608

### DIFF
--- a/appserver/pom.xml
+++ b/appserver/pom.xml
@@ -107,7 +107,7 @@
 
         <!-- JPA -->
         <javax-persistence-api.version>2.2.0</javax-persistence-api.version>
-        <eclipselink.version>2.7.1.payara-p1</eclipselink.version>
+        <eclipselink.version>2.7.2.payara-p1</eclipselink.version>
 
         <!-- JTA -->
         <javax.transaction-api.version>1.2.1</javax.transaction-api.version>


lprimak:
jenkins test

PR #2912, opened 2018-07-01
PAYARA-2617 update grizzly

lprimak:


lprimak:
jenkins test

lprimak:
They have the version at 2.4.4-SNAPSHOT so it seems more intuitive to have 2.4.4

MattGill98:
I personally think that it should be a new 2.4.3 patch version since it's built from version 2.4.3. You're right that it's tagged so it's just a semantic argument, but regardless I'd rather see it changed. Could we get a third opinion?

lprimak:
Sorry, I still don't see how it's 2.4.3 when it's tagged 2.4.4-SNAPSHOT?

MattGill98:
No that's fine, I'm not always the best at explaining! I just mean since `2.4.4` hasn't been released, a `2.4.4.patch-version` would imply that it has. By making the version `2.4.3.patch-version`, we're essentially making our own 'full release' built from the version `2.4.3`. The same way that Payara having version `5.183-SNAPSHOT` means that the version is `5.182` (with some `SNAPSHOT` changes on top).

lprimak:
This is included with other @MattGill98 Grizzly chages

PR #2925, opened 2018-07-03
PAYARA-2897 Moved generic JASPIC code to provider project

arjantijms:
Refactored client code and improved some naming


arjantijms:
Jenkins test please

appserver/security/jaspic-provider-framework/src/main/java/com/sun/jaspic/services/RegistrationWrapperRemover.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 1997-2012 Oracle and/or its affiliates. All rights reserved.
```
MattGill98: Newly added classes shouldn't have an oracle header, unless it's been moved from somewhere?

arjantijms: It's "sort of" moved. The method is extracted from an existing Oracle copyrighted interface into its own interface. I think the argument can go both ways, it's both a moved method and a new class. I meditated over this for a while, but in the end I decided it was more about a method being moved, and less about a class being created from scratch, hence the Oracle header.

MattGill98: That's fair enough then, I just wanted to make sure it wasn't something forgotten about!

PR #2928, opened 2018-07-04
PAYARA-2878 Applications aren't Deployed to an Instance with Autobinding Enabled

jGauravGupta:


jGauravGupta:
jenkins test please

PR #2937, opened 2018-07-06
PAYARA-2733 Add reasonable default timeouts for Jersey client

jGauravGupta:


jGauravGupta:
jenkins test please

appserver/admin/production_domain_template/src/main/resources/config/domain.xml
```diff
@@ -243,6 +243,8 @@
         <thread-pool name="admin-thread-pool" max-thread-pool-size="15" min-thread-pool-size="1" max-queue-size="256"></thread-pool>
       </thread-pools>
       <system-property name="fish.payara.classloading.delegate" value="false"/>
+      <system-property name="jersey.config.client.readTimeout" value="60000"/>
```
OndrejM: I don't see what was changed in the Jersey patch - are these system properties really necessary to turn on the timeout? These timeouts should be the default if no system property specified so these system properties shouldn't be necessary in domain.xml.

OndrejM: I mean, Payara Server should set the system properties to default values automatically if they're not already set. The solution shouldn't rely on setting the properties in the domain.xml

jGauravGupta: Jersey Client patch is changed [1] to provide the MP Config support. As Jersey Client support many other properties like [2] : 
- jersey.config.client.async.threadPoolSize"
- jersey.config.client.backgroundScheduler.threadPoolSize

So developer can also update the configuration using MP Config way. I agree moving these config from global properties to network-config (or application level) make more sense for first class jersey support .

Why domain.xml : 
These configuration are applied  only on production/micro domain and skipped the domain1. As domain1 is intended for development which includes the debugging with indefinite timeout.


[1]  :  https://github.com/payara/patched-src-jersey/pull/1/files
[2]  :  https://github.com/eclipse-ee4j/jersey/blob/master/core-client/src/main/java/org/glassfish/jersey/client/ClientProperties.java#L64

OndrejM: > Why domain.xml :
Ok, I accept.

jGauravGupta:
Jenkins test please

PR #2965, opened 2018-07-17
PAYARA-2871 Improve listing batch jobs memory usage in admin console

MeroRai:


Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

appserver/admingui/full/src/main/java/fish/payara/full/admingui/handlers/BatchHandlers.java
```diff
@@ -0,0 +1,163 @@
...
+import com.sun.jsftemplating.layout.descriptors.handler.HandlerContext;
+import org.glassfish.admingui.common.util.GuiUtil;
+
+/**
+ *
+ * @author Susan Rai
+ */
+public class BatchHandlers {
+
+    private static final int DEFAULT_OFFSET_VALUE = 20;
```
MattGill98: Small thing, but for readability can this be renamed to `DEFAULT_OFFSET_INCREMENT` or something like that? It doesn't function as a default offset, but as an increment value

appserver/admingui/full/src/main/resources/batch/assets/js/batch.js
```diff
@@ -0,0 +1,53 @@
...
+ * recipient has the option to distribute your version of this file under
+ * either the CDDL, the GPL Version 2 or to extend the choice of license to
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+
+<f:verbatim>
+    <script type="text/javascript">
```
MattGill98: This is an unusual way of adding CSS. Usually this would be done by using CSS selectors carefully, is this not possible here?

appserver/batch/glassfish-batch-commands/src/main/java/org/glassfish/batch/ListBatchJobsProxy.java
```diff
@@ -79,24 +79,39 @@
     @Param(primary = true, optional = true)
     String jobName;
 
+    @Param(name = "offset", optional = true, defaultValue = "0")
+    String offSetValue;
```
MattGill98: Make sure to add validation to these to prevent negative errors etc! May require adding to both the proxy and the main command. Will probably need a rework sometime in the future, but I would just validate both for now.

appserver/admingui/full/src/main/resources/batch/assets/js/batch.js
MeroRai: I have looked at different ways to do it, but IMO this is the cleanest way and makes it easily maintainable too.

MeroRai:
Jenkins test please

PR #2974, opened 2018-07-20
PAYARA-2923 Fix OpenTracing IllegalStateException

Pandrex247:
Fixes an IllegalStateException that can occur when a JAX-RS method of an app with no CDI beans is called with request tracing enabled.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
+++ b/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
@@ -112,6 +112,8 @@ public class JaxrsContainerRequestTracingFilter implements ContainerRequestFilte
                 beanManager = CDI.current().getBeanManager();
             } catch (IllegalStateException ise) {
                 // *Should* only get here if CDI hasn't been initialised, indicating that the app isn't using it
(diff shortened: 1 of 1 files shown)


appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
```diff
@@ -153,9 +166,21 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
-                // Get the traced annotation from the method
-                Traced tracedAnnotation = OpenTracingCdiUtils.getAnnotation(CDI.current().getBeanManager(),
-                        Traced.class, resourceInfo);
+                // Check if CDI has been initialised by trying to get the BeanManager
+                BeanManager beanManager = null;
+                try {
+                    beanManager = CDI.current().getBeanManager();
+                } catch (IllegalStateException ise) {
+                    // *Should* only get here if CDI hasn't been initialised, indicating that the app isn't using it
+                    logger.log(Level.FINE, "Error getting Bean Manager, presumably due to this application not using CDI",
```
MattGill98: Is this log message not going to get logged twice?

Pandrex247: Yes, but it is possible for the two filter methods to execute on separate threads.

Pandrex247:
Jenkins test please

PR #2981, opened 2018-07-23
PAYARA-2926 Cookie session persistence type does not throw exception after maxSessions reached (Payara4)

MarkWareham:


MarkWareham:
Jenkins quick test Payara 4 please

PR #2985, opened 2018-07-24
PAYARA-1714 Added command to generate bash autocomplete

Cousjava:
This creates a file that when run adds glassfish/bin to the path and creates autocompletion of asadmin commands


Cousjava:
Jenkins test please

Cousjava:
@MattGill98 Because this command requires the server to be running

PR #2987, opened 2018-07-24
QA-4 Set up sonar PR analysis

MarkWareham:
The intention with this is to have local Jenkins run a job to analyse the PR and decorate the GitHub PR. This file is the Jenkins pipeline (a description of what the job should do).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ pipeline {
     stages {
         stage("Analyse") {
             agent {
-                label "sonarqube"
+                label "sonar"
             }
             tools {
                 jdk "zulu-8"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,60 @@
+pipeline {
+    agent none
+    stages {
+        stage("Analyse") {
+            agent {
+                label "sonarqube"
+            }
+            tools {
+                jdk "zulu-8"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,60 @@
...
+            }
+            steps {
+                echo "Analysing"
+                checkoutAndBuildSource()
+            }
+        }
+    }
+   post {
+        always {
+            hipchatSend (color: 'GRAY', credentialId: 'hipchat2', textFormat: false, message: "${env.JOB_NAME} build # ${env.BUILD_NUMBER} Completed. See <a href='${BUILD_URL}'>Jenkins</a>")
```
MattGill98: I'm not sure what's in the job name, but will this message contain any JIRA specific reference? I.e. the PR name or something?

MarkWareham: Good point, that would be handy, but then I don't think the hipchat message is even necessary for PR analysis. It would get annoying quickly. Maybe in the future it could be set up to message the PR author.

MarkWareham:
Accidentally deleted branch.

MarkWareham:
I explained what it does in the first comment. The hipchat block was just an extra afterwards.

MarkWareham:
Jenkins test please

Issue #2986, opened 2018-07-24
Microprofile Config fills missing required String property with a default value


IrminAuwerda:
# Description #
----------

In our project we utilize the microprofile-config part of the Microprofile. The goal is to have the application service fail startup if required properties are missing.

This works for properties of any class except String properties. If a String property is missing the application service starts up normally and fills the property with `org.eclipse.microprofile.config.configproperty.unconfigureddvalue`.

Example in code:

```
    @Inject
    @ConfigProperty(name = "pixelware.user")
    private String user;
```
When not provided the field `user` will have the value `org.eclipse.microprofile.config.configproperty.unconfigureddvalue`.

## Expected Outcome

The expected outcome is that the Payara server will notice the config property is missing and give an error and startup and shutdown.

## Current Outcome

The current outcome is that the Payara server starts up normally and the field `user` has the value `org.eclipse.microprofile.config.configproperty.unconfigureddvalue`.

## Steps to reproduce (Only for bug reports) 

Create a class in which you define a property as a String property. Inject it with ConfigProperty.
Start the domain and print out the property after startup.
The result is that the property has the value `org.eclipse.microprofile.config.configproperty.unconfigureddvalue`.

### Samples

We found the problem(s) in Payara:

1. Payara version 4.1.2.181 is missing a StringConverter completely.
2. Payara version 5.182 has a StringConverter:
```
public class StringConverter implements Converter<String>{

    @Override
    public String convert(String value) {
        if (value == null || value.equals(ConfigProperty.UNCONFIGURED_VALUE)) return null;
        return value;
    }   
}
```
But in the class PayaraConfig.java there is this piece of code that will prevent it from ever being executed:
```
private <T> T convertString(String value, Class<T>  propertyType) {
        if (String.class.equals(propertyType)) {
            return (T) value;
        }
         ......
```

### Solution

Remove the specific check on the String class from the convertString(...) method in the PayaraConfig.java:
```
private <T> T convertString(String value, Class<T>  propertyType) {
        if (String.class.equals(propertyType)) { //REMOVE
            return (T) value;           //REMOVE
        }                                 //REMOVE
         ......
```

## Context (Optional)

Now we have to create a "ConfigString" class for String properties and include our own converter to make it use. It is also highly unexpected behaviour since the Microprofile Config spec states specifically that it should fail on startup on a missing required property.

## Environment ##

- **Payara Version**: 4.1.2.181 and 5.182 (tested both)
- **Edition**: Micro
[...]

PR #2989, opened 2018-07-25
Fix for #2986 Missing String ConfigProperty should cause DeploymentException

IrminAuwerda:
When the ConfigProperty is used with Injection the PayaraConfig is always called with a default value (org.eclipse.microprofile.config.configproperty.unconfigureddvalue).

The logic to notice that no property is found requires the StringConverter to throw the DeploymentException. The cast to String and return before Converters are determined prevents this.

Removed this line of code.

Fixes issue #2986.


mulderbaba:
Jenkins test please

PR #2994, opened 2018-07-27
PAYARA-2834 OpenID Connect support

jGauravGupta:


mulderbaba:
Jenkins test please

jGauravGupta:
Jenkins test please

mulderbaba:
Jenkins test please

jGauravGupta:
Jenkins test please

jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
```diff
@@ -190,24 +187,28 @@ public OpenIdAuthenticationMechanism setConfiguration(OpenIdAuthenticationDefini
...
-            if (nonNull(recievedState) && recievedState.equals(state.getState())) {
-                // (3) Successful Authentication Response : redirect_uri?code=abc&state=123
-                return validateAuthorizationCode(request, httpMessageContext);
-            } else if (nonNull(recievedState) && !recievedState.equals(state.getState())) {
-                LOGGER.fine("Inconsistent recieved state, value not matched");
-                return httpMessageContext.notifyContainerAboutLogin(NOT_VALIDATED_RESULT);
+            OpenIdState expectedState = stateController.get(configuration, httpContext);
+
+            if (nonNull(recievedState) && nonNull(expectedState)) {
+                if (expectedState.equals(recievedState)) {
```
sgflt: I think [Sonar](https://sonarcloud.io/project/issues?id=org.glassfish.main%3Aglassfish-main-aggregator&open=AWoN7ILp9zLXaOU8UM1S&resolved=false&sinceLeakPeriod=true&types=BUG) is right about this comparison String.equals(OpenIdState).

svendiedrichsen: I don't think so. Please look at the OpenIdState.equals implementation.

sgflt: You are right. False positive in this case, but the contract of equals is broken.

Also [javadoc for Object](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#equals-java.lang.Object-) defines a symmetric relation.

> It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if and only if y.equals(x) returns true.

This specific implementation of equals possibly opens a space for mistakes when equals would be called on String with OpenIdState as a parameter.

Issue #3007, opened 2018-07-30
PayaraMicroDeployableContainer should allow to select the java program


dmatej:
It always uses a java program from the operationg system, no respect to javaHome or AS_JAVA or JAVA_HOME or some configuration setting. But most systems these days have already the JDK10 as default, that's why these tests fail: https://github.com/javaee-samples/javaee8-samples/issues/20#issuecomment-408889621

https://github.com/payara/Payara/blob/43cf0c68d2d38fa6f308fd1c62fa9c3525f5aaa2/appserver/extras/arquillian-containers/payara-micro/src/main/java/fish/payara/arquillian/container/payara/managed/PayaraMicroDeployableContainer.java#L165-L168

MattGill98:
Hi @dmatej,

Thanks for raising this issue. I've raised internal issue `PAYARA-2953` to fix this.

Kind regards,

Matt

dmatej:
It works! :-)
Would you release a version, please?

MattGill98:
Hi @dmatej,

A new version has been released here: https://github.com/payara/Payara_PatchedProjects/pull/179. The new version should be now accessible from the Java EE 8 Samples suite. I'll close this issue, although feel free to reopen it if you have any queries.

Hope this helps!

Kind regards,

Matt

PR #3009, opened 2018-07-31
PAYARA-2953 Payara Micro Arquillian Container Doesn't Use Java Home

MattGill98:
Now selects the Java version used to run the Arquillian container. Addresses https://github.com/payara/Payara/issues/3007.


mulderbaba:
Jenkins test please

PR #3031, opened 2018-08-09
PAYARA-2924 NPE on JAX-RS Async-Server EE7 Sample

Pandrex247:
If request tracing is turned on when running this sample, a NPE is thrown.
This is caused by the return happening on a separate thread, so the container filter gets a NPE when trying to get the access active span (as these are thread local).

This PR fixes this by saving the span context as an attribute of the InvocationContext object created by managed pools so that the span context can be propagated when setting up the thread context.

Threads from managed pools created external to Payara will not be traced however, such as those created by Jersey to handle timeouts.


Pandrex247:
This also adds the ability for OpenTracing to create traces itself (rather than requiring a trace to already be running), and includes a couple bits of cleanup

Pandrex247:
Error in TCK, will recreate once resolved.

PR #3048, opened 2018-08-15
PAYARA-2931 Resolved memory leak

Cousjava:
from sessions being stored in memory with file persistence


Cousjava:
Jenkins test please

PR #2980, opened 2018-07-23
PAYARA-2926 Cookie session persistence type does not throw exception …

MarkWareham:
…after maxSessions reached

PR #3059, opened 2018-08-18
PAYARA-2851 - Payara4

smillidge:
MicroProfile Config 1.3 for Payara 4.x passes MP Config 1.3 TCK


smillidge:
jenkins quick test payara 4

smillidge:
#3048 #2980 #2874 are all related so this will need a retest after 183 release.

PR #3060, opened 2018-08-18
PAYARA-3027 py.resolveDynamicPort needs to take into account that the…

smillidge:
… config may not be the DAS config

Signed-off-by: Steve Millidge <steve.millidge@payara.fish>


smillidge:
jenkins test please

PR #3066, opened 2018-08-20
PAYARA-2871 Improve listing batch jobs memory usage in admin console - Payara4

MeroRai:


MeroRai:
jenkins quick test payara 4 please

PR #3079, opened 2018-08-22
PAYARA-3036 Undo package name changes for Jersey internal classes

jGauravGupta:


jGauravGupta:
jenkins quick test payara 4 please

PR #3108, opened 2018-08-31
PAYARA-3078 Removed unused security-dol module

Cousjava:
This module is never built or used by Payara


Cousjava:
Jenkins test please

Pandrex247:
Does this have a Jira number?

Issue #3110, opened 2018-08-31
ConcurrentModificationException when Slow SQL Logger is enabled


silvafabio:
# Description #
----------

When "Slow Query Log Threshold" is enabled (in my case 3 seconds) on connection pool, all queries that exceeds the threshold throws an java.util.ConcurrentModificationException on server.log. When is disabled the exceptions in server.log go away. 

## Expected Outcome

Section "Example Trace" at https://docs.payara.fish/documentation/payara-server/advanced-jdbc/slow-sql-logger.html

## Current Outcome

Stack Trace:
[...]

PR #3114, opened 2018-09-02
PAYARA-3059 don't deregister slow sql logger on every call and make l…

smillidge:
…isteners collection a concurrent hashmap


smillidge:
Likely fixes #3110

jGauravGupta:
jenkins test please

PR #3126, opened 2018-09-06
PAYARA-3075 Tie Bulkheads and CircuitBreakers to Bean instances

Pandrex247:
Previously these were tied to the application, so would be shared across all bean instances.


smillidge:
jenkins test please

PR #3130, opened 2018-09-06
PAYARA-3078 Removed unused security-dol module (payara4)

Cousjava:
This module is never built or used by Payara


Cousjava:

PR #3141, opened 2018-09-10
PAYARA-2912 JBatch not working with PostgreSQL

Cousjava:


Cousjava:
Jenkins test please

appserver/batch/glassfish-batch-commands/src/main/java/org/glassfish/batch/ListBatchJobs.java
```diff
@@ -521,8 +521,8 @@ private boolean isSimpleMode() {
         String columnID = "jobinstanceid";
         try (Connection connection = dataSource.getConnection()) {
             try (PreparedStatement preparedStatement = connection.prepareStatement(query)) {
-                preparedStatement.setString(1, parameter1);
-                preparedStatement.setString(2, parameter2);
+                preparedStatement.setInt(1, Integer.parseInt(parameter1));
```
smillidge: You will need to test this change with other database servers

Cousjava: Tested with PostgreSQL, Derby and MySQL

MeroRai: I tested it with Derby

Cousjava: Also tested with H2

Pandrex247: Oracle DB is the last big'un to test against then

PR #3142, opened 2018-09-10
PAYARA-2975 'metrics.xml' parsed and metadata re-registered on every call to '/metrics' endpoint

jGauravGupta:


jGauravGupta:
jenkins test please

jGauravGupta:
Hi @Cousjava,

The lazy loaded mbeans are now stored in `unresolvedBaseMetadataList` & `unresolvedVendorMetadataList` for further evaluation so it will not be ignored.

Cousjava:
@jGauravGupta But they will only be resolved to one MBean, in cases where they are for multiple when it has a wildcard it will only resolve the first one. I have tested this.

jGauravGupta:
jenkins test please

PR #3153, opened 2018-09-12
PAYARA-3059 Don't deregister slow sql logger on every call and make listeners collection a concurrent hashmap - Payara4

Pandrex247:
PAYARA-3059 don't deregister slow sql logger on every call and make l…


Pandrex247:

Pandrex247:

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/util/SQLTraceDelegator.java
```diff
@@ -101,7 +101,7 @@ public void setModuleName(String moduleName) {
      */
     public void registerSQLTraceListener(SQLTraceListener listener) {
         if (sqlTraceListeners == null) {
-            sqlTraceListeners = new HashMap<>();
+            sqlTraceListeners = new ConcurrentHashMap<>();
```
sgflt: Shouldn't be also type of variable changed to interface ConcurrentMap to enforce implementations that support parallelism?

PR #3175, opened 2018-09-18
PAYARA-3053 Created JDK11 domain template

Cousjava:
This does not include any JDK9+ specific options


mulderbaba:
Jenkins test please

PR #3177, opened 2018-09-18
PAYARA-3109 OpenAPI doesn't support if schema type is array

MeroRai:


MeroRai:
Jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -283,7 +283,7 @@ public static SchemaType getSchemaType(Class<?> type) {
                 || type == float.class || type == Float.class || type == double.class || type == Double.class) {
             return SchemaType.NUMBER;
         }
-        if (type.isArray()) {
+        if (type.isArray() || Iterable.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type)) {
```
MattGill98: That second addition is redundant, Collections are already iterable

MeroRai:
@MattGill98, It passes the TCK

PR #3190, opened 2018-09-23
PAYARA-2977 The resource type of a JDBC connection pool can't be changed in the second step of creation

Ykus:
Fixed problem during creation a JDBC connection pool, when resource type was not specified and the value was empty string instead of null.


Cousjava:
Jenkins test please

PR #3191, opened 2018-09-24
PAYARA-3117 made deregistring non-existent thread message fine

Cousjava:


Cousjava:
Jenkins test please

smillidge:
Can you try and work out why there is a double message in the first place?

Cousjava:
This is due to the thread being registered twice if there is a contexthandle generating a proxy, but there is no way at https://github.com/payara/Payara/blob/727a16105d1cd7c4dea3c7e86d9cd09edbc44000/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java#L240 to find out if the Runnable is being proxied at that point

smillidge:
Have you reviewed the stack at each registration in a debugger to understand if this can be prevented for the case when there is a proxy?

Cousjava:
@smillidge Yes, I said in my previous comment there is no way to tell in the code at the (un)registration point if it is a proxy or not.

PR #3193, opened 2018-09-25
PAYARA-3125 Add lowest possible priority to top-level exception mapper

Pandrex247:
This should help prevent clashes with other top-level exception mappers as it will defer to using the other (assuming it has a higher priority).

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingExceptionMapper.java
+++ b/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingExceptionMapper.java
@@ -39,6 +39,7 @@
  */
 package fish.payara.microprofile.opentracing.jaxrs;
 
+import javax.annotation.Priority;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.StatusType;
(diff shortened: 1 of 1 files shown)


Pandrex247:
From a black box POV the ordering seemed to work for me - either that or I was tremendously lucky

MattGill98:
I think you may have been lucky, I seem to get unpredictable behaviour. Having said that, it seems to be heavily weighted in the favour of the custom mapper. Would someone be able to verify this?

Pandrex247:
@MattGill98 I ran a test where I redeployed and forced an exception on an example application which contained an exception mapper of throwable against this branch for over an hour (the redeploy and poke would take ~2-3 seconds, so roughly 1200 requests) - I didn't hit the OpenTracing exception mapper once.

In comparison against base master with the same conditions, I hit the OpenTracing exception mapper roughly 9 times out of 10.

So either I should start getting lottery tickets, or this fix works as expected.

arjantijms:
Jenkins test please

MattGill98:
In your reproducer, I believe the JAX-RS endpoint should have a CDI annotation such as `@RequestScoped`. If you're not able to reproduce this, I would suggest investigating this together at some point - one of us must be doing something wrong!

Pandrex247:
@MattGill98 You're correct! With a CDI annotation on the endpoint I very infrequently get the wrong exception mapper - I'll look into it

Pandrex247:
@MattGill98 You're correct! With a CDI annotation on the endpoint I very infrequently get the wrong exception mapper - I'll look into it

Pandrex247:
@MattGill98 You're correct! With a CDI annotation on the endpoint I very infrequently get the wrong exception mapper - I'll look into it

Pandrex247:
@MattGill98 You're correct! With a CDI annotation on the endpoint I very infrequently get the wrong exception mapper - I'll look into it

PR #3209, opened 2018-09-27
PAYARA-3119 NullPointerException when starting Jersey/EJB Containers in Order

jGauravGupta:
Depends on :
https://github.com/payara/patched-src-jersey/pull/3


jGauravGupta:
jenkins test please

jGauravGupta:
Locking is added on EJB container instance and interceptor registered synchronously from Jersey (or other modules) ComponentProvider during EJB Container start phase.  So it should not have impact on performance.

PR #3213, opened 2018-09-28
PAYARA-3122 Managed Scheduled Executor Service still executes tasks after undeployment

jGauravGupta:
Depends on :
https://github.com/payara/patched-src-cu-ri/pull/1


jGauravGupta:
jenkins test please

PR #3215, opened 2018-09-29
PAYARA-3143 don't set component id as it is being captured anyway a…

### LINKED ISSUE #3198: CDI Events in Data Grid using Payara Micro > 181 PAYARA-3143
# Description #
----------

In my application, I send/receive CDI events across a cluster of identical nodes, similar to the example given here:https://blog.payara.fish/microservice-choreography-with-the-payara-micro-clustered-cdi-event-bus
This worked under Payara Micro up to version 5.181, but the same code no longer works when deployed on versions 182 and 183.
I am using the default configuration for domain1 with each version.

## Expected Outcome

CDI Events should be sent/received across the cluster, same as in version 181, and using the default domain1 configuration.

## Current Outcome

I have not yet determined if it is the @Outbound event not being sent or the @Inbound not being received, however somewhere it no longer works as before.

## Steps to reproduce

Create an application with an Application Scoped Bean, deploy it and trigger the sendNotification method. 

```
package com.example;
import java.io.Serializable;

import javax.enterprise.context.ApplicationScoped;
import javax.enterprise.event.Event;
import javax.enterprise.event.Observes;
import javax.inject.Inject;
import javax.inject.Named;

import com.example.RemoteEvent;

import fish.payara.micro.cdi.Inbound;
import fish.payara.micro.cdi.Outbound;

@Named
@ApplicationScoped
public class PushBean implements Serializable {
	
	private static final long serialVersionUID = 1L;
    
    @Inject
    @Outbound
    private Event<RemoteEvent> event;
    

    
    /**
     * On new local notification.
     * @param urn The URN
     */
    public void sendNotification( String urn) {
        
    	try {
        	RemoteEvent remoteEvent = new RemoteEvent();
            remoteEvent.setUrn(urn);
            
            event.fire(remoteEvent);
        }
        catch (Exception exception) {
        	
        }
        
    }
    
    /**
     * On new remote event.
     * @param remoteEvent The RemoteEvent 
     */
    public void onNewRemoteEvent(@Observes @Inbound RemoteEvent remoteEvent) {
        
    System.out.println(remoteEvent);
    }
}
```

## Environment ##

- **Payara Version**: 5.183
- **Edition**: Micro
- **JDK Version**:  8 OpenJDK
- **Operating System**:  Linux
- **Database**:  MariaDB

smillidge:
…nd this is causing the captured invocation to be set to null.

Fixes #3198

### DIFF
--- a/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java
+++ b/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java
@@ -18,7 +18,7 @@
  * GPL Classpath Exception:
  * The Payara Foundation designates this particular file as subject to the "Classpath"
  * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
- * file that accompanied this code.
+ * file that accompanied ex.getCause()this code.
  *
  * Modifications:
  * If applicable, add the following below the License Header, with the fields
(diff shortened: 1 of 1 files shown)


MeroRai:
jenkins test please

PR #3227, opened 2018-10-03
PAYARA-3154 Add .factorypath to .gitignore.

MattGill98:
Eclipse and VScode use this file for annotation processing. It should be removed from the codebase.


Cousjava:
Jenkins test please

PR #3232, opened 2018-10-03
PAYARA-3109 OpenAPI doesn't support if schema type is array - Payara4

MeroRai:


jGauravGupta:

PR #3268, opened 2018-10-14
PAYARA-2692 renamed set-config-secrets-dir to set-config-dir same for…

smillidge:
… the get commands


jGauravGupta:
jenkins test please

PR #3320, opened 2018-10-25
PAYARA-3062 Changed datasource classnames to match with MySQL Connector/J 8

Cousjava:
This will mean that creating a datasource in the admin console for MySQL 5 will not work without editing it


Cousjava:
Jenkins test please

PR #3323, opened 2018-10-26
PAYARA-3072 PAYARA-3073 deprecate functions and classes removed in JD…

Cousjava:
…K11 branch for endorsed dirs


Cousjava:
Jenkins test please

PR #3326, opened 2018-10-26
PAYARA-3199 NPE in MicroProfile Metrics on Payara Micro

Pandrex247:
Move onto one line to prevent stream closing.


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -189,10 +189,9 @@ public void reregisterMetadataConfig() {
     }
 
     private MBeanMetadataConfig getConfig() {
-        
-        InputStream defaultConfig = MetricsHelper.class.getResourceAsStream("/metrics.xml");
-        MBeanMetadataConfig config = JAXB.unmarshal(defaultConfig, MBeanMetadataConfig.class);
-        
+        MBeanMetadataConfig config = JAXB.unmarshal(
+                MetricsHelper.class.getResourceAsStream("/metrics.xml"), 
+                MBeanMetadataConfig.class);
```
svendiedrichsen: Is it necessary to load it from classpath each time this method is called? Or can it be stored as member and reused?

arjantijms: This seems peculiar indeed. Hard to see how this would exactly prevent an NPE. Do you have a stack trace that shows the original NPE this allegedly would prevent?

Pandrex247: I couldn't explain it either, there's nothing evident as to why it would fix it. I'll try to reproduce the stack trace off of the current master, though from memory I don't think this actually threw the stack trace - it would fail out here and swallow the exception as this happens on a executor service thread; the NPE would get thrown later on when trying to access the metrics page due to the failed initialisation.

Pandrex247: Yep, can still reproduce, just build the latest master, start a Payara Micro instance, and hit _localhost:8080/metrics_
```
[2018-11-02T14:41:07.156+0000] [] [WARNING] [] [javax.enterprise.web] [tid: _ThreadID=27 _ThreadName=http-thread-pool::http-listener(1)] [timeMillis: 1541169667156] [levelValue: 900] [[
  StandardWrapperValve[microprofile-metrics-resource]: Servlet.service() for servlet microprofile-metrics-resource threw exception
        at fish.payara.microprofile.metrics.MetricsService.initMetadataConfig(MetricsService.java:169)
        at fish.payara.microprofile.metrics.MetricsService.reregisterMetadataConfig(MetricsService.java:188)
        at fish.payara.microprofile.metrics.rest.MetricsResource.processRequest(MetricsResource.java:93)
        at fish.payara.microprofile.metrics.rest.MetricsResource.doGet(MetricsResource.java:180)
```

Pandrex247: The same line of code works fine in server - it's specific to Micro

Pandrex247: Hmm, still getting the failure even with my change applied now.
What did I do to make it work then? Closing for now

svendiedrichsen:
@Pandrex247 Are there different classloaders involved comparing micro to full?

Issue #3085, opened 2018-08-26
Docker secrets and @DataSourceDefinitions in Payara Micro


tofflos:
# Description #
----------

I'm using the Payara Micro Docker image and injecting a Docker secret into the file `/run/secrets/database-password`. I would like to use that secret in a `@DataSourceDefinition` annotation.
```Java
@DataSourceDefinition(name = "java:app/MyDataSource",
        minPoolSize = 4,
        maxPoolSize = 8,
        className = "org.postgresql.ds.PGPoolingDataSource",
        url = "${dataSource.url}",
        user = "${dataSource.user}",
        password = "${ALIAS=dataSource.password}"
```
If I was using Payara Server I would run `asadmin --passwordfile /run/secrets/database-password create-password-alias dataSource.password` but I don't believe any of the --postbootcommandfile, --postdeploycommandfile and --prebootcommandfile options support the --passwordfile option?

I noticed there was a --secretsdir option but it seems restricted to Microprofile Config and I'm unsure how to get it from there into my `@DataSourceDefinition`.

## Expected Outcome

Perhaps make `${ALIAS=dataSource.password}` look in the --secretsdir or if it's already possible provide an example in the Payara Micro Docker image description at https://hub.docker.com/r/payara/micro?

## Current Outcome

Without a mechanism in place it becomes difficult to use Dockers recommended practices for managing secrets when connecting to datasources. All of the examples I've seen so far are either using hardcoded credentials or passing environment variables.

## Environment ##

- **Payara Version**: 5.182
- **Edition**: Micro
- **JDK Version**: 8 u151 - OpenJDK
- **Operating System**: Linux
- **Database**:

tofflos:
Another solution I've seen floating around is providing a non-standard CDI producer class that simplifies the mapping from Microprofile Config. Both Hammock https://github.com/hammock-project/hammock/wiki/JPA-Support and Meecrowave http://openwebbeans.apache.org/meecrowave/meecrowave-jpa/index.html do this but I'm unsure how well this would fit in with Payara.

lprimak:
You can use environment variable instead of ```ALIAS=``` you can use ```password = "${ENV=dataSource.password}"``` or something similar

tofflos:
I'm aware and I was able to make that work but it's less secure.

> Note: Docker secrets do not set environment variables directly. This was a conscious decision, because environment variables can unintentionally be leaked between containers (for instance, if you use --link).

[...]

PR #3334, opened 2018-10-29
PAYARA-3085 Support global configuration during application deployment

edthorne:
@smillidge I don't have a Github repo handy to reproduce the issue. I can however offer this PR as a solution. I've tested this locally and here's my understanding of the problem:

* During application deployment, the application's class loader has not yet been assigned.
* The application exists in the application registry during deployment.
* The check at `ConfigProviderResolverImpl:169` passes because the null class loader is passed to the `equals` method.
* The check at `ConfigProviderResolverImpl:202` attempts to pass the current loader to the application classloader's `equals` method, resulting in an NPE.

I've simply added a null check before calling `equals`. In hindsight, reversing the comparison to match line 169 would be sufficient to solve the problem.


edthorne:
Related to #3085

smillidge:
Thanks for the contribution with these changes does the deployment work?

edthorne:
The deployment works as expected with the secrets mapped into files by the container and referenced in the `glassfish-resources.xml` contained in the WAR file using `${MPCONFIG=db_user}`.

As I was typing up the PR description, I noticed that the code could be further simplified to match a check earlier in the method. I thought I'd let the review decide if it needed to change and if my analysis was correct.

Pandrex247:
Jenkins test please

ischuraev:
I'd like to add another description how you could get a NPE in ConfigProviderResolverImpl:202. We experienced the NPE-problem in Payara 5.183 in combination with versioned applications.
The fix in this pull request solves the problem with NPE. In my opinion the "search in the app registry" in lines 200+ shouldn't be executed at all, at least in my case.

I'll try to describe steps to reproduce the problem here. If it is not enough, I can open a new issue with more details
1) All applications on the affected server are versioned.
2) Deployment of Applications. Restart of Payara-Server.
3) Call of JerseyClient ClientBuilder.newClient().get() in an application -> the call works.
4) redeployment of an application. It should not be the application from step 3.
5) repeat step 3. -> NPE in ConfigProviderResolverImpl:202

This NPE is not a problem anymore because this pull request fixed it. I think, that the real problem is in the lines 181-184
```
        } else {
            String appName = currentInvocation.getAppName();
            appInfo = applicationRegistry.get(appName);
        }
```
Our application has the name "cms:version". _appName_-variable is resolved to "cms" and _appInfo_ is _null_ as result. If we fill _appName_ with _currentInvocation.getRegistrationName()_ instead,  then _appInfo_ isn't _null_ anymore and in my case "_appInfo != null && appInfo.getAppClassLoader().equals(loader)_" equals to _true_.

PR #3338, opened 2018-10-30
Remove Unnecessary Dependency

Pandrex247:
These modules don't actually need the MicroProfile API as they're not injecting the Tracer.


smillidge:
jenkins test please

PR #3345, opened 2018-10-31
PAYARA-3223 Null check instead of throw/catch NPE

arjantijms:


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/ContainerMapper.java
```diff
@@ -177,23 +177,30 @@ public void service(final Request request, final Response response) throws Excep
...
-                        // Just try to start a normal trace
+                    if (propagatedTraceIdHeader == null) {
+                        // No header with ID provided: just try to start a normal trace
                         requestTracing.startTrace("processContainerRequest");
+                    } else {
+                        UUID propagatedTraceId = UUID.fromString(propagatedTraceIdHeader);
+                        
+                        try {
+                            // Try to get the propagated parent ID if there is one
+                            UUID propagatedParentId =
```
svendiedrichsen: This does do it the same way, doesn't it?

Pandrex247:
Jenkins test please

arjantijms:
Jenkins test please

PR #3368, opened 2018-11-02
PAYARA-3232 LocalTxConnectionEventListener Not Thread Safe

MattGill98:


appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
```diff
@@ -82,7 +83,7 @@
     private ResourceHandle localHandle_;
     private JavaEETransaction associatedTransaction;
 
-    private static Hashtable listenerTable = new Hashtable();
+    private static Map<ManagedConnection, ?> listenerTable = new Hashtable<>();
```
arjantijms: Might consider using a ConcurrentHashMap here.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
```diff
@@ -36,100 +36,138 @@
...
+    private final Map<Object, ResourceHandle> associatedHandles;
+
+    private final ResourceHandle resource;
 
-    private ResourceHandle resource;
-        
     public LocalTxConnectionEventListener(ResourceHandle resource) {
         this.resource = resource;
-        this.associatedHandles = new IdentityHashMap(10);
+        this.associatedHandles = new IdentityHashMap<>(10);
```
svendiedrichsen: Would it be a sufficient enough synchronization to use Collections.synchronizedMap(new IdentityMap(10)) instead of synchronizing the whole listener?

MattGill98: The reason I've not used a synchronized map and instead synchronized specific blocks is that I believe the poolManager.resourceClosed method needs to be run in the same atomic action, and a synchronized map wouldn't do this.

arjantijms:
Jenkins test please

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ConnectorXAResource.java
```diff
@@ -273,32 +274,18 @@ private JavaEETransaction getCurrentTransaction() throws SystemException {
     private void resetAssociation() throws XAException{
         try {
             ResourceHandle handle = getResourceHandleForFinalize();
+            LocalTxConnectionEventListener l = (LocalTxConnectionEventListener) handle.getListener();
```
MarkWareham: name doesn't conform to code conventions

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/listener/LocalTxConnectionEventListener.java
svendiedrichsen: Looks good then.

arjantijms:
Jenkins test please

PR #3390, opened 2018-11-07
PAYARA-2940 Secure Admin Support Portal Fix

MattGill98:
Check JIRA for dependent PRs.

This PR basically does the following:

- Converts index.html for support portal Angular app to JSF page. This allows the special admin indicator token to be fetched from the domain.xml when the page is loaded, and passed to the angular code.
- Passes this token to the Angular code, which allows it to be sent with every REST request. This means that the user doesn't need to authenticate multiple times when browsing the support portal.

The main things I need checked are:

- There are a couple of styling problems that appear to occur for me. I've checked and no styles appear to have been changed. All I can see is that the stylesheet order is inconsistent with the previous HTML solution. This would have been a problem anyway, but may have been revealed by this PR. If so this PR will raise a new issue for @diturner.
- I think the token method I've used is perfectly secure, but this needs verifying by someone else.


appserver/admingui/payara-support/src/main/resources/support/support.jsf
```diff
@@ -1,8 +1,48 @@
...
+ * recipient has the option to distribute your version of this file under
+ * either the CDDL, the GPL Version 2 or to extend the choice of license to
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+-->
+
 <f:verbatim>
     <iframe
```
MeroRai: iFrame is generally a bad idea. Is it possible to use AJAX?

arjantijms:
Jenkins test please

appserver/admingui/payara-support/src/main/resources/support/support.jsf
MattGill98: I checked, but without the IFrame the angular scripts will attempt to send using a POST request, which exceeds the max post form size. The IFrame gets around this.

arjantijms:
@di can you take a look at this too?

PR #3406, opened 2018-11-11
PAYARA-3249 Move Timer Task in DynamicReloader to Payara Executor Service

smillidge:
Part of the thread usage reduction task. Moved scheduled dynamicreloader to the general Payara Executor Service.


arjantijms:
Jenkins test please

PR #3413, opened 2018-11-12
Temporarily disable most JDK7 tests

Pandrex247:
Leaves a JDK7 build and quicklook test just to some quick checks for compile errors.


Pandrex247:
Broken test is fixed in another PR

PR #3414, opened 2018-11-12
PAYARA-3247 Add NPN Version Switch

MattGill98:
These changes include several NPN jars, and code to dynamically select one when the asadmin start script runs. Ran manual tests:

- `start-domain` and then ping https://localhost:8181/ with OpenJDK 8u72 through to 8u192.
- Starting 8u192 instance from 8u181 DAS admin console.
- Payara Micro and ping https://localhost:8181/ on 8u192 (Shouldn't be any different but who knows).


MattGill98:
jenkins test please

lprimak:
Any chance this all can be avoided just by using version support that now exists in the jvm-options tag in domain.xml?

MattGill98:
I've switched out the hard coded logic for the version support in the domain.xml, and fixed a bug that existed related to subversion comparison. Thanks @lprimak for pointing that out!

MattGill98:
@smillidge the new changes should make this a bit more maintainable, as it no longer affects any Java 9+ versions. The only reason we'd need to update this would be for new 8u192+ breaking changes.

MattGill98:
jenkins test please

eduarddrenth:
with todays update to openjdk version "1.8.0_252" this popped up again for 5.191. Not realy a problem for me, only sometimes use 5.191 locally, justletting you know

Issue #3087, opened 2018-08-27
Multithreaded calls to remote EJB may result in LinkageError


svendiedrichsen:
# Description #
----------

When remote EJB methods are called with multiple threads a LinkageError may occur.

## Expected Outcome

Remote EJB methods may be called without LinkageError.

## Current Outcome

```
Caused by: java.lang.LinkageError: loader &#40;instance of  org/glassfish/web/loader/WebappClassLoader&#41;: attempted  duplicate class definition for name: &quot;de/edict/eoc/server/proxy/_CasinoEnvironmentConstructorApi5_Remote&quot;
	at sun.misc.Unsafe.defineClass&#40;Native Method&#41;
	at org.glassfish.pfl.basic.reflection.BridgeBase.defineClass&#40;BridgeBase.java:227&#41;
	at org.glassfish.pfl.dynamic.codegen.impl.CodeGeneratorUtil.makeClass&#40;CodeGeneratorUtil.java:70&#41;
	at org.glassfish.pfl.dynamic.codegen.spi.Wrapper._generate&#40;Wrapper.java:1083&#41;
	at org.glassfish.pfl.dynamic.codegen.spi.Wrapper._generate&#40;Wrapper.java:1067&#41;
	at com.sun.ejb.EJBUtils.generateAndLoad&#40;EJBUtils.java:599&#41;
	at com.sun.ejb.EJBUtils.loadGeneratedRemoteBusinessClasses&#40;EJBUtils.java:513&#41;
	at com.sun.ejb.EJBUtils.loadGeneratedRemoteBusinessClasses&#40;EJBUtils.java:471&#41;
	at com.sun.ejb.containers.RemoteBusinessWrapperBase.readObject&#40;RemoteBusinessWrapperBase.java:122&#41;
	at sun.reflect.NativeMethodAccessorImpl.invoke0&#40;Native Method&#41;
	at sun.reflect.NativeMethodAccessorImpl.invoke&#40;NativeMethodAccessorImpl.java:62&#41;
	at sun.reflect.DelegatingMethodAccessorImpl.invoke&#40;DelegatingMethodAccessorImpl.java:43&#41;
	at java.lang.reflect.Method.invoke&#40;Method.java:498&#41;
	at com.sun.corba.ee.impl.io.IIOPInputStream.invokeObjectReader&#40;IIOPInputStream.java:1813&#41;
[...]

Issue #2763, opened 2018-05-24
Error: Injection of remote EJB into a MDB in clustered mode


LumnitzF:
# Description #
----------

I have a MDB, which uses a remote EJB. The remote EJB can easily be injected when running as standalone, however in clustered mode, an Exception occurs.

## Expected Outcome

The remote EJB gets in injected.

## Current Outcome

An IllegalStateException is thrown, as result of a LinkageError. 
[Here](https://github.com/payara/Payara/files/2035204/stackTrace.txt) is the full stack trace.

However I think this is the important part:
`java.lang.LinkageError: loader (instance of  org/glassfish/web/loader/WebappClassLoader): attempted  duplicate class definition for name: "com/sun/ejb/codegen/GenericEJBHome_Generated"`

or 

`java.lang.LinkageError: loader (instance of  org/glassfish/web/loader/WebappClassLoader): attempted  duplicate class definition for name: "fmg/stp/backendservices/push/api/_PushServiceRemote_Remote"`

where `PushServiceRemote` is the remote interface.

Both Exceptions occur multiple times, and I can't really say how they connect.

### Samples

I've included the classes. 
[RequiredClasses.zip](https://github.com/payara/Payara/files/2035253/RequiredClasses.zip)

## Environment ##

- **Payara Version**: 4.1.2.174
- **Edition**: Full
- **JDK Version**: OpenJDK Runtime Environment (build 1.8.0_161-b14)
- **Operating System**: Linux

smillidge:
Can you build a test case that demonstrates the issue. Perhaps as a simple maven project on GitHub?

frl7082:
OK, I'll try to provide an example for you

lprimak:
Also, have you tried this on .181 version? Please do that as a first step as well

frl7082:
So I digged into this a bit more.

I've created a test case for this: [link](https://github.com/frl7082/Issue-2763)

In order to reproduce:

1. asadmin start-domain domain1
2. create a new config from default-config. I've named it cluster-config
3. create a new cluster:
3.1 goto "clusters"
3.2 "New..."
3.3 enter a cluster name and use the previously created config as config (referenced)
3.4 create two new instances with equal weight
3.5 "OK"
4. Start the created cluster
5. Create a JMS TopicConnectionFactory "jms/LogToInfoTopicConnectionFactory" and enable it on the cluster
6. Create a new JMS Topic "jms/LogToInfoTopic" and enable it on the cluster
[...]

PR #3422, opened 2018-11-14
PAYARA-3265 Fixing linkage errors with generated classes (Part 2)

svendiedrichsen:
Refactoring the previously added synchronization code to prevent LinkageError also when generating EJBHome and Webservice SEI classes.

Extends the fix for #3087 and #2763 to EJBHome and Webservice SEI class generation.


arjantijms:
Jenkins test please

PR #3433, opened 2018-11-18
PAYARA-3265 Openapi using @Operation(hidden = true)

Ykus:
I don't know, if this is the best solution. So I am open to any suggestions or comments.


appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
```diff
@@ -288,6 +289,7 @@ private OpenAPI buildDocument() throws OpenAPIBuildException {
                 openapi = new ApplicationProcessor(classes).process(openapi, appConfig);
                 openapi = new BaseProcessor(baseURLs).process(openapi, appConfig);
                 openapi = new FilterProcessor().process(openapi, appConfig);
+                openapi = new PathsProcessor().process(openapi, appConfig);
```
MattGill98: The extra processor here adds unneeded technical overhead in my opinion. I would personally add this functionality to the BaseProcessor class.

Ykus:
Fixed. Thank you.

MattGill98:
Thanks for those changes @Ykus. I've tested with the MicroProfile OpenAPI TCK and it still passes. Looks all good to me!

MattGill98:
jenkins test please

MattGill98:
jenkins test please

PR #3435, opened 2018-11-19
5.191-SNAPSHOT Increment Version Numbers

MattGill98:


MattGill98:
jenkins test please

appserver/extras/payara-micro/payara-micro-boot/src/main/resources/MICRO-INF/domain/branding/glassfish-version.properties
```diff
@@ -2,7 +2,7 @@ product_name=@@@PRODUCT@@@
 brief_product_name=@@@PRODUCT@@@
 abbrev_product_name=@@@PRODUCT@@@
 major_version=5
-minor_version=184-SNAPSHOT
+minor_version=191184-SNAPSHOT
```
Pandrex247: Sacré bleu!

MattGill98: Good spot, these are both caused by a bug in the release script for community releases! I'll fix the release script after this.

PR #3439, opened 2018-11-20
Reverted System Specific Path Variables

MattGill98:


nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/CreateNodeConfigCommand.java
```diff
@@ -104,7 +104,7 @@ public void execute(AdminCommandContext context) {
             Map<String, String> systemPropsMap = new HashMap<String, String>((Map) (System.getProperties()));
             resolver = new TokenResolver(systemPropsMap);
             String resolvedInstallDir = resolver.resolve(installdir);
-            File actualInstallDir = new File(resolvedInstallDir + File.separatorChar + NodeUtils.LANDMARK_FILE);
+            File actualInstallDir = new File(resolvedInstallDir + "/" + NodeUtils.LANDMARK_FILE);
```
svendiedrichsen: This should be File.pathSeparator not separatorChar. On Windows the char \ would be interpreted as an escape char. The string pathSeparator is "\\" which in concatenation results in the desired result.

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ServerSynchronizer.java
```diff
@@ -676,6 +676,6 @@ private File fileOf(File base, String uri) throws URISyntaxException {
         // have to use string concatenation to combine the relative URI
         // with the base URI because URI.resolve() ignores the last
         // component of the base URI
-        return new File(new URI(base.toURI().toString() + File.separatorChar + uri));
+        return new File(new URI(base.toURI().toString() + "/" + uri));
```
svendiedrichsen: Here an URI is created. IMO URIs use / to separate the path even on Windows.

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/CreateNodeConfigCommand.java
MattGill98: This is a revert to fix the release branches. Feel free to make a separate PR improving this, but this PR is simply to revert broken changes.

Pandrex247: We've tested this and we're fairly certain it should actually just be a "/" - at this particular point in time it doesn't actually seem to be reading it as a "proper" Windows path and it should be using /'s. We came across this bug specifically because using the Windows native file separator of "\\" breaks it at this point!

Also File.pathSeparator is for splitting up multiple paths, so is a _;_ or a _:_ depending on your system ;)

Pandrex247: I've just tested and it works fine using _File.separator_ **or** plain "/" on Windows. So while it's probably "more correct" to use _File.separator_, practically it doesn't seem to make a difference.
As Matt said though this is PR is simply for reverting a change rather than editing stuff - we can change it later.

PR #3451, opened 2018-11-27
PAYARA-3262-Dont-create-a-new-response-in-the-JaxrsContainerExceptionMapper

mulderbaba:


mulderbaba:
jenkins test

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingExceptionMapper.java
```diff
@@ -58,24 +57,18 @@
...
+            return ((WebApplicationException) exception).getResponse();
         }
 
         Response.ResponseBuilder responseBuilder = Response.status(status);
-        if (location != null) {
-            responseBuilder.location(location);
-        }
+
         // If the status is a server error, attach it as an entity, otherwise just return a response
         if (status.getFamily() == Response.Status.Family.SERVER_ERROR) {
```
MattGill98: I'm not sure this fix works. Does it pass the OpenTracing TCK? If not, I believe that the response entity and status still need changing. Everything else should be carried across.

PR #3470, opened 2018-12-03
PAYARA-3340 upgrade org.apache.felix.configadmin to 1.9.10

mulderbaba:


Cousjava:
This appears to be identical to #3469

mulderbaba:
@Cousjava no they are different, they update the version of the component pointed out in the title. they share the same commit on creating properties for versions, that should be the reason of your assumption.

arjantijms:
jenkins test please

PR #3481, opened 2018-12-04
PAYARA-3298 Upgrade javax.el to 3.0.1-b10

jGauravGupta:


jGauravGupta:
jenkins test please

appserver/pom.xml
```diff
@@ -619,8 +619,8 @@
                                 <version>${javax.el.version}</version>
                             </artifact>
                             <jarType>impl</jarType>
-                            <specVersion>3.0</specVersion>
-                            <implVersion>3.0.0</implVersion>
+                            <specVersion>${javax.el-api.version}</specVersion>
```
arjantijms: Isn't this the actual spec version, not the "spec artefact build version"?

E.g. spec should remain at 3.0, since that's the latest spec.

jGauravGupta:
jenkins test please

PR #3486, opened 2018-12-04
PAYARA-3303 Upgrade jstl-api to 1.2.2, jstl-impl to 1.2.5

jGauravGupta:


jGauravGupta:
jenkins test please

appserver/pom.xml
```diff
@@ -396,8 +396,8 @@
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>api</jarType>
-                            <specVersion>1.2</specVersion>
-                            <specImplVersion>1.2.1</specImplVersion>
+                            <specVersion>${jstl-api.version}</specVersion>
```
arjantijms: I think spec version should stay at 1.2. 1.2.2 seems to be the implementation artefact of the 1.2 spec.

jGauravGupta:
jenkins test please

PR #3511, opened 2018-12-11
PAYARA-3258 Disabled Path Based Annotations For Interfaces - Payara4

Pandrex247:
Reverts payara/Payara#3510


appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -91,6 +91,7 @@
 import fish.payara.microprofile.openapi.api.visitor.ApiVisitor;
 import fish.payara.microprofile.openapi.api.visitor.ApiVisitor.VisitorFunction;
 import fish.payara.microprofile.openapi.api.visitor.ApiWalker;
+import javassist.Modifier;
```
Pandrex247: Is this correct?

PR #3520, opened 2018-12-11
PAYARA-3358-asadmin-command-stop-domains-no-longer-works

mulderbaba:


mulderbaba:
jenkins test

arjantijms:
@mulderbaba Can you test with the StopAllDomains command? Fear is that StopAllDomains might nag for a parameter now.

PR #3533, opened 2018-12-19
PAYARA-546 Allow setting SO_KeepAlive on the server side.

Pandrex247:
Set via a SystemProperty on all IIOP listeners.
Follow up PR (https://github.com/payara/Payara/pull/3534) is a provisionary improvement that needs a fine-grained review as I'm not convinced that it'll work on multi-homed environments.


appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -105,10 +106,9 @@
     private static final String PERSISTENT_SSL = "PERSISTENT_SSL";
 
     private static final int BACKLOG = 50;
-    
+
     private static final String SO_KEEPALIVE = "fish.payara.SOKeepAlive";
```
lprimak: I think the case of this property is so complicated, it will lead to issues.
May I suggest ```fish.payara.so-keepalive```

lprimak:
jenkins test

PR #3534, opened 2018-12-19
PAYARA-546 Allow setting SO_KeepAlive on the server side

Pandrex247:
Expansion on https://github.com/payara/Payara/pull/3533 that allows setting SO_KEEPALIVE on the server side, and also allows configuration on a per-listener basis.

This won't work on multi-homed environments - will look at adding this later.


appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -105,10 +106,9 @@
     private static final String PERSISTENT_SSL = "PERSISTENT_SSL";
 
     private static final int BACKLOG = 50;
-    
+
     private static final String SO_KEEPALIVE = "fish.payara.SOKeepAlive";
```
lprimak: Same here. Suggest property to be ```fish.payara.so-keepalive``` to prevent case sensitivity issues

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -651,34 +618,82 @@ private Socket createSSLSocket(String host, int port)
...
+    private void enableSOKeepAliveAsRequired(Socket socket) throws SocketException {
+        boolean shouldSet = false;
+
+        // For each listener, find one with a matching port
+        for (IiopListener iiopListener : IIOPUtils.getInstance().getIiopService().getIiopListener()) {
+            if (Integer.valueOf(iiopListener.getPort()) == socket.getPort()) {
+                // Check for the property globally before checking on the specific listener
+                if (Boolean.getBoolean(SO_KEEPALIVE)) {
+                    // If it's present globally, check if we've actually set the property on the listener
+                    if (iiopListener.getPropertyValue(SO_KEEPALIVE) != null) {
```
lprimak: This if() statement isn't necessary
Boolean.getBoolean(null) will always return false, null check isn't necessary

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -651,34 +618,82 @@ private Socket createSSLSocket(String host, int port)
...
+                            shouldSet = true;
+                        }
+                    } else {
+                        // If the property wasn't set on the listener, go with the global setting
+                        shouldSet = true;
+                    }
+                    break;
+                } else {
+                    // If it wasn't set globally, just check if it's set on the listener
+                    if (iiopListener.getPropertyValue(SO_KEEPALIVE) != null
```
lprimak: same here, null check isn't necessary

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -651,34 +618,82 @@ private Socket createSSLSocket(String host, int port)
...
+     *
+     * @param socket The socket to potentially enable SO_KEEPALIVE on
+     * @throws SocketException If there was an error enabling SO_KEEPALIVE on the socket.
+     */
+    private void enableSOKeepAliveAsRequired(Socket socket) throws SocketException {
+        boolean shouldSet = false;
+
+        // For each listener, find one with a matching port
+        for (IiopListener iiopListener : IIOPUtils.getInstance().getIiopService().getIiopListener()) {
+            if (Integer.valueOf(iiopListener.getPort()) == socket.getPort()) {
```
lprimak: Not quite sure, but I think you can do the address check so the socket's address is compared to iiopListener's address. So if you have multi-homed scenario, the different addresses can be compared

Pandrex247: It's not quite as simple as a quick check: a) it comes out in String format so the only way to compare is literally b) the values returned by the socket and the listener are typically different (e.g. 0.0.0.0 vs. 127.0.0.1, or with extra additional network address gubbins) - we'd have to add functionality for resolving the addresses ourselves.

I think that's enough of a task to be its own Jira.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
Pandrex247: The null check is for checking if it's actually set vs. checking if it's false. I *want* it to tell me if it's null, rather than telling me if it's false - I'm using the null check essentially as a presence check.

The reasoning for this is that value can be set in two places, globally via system property (`Boolean.getBoolean`) and specifically on the listener itself (`iiopListener.getPropertyValue`). So I'm checking what the value is globally, and then overriding that *if* it's set on the specific listener - if I just used the Boolean value it would override the global value with a "false" even if it hadn't been set.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
Pandrex247: Changed to `fish.payara.SO_KEEPALIVE` as this is the most similar to how the property is used elsewhere.

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Keeps failing on the ejb_remoteview test - either something has changed since I made the PR or my luck is awful and that intermittent test just keeps failing.

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
This passes locally - it is simply failing on the intermittent test now.

Pandrex247:
Jenkins test please

PR #3548, opened 2019-01-04
PAYARA-3164 Variable substitution in post boot command file doesn't work in Payara Micro

jGauravGupta:


jGauravGupta:
Jenkins test please

nucleus/core/bootstrap/src/main/java/fish/payara/boot/runtime/BootCommands.java
```diff
@@ -75,27 +82,49 @@ public void parseCommandScript(File file) throws IOException {
...
-                commandStr = br.readLine();
+                commandStr = reader.readLine();
             }
         } catch (IOException ex) {
-            Logger.getLogger(BootCommands.class.getName()).log(Level.SEVERE, null, ex);
+            LOG.log(SEVERE, null, ex);
         }
     }
     
+    private String getEnvironmentSubstitution(String value) {
```
MattGill98: This seems to work, but could it not use the `TranslatedConfigView`?

nucleus/core/bootstrap/src/main/java/fish/payara/boot/runtime/BootCommands.java
```diff
@@ -59,7 +61,12 @@
  */
 public class BootCommands {
 
-    private List<BootCommand> commands;
+    private final List<BootCommand> commands;
+
+    private static final Pattern ENV_VAR_PATTERN = Pattern.compile("([^\\$]*)\\$\\{([^\\}]*)\\}([^\\$]*)");
+    private static final int MAX_SUBSTITUTION_DEPTH = 100;
+
+    private static final Logger LOG = Logger.getLogger(BootCommands.class.getName());
```
MattGill98: Just to stay inkeeping with current conventions, could this be called `LOGGER` please?

nucleus/core/bootstrap/src/main/java/fish/payara/boot/runtime/BootCommands.java
jGauravGupta: Updated the boot commands to use TranslatedConfigView where it requires the pattern ${ENV=key} format for environment variable.

jGauravGupta:
Jenkins test please

PR #3551, opened 2019-01-06
PAYARA-3348-Resources-and-Properties-Tabs-have-wrong-name-when-viewing-the-Healthcheck-Checker-Tab

mulderbaba:


mulderbaba:
jenkins test

PR #3578, opened 2019-01-08
PAYARA-3414 Upgrade javax.el to 3.0.1-b11

jGauravGupta:


jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

Cousjava:
Would it not be better to wait for the Jakarta version to be released and upgrade to that instead, as it is due to be released shortly?

arjantijms:
>Would it not be better to wait for the Jakarta version to be released and upgrade to that instead, as it is due to be released shortly?

Perhaps, but doesn't hurt too much to update to this in the mean time.

PR #3582, opened 2019-01-09
PAYARA-3420 Upgrade jsftemplating to 2.1.3

jGauravGupta:


Cousjava:
Would it not be better to wait for the Jakarta version to be released and upgrade to that instead, as it is due to be released shortly?

jGauravGupta:
Ok @Cousjava, I will wait for other Java EE components to released under EE4J.

jGauravGupta:
jenkins test please

arjantijms:
@Cousjava We can still accept this one, to have some pre-testing during the cycle.

PR #3631, opened 2019-01-18
Sonar critical no1: division by zero in MachineMemoryUsageHealthCheck

jbee:
Mainly created this PR as a test to get into the routines. Picked a random critical bug reported by sonar.

This is a change of behaviour and the `catch` for `ArithmeticException` will no longer be used but a total of zero computes to a percentage of zero. Not sure if this is what we want but it seams more sensible to me.

Also cleaned some minor warnings that don't change of behaviour:
* added `static` modifiers (where possible to `private` methods only), 
* extracted methods to reduce complexity, 
* moved local vars into blocks to reduce scope, 
* removed unnecessary else (as by our style guide)

### DIFF
--- a/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseThresholdHealthCheck.java
+++ b/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseThresholdHealthCheck.java
@@ -76,13 +76,14 @@ public abstract class BaseThresholdHealthCheck<O extends HealthCheckWithThreshol
     protected HealthCheckResultStatus decideOnStatusWithRatio(Double percentage) {
         if (percentage > options.getThresholdCritical()) {
             return HealthCheckResultStatus.CRITICAL;
-        } else if (percentage > options.getThresholdWarning()) {
-            return HealthCheckResultStatus.WARNING;
-        } else if (percentage >= options.getThresholdGood()) {
-            return HealthCheckResultStatus.GOOD;
-        } else {
-            return HealthCheckResultStatus.FINE;
         }
+        if (percentage > options.getThresholdWarning()) {
+            return HealthCheckResultStatus.WARNING;
+        }
+        if (percentage >= options.getThresholdGood()) {
+            return HealthCheckResultStatus.GOOD;
+        }
+		return HealthCheckResultStatus.FINE;
     }
 
     @Override
--- a/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
+++ b/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
@@ -91,7 +91,7 @@ public class HeapMemoryUsageHealthCheck extends BaseThresholdHealthCheck<HealthC
         return result;
     }
 
-    private Double calculatePercentage(MemoryUsage usage) {
+    private static Double calculatePercentage(MemoryUsage usage) {
         if (usage.getMax() > 0) {
             return Math.floor(((double)usage.getUsed() / (double)usage.getMax()) * 100);
         }
--- a/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
+++ b/nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
@@ -91,110 +91,111 @@ public class MachineMemoryUsageHealthCheck extends BaseThresholdHealthCheck<Heal
 
     @Override
     public HealthCheckResult doCheck() {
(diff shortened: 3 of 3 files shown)


nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
```diff
@@ -91,7 +91,7 @@ public HealthCheckResult doCheck() {
         return result;
     }
 
-    private Double calculatePercentage(MemoryUsage usage) {
+    private static Double calculatePercentage(MemoryUsage usage) {
```
svendiedrichsen: Shouldn't this method be moved to class MemoryUsage as an instance method?

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
```diff
@@ -91,110 +91,111 @@ public String getDescription() {
...
+	}
+
+    private static boolean isLinux() {
         String osName = System.getProperty("os.name");
-        if (osName.startsWith("Linux") ||
+        return osName.startsWith("Linux") ||
                 osName.startsWith("FreeBSD") ||
                 osName.startsWith("OpenBSD") ||
                 osName.startsWith("gnu") ||
                 osName.startsWith("gnu/kfreebsd") ||
```
svendiedrichsen: Isn't this line redundant?

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
jbee: In principle this could be better but here we have no choice as this is `java.lang.management.MemoryUsage`.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
jbee: You are right. It should be ok to remove it without change of behaviour.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
```diff
@@ -91,110 +91,111 @@ public String getDescription() {
...
+		                memReclaimable = parseMemInfo(parts);
+		            }
+		        }
+		    }
+
+		    if (!memAvailableFound) {
+		        memAvailable = memFree + memActiveFile + memInactiveFile + memReclaimable;
+		    }
+
+		    double usedPercentage = memTotal == 0L ? 0d :((double) (memTotal - memAvailable) / memTotal) * 100;
```
jbee: Diff can out a bit unfortunate. I moved the local vars and extracted the main level if-else block into methods on their own. Other than that I did not change the code except for the addition of `memTotal == 0L ? 0d :` here to handle to zero case and avoid div-by-zero.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
svendiedrichsen: Haha, sorry. I only read the diff and didn't know about this class.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseThresholdHealthCheck.java
```diff
@@ -76,13 +76,14 @@ public HealthCheckWithThresholdExecutionOptions constructThresholdOptions(Thresh
...
+        if (percentage > options.getThresholdWarning()) {
             return HealthCheckResultStatus.WARNING;
-        } else if (percentage >= options.getThresholdGood()) {
+        }
+        if (percentage >= options.getThresholdGood()) {
             return HealthCheckResultStatus.GOOD;
-        } else {
-            return HealthCheckResultStatus.FINE;
         }
+		return HealthCheckResultStatus.FINE;
```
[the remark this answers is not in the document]
jbee: Tabs spotted!

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/MachineMemoryUsageHealthCheck.java
jbee: Removed the `gnu/kfreebsd` as it is included in `gnu`.

PR #3632, opened 2019-01-18
PAYARA-3438: Improve usability MP Health endpoint as readinessProbe

rdebusscher:
… for k8s

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
+++ b/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
@@ -205,7 +205,7 @@ public class HealthCheckService implements EventListener, ConfigListener {
         // No applications (yet), server is not ready.
         if (applicationsLoaded.isEmpty()) {
             // Application is not yet deployed
-            healthCheckResponses.add(HealthCheckResponse.builder().down().build());
+            healthCheckResponses.add(HealthCheckResponse.builder().name("No Application deployed").down().build());
         }
 
         // If we haven't encountered an exception, construct the JSON response
--- a/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
+++ b/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  * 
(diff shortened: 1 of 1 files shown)


svendiedrichsen:
@rdebusscher Looks as if this only works if you have a single app to deploy or how do you specify the target number of deployed applications?

rdebusscher:
@svendiedrichsen Yes it works optimally for a single app in Payara Micro. Having multiple apps on 1 Payara Micro doesn't make sense if you like to optimize the scale for each app differently.

svendiedrichsen:
@rdebusscher But the MP Health endpoint ist available in Payara full as well. And you can run this with multiple apps in docker/k8s. This can't be used in such an environment for a readiness probe then.

How about providing a system property at server startup with the expected number of apps? This could if absent default to 1. Otherwise why bother with a list of app names? This could be a boolean.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -191,6 +202,12 @@ public void performHealthChecks(HttpServletResponse response) throws IOException
             }
         }
 
+        // No applications (yet), server is not ready.
+        if (applicationsLoaded.isEmpty()) {
+            // Application is not yet deployed
+            healthCheckResponses.add(HealthCheckResponse.builder().down().build());
```
Pandrex247: Is this valid? It should at least have a name.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -112,6 +114,15 @@ public void event(Event event) {
             if (appInfo != null) {
                 healthChecks.remove(appInfo.getName());
                 applicationClassLoaders.remove(appInfo.getName());
+                applicationsLoaded.remove(appInfo.getName());
+            }
+        }
+
+        // Keep track of all deployed applications.
+        if (event.is(Deployment.APPLICATION_STARTED)) {
+            ApplicationInfo appInfo = Deployment.APPLICATION_STARTED.getHook(event);
```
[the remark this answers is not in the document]
rdebusscher: Application_loaded is too early, the endpoints are not available at that time.
Application must be fully operational

smillidge:
What do we need to do to get this in?

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
rdebusscher: I can't find anything in the spec that the name is required. But I added a name.

arjantijms:
Jenkins test

arjantijms:
Jenkins test

PR #3647, opened 2019-01-24
PAYARA-3164 Variable substitution in post boot command file doesn't work in Payara Micro (Payara 4.x)

jGauravGupta:


jGauravGupta:
jenkins test please

PR #3650, opened 2019-01-24
PAYARA-3294 Upgrade javax.mail to 1.6.2 (Payara 4.x)

jGauravGupta:


jGauravGupta:
jenkins test please

jGauravGupta:
You are right, but it is compatible with JDK 7.

PR #3663, opened 2019-01-29
PAYARA-1896 Unified asadmin command to set health check notifier configuration

jbee:
This PR is only one part of the effort of PAYARA-1896 split for easier review.

Goal was to combine the various asadmin commands to set individual health check notifier configurations into one command. To not break existing commands these were deprecated and a new command `set-healthcheck-service-notifier-configuration` was added.

The working principle of the new service is largely based on the `BaseHealthCheckNotifierConfigurer` that was the base class for the various specific commands.

The new command has an additional parameter to specify the type of notifier the setting should apply to. The list of possible values is based on the enum `NotifierType`. The code is entirely written so newly added types of notifiers do not require changes to this service/command.

Like the `BaseHealthCheckNotifierConfigurer` this implementation uses a default value for `noisy` resolved from the corresponding `NotifierConfiguration`. Unlike the
`BaseHealthCheckNotifierConfigurer` implementation this value is not resolved by running another command and resolving the flag from the `ActionReport`. Instead the corresponding `NotifierConfiguration` instance is directly resolved from the `NotificationServiceConfiguration` so that the flag can be read directly using `NotifierConfiguration#getNoisy()`. This allows services to be less entangled and reliant upon each other which should result in a more stable behaviour. 

The new command also prints feedback on the changes to the config. It will only apply and report actual changes, that is a setting changing from `true` to `false` or vice versa.

Example session:
```
Command set-healthcheck-service-notifier-configuration executed successfully.
asadmin> set-healthcheck-service-notifier-configuration --enabled=false --notifier=jms --noisy=true
jms.noisy was false set to true

Command set-healthcheck-service-notifier-configuration executed successfully.
asadmin> set-healthcheck-service-notifier-configuration --enabled=true --notifier=jms --noisy=false
jms.enabled was false set to true
jms.noisy was true set to false

Command set-healthcheck-service-notifier-configuration executed successfully.
asadmin> set-healthcheck-service-notifier-configuration --enabled=true --notifier=jms --noisy=false
Command set-healthcheck-service-notifier-configuration executed successfully.
asadmin> 
```

PR #3679, opened 2019-02-01
PAYARA-1896 Unified asadmin command to set health check service configuration

jbee:
PR with the second part of PAYARA-1896. Again none of the existing commands was changed for backwards compatibility. They were just deprecated for later removel and new commands were added.

`ListHealthCheckServices` is mostly an alias to `HealthCheckServiceLister`. As the command has no parameters the call is directly forwarded.

`SetHealthCheckConfiguration` is largely based on `HealthCheckConfigurer`. For more consistent API some parameter names were changed from camel case to lower case with dash. The service still enables/disables the log notifier in union with enabling and disabling the main config. This now delegates to the new `set-healthcheck-notifier-configuration` command added by #3663.

`HealthCheckServiceConfigureCheckerWithThresholdsCommand.java`, `HealthCheckServiceConfigurer.java`, `HealthCheckServiceThresholdConfigurer.java`, `HoggingThreadsConfigurer.java` and `StuckThreadsConfigurer.java` are combined to the new `SetHealthCheckServiceConfiguration` command. Some parameter names were extended to be clear in the new combined context. The targeted service is no longer given through the name as specified in the `@Service` annotation of the different implementations of `BaseHealthCheck`. Instead this is now based on the `CheckerType` enum (similar to notifiers). To ease providing the otherwise very long names one now can also give the name by the first letters of the word (luckily this combination is unique for all constants). So any of the following is valid:
```
--service=GARBAGE_COLLECTOR 
--service=garbage-collector
--service=GC
--service=gc
```
Generally the new command does not assume a fixed set of services so it automatically works for new services and "updates" when services are removed.

Equally similar to the new notifier configuration command this command now prints the actual changes of configuration, for example:
```
asadmin> set-healthcheck-service-configuration --service=CP --enabled=false --threshold-warning=50
connection-pool.threshold-warning was 60 set to 50
```


nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+    private void updateLogNotifier(AdminCommandContext context) {
+        CommandRunner runner = serviceLocator.getService(CommandRunner.class);
+        ActionReport subReport = context.getActionReport().addSubActionsReport();
+        CommandRunner.CommandInvocation inv = runner.getCommandInvocation("set-healthcheck-service-notifier-configuration", subReport, context.getSubject());
+        ParameterMap params = new ParameterMap();
+        params.add("target", target);
+        params.add("dynamic", String.valueOf(dynamic));
+        params.add("enabled", String.valueOf(enabled));
+        params.add("notifier", NotifierType.LOG.name().toLowerCase());
+        // noisy will default to the notifier's config when not set so we do not set it as this is what we want
```
jbee: I intentionally removed the preset of `noisy` done in the old command as the called service does just have that exact logic if the parameter is not supplied so I think its better to not supply it and have the called command take care of providing the default.

jbee:
jenkins test please

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
```diff
@@ -0,0 +1,402 @@
...
+            report.appendMessage("\n");
+        }
+    }
+
+    private <C extends Checker> Checker updateProperties(Checker config, Class<C> type) throws PropertyVetoException {
+        updateProperty(config, "enabled", config.getEnabled(), String.valueOf(enabled), Checker::setEnabled);
+        updateProperty(config, "time", config.getTime(), time, Checker::setTime);
+        updateProperty(config, "time-unit", config.getUnit(), unit, Checker::setUnit);
+        if (HoggingThreadsChecker.class.isAssignableFrom(type)) {
+            HoggingThreadsChecker htConfig = (HoggingThreadsChecker) config;
```
Pandrex247: `htConfig` doesn't confirm to our code style.
It should be something more like `hoggingThreadsConfig`.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
```diff
@@ -0,0 +1,402 @@
...
+            updateProperty(htConfig, "hogging-threads-threshold", htConfig.getThresholdPercentage(), 
+                    hogginThreadsThreshold, HoggingThreadsChecker::setThresholdPercentage);
+            updateProperty(htConfig, "hogging-threads-retry-count", htConfig.getRetryCount(), 
+                    hogginThreadsRetryCount, HoggingThreadsChecker::setRetryCount);
+        }
+        if (StuckThreadsChecker.class.isAssignableFrom(type)) {
+            StuckThreadsChecker stConfig = (StuckThreadsChecker) config;
```
Pandrex247: Variable name doesn't match code style

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
```diff
@@ -0,0 +1,402 @@
...
+            updateProperty(stConfig, "stuck-threads-threshold", stConfig.getThreshold(), 
+                    stuckThreadsThreshold, StuckThreadsChecker::setThreshold);
+            updateProperty(stConfig, "stuck-threads-threshold-unit", stConfig.getThresholdTimeUnit(), 
+                    stuckThreadsThresholdUnit, StuckThreadsChecker::setThresholdTimeUnit);
+        }
+        return config;
+    }
+
+    private void updateProperty(ThresholdDiagnosticsChecker config, final String name, final String value)
```
Pandrex247: Again, a small bit of Javadoc explaining that this updateProperty method is specifically for one purpose.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
```diff
@@ -0,0 +1,402 @@
...
+
+        if (serviceType == null) {
+            String values = Arrays.asList(CheckerType.values())
+                    .stream().map(t -> t.name().toLowerCase().replace('_', '-')).collect(Collectors.joining(", "));
+            report.setMessage("No such service: " + serviceName + ".\nChoose one of: " + values
+                    + ".\nThe name can also be given in short form consisting only of the first letters of each word.");
+            report.setActionExitCode(ActionReport.ExitCode.FAILURE);
+            return;
+        }
+        serviceName = serviceType.name().toLowerCase().replace('_', '-');
```
Pandrex247: This could be moved into the if below, as it isn't used anywhere else

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/admin/StuckThreadsConfigurer.java
```diff
@@ -43,6 +43,7 @@
 import com.sun.enterprise.util.LocalStringManagerImpl;
 import fish.payara.nucleus.healthcheck.HealthCheckService;
 import fish.payara.nucleus.healthcheck.admin.HealthCheckServiceConfigurer;
+import fish.payara.nucleus.healthcheck.admin.SetHealthCheckServiceConfiguration;
```
Pandrex247: Import just to link?

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+    @Param(name = "dynamic", optional = true, defaultValue = "false")
+    private boolean dynamic;
+
+    @Param(name = "target", optional = true, defaultValue = "server-config")
+    private String target;
+
+    @Param(name = "enabled")
+    private boolean enabled;
+
+    @Param(name = "trace-enabled", optional = true)
```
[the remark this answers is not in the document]
jbee: I was writing the documentation and noticed I wasn't happy with the name either. That said I also don't like the old name much better. If I get this feature right it allows to keep previous runs of the health check service. The entries can be limited by two factors: The number (size) and the age (timeout). I think both these are really confusingly named. From the user's perspective I think of this as `--keep-results`, `max-kept-results`, `max-result-age`. Do you think this would be an improvement or do you rather want to go back to old names with `historical`. In that case should they even use camel case instead of dash?

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/admin/StuckThreadsConfigurer.java
jbee: Yeah, the `StuckThreadsHealthCheck` is in another package (and module) so it ends up like this.

jbee:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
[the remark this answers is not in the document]
Pandrex247: I think I'd still leave it with the old name for now.
It is something possibly worth coming back to though, as your argument of defining the number of results to keep and for how long does have some merit over defining the size of the trace store.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
jbee: If you refer to the `serviceName` assignment this has another function. Since the user is allowed to give the service name in various ways I updated the name here back from the type so it is printed uniformly independent of how the user was given the service name. I added a comment to make this clear.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckServiceConfiguration.java
jbee: Added.

jbee:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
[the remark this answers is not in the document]
jbee: Done.

Pandrex247:
jenkins test please

PR #3682, opened 2019-02-02
PAYARA-1165 allow setting of context root on Payara Micro deployments

smillidge:
This PR allows to set the context path in two ways:

1. use `--contextroot`
```bash
java -jar payara-micro.jar --deploy myapp.war --contextroot mycontext
```
2. use `--deploy` with `:root` after the file name
```bash
java -jar payara-micro.jar --deploy myapp.war:mycontext
```
Both examples configure a effective root context of `/mycontext/`

When both are specified the root given with `--deploy` takes precedence. That means `--contextroot` works as a default or fallback for all applications where no root context was specified explicitly.

```bash
java -jar payara-micro.jar --deploy app1.war:foo --deploy app2.war --contextroot bar
```
This will deploy `app1` in `/foo/` and `app2` in `/bar/`.

Note: `--deploy app.war:/` is the same as `--deploy app.war` and even `--deploy app.war:` which all deploy in the root `/`.
That means a leading `/` is not necessary but can be given. In contrast a tailing `/` should not be given and seems to not work. While the configuration keeps it as `"Context Root": "/foo/"` browsing to `/foo` or `/foo/` forwards to `/foo//` which fails to resolve to the expected page.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/DeploymentFileValidator.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/DeploymentFileValidator.java
@@ -1,0 +1,62 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2019 Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://github.com/payara/Payara/blob/master/LICENSE.txt
+ * See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ * GPL Classpath Exception:
+ * The Payara Foundation designates this particular file as subject to the "Classpath"
+ * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ * file that accompanied this code.
+ *
+ * Modifications:
+ * If applicable, add the following below the License Header, with the fields
+ * enclosed by brackets [] replaced by your own identifying information:
+ * "Portions Copyright [year] [name of copyright owner]"
+ *
+ * Contributor(s):
+ * If you wish your version of this file to be governed by only the CDDL or
+ * only the GPL Version 2, indicate your decision by adding "[Contributor]
+ * elects to include this software in this distribution under the [CDDL or GPL
+ * Version 2] license."  If you don't indicate a single choice of license, a
+ * recipient has the option to distribute your version of this file under
+ * either the CDDL, the GPL Version 2 or to extend the choice of license to
+ * its licensees as provided above.  However, if you add GPL Version 2 code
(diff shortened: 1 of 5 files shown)


lprimak:
jenkins test

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/DeploymentFileValidator.java
```diff
@@ -0,0 +1,62 @@
...
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+package fish.payara.micro.cmd.options;
+
+/**
+ *
+ * @author steve
```
Pandrex247: Full name or GitHub username please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1181,11 +1186,22 @@ private void scanArgs(String[] args) {
...
                     case deploy:
-                        File deployment = new File(value);
+                        // check for context root definition
                         if (deployments == null) {
                             deployments = new LinkedList<>();
+                            contextRoots = new Properties();
+                        }                        
+                        String fileName = value;
+                        String deployContext = null;
+                        if (value.contains(":")) {
```
Pandrex247: Would this work on Windows with fully qualified paths?
`--deploy C:\Users\AndrewPielage\Downloads\blah.war:/myApp`

jbee: Having looked into it a bit since I should do a short video on the usage for marketing I would say if it is valid to use a qualified path like `C:\...` the `value.contains(":")` check is not sufficient and the splitting of the root-context below also has to be adopted since `indexOf` could wrongly point to the `C:`.

rdebusscher:
@smillidge --contextroot is ignored when using --deployFromGAV.

Changing the contextroot logic in **_fish.payara.micro.impl.PayaraMicroImpl#getGAVURLs_** to the following statements fixes this.

```
                String contextRoot = artefactMapEntry.getKey();
                if (this.contextRoot == null) {
                    if ("ROOT".equals(contextRoot)) {
                        contextRoot = "/";
                    }
                } else {
                    contextRoot = this.contextRoot;
                }

```

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
smillidge: OK on windows maybe we need to use the ; as separator

smillidge:
@rdebusscher good catch I will fix that

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
smillidge: Now uses File.pathSeparator to split

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/cmd/options/DeploymentFileValidator.java
smillidge: Done

smillidge:
Now works with deploy from Gav and uses File.pathseparator so will work on windows

smillidge:
jenkins test please

Issue #3630, opened 2019-01-18
Managed Scheduled Executor Service doesn't execute tasks from versioned applications PAYARA-3451


ischuraev:
# Description   ###
----------
An attempt to start a Runnable-task from a versioned applications leads to an exception like "Caused by: javax.enterprise.concurrent.AbortedException: Module minimalejb is disabled". A deployment of the same application with default name doesn't cause exceptions and the the task is executed. Probably the problem is caused by changes in #3213 .

## Expected  Outcome 
The task is executed.

## Current Outcome
An Exception like 

> [2019-01-18T13:02:45.001+0100] [Payara 5.184] [SCHWERWIEGEND] [] [] [tid: _ThreadID=332 _ThreadName=__ejb-thread-pool14] [timeMillis: 1547812965001] [levelValue: 1000] [[
>   java.util.concurrent.ExecutionException: javax.enterprise.concurrent.AbortedException: Module minimalejb is disabled
> 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
> 	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
> 	at de.xqueue.eagle.minimalejb.NewSessionBean.businessMethod(NewSessionBean.java:32)
> 	at sun.reflect.GeneratedMethodAccessor492.invoke(Unknown Source)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.glassfish.ejb.security.application.EJBSecurityManager.runMethod(EJBSecurityManager.java:953)
> 	at org.glassfish.ejb.security.application.EJBSecurityManager.invoke(EJBSecurityManager.java:1017)
> 	at com.sun.ejb.containers.BaseContainer.invokeBeanMethod(BaseContainer.java:4884)
> 	at com.sun.ejb.EjbInvocation.invokeBeanMethod(EjbInvocation.java:656)
> 	at com.sun.ejb.containers.interceptors.AroundInvokeChainImpl.invokeNext(InterceptorManager.java:836)
> 	at com.sun.ejb.EjbInvocation.proceed(EjbInvocation.java:608)
> 	at com.sun.ejb.containers.interceptors.SystemInterceptorProxy.doCall(SystemInterceptorProxy.java:163)
[...]

PR #3693, opened 2019-02-05
PAYARA-3451 Managed Scheduled Executor Service doesn't execute tasks from versioned applications

jGauravGupta:
Github issue : #3630

### DIFF
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/ComponentInvocation.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/ComponentInvocation.java
@@ -277,10 +277,14 @@ public class ComponentInvocation implements Cloneable {
 
     /**
      * Returns the unique registration name of application.
+     *
      * @return
      */
     public String getRegistrationName() {
-        return registrationName;
+        if (registrationName != null) {
+            return registrationName;
+        }
+        return getAppName();
     }
 
     public void setRegistrationName(String registrationName) {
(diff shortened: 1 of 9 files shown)


jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/ComponentInvocation.java
```diff
@@ -271,6 +275,22 @@ public void setModuleName(String moduleName) {
...
+    /**
+     * Returns the unique registration name of application.
+     *
+     * @return
+     */
+    public String getRegistrationName() {
+        if (registrationName != null) {
+            return registrationName;
+        }
+        return getAppName();
```
svendiedrichsen: I think this is somewhat suprising to the caller to get the AppName when they are expecting the RegistrationName. IMHO this should be handled by the caller.

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

jGauravGupta:
`appName` is duplicated for versioned application but `registrationName` is unique id (versioned name) for deployed versioned application. I covered each client invocation to make sure non-null value.

`PAYARA-3451` ticket can be fixed by either caching `registrationName` value in `ComponentInvocation` or fetching it's value via Jndi in `ContextSetupProviderImpl#setup`.

jGauravGupta:
To showcase the alternate fix to fetch `registationName` value via Jndi, I have created the PR #3707 .

Although PR #3693 involves changes in many components but caching the unique id.

PR #3702, opened 2019-02-07
PAYARA-3344 Added System Property for Shutdown Grace Period

jbee:
Initially the plan was to use `domain.xml` based configuration for the grace period setting. I looked into it but it turned out to be overly complicated due to module dependencies and access limitations. So I went for the simpler option using `System` property `fish.payara.shutdowngrace` instead. The period is given is milliseconds. E.g. `-Dfish.payara.shutdowngrace=1000` for 1 second grace period.

After a first look into where to add the waiting it seamed most appropriate on the outermost level, like the `PayaraMicroImpl#shutdown` and equivalent methods. After deeper investigation on possible paths during shutdown I found that the common abstraction to server, micro and embedded that is called independent on how the shutdown was initiated should be the `GlassFish` abstraction. While there are a handful of implementations some of them are wrappers that delegate or extensions that extend. I think there are only 2 relevant implementations: `GlassFishImpl` and `MicroGlassFish`. As `MicroGlassFish` has access to `GlassFishImpl` it seamed a good way to share the code by making it a static method in `GlassFishImpl`. Surely no ideal place for common code but I could not find a better place in a common utility class. I believe that this modification includes embedded as embedded seams to wrap one of the other implementations.

In case of thread interruption or illegal value the grace period is zero and the shutdown continues as usual.

While testing the feature for payara micro I got exceptions logged during or after the grade period caused by further calls to `stop`/`dispose` as those were not `synchronized` for micro.
```java
Cjava.lang.IllegalStateException: Already in STOPPING state.
	at fish.payara.micro.boot.runtime.MicroGlassFish.stop(MicroGlassFish.java:80)
	at fish.payara.micro.boot.runtime.MicroGlassFish.dispose(MicroGlassFish.java:96)
	at fish.payara.micro.impl.PayaraMicroImpl.bootStrap(PayaraMicroImpl.java:1045)
	at fish.payara.micro.impl.PayaraMicroImpl.main(PayaraMicroImpl.java:200)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at fish.payara.micro.boot.loader.MainMethodRunner.run(MainMethodRunner.java:48)
	at fish.payara.micro.boot.loader.Launcher.launch(Launcher.java:107)
	at fish.payara.micro.boot.loader.Launcher.launch(Launcher.java:70)
	at fish.payara.micro.boot.PayaraMicroLauncher.main(PayaraMicroLauncher.java:79)
	at fish.payara.micro.PayaraMicro.main(PayaraMicro.java:397)
	at fish.payara.micro.boot.runtime.MicroGlassFish.stop(MicroGlassFish.java:85)
	at fish.payara.micro.boot.runtime.MicroGlassFish.dispose(MicroGlassFish.java:96)
	at fish.payara.micro.impl.PayaraMicroImpl$1.run(PayaraMicroImpl.java:1600)
```
Since the `GlassFishImpl` does synchronize these methods it seams logical that micro should do this as well. I applied it and tested it and it seams to prevent the problem.

As requested I also added a command line option for payara micro: `--shutdowngrace <duration-ms>`.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2016 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016-2019 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
@@ -39,6 +39,7 @@
  */
 package fish.payara.micro.boot.runtime;
 
+import com.sun.enterprise.glassfish.bootstrap.GlassFishImpl;
 import com.sun.enterprise.module.bootstrap.ModuleStartup;
 import java.util.Properties;
 import org.glassfish.embeddable.CommandRunner;
@@ -74,19 +75,20 @@ public class MicroGlassFish implements GlassFish {
     }
 
     @Override
-    public void stop() throws GlassFishException {
+    public synchronized void stop() throws GlassFishException {
         if (status == Status.STOPPED || status == Status.STOPPING || status == Status.DISPOSED) {
             throw new IllegalStateException("Already in " + status + " state.");
         }
-        
+
         status = Status.STOPPING;
+        GlassFishImpl.sleepShutdownGracePeriod();
         kernel.stop();
         habitat.shutdown();
         status = Status.STOPPED;
     }
 
     @Override
-    public void dispose() throws GlassFishException {
+    public synchronized void dispose() throws GlassFishException {
         if (status == Status.DISPOSED) {
             throw new IllegalStateException("Already disposed.");
         } else if (status != Status.STOPPED) {
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
- * Portions Copyright [2017] Payara Foundation and/or affiliates
+ * Portions Copyright [2017-2019] Payara Foundation and/or affiliates
  */
 
 package com.sun.enterprise.glassfish.bootstrap;
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
@@ -47,12 +47,9 @@ import org.glassfish.embeddable.CommandRunner;
 import org.glassfish.embeddable.Deployer;
 import org.glassfish.embeddable.GlassFish;
 import org.glassfish.embeddable.GlassFishException;
-import org.glassfish.grizzly.threadpool.Threads;
 import org.glassfish.hk2.api.ServiceLocator;
 
 import java.util.Properties;
-
-import javax.swing.plaf.SliderUI;
 
 /**
  * @author Sanjeeb.Sahoo@Sun.COM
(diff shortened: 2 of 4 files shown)


appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
```diff
@@ -74,19 +75,20 @@ public void start() throws GlassFishException {
...
 
     @Override
-    public void stop() throws GlassFishException {
+    public synchronized void stop() throws GlassFishException {
         if (status == Status.STOPPED || status == Status.STOPPING || status == Status.DISPOSED) {
             throw new IllegalStateException("Already in " + status + " state.");
         }
-        
+
         status = Status.STOPPING;
```
jbee: Question is: after status is `STOPPING` will this prevent accepting new requests? I think the intention is to finish the ongoing ones but I guess we do not want to accept new ones.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
```diff
@@ -87,10 +90,24 @@ public synchronized void stop() throws GlassFishException {
             throw new IllegalStateException("Already in " + status + " state.");
         }
         status = Status.STOPPING;
```
jbee: Same here: do we prevent accepting new requests effectively?

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
```diff
@@ -47,10 +47,13 @@
 import org.glassfish.embeddable.Deployer;
 import org.glassfish.embeddable.GlassFish;
 import org.glassfish.embeddable.GlassFishException;
+import org.glassfish.grizzly.threadpool.Threads;
 import org.glassfish.hk2.api.ServiceLocator;
 
 import java.util.Properties;
 
+import javax.swing.plaf.SliderUI;
```
Cousjava: Why is swing being imported?

jbee:
jenkins test please

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
jbee: My bad. Must have auto-completed wrong class and forgotten to remove it.

jbee:
jenkins test please

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
```diff
@@ -87,10 +87,24 @@ public synchronized void stop() throws GlassFishException {
             throw new IllegalStateException("Already in " + status + " state.");
         }
         status = Status.STOPPING;
+        sleepShutdownGracePeriod();
         gfKernel.stop();
         status = Status.STOPPED;
     }
 
+    public static void sleepShutdownGracePeriod() {
+        String shutdowngrace = System.getProperty("fish.payara.shutdowngrace");
```
Pandrex247: Would probably be a little cleaner to have this as a final String variable or enum rather than just specified here.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/MicroGlassFish.java
Pandrex247: I don't follow what the question is here?
There's a check literally above this that prints an error if multiple things try to stop the server.

jbee: What I wanted to point out is that before the wait only the `status` was changed. No internal processing has been signalled to stop by that point. I would imagine that it could very well be the case that new request still are accepted and start to process. If the goal with this feature is to give time to finish started requests it seams only consequent not to accept new ones. I would assume that some parts of the server needs to be stopped or flagged to get this behaviour.

smillidge: Searching the source I don't see that status is used much other than to print it into the admin console. Perhaps we could use this status variable better to stop processing etc.However that would be a different JIRA.

jbee:
> It would be nice to have a command line flag for Payara Micro that just set this system property
I'll add one.

nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
jbee: looking into it...

jbee:
jenkins test please

PR #3707, opened 2019-02-08
PAYARA-3451 Managed Scheduled Executor Service doesn't execute tasks from versioned applications (Fix-2)

jGauravGupta:


jGauravGupta:
jenkins test please

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -344,6 +346,19 @@ private boolean isApplicationEnabled(String appId) {
         }
         return result;
     }
+    
+    private boolean isApplicationAvailable(String appId) {
+        boolean result = false;
+        if (appId != null) {
+            Application app = applications.getApplication(appId);
+            if (app != null) {
+                result = true;
```
MarkWareham: Could just directly return here

jGauravGupta:
This PR is not intended to be merge,  just to showcase the alternate solution of #3693 .

Pandrex247:
We can close the PR and the diff will still be visible :)

PR #3713, opened 2019-02-11
PAYARA-3490 Code level improvements for payara micro argument and property parsing 

jbee:
Adding an option (see #3702) I noticed some duplication in the argument parsing of payara micro.
This PR does some smaller code level improvements.

Mainly the reoccurring `try`-`catch` block to parse arguments to `TimeUnit`, `int` or `long` was extracted to a common helper method.

IDE warned me that the property parsing method unnecessarily would use wrapper types. While the methods all returned wrappers the fields the values were assigned to were all primitive. So I switched those to primitive.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -91,6 +91,7 @@ import java.io.FileNotFoundException;
 import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
 
@@ -1209,7 +1210,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
                         autoBindRange = Integer.parseInt(value);
                         break;
                     case enablehealthcheck:
-                        enableHealthCheck = Boolean.valueOf(value);
+                        enableHealthCheck = Boolean.parseBoolean(value);
                         break;
                     case deployfromgav:
                         if (GAVs == null) {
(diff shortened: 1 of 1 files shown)


appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1285,55 +1266,29 @@ else if (requestTracing[0].matches("\\D+")) {
...
                         enableRequestTracingAdaptiveSampling = true;
-                        try {
-                            TimeUnit.valueOf(value.toUpperCase());
-                            requestTracingAdaptiveSamplingTimeUnit = value.toUpperCase();
-                        } catch (IllegalArgumentException e) {
-                            LOGGER.log(Level.WARNING, "{0} is not a valid value for --requestTracingAdaptiveSamplingTimeUnit", value);
-                            throw e;
-                        }
+                        requestTracingAdaptiveSamplingTimeUnit = parseTimeUnit(value,
+                                "value for --requestTracingAdaptiveSamplingTimeUnit").name();
```
jbee: Note that this is a slight change in behaviour since the `parseTimeUnit` helper will also allow for the short forms (like `ns`) which originally was not done for this particular variable (`requestTracingAdaptiveSamplingTimeUnit`) while other variable holding `TimeUnit` names did this.

jbee:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -2566,30 +2525,28 @@ private Boolean getBooleanProperty(String value, String defaultValue) {
...
         if (property == null) {
             return defaultValue;
-        } else {
-            return Integer.decode(property);
         }
+        return Integer.decode(property).intValue();
     }
 
-    private Long getLongProperty(String value, Long defaultValue) {
+    private static long getLongProperty(String value, long defaultValue) {
```
Pandrex247: Why have you made all of these get methods static? This just seems like a change rather than an improvement.

jbee: The principle of least surprise. A member method (even private) can have dependencies and side effects to the instance that I would need check and consider. A static method cannot have those, hence will not surprise me in that way. Secondly it makes clear that the method might not be best located in this class as it does not depend on the class as a context. Helps to identify code that potentially should be in another class (one of its arguments) or in a utility to allow reuse without causing spaghetti dependencies. Sure its no big deal here, it is more of a thing that if you keep doing it the codebase get easier to reason about and less entangled.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
jbee: Doc PR: https://github.com/payara/Payara-Server-Documentation/pull/479

PR #3715, opened 2019-02-11
PAYARA-3028 Fix remote monitoring of instances from the admin console

Pandrex247:
Fixes retrieving the monitoring data of remote instances from the DAS.
When using the module / bundle classloader, it fails to find any of the Jersey services (despite them being present). This change takes inspiration from [here](https://github.com/payara/Payara/blob/master/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/RestAdapter.java#L299) to switch the classloader to the common one, which *can* find the Jersey services,

### DIFF
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/RestAdapter.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/RestAdapter.java
@@ -178,8 +178,25 @@ public abstract class RestAdapter extends HttpHandler implements ProxiedRestAdap
                     adapter = exposeContext();
                     RestLogging.restLogger.log(Level.INFO, RestLogging.REST_INTERFACE_INITIALIZED, context);
                 }
-                //delegate to adapter managed by Jersey.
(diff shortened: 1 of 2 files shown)


nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/adapter/RestAdapter.java
```diff
@@ -178,8 +178,25 @@ public void service(Request req, Response res) {
                     adapter = exposeContext();
                     RestLogging.restLogger.log(Level.INFO, RestLogging.REST_INTERFACE_INITIALIZED, context);
                 }
-                //delegate to adapter managed by Jersey.
-                adapter.service(req, res);
+
+                ClassLoader originalContextClassLoader = Thread.currentThread().getContextClassLoader();
+                boolean monitoring = false;
```
MattGill98: Just a small OCD change, but could this line be simplified?
```suggestion
                boolean monitoring = context.equals("/monitoring");
```

Pandrex247:
Jenkins test please

PR #3727, opened 2019-02-13
PAYARA-3478 Fix Port Usage and Logic in PAYARA-546

Pandrex247:
Fixes some silly mistakes introduced in the fix for PAYARA-546: https://github.com/payara/Payara/pull/3534

### DIFF
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
@@ -650,22 +650,26 @@ public class IIOPSSLSocketFactory implements ORBSocketFactory {
 
             // For each listener, find one with a matching port
             for (IiopListener iiopListener : IIOPUtils.getInstance().getIiopService().getIiopListener()) {
-                if (Integer.valueOf(iiopListener.getPort()) == socket.getPort()) {
+                if (Integer.valueOf(iiopListener.getPort()) == socket.getLocalPort()) {
                     // Check for the property globally before checking on the specific listener, giving precedence to the
                     // new property
                     if ((System.getProperty(SO_KEEPALIVE) == null && Boolean.getBoolean(SO_KEEPALIVE_DEPRECATED))
                             || Boolean.getBoolean(SO_KEEPALIVE)) {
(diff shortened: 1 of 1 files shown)


Pandrex247:
Jenkins test please

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -650,22 +650,26 @@ private void enableSOKeepAliveAsRequired(Socket socket) throws SocketException {
...
+                if (Integer.valueOf(iiopListener.getPort()) == socket.getLocalPort()) {
                     // Check for the property globally before checking on the specific listener, giving precedence to the
                     // new property
                     if ((System.getProperty(SO_KEEPALIVE) == null && Boolean.getBoolean(SO_KEEPALIVE_DEPRECATED))
                             || Boolean.getBoolean(SO_KEEPALIVE)) {
-                        // Check if we should override the global value
-                        if (soKeepAliveEnabledOnIiopListener(iiopListener)) {
-                            shouldSet = true;
+                        // Check if the property has been set on the listener
+                        if (soKeepAlivePropertyPresentOnIiopListener(iiopListener)) {
```
MarkWareham: You have a "present" and "enabled" method. Yet they are always called together. Why not put the "present" check into the "enabled" check method

Pandrex247: Because I need to check separately for if the property is present, and enabled. What you've described is what the `soKeepAliveEnabledOnIiopListener` method actually used to do before.

At this point in the method, I've already determined that the property has been set globally, so what I'm checking for is if there's an override. So what happens here is it checks if the property has been set on the listener, and **if it has**, check what the value of that override is. If the property **isn't** present on the listener, it should roll back to the global setting because there's no override.

If I had them both together, I can't differentiate between the presence of the override, and what the value of that override actually is: It's not an _AND_ relationship, it's more of a "_THEN IF_" relationship.

MarkWareham: I see. Apologies. I missed the "else"

PR #3747, opened 2019-02-19
PAYARA-3318 Restrict permitted HTTP method for FORM based auth

jbee:
The ticket requested configuration either based on a system property or as part of a specific configuration.

I decided to use the system property as this communicates clearly that this is a general "global" on/off switch that applies to all instances and applications. To me this is the most sensible way to implement this feature as it prevents unintentionally permitting any HTTP method for some instances or apps due to using wrong configuration scope.

If customers want to apply the system property cluster wide there is already a mechanism build into the admin GUI that allows to do so: `server => Properties => System Properties (Add Property)`. This also allows to set or change the property after startup. 

For the property name I decided for `fish.payara.permittedFormBasedAuthHttpMethods` as we (mostly) use the `fish.payara.` with no further dot levels.

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
@@ -109,6 +109,11 @@ public class FormAuthenticator extends AuthenticatorBase {
     // -------------------------------------------------- Instance Variables
 
     /**
+     * Property that can be set to restrict the HTTP methods permitted when doing a FORM based authentication.
+     */
+    private static final String PERMITTED_FORM_BASED_AUTH_HTTP_METHODS_PROPERTY = "fish.payara.permittedFormBasedAuthHttpMethods";
+
+    /**
      * Descriptive information about this implementation.
      */
     protected static final String info = "org.apache.catalina.authenticator.FormAuthenticator/1.0";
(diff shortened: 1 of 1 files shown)


appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
```diff
@@ -148,6 +153,12 @@ public boolean authenticate(HttpRequest request, HttpResponse response, LoginCon
         
         // Is this the action request from the login page?
         boolean loginAction = requestURI.startsWith(contextPath) && requestURI.endsWith(FORM_ACTION);
+        if (loginAction) {
```
arjantijms: It's a bit easier to see what's going on if you combine the tests:

```java
if (loginAction && !isPermittedHttpMethod(hreq.getMethod())
```

appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/FormAuthenticator.java
```diff
@@ -148,6 +153,12 @@ public boolean authenticate(HttpRequest request, HttpResponse response, LoginCon
...
+            if (!isPermittedHttpMethod(hreq.getMethod())) {
+                forwardToErrorPage(request, response, config);
```
arjantijms: I think you maybe just want to sent a `deny` here. The error page is often tailored for wrong input (i.e. credentials did not validate).

jbee: Changed to deny.

jbee:
jenkins test please

PR #3749, opened 2019-02-19
PAYARA-3499 Optimisation of ASClassLoaderUtil

Cousjava:


Cousjava:
Jenkins test please

nucleus/deployment/common/src/main/java/org/glassfish/loader/util/ASClassLoaderUtil.java
```diff
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2019] Payara Foundation and/or affiliates
```
Pandrex247: That copyright text isn't quite right.
a) missing square brackets
b) missing "its"

PR #3750, opened 2019-02-19
PAYARA-3261 OpenTracing 1.3

Pandrex247:
Provides support for OpenTracing 1.3

* Skip Patterns
* Operation Name Providers
* MicroProfile Endpoints no longer Traced
* Extra Exception Logging
* MicroProfile Rest Client Support

Requires a patch to Jersey to account for changes in MicroProfile Rest Client, PR here: https://github.com/payara/patched-src-jersey/pull/12


appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
```diff
@@ -125,11 +130,11 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                 tracedAnnotation = OpenTracingCdiUtils.getAnnotation(beanManager, Traced.class, resourceInfo);
             }
             
-            // If there is no annotation, or if there is an annotation and a config override indicating that we should
-            // trace the method...
-            if (tracedAnnotation == null || (boolean) OpenTracingCdiUtils.getConfigOverrideValue(
-                    Traced.class, "value", resourceInfo, boolean.class)
-                    .orElse(tracedAnnotation.value())) {
+            // If there is no matching skip pattern and no traced annotation, or if there is there is no matching skip
+            // pattern and a traced annotation set to true (via annotation or config override)
```
MarkWareham: This method is quite complex. Could use the logical seperation by the comments as points for extracting methods. Just an observation, not a request to change now.

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
```diff
@@ -235,26 +241,111 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
+                        operationName += "/" + classLevelAnnotation.value();
+                    }
+
+                    if (methodLevelAnnotation.value().startsWith("/")) {
+                        operationName += methodLevelAnnotation.value();
+                    } else {
+                        operationName += "/" + methodLevelAnnotation.value();
+                    }
+
+                    return operationName;
```
MattGill98: What's the behaviour here if a class level annotation exists but a method level one doesn't?

nucleus/pom.xml
```diff
@@ -257,12 +257,12 @@
         
         <!-- Microprofile -->
         <microprofile-config.version>1.3</microprofile-config.version>
-        <microprofile-opentracing.version>1.1</microprofile-opentracing.version>
+        <microprofile-opentracing.version>1.3</microprofile-opentracing.version>
```
MattGill98: This is actually 1.3.1 compliant as well, is there anything stopping us updating this for the sake of clarity?

MarkWareham:
> Note that on testing this PR a REST client TCK test fails. Is this relevant?
> 
> ```
> [ERROR] Tests run: 66, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 104.452 s <<< FAILURE! - in TestSuite
> [ERROR] canInvokeMethodOnSubResourceInterface(org.eclipse.microprofile.rest.client.tck.SubResourceTest)  Time elapsed: 0.296 s  <<< FAILURE!
> java.lang.UnsupportedOperationException: Not a resource method.
> 	at org.eclipse.microprofile.rest.client.tck.SubResourceTest.canInvokeMethodOnSubResourceInterface(SubResourceTest.java:54)
> ```

Not seen before, so perhaps a legitimate failure

appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/decorators/JaxrsInvocationBuilderDecorator.java
```diff
@@ -309,5 +317,23 @@ public Response method(String name, Entity<?> entity) {
...
     }
-    
+
+    /**
+     * Instruments this InvocationBuilder instance with OpenTracing
+     */
+    private void instrumentInvocationBuilder() {
+        // Get the ServiceLocator and OpenTracing services
+        ServiceLocator serviceLocator = Globals.getDefaultBaseServiceLocator();
+        OpenTracingService openTracing = serviceLocator.getService(OpenTracingService.class);
```
mulderbaba: do we need to access the service at every call of `instrumentInvocationBuilder`?

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
Pandrex247: At this specific point? You wouldn't get in to the if block:
Line 264: `classLevelAnnotation != null && methodLevelAnnotation != null`

nucleus/pom.xml
Pandrex247: It only got released 2 days ago, I didn't know it was out :P

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
MattGill98: Yeah I know it won't get into that block, I should really have commented a few lines down! I just mean does this method in general handle that scenario as expected? Because I don't see any TCK tests for that in particular so it may not come up!

Pandrex247: Atm it requires that both be set, otherwise it reverts to using the class-method style.
And no the TCK doesn't check for this behaviour and the spec doesn't state what should happen.

I'll check if just using the Path annotation from the class breaks the TCK.

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JaxrsContainerRequestTracingFilter.java
Pandrex247: In reference to the if block? Or the method itself? Because if the latter it's not that complex? Everything above the if is just variable initialisation.

Pandrex247:
MicroProfile Rest Client should now be fixed - my fix to the web decorator revealed a flaw in the web decorator which I then had to fix.

Pandrex247:
Jenkins test please

Pandrex247:
D'oh! Test failed because Jersey hasn't been patched yet.

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

PR #3753, opened 2019-02-20
PAYARA-2206 Remove registration and installer modules

Pandrex247:
Whilst removing the registration module I kind of also had to gut the installer module, which we also don't use, so I removed it as well.


mulderbaba:
jenkins test

PR #3759, opened 2019-02-21
PAYARA-3507 Set executor core pool size to 0 and queue to max on Micro

Pandrex247:
With the current settings, Micro will only create 2 threads and queue 20000 tasks before expanding the pool size any further. With these settings, Micro will create up to 200 threads _as required_ before queuing tasks.


Pandrex247:
Jenkins test please

PR #3760, opened 2019-02-21
PAYARA-3508 Throw informative error in ConnectionPool.getResouce if Interrupted

Cousjava:
This instead of throwing a NPE a short while later


Cousjava:
Jenkins test please

PR #3765, opened 2019-02-22
PAYARA-3511 Don't try to cast a String to a ChronoUnit

Pandrex247:
Should fix https://groups.google.com/forum/#!topic/payara-forum/dA9T4rGP91Q

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/CircuitBreakerInterceptor.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/CircuitBreakerInterceptor.java
@@ -214,15 +214,9 @@ public class CircuitBreakerInterceptor implements Serializable {
         long delay = (Long) FaultToleranceCdiUtils.getOverrideValue(
                 config, CircuitBreaker.class, "delay", invocationContext, Long.class)
                 .orElse(circuitBreaker.delay());
-
-        ChronoUnit delayUnit = circuitBreaker.delayUnit();
-        // Check for an override
-        Optional<String> delayUnitOverride = FaultToleranceCdiUtils.getOverrideValue(
-                config, CircuitBreaker.class, "delayUnit", invocationContext, String.class);
-        if (delayUnitOverride.isPresent()) {
-            delayUnit = ChronoUnit.valueOf(delayUnitOverride.get());
-        }
-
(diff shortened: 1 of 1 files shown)


Pandrex247:
Jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/CircuitBreakerInterceptor.java
```diff
@@ -214,10 +214,15 @@ private Object circuitBreak(InvocationContext invocationContext) throws Exceptio
...
                 config, CircuitBreaker.class, "delay", invocationContext, Long.class)
                 .orElse(circuitBreaker.delay());
-        // Look for a String and cast to ChronoUnit - Use the Common Sense Convertor
-        ChronoUnit delayUnit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
-                config, CircuitBreaker.class, "delayUnit", invocationContext, String.class)
-                .orElse(circuitBreaker.delayUnit());
+
+        ChronoUnit delayUnit = circuitBreaker.delayUnit();
+        // Check for an override
+        Optional<String> delayUnitOverride = FaultToleranceCdiUtils.getOverrideValue(
```
OndroMih: Why all this complex code? I think it's enough to use ChronoUnit.class instead of String.class as a parameter to `getOverrideValue`:

```
ChronoUnit delayUnit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
                config, CircuitBreaker.class, "delayUnit", invocationContext, ChronoUnit.class)
                .orElse(circuitBreaker.delayUnit());
```

It would call config.getOptionalValue(..., ChronoUnit.class) which is valid because Config has a default converter for enums (in general, for all classes that provide static valueOf(String) method).

Pandrex247: I'm fairly certain that it used to be that, but then config changed and I had to use string. I was lead to believe that this wasn't a bug and was intentional. I'll check when I have some time

Pandrex247: You were right, maybe I misremembered. Changes made :)

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

PR #3768, opened 2019-02-22
PAYARA-3512 Check for failOn override on correct annotation

Pandrex247:
For some reason the CircuitBreaker interceptor was checking for overrides of the failOn parameter on the Retry annotation - presumably a copy-pasta bug as I can't immediately see why it should do this.


Pandrex247:
Jenkins test please

PR #3785, opened 2019-02-25
PAYARA-3508 Throw informative error in ConnectionPool.getResouce if Interrupted - Payara4

Cousjava:
…nterrupted

This instead of throwing a NPE a short while later


Pandrex247:
Jenkins test please

PR #3798, opened 2019-02-26
PAYARA-3666 Fixes for issues found by tests for new health check as-admin commands

jbee:
Several issues were found by the tests (separate PR as soon as I know where they should go and how to integrate them in the build/test execution) created for new health check asadmin commands.

I annotated changes with an explanation.


nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/HealthCheckService.java
```diff
@@ -208,13 +209,16 @@ public void bootstrapHealthCheck() {
     /**
      * Starts all notifiers that have been enable with the healthcheck service.
      */
-    public void bootstrapNotifierList() {
+    public synchronized void bootstrapNotifierList() {
```
jbee: I think this should be `synchronized` to be consistent in case of concurrent service reboots (e.g. possible by concurrent admin commands).

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/HealthCheckService.java
```diff
@@ -208,13 +209,16 @@ public void bootstrapHealthCheck() {
...
         notifierExecutionOptionsList = new ArrayList<>();
         if (configuration.getNotifierList() != null) {
             for (Notifier notifier : configuration.getNotifierList()) {
                 ConfigView view = ConfigSupport.getImpl(notifier);
                 NotifierConfigurationType annotation = view.getProxyType().getAnnotation(NotifierConfigurationType.class);
-                notifierExecutionOptionsList.add(executionOptionsFactoryStore.get(annotation.type()).build(notifier));
+                NotifierExecutionOptionsFactory<Notifier> factory = executionOptionsFactoryStore.get(annotation.type());
+                if (factory != null) {
```
jbee: The factory can be null depending on prior changes to the notifier configurations and reboots occurring.

I'm not 100% sure this is the right fix. Maybe the factory should never be null but it solves the immediate problem.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/HealthCheckService.java
```diff
@@ -300,7 +304,8 @@ public void shutdownHealthCheck() {
     }
 
     public BaseHealthCheck getCheck(String serviceName) {
-        return registeredTasks.get(serviceName).getCheck();
+        HealthCheckTask task = registeredTasks.get(serviceName);
+        return task == null ? null : task.getCheck();
```
jbee: Here NPE occurred in test scenarios if there was no task for the `serviceName`. Returning `null` allows the caller to take care of the case.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
```diff
@@ -120,7 +120,7 @@
     private String target;
 
     @Param(name = "enabled")
-    private boolean enabled;
+    private Boolean enabled;
```
jbee: While `boolean` does work mostly as expected it has one unexpected behaviour: The `enabled` parameter effectively becomes optional as HK2 does not complain about a missing parameter and simply will use `false`.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/domain/NotifierExecutionOptionsFactoryStore.java
```diff
@@ -56,7 +57,7 @@
 public class NotifierExecutionOptionsFactoryStore {
 
     private Map<NotifierType, NotifierExecutionOptionsFactory> execOptionsFactoryStore =
-            new HashMap<NotifierType, NotifierExecutionOptionsFactory>();
+            new ConcurrentHashMap<NotifierType, NotifierExecutionOptionsFactory>();
```
jbee: If I get HK2 semantics right this map could be accessed and modified by multiple threads so it should be thread safe.

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/admin/SetHealthCheckConfiguration.java
pdudits: What is the behavior with boxed boolean?

If it is `null` when parameter is not present, then I see a `NullPointerException` in `updateConfig` and `updateLogNotifier`

jbee: This is confusing HK2 magic. The parameter is not optional thus you only ever get so actually running the command if it is provided. But if you make it `boolean` instead of `Boolean` it behaves as if the parameter is optional with default `false`. This is why I changed this back to `Boolean` as we want the console to ask you for the `enabled` parameter if you forget to provide it.

jbee:
PR for 4.x https://bitbucket.org/payara/payara/pull-requests/606/

PR #3807, opened 2019-03-01
QA-190 Tests for added cmd line option to disable hazelcast port auto increment

jbee:
Review should wait until PR #3803 and PR #3798 are merged as they share common tree that otherwise makes this PR appear bigger than it actually is.


jbee:
Moving to samples.

PR #3808, opened 2019-03-01
QA-191 Tests for added shutdown grace period system property

jbee:
Review should wait until PR #3803, PR #3798 and PR #3807 are merged as they share common tree that otherwise makes this PR appear bigger than it actually is.


jbee:
Moving to samples.

PR #3810, opened 2019-03-04
QA-192 POC to launch micro via cmd line args runner programmatically

jbee:
Using the programmatic API of `PayaraMicro` to bootstrap a micro instance has shown to have classloading isses when deploying a simple war with a JSP page. This is not the case when starting the server via the `PayaraMicroLauncher`. The problem encountered illustrated that two code paths to do the same thing is problematic. This lead to the idea to transition to one common code path of `PayaraMicroLauncher` (as this is the working one).  This should not mean that the programmatic bootrapping **API** (as provided by `PayaraMicro`)  should be abandoned, just that this API can be nothing more then a builder API to assemble the `String[]` arguments passed to the launcher. From there is is a common code path.

This "prove-of-concept" PR only changes enough to allow using the `PayaraMicroLauncher` to run payara micro programmatically and receive the instance as possible via `PayaraMicro`.

The first change is to add an "overloaded" version of a _main_ method(s) used to start the server that returns the started `PayaraMicroBoot` instance. To avoid conflicts for tools using the `main` method the new method is not called _main_ but `create` and the main method is left untouched signature wise.

The second small change is that `PayaraMicro` now implements `PayaraMicroBoot`. Thereby both programmatic ways return `PayaraMicroBoot` instance. Judging from the code I believe this was always planned but simply was forgotten and by chance two of the setters did not return the instance for chaining (but `void`).

With these changes it effectively becomes possible to do:
```java
PayaraMicroBoot instance = PayaraMicroLauncher.create(args);
```
And it shows that using `PayaraMicro` and `PayaraMicroLauncher` conceptually yields the same `PayaraMicroBoot` instance.

This should be understood as a first step. Following steps can gradually remove one code path and change the implementation of the fluent API to assemble the argument list passed to the launcher.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/PayaraMicro.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/PayaraMicro.java
@@ -51,7 +51,7 @@ import java.util.logging.Logger;
  *
  * @author steve
  */
-public class PayaraMicro {
+public class PayaraMicro implements PayaraMicroBoot {
     
     
     private PayaraMicroBoot wrappee;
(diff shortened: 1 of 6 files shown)


jbee:
jenkins test please

PR #3812, opened 2019-03-04
QA-192 New module for tests based on payara-micro.jar

jbee:
This PR provides the module setup and test infrastructure common to all tests added based on the `payara-micro.jar` used by several test PRs.

This PR itself is dependent on #3810 to compile as the `PayaraMicroServer` utility makes used of the programmatic bootstrapping using the launched added by #3810.

This PR adds JUnit based integration test setup that make use of the `payara-micro.jar` generated during the build. The tests use the generated jar as dependency to run Payara Micro. The provided  `AsAdminIntegrationTest` base class makes testing _as-admin_ commands straight forward.

This has the benefit of a simple setup that works on command line and in the IDE as any other JUnit based test. The setup also makes sure we test the actual delivery artefact.
Because of the "dependency" on the generated `payara-micro.jar` the tests fit best in the `Payara` repository as they can reference the jar with a relative path. This also has the benefit of running the tests during normal build.

To run in the `integration-test` phase of the build the module configures test classes annotated with `@Category(IntegrationTest.class)` to be run by failsafe and not surefire. This has the benefit of visibility/discoverability and automatic inheritance from base test classes.

Due to the packaging of Payara Micro the internal classes are not accessible at compile time so that reflection is required to check internal state. This PR also adds some basic utilities for this that can be reused for further tests based on the `payara-micro.jar`.

### DIFF
--- a/appserver/tests/pom.xml
+++ b/appserver/tests/pom.xml
@@ -54,6 +54,10 @@
     <packaging>pom</packaging>
     <name>GlassFish Tests related modules</name>
 
+    <modules>
+        <module>test-micro-jar</module>
+    </modules>
+
     <profiles>
         <profile>
             <id>ci</id>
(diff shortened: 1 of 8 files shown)


jbee:
Moving to samples.

PR #3813, opened 2019-03-05
PAYARA-3172: Improve support for building in parallel

pdudits:
There were implicit dependencies between domain templates and
nucleus-domain.

They made dependency:unpack plugin to fail, as the artifact could not be found in local repo


pdudits:
jenkins test please

PR #3819, opened 2019-03-07
PAYARA-3646 ejb rest client automation

cubastanley:
Updated all ejb-rest-remoting POMs to use the term 'HTTP' instead of 'REST'


cubastanley:
@smillidge Not by the looks of things. I've done a full refactor of the packages so everywhere that needs updating has been and there aren't any build faults with the packages being named this way (i.e. server builds and compiles correctly) Are there more testing methods that should be carried out?

arjantijms:
> @smillidge Not by the looks of things.

It theoretically breaks a small thing in the public API, namely in the published constants like `fish.payara.ejb.rest.client.RemoteEJBContextFactory.FISH_PAYARA_CONNECT_TIMEOUT`. If customer code would use that, it would break after this change.

I think it's okay though, as the feature has been released in tech preview.

mulderbaba:
payara test

Pandrex247:
Jenkins test please

Pandrex247:
Would we be able to have both old and new factories and constants for a single release so that the shift isn't quite so abrupt, or are they just going to trip each other up?

Pandrex247:
Jenkins test please

Pandrex247:
_DO NOT MERGE_ label added till convo resolved

Pandrex247:
Jenkins test please

PR #3826, opened 2019-03-12
PAYARA-3540 Fix javadoc compilation errors

Cousjava:
By ensuring none-javadoc stuff is not done when running the javadoc profile


Cousjava:
Using aggregate-no-fork instead of aggregate makes no difference, the errors come from doing aggregate-jar. @Pandrex247 I am only using profiles to work around the problematic modules, the problem is there are so many and if I specify a module not in a profile that it will always run.

Pandrex247:
I still think is isn't the best solution. I'd rather we spent more time fixing it properly than smacking it with a hammer till it behaves like this.

Cousjava:
Closing in favour of #3864

PR #3827, opened 2019-03-12
PAYARA-3466 Open API 1.1

jbee:
Summary:

* Add generics to `Extensible` hierarchy (added in 1.1)
* Implemented new methods (mostly explicit type-safe add/remove/contains)
* Replaced usage of deprecated methods with type-safe replacements added in 1.1
* Removed methods now implemented as `default` methods in the API interfaces
* Added processing of `@Parameters` (missed both by TCK and us in 1.0)
* Added processing of `@Extensions` (missed both by TCK and us in 1.0)
* Added `null` check to `addX` methods (I think 1.1 addec this semantic or 1.0 TCK missed to check)
* Added _copy constructor_ for `Map` model types for their "copy getter" (see API javadoc) 
* Extracted a common base type for `Extensible` `Map`s (share impl.)
* Some more specific local fixes for minor deviations from the spec since 1.0 that TCK 1.1 now found
* Fixed some serialisation issues (present since 1.0)

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
@@ -49,7 +49,6 @@ import fish.payara.microprofile.openapi.impl.processor.BaseProcessor;
 import fish.payara.microprofile.openapi.impl.processor.FileProcessor;
 import fish.payara.microprofile.openapi.impl.processor.FilterProcessor;
 import fish.payara.microprofile.openapi.impl.processor.ModelReaderProcessor;
-import fish.payara.nucleus.executorservice.PayaraExecutorService;
 import java.beans.PropertyChangeEvent;
 import java.net.InetAddress;
 import java.net.MalformedURLException;
@@ -58,7 +57,6 @@ import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Deque;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedDeque;
@@ -104,9 +102,6 @@ public class OpenApiService implements PostConstruct, PreDestroy, EventListener,
 
     @Inject
     private OpenApiServiceConfiguration config;
-
-    @Inject
-    private PayaraExecutorService executor;
 
     @Inject
     private ServerEnvironment environment;
@@ -189,7 +184,7 @@ public class OpenApiService implements PostConstruct, PreDestroy, EventListener,
         if (mappings.isEmpty() || !isEnabled()) {
             return null;
         }
-        return (OpenAPI) mappings.peekLast().getDocument();
+        return mappings.peekLast().getDocument();
     }
 
     /**
@@ -224,7 +219,7 @@ public class OpenApiService implements PostConstruct, PreDestroy, EventListener,
      * @return a list of all loadable classes in the archive.
      */
     private static Set<Class<?>> getClassesFromArchive(ReadableArchive archive, ClassLoader appClassLoader) {
-        return Collections.list((Enumeration<String>) archive.entries()).stream()
+        return Collections.list(archive.entries()).stream()
                 // Only use the classes
                 .filter(x -> x.endsWith(".class"))
                 // Remove the WEB-INF/classes and return the proper class name format
@@ -262,12 +257,12 @@ public class OpenApiService implements PostConstruct, PreDestroy, EventListener,
         private final OpenApiConfiguration appConfig;
         private volatile OpenAPI document;
 
-        private OpenApiMapping(ApplicationInfo appInfo) {
+        OpenApiMapping(ApplicationInfo appInfo) {
             this.appInfo = appInfo;
             this.appConfig = new OpenApiConfiguration(appInfo.getAppClassLoader());
         }
 
-        private ApplicationInfo getAppInfo() {
+        ApplicationInfo getAppInfo() {
             return appInfo;
         }
 
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -42,6 +42,8 @@ package fish.payara.microprofile.openapi.impl.model;
 import static fish.payara.microprofile.openapi.impl.model.util.ModelUtils.isAnnotationNull;
 import static fish.payara.microprofile.openapi.impl.model.util.ModelUtils.mergeProperty;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -127,4 +129,34 @@ public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensi
         return value;
     }
 
+    @Override
+    public String toString() {
+        StringBuilder str = new StringBuilder();
+        toString(str, "\t");
+        return str.toString();
+    }
+
+    void toString(StringBuilder str, String indent) {
+        str.append(getClass().getSimpleName());
+        Class<?> type = getClass();
+        while (type != Object.class) {
+            for (Field field : type.getDeclaredFields()) {
+                if (!Modifier.isStatic(field.getModifiers()) && !field.isSynthetic()) {
+                    str.append("\n").append(indent).append(field.getName()).append(": ");
+                    try {
+                        field.setAccessible(true);
+                        Object value = field.get(this);
+                        if (value instanceof ExtensibleImpl) {
+                            ((ExtensibleImpl<?>)value).toString(str, indent + "\t");
+                        } else {
+                            str.append(value);
+                        }
+                    } catch (IllegalArgumentException | IllegalAccessException e) {
+                        str.append("<failure:").append(e.getMessage()).append(">");
+                    }
+                }
+            }
+            type = type.getSuperclass();
+        }
+    }
 }
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -49,7 +49,7 @@ import java.util.Map;
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.models.Extensible;
 
-public abstract class ExtensibleImpl implements Extensible {
+public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensible<T> {
 
     protected Map<String, Object> extensions = new HashMap<>();
 
@@ -58,9 +58,16 @@ public abstract class ExtensibleImpl implements Extensible {
         return extensions;
     }
 
+    @SuppressWarnings("unchecked")
     @Override
-    public void addExtension(String name, Object value) {
+    public T addExtension(String name, Object value) {
         extensions.put(name, value);
+        return (T) this;
+    }
+
+    @Override
+    public void removeExtension(String name) {
+        extensions.remove(name);
     }
 
     @Override
@@ -68,7 +75,7 @@ public abstract class ExtensibleImpl implements Extensible {
         this.extensions = extensions;
     }
 
-    public static void merge(Extension from, Extensible to, boolean override) {
+    public static void merge(Extension from, Extensible<?> to, boolean override) {
         if (isAnnotationNull(from)) {
             return;
         }
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -51,17 +51,15 @@ import java.util.logging.Logger;
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.models.Extensible;
 
-import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensible<T> {
 
     private static final Logger LOGGER = Logger.getLogger(ExtensibleImpl.class.getName());
-    
+
     protected Map<String, Object> extensions = new LinkedHashMap<>();
 
-    @JsonAnyGetter
     @Override
     public Map<String, Object> getExtensions() {
         return extensions;
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -46,6 +46,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.logging.Logger;
 
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
@@ -71,19 +72,26 @@ public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensi
     @Override
     public T addExtension(String name, Object value) {
         if (value != null) {
-            extensions.put(name, value);
+            extensions.put(extensionName(name), value);
         }
         return (T) this;
     }
 
     @Override
     public void removeExtension(String name) {
-        extensions.remove(name);
+        extensions.remove(extensionName(name));
     }
 
     @Override
     public void setExtensions(Map<String, Object> extensions) {
-        this.extensions = extensions;
+        this.extensions.clear();
+        for (Entry<String, Object> entry : extensions.entrySet()) {
+            this.extensions.put(extensionName(entry.getKey()), entry.getValue());
+        }
+    }
+
+    public static String extensionName(String name) {
+        return name.startsWith("x-") ? name : "x-" + name;
     }
 
     public static void merge(Extension from, Extensible<?> to, boolean override) {
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -51,6 +51,7 @@ import java.util.logging.Logger;
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.models.Extensible;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
@@ -58,6 +59,7 @@ public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensi
 
     private static final Logger LOGGER = Logger.getLogger(ExtensibleImpl.class.getName());
 
+    @JsonIgnore
     protected Map<String, Object> extensions = new LinkedHashMap<>();
 
     @Override
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
@@ -42,24 +42,20 @@ package fish.payara.microprofile.openapi.impl.model;
 import static fish.payara.microprofile.openapi.impl.model.util.ModelUtils.isAnnotationNull;
 import static fish.payara.microprofile.openapi.impl.model.util.ModelUtils.mergeProperty;
 
-import java.io.IOException;
-import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
-
-import javax.json.Json;
 
 import org.eclipse.microprofile.openapi.annotations.extensions.Extension;
 import org.eclipse.microprofile.openapi.models.Extensible;
 
-import com.fasterxml.jackson.core.JsonFactory;
-import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 public abstract class ExtensibleImpl<T extends Extensible<T>> implements Extensible<T> {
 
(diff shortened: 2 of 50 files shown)


appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
```diff
@@ -42,50 +42,81 @@
...
             to.getExtensions().put(from.name(), value);
         }
     }
 
-    public static Object convertExtensionValue(String value) {
+    public static Object convertExtensionValue(String value, boolean parseValue) {
         if (value == null) {
             return null;
         }
+        if (parseValue) {
```
jbee: New feature in 1.1 where value should be parsed to some other type than `String`.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
```diff
@@ -98,4 +129,34 @@ public static Object convertExtensionValue(String value) {
         return value;
     }
 
+    @Override
+    public String toString() {
```
jbee: Added this to ease debugging by providing a helpful `toString` for the model.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/ContentImpl.java
```diff
@@ -76,16 +102,19 @@ public static void merge(org.eclipse.microprofile.openapi.annotations.media.Cont
...
+            EncodingImpl.merge(encoding, to.getMediaType(typeName).getEncoding(), override, currentSchemas);
         }
 
         // Merge examples
         for (ExampleObject example : from.examples()) {
-            ExampleImpl.merge(example, to.get(typeName).getExamples(), override);
+            ExampleImpl.merge(example, to.getMediaType(typeName).getExamples(), override);
+        }
+        if (!from.example().isEmpty()) {
+            to.getMediaType(typeName).setExample(from.example());
```
jbee: `example`  is a new property in 1.1

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/security/ScopesImpl.java
```diff
@@ -39,37 +39,44 @@
...
+    }
+
+    public ScopesImpl(Map<String, ? extends String> scopes) {
+        super(scopes);
+    }
 
     @Override
     public Scopes addScope(String name, String item) {
-        this.put(name, item);
+        this.put(name, item); // this DOES accept null!
```
jbee: This seems inconsistent but this add method explicitly allows for null while usually the add methods do not allow for null or should ignore it.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/security/SecurityRequirementImpl.java
```diff
@@ -45,22 +45,31 @@
...
+    }
+
+    public SecurityRequirementImpl(Map<? extends String, ? extends List<String>> items) {
+        super(items);
+    }
+
     @Override
     public SecurityRequirement addScheme(String name, String item) {
-        this.put(name, Arrays.asList(item));
+        this.put(name, item == null ? new ArrayList<>() : Arrays.asList(item));
```
jbee: `null` should become empty list

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/security/SecurityRequirementImpl.java
```diff
@@ -45,22 +45,31 @@
...
         return this;
     }
 
     @Override
     public SecurityRequirement addScheme(String name, List<String> item) {
-        this.put(name, item);
+        this.put(name, item == null ? new ArrayList<>() : item);
```
jbee: `null` should become empty list

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/servers/ServerImpl.java
```diff
@@ -80,26 +79,23 @@ public void setDescription(String description) {
...
-        setDescription(description);
-        return this;
-    }
-
     @Override
     public ServerVariables getVariables() {
-        return variables;
+        return variables instanceof ServerVariables || variables == null 
+                ? (ServerVariables) variables
+                : new ServerVariablesImpl(variables);
```
jbee: FYI: `ServerVariables` got "half-removed". It is deprecated and should be removed in the future. In some places it already was replaced with a `Map<String, SeverVariable>` what causes changes here.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -427,6 +421,18 @@ public static boolean isAnnotationNull(Annotation annotation) {
...
+    }
+
+    public static Boolean mergeProperty(boolean current, Boolean offer, boolean override) {
+        return mergeProperty(Boolean.valueOf(current), offer, override);
+    }
+
+    public static Boolean mergeProperty(boolean current, boolean offer, boolean override) {
+        return mergeProperty(Boolean.valueOf(current), Boolean.valueOf(offer), override);
+    }
+
```
jbee: FYI: I added this overloaded variants to avoid boxing/unboxing warnings. The module is almost free of warnings and it would be nice to get there.

appserver/payara-appserver-modules/microprofile/openapi/src/test/java/fish/payara/microprofile/openapi/resource/classloader/ApplicationClassLoader.java
```diff
@@ -106,10 +106,10 @@ public ApplicationClassLoader(Application app) {
...
         int size = stream.available();
         byte buff[] = new byte[size];
-        DataInputStream in = new DataInputStream(stream);
-        // Reading the binary data
-        in.readFully(buff);
-        in.close();
+        try (DataInputStream in = new DataInputStream(stream)) {
+            // Reading the binary data
+            in.readFully(buff);
+        }
```
jbee: FYI: fixed potential resource leak.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -513,111 +516,118 @@ public void visitOpenAPI(OpenAPIDefinition definition, AnnotatedElement element,
...
         if (element instanceof Class) {
+            visitSchemaClass(schema, (Class<?>) element, context);
+        }
+        if (element instanceof Field) {
+            visitSchemaField(schema, (Field) element, context);
+        }
+        if (element instanceof java.lang.reflect.Parameter) {
+            visitSchemaParameter(schema, (java.lang.reflect.Parameter) element, context);
+        }
+    }
```
jbee: FYI: extracted the cases into one method each to make this more readable.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -626,22 +636,32 @@ public void visitSchema(Schema schema, AnnotatedElement element, ApiContext cont
 
     @Override
     public void visitExtension(Extension extension, AnnotatedElement element, ApiContext context) {
-        if (extension.name() != null && !extension.name().isEmpty() && extension.value() != null
-                && !extension.value().isEmpty()) {
+        String value = extension.value();
+        String name = extension.name();
+        if (name != null && !name.isEmpty() && value != null
+                && !value.isEmpty()) {
+            Object parsedValue = ExtensibleImpl.convertExtensionValue(value, extension.parseValue());
```
jbee: FYI: here we were missing applying conversion via `convertExtensionValue`.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -705,47 +725,23 @@ public void visitAPIResponses(APIResponses apiResponses, AnnotatedElement elemen
...
-                String matchingMethodParamName = ModelUtils.getParameterName(matchingMethodParameters.get(0));
-                // Find the matching operation parameter
-                for (org.eclipse.microprofile.openapi.models.parameters.Parameter operationParam : context
-                        .getWorkingOperation().getParameters()) {
-                    if (operationParam.getName().equals(matchingMethodParamName)) {
-                        matchedParam = operationParam;
-                    }
-                }
-            }
+            matchedParam = findOperationParameterFor(parameter, (Method) element, context);
```
jbee: FYI: extracted the cases to make it more readable.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -760,12 +756,63 @@ public void visitParameter(Parameter parameter, AnnotatedElement element, ApiCon
...
+                    .collect(Collectors.toList());
+            // If there is more than one match, filter it further
+            ParameterIn in = parameter.in();
+            if (matchingMethodParameters.size() > 1 && in != null && in != ParameterIn.DEFAULT) {
+                // Remove all parameters of the wrong input type
+                matchingMethodParameters
+                .removeIf(x -> ModelUtils.getParameterType(x) != In.valueOf(in.name()));
+            }
+            if (matchingMethodParameters.isEmpty()) {
+                return null;
```
jbee: FYI: This is a fix I added. 1.0 TCK simply had no case that would otherwise fail the next line where we always get the first entry of the list. But some cases do not have a matching parameter.

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
pdudits: I'm afraid it is much more complicated than that, the spec says [any valid JSON value](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#specificationExtensions), which entails for exable:

* leading and trailing spaces on lists and objects
* strings containing dots while not being numbers
* NaN being a double

I think the best approach is to always pass it to Jackson, and then decide the return value on the type of JsonNode.

If it fails, it is a string

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
```diff
@@ -83,16 +90,33 @@ public static void merge(Extension from, Extensible<?> to, boolean override) {
...
+    public static Object convertExtensionValue(String value, boolean parseValue) {
         if (value == null) {
             return null;
         }
+        if (parseValue) {
+            if (value.startsWith("{") || value.startsWith("[")) { // JSON object + array
+                try {
+                    return new ObjectMapper().readTree(value);
+                } catch (Exception e) {
+                    return value;
```
pdudits: Logging the exception on debug-level might help the developers that didn't expect the result to be string. Actually, if they explicitly set `parseValue=true`, they likely do not expect string back at all. So it's likely worth a warning.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -760,12 +756,63 @@ public void visitParameter(Parameter parameter, AnnotatedElement element, ApiCon
...
 
+    private static org.eclipse.microprofile.openapi.models.parameters.Parameter findOperationParameterFor(Parameter parameter,
+            Method annotated, ApiContext context) {
+        String name = parameter.name();
+        // If the parameter reference is valid
+        if (name != null && !name.isEmpty()) {
+            // Get all parameters with the same name
+            List<java.lang.reflect.Parameter> matchingMethodParameters = Arrays
+                    .asList(annotated.getParameters()).stream()
+                    .filter(x -> name.equals(ModelUtils.getParameterName(x)))
```
jbee: FYI: swapped the equals because `ModelUtils.getParameterName(x)` can be null causeing NPE while `name` is from an annotation and cannot be null.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
jbee: Turn out one can't rely on the API javadoc which listed the 4 cases I was handling. 
Good suggestion. Will change it that way.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -313,12 +313,6 @@ public static SchemaType getParentSchemaType(SchemaType type1, SchemaType type2)
         if (type2 == null) {
             return type1;
         }
-        if (type1 == SchemaType.OBJECT || type2 == SchemaType.OBJECT) {
-            return SchemaType.OBJECT;
-        }
-        if (type1 == SchemaType.STRING || type2 == SchemaType.STRING) {
-            return SchemaType.STRING;
-        }
```
jbee: FYI: the above 2 cases aren't required as the default `return type1` (below, not visible) will effectively do the same while the other check (`type1 != type2`) done will not match.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
```diff
@@ -262,12 +257,12 @@ private static String getContextRoot(ApplicationInfo appInfo) {
         private final OpenApiConfiguration appConfig;
         private volatile OpenAPI document;
 
-        private OpenApiMapping(ApplicationInfo appInfo) {
+        OpenApiMapping(ApplicationInfo appInfo) {
```
MattGill98: Is there any reason for this?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleTreeMap.java
```diff
@@ -0,0 +1,48 @@
...
+
+    protected ExtensibleTreeMap() {
+        super();
+    }
+
+    protected ExtensibleTreeMap(Map<String, ? extends V> items) {
+        super(items);
+    }
+
+    @JsonAnyGetter
```
MattGill98: If possible I'd like this in the mixin for consistency

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
MattGill98: Is this correct? As I see it the results have changed:

| type1  	| type2  	| old parent 	| new parent 	|
|--------	|--------	|------------	|------------	|
| OBJECT 	| STRING 	| OBJECT     	| STRING     	|
| OBJECT 	| NUMBER 	| OBJECT     	| STRING     	|

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -405,7 +399,7 @@ public static boolean isAnnotationNull(Annotation annotation) {
                             return false;
                         } else if (value instanceof Collection && !Collection.class.cast(value).isEmpty()) {
                             return false;
-                        } else if (value instanceof Boolean && Boolean.class.cast(value)) {
+                        } else if (value instanceof Boolean && ((Boolean)value).booleanValue()) {
```
MattGill98: I don't particularly like this new syntax, but this comment isn't exactly a blocker so it's up to you

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -404,8 +407,8 @@ public void visitFormParam(FormParam param, AnnotatedElement element, ApiContext
 
             // Set the request body type accordingly.
             context.getWorkingOperation().getRequestBody().getContent()
-                    .get(javax.ws.rs.core.MediaType.WILDCARD).getSchema()
-                    .setType(formSchemaType);
+            .getMediaType(javax.ws.rs.core.MediaType.WILDCARD).getSchema()
```
MattGill98: This is no standard formatting, could this be reverted?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -490,8 +493,8 @@ private SchemaImpl getArraySchema(AnnotatedElement element) {
         arraySchema.setType(ModelUtils.getSchemaType((Class<?>) parameterizedType.getActualTypeArguments()[0]));
         return arraySchema;
     }
-    
-    private String getDefaultValueIfPresent(AnnotatedElement element) {
+
+    private static String getDefaultValueIfPresent(AnnotatedElement element) {
```
MattGill98: Everywhere you've done this is a really positive refactor!

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
```diff
@@ -262,12 +257,12 @@ private static String getContextRoot(ApplicationInfo appInfo) {
...
             this.appInfo = appInfo;
             this.appConfig = new OpenApiConfiguration(appInfo.getAppClassLoader());
         }
 
-        private ApplicationInfo getAppInfo() {
```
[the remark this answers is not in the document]
jbee: Its not 1.1. related. If the method is private it is not accessible for the caller what java mitigates with synthetic methods added that forward to the private method. I saw no reason why we have to force this being private when default visibility can avoid synthetic methods.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
jbee: Short: Synthetic methods. Long explanation given in another comment.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleTreeMap.java
jbee: You mean moving the class into the `mixin` package?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
jbee: I think it becomes clear if you look at the fill method:
```java
if (type1 == null && type2 == null) {
            return null;
        }
        if (type1 == null) {
            return type2;
        }
        if (type2 == null) {
            return type1;
        }
        // here we removed the logic that would return OBJECT of both are OBJECT or STRING if both are STRING
        if (type1 != type2) {
            return SchemaType.STRING;
        }
        return type1;
}
```
So the removed logic returns the common type of both are the same. If we remove it it now falls through, first tested if the are not the same (which for same has no effect) what brings us to do `return type1` (which for any same pair is the common type.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
jbee: While it might not be prettier textual it has the benefit of the compiler helping you out as it can apply type analysis to tell you if this can or cannot work. Using `X.class.cast` which accepts `Object` is valid for any input and the compiler will never complain. This makes this tricky when changes are applied as you don't get compile errors for changes so you tend to overlook needed updates to this style of casting.

The other issue is that it tends to be overused as it looks like a usual method call. Instead of casting once and working type safe from there variables are casted multiple times essentially abandoning static typing for dynamic typing.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
jbee: Sure. Not sure what happened there. IDE "helped" me or something.

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleTreeMap.java
jbee: I had a look and it seemed the mixin contained all the JSON output tweaking (not just for `Extensible`) so I decided to continue this practice and added all tweaks to it (and removed them from the concrete classes).

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleTreeMap.java
```diff
@@ -0,0 +1,100 @@
...
+    public final void removeExtension(String name) {
+        this.extensions.remove(extensionName(name));
+    }
+
+    /**
+     * Custom {@link JsonSerializer} that adds both the extended {@link TreeMap} entries as well as the
+     * {@link ExtensibleTreeMap#extensions} map to the output object unless the value represents a {@link Reference} in
+     * which case only the {@link Reference#getRef()} is added.
+     */
+    static class ExtensibleTreeMapSerializer extends JsonSerializer<ExtensibleTreeMap<?,?>> {
```
jbee: FYI: This fixes two defects in the JSON/YAML serialisation present since 1.0 where properties on `Map` extending types would be ignored as they are handled by the map serializer. Thereby extensions would not be added and in case of a reference an empty object was rendered instead of the reference.

jbee:
jenkins test please

PR #3844, opened 2019-03-15
PAYARA-3684 Security Cleaning sweep 5

arjantijms:
Mostly factored out functionality from Realm. Notice copyright for new files, as they are not new code but split out from existing files.

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/login/LDAPLoginModule.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/login/LDAPLoginModule.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2018-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2018] [Payara Foundation and/or its affiliates]
 package com.sun.enterprise.security.auth.login;
 
 import static com.sun.enterprise.security.auth.realm.ldap.LDAPRealm.MODE_FIND_BIND;
@@ -72,6 +72,7 @@ import com.sun.enterprise.security.auth.realm.ldap.LDAPRealm;
  *   (having substituted %d for the user DN previously found). From the matched entry(ies) all the values of group-target
  *   are taken as group names in which the user has membership. If no entries are found, the group membership is empty.
  * </ol>
+ *
  *
  */
 public class LDAPLoginModule extends BasePasswordLoginModule {
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/login/LDAPLoginModule.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/login/LDAPLoginModule.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2018-2019] [Payara Foundation and/or its affiliates]
 package com.sun.enterprise.security.auth.login;
 
 import static com.sun.enterprise.security.auth.realm.ldap.LDAPRealm.MODE_FIND_BIND;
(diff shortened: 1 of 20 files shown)


arjantijms:
Jenkins test please

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/Realm.java
```diff
@@ -60,108 +60,42 @@
...
  * @author Harish Prabandham
  * @author Harpreet Singh
  * @author Jyri Virkki
  * @author Shing Wai Chan
  *
+ * @see java.security.Principal
  */
 @Contract
-public abstract class Realm implements Comparable<Realm> {
+public abstract class Realm extends BaseRealm implements Comparable<Realm> {
```
pdudits: I think this badly affects binary compatibility - hierarchy changed, static methods changed their containing classes. Realms written against previous API will fail to run after merging.

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/BaseRealm.java
```diff
@@ -0,0 +1,276 @@
...
+import java.util.Properties;
+
+/**
+ * This class is the base for all Payara Realm classes, and contains common
+ * object state such as the realm name, general properties and the assign groups.
+ * 
+ * @author Arjan Tijms
+ *
+ */
+public abstract class BaseRealm extends AbstractRealm implements Comparable<Realm> {
```
pdudits: What happens with `com.sun.enterprise.security.BaseRealm`? We shouldn't remove it, as it may be in use by users, and having two classes with similar name is confusing.

arjantijms: Good point, let's think of another name for that one then.

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/Realm.java
smillidge: I'm not keen on breaking binary compatibility for some slight refactoring

arjantijms: It's not breaking binary compatibility is it? No static methods moved at all. Only the private state and instance methods have been moved, which are inherited by Realm.

pdudits: I've generated [revapi](https://revapi.org/) report against 5.191, it turns out, that it is ok. The only breaking change was caused by replacement of `javax.security.auth.x500.X500Principal` in method parameters, which was in different PR and is deprecated/for removal in JDK.

The diff was too big to apply in my head :)

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/BaseRealm.java
arjantijms: After a chat settled on `AbstractStatefulRealm`. Thx again @pdudits

arjantijms:
Jenkins test please

arjantijms:
Jenkins test please

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/login/LDAPLoginModule.java
```diff
@@ -73,7 +73,6 @@
  *   are taken as group names in which the user has membership. If no entries are found, the group membership is empty.
  * </ol>
  *
- *
```
MattGill98: Is this the change that requires a copyright header? If so I wouldn't change the file at all. This comment applies to quite a number of the touched files.

arjantijms: Okay, fair point

arjantijms:
Jenkins clean test

arjantijms:
Jenkins test please

PR #3852, opened 2019-03-19
PAYARA-3466 Open API 1.1 - Tests

jbee:
This PR adds more tests for the Open API module motivated by the defects reported earlier and found during update to Open API 1.1.

A particular focus area is the JSON rendering. Being annotation guided semi-automatic the output wasn't always intuitive and often incorrect. There were problems both with missing elements and elements that should never have been in the output. 

To verify the JSON output from our model against the Open API spec (as outlined here https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md ) I created a model for the spec structure (see `NoteType`) and a `OpenApiValidator` that can verify the structure of a Open API document given as `JsonNode`. As this is no trivial operation in itself the validator itself needed testing to verify its correctness. For that reason I added example JSON snippets from the spec which are expected to pass the verification. So none of the `.json` files needs serious review.

The likely correct structural Open API document validator is then used to validate documents created for existing and new tests. 

A common base class `OpenApiApplicationTest` was extracted that bootstraps the applications with the extras of the class under test and verifies the structure of created document so that the individual tests only have to focus on verifying the contents of the document.
The same was done with `OpenApiBuilderTest` as a base for tests using the model fluent API to setup the basis of the document.

A larger change in the annotation processing was needed to handle annotations as if they are inherited from super-classes and interfaces. The util `AnnotationInfo` takes care of this aggregation for each type.


appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
```diff
@@ -147,26 +148,57 @@ public String toString() {
...
+
+    private static boolean isNonEmpty(Object value) {
+        if (value == null) {
+            return false;
+        }
+        if (value instanceof Object[]) {
+            return ((Object[])value).length > 0;
+        }
+        if (value instanceof Collection) {
+            return ((Collection<?>) value).size() > 0;
```
svendiedrichsen: Rather `!isEmpty()`

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/ExtensibleImpl.java
```diff
@@ -147,26 +148,57 @@ public String toString() {
...
+        }
+        if (value instanceof Map) {
+            return ((Map<?, ?>) value).size() > 0;
```
svendiedrichsen: Same here.

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -416,15 +391,17 @@ public static boolean isAnnotationNull(Annotation annotation) {
                         } else if (String.class.isAssignableFrom(value.getClass()) && !value.toString().isEmpty()) {
                             return false;
                         } else if (value instanceof Annotation) {
-                            allNull = isAnnotationNull((Annotation) value);
+                            if (!isAnnotationNull((Annotation) value)) {
+                                return false;
```
jbee: FYI: This was a logical slip making an annotation only "empty" if all its fields were "empty" but logically it should be "not all" or "any non empty" makes it "non empty".

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -569,37 +546,39 @@ else if (newValue instanceof Constructible) {
     }
 
     public static String getResourcePath(GenericDeclaration declaration, Map<String, Set<Class<?>>> resourceMapping) {
```
jbee: FYI: split this in a method for `Class` and `Method` each which now use `AnnotationInfo` to check annotation so "inherited" annotations are taken into account.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -150,6 +154,26 @@
      */
     public ApplicationProcessor(Set<Class<?>> appClasses) {
         this.classes = appClasses;
+        this.classes.removeIf(cls -> cls.isInterface() || Modifier.isAbstract(cls.getModifiers()));
+        if (classes != null) {
+            addInnerClasses();
```
jbee: FYI: Added this here as an after-fact since it allows to only have one place that both works in full integration setting or production setting as well as our test setup. This makes this effective in our testing and keeps the test setup in line what will happen in production.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -150,6 +154,26 @@
      */
     public ApplicationProcessor(Set<Class<?>> appClasses) {
         this.classes = appClasses;
+        this.classes.removeIf(cls -> cls.isInterface() || Modifier.isAbstract(cls.getModifiers()));
```
jbee: FYI: TCK had some cases where we should ignore given types that do not represent actual endpoints so I added this filter.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -1012,19 +1034,22 @@ private static String getContentType(String name) {
...
+    private org.eclipse.microprofile.openapi.models.media.Schema createSchema(ApiContext context, Class<?> type, Type genericType) {
         org.eclipse.microprofile.openapi.models.media.Schema schema = new SchemaImpl();
         SchemaType schemaType = ModelUtils.getSchemaType(type);
         schema.setType(schemaType);
 
         // Set the subtype if it's an array (for example an array of ints)
         if (schemaType == SchemaType.ARRAY) {
-            Class<?> subType = type.getComponentType();
-            org.eclipse.microprofile.openapi.models.media.Schema subSchema = schema;
-            while (subType != null) {
```
jbee: FYI: Changed this from looping to recursion (see `schema.setItems` to the right) since the return type could be any combination of collections, arrays and simple types. So this now supports collections is well which required to carry the full generic type through the methods involved.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/FilterProcessor.java
```diff
@@ -80,10 +80,18 @@
      */
     private OASFilter filter;
 
+    public FilterProcessor() {
+        this(null);
+    }
+
+    public FilterProcessor(OASFilter filter) {
+        this.filter = filter;
```
jbee: FYI: Added for testing.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/FilterProcessor.java
```diff
@@ -143,7 +151,8 @@ private Object filterObject(Object object) {
             }
 
             // If the object is a model item
-            if (object.getClass().getPackage().getName().startsWith(OpenAPIImpl.class.getPackage().getName())) {
+            Package pkg = object.getClass().getPackage();
```
jbee: FYI: Package can be `null` in special cases. Got that error ;)

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -153,112 +153,82 @@ public void accept(ApiVisitor visitor) {
...
         processAnnotations(Produces.class, visitor::visitProduces);
         processAnnotations(Consumes.class, visitor::visitConsumes);
+
+        // OpenAPI response types
         processAnnotations(RequestBody.class, visitor::visitRequestBody);
     }
 
-    private <A extends Annotation, A2 extends Annotation, E extends AnnotatedElement> void processAnnotations(
-            Class<A> annotationClass, VisitorFunction<A, E> annotationFunction, Class<A2> altClass,
-            VisitorFunction<A2, E> altFunction, boolean allowInterfaces) {
```
jbee: FYI: Interface exclusion are now handled elsewhere since interfaces are not disallowed generally since one can put annotations on the interface to make them take effect. With this this flag was no longer needed.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -133,8 +133,8 @@ public OpenApiWalker(OpenAPI api, Set<Class<?>> allowedClasses, Map<String, Set<
     @Override
     public void accept(ApiVisitor visitor) {
```
jbee: FYI: I change the `processAnnotations` to always only processing one annotation at a time as processing two (+"alt") was quite confusing to understand and how to change it to also handle "inherited" annotations. The alternative(s) are now stated as last argument so they can be taking into account when deciding if annotation on method level should inherit from class level. This now is their only function so this semantic now becomes more apparent.

appserver/payara-appserver-modules/microprofile/openapi/src/test/java/fish/payara/microprofile/openapi/test/app/TestApplication.java
```diff
@@ -39,24 +39,10 @@
...
 import javax.ws.rs.core.Application;
 
-import fish.payara.microprofile.openapi.test.app.application.ResponseTest;
-import fish.payara.microprofile.openapi.test.app.application.RootPathTest;
-
 @ApplicationPath("/test")
 public class TestApplication extends Application {
 
-    @Override
-    public Set<Class<?>> getClasses() {
```
jbee: FYI: Now passed to the test setup to be more flexible and isolated per test.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/SchemaImpl.java
```diff
@@ -354,6 +357,7 @@ public void removeProperty(String key) {
         this.properties.remove(key);
     }
 
+    @JsonProperty("additionalProperties")
```
MattGill98: Isn't this the default name?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/AnnotationInfo.java
```diff
@@ -0,0 +1,201 @@
...
+            return getAnnotation(annotationType, (Parameter) element);
+        }
+        return null;
+    }
+
+    public boolean isAnnotationPresent(Class<? extends Annotation> annotationType) {
+        return getAnnotation(annotationType) != null;
+    }
+
+    public boolean isAnnotationPresent(Class<? extends Annotation> annotationType, Field field) {
```
MattGill98: These could do with documenting, it's not clear to me from the signature what this method does

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
MattGill98: I'm not quite sure I follow. What's the difference between empty == all fields empty and not empty == at least one non-empty field?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -395,7 +371,6 @@ public static boolean isAnnotationNull(Annotation annotation) {
         if (annotation == null) {
```
MattGill98: Might it now be possible to move this static method into the new AnnotationInfo class?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -150,6 +154,26 @@
...
+    }
+
+    private void addInnerClasses() {
+        List<Class<?>> topLevelClasses = new ArrayList<>(classes);
+        for (Class<?> topLevelClass : topLevelClasses) {
+            addInnerClasses(topLevelClass);
+        }
+    }
+
+    private void addInnerClasses(Class<?> topLevelClass) {
```
MattGill98: I don't think the application processor cares about this. Is it possible to put this in the walker?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -153,112 +153,82 @@ public void accept(ApiVisitor visitor) {
         processAnnotations(FormParam.class, visitor::visitFormParam);
 
         // All other OpenAPI annotations
-        processAnnotations(Schema.class, visitor::visitSchema, true);
-        processAnnotations(Server.class, visitor::visitServer, Servers.class, visitor::visitServers, true);
-        processAnnotations(Extensions.class, visitor::visitExtensions, true);
-        processAnnotations(Extension.class, visitor::visitExtension, true);
+        processAnnotations(Schema.class, visitor::visitSchema);
```
MattGill98: I've only just noticed that Schema is processed twice. There may be a good reason for that, but out of interest does the TCK pass with either of them removed?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/SchemaImpl.java
jbee: Yes, I think I forgot to remove this after trying some things with jackson. Should be possible to remove.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
jbee: Possible yes, I'd think doing so conflates the classes purpose with that of a utility around annotations which it is not (yet) so I'd not do that unless you think this still is nicer than having it in `ModelUtils`. `AnnotationInfo` is used as one instance per type with state so a general annotation helper fits name wise but not so much context wise. WDYT?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
jbee: The logic that was in place did assign to the `allNull` variable for every property that had an annotation type. So if an annotation would have multiple properties of such types the last of them would effectively make the result of them "all" being null/empty which is wrong logically. The correct logic is all of them should be empty for an annotation to be considered empty which is same as any being non-empty means its non empty.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
jbee: No, this was unintentional. Must have pasted this wrongly.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
MattGill98: That's a fair comment, I'd just be keen to remove things from ModelUtils whenever I see a chance ;) The class exists just because I perceived it as necessary, but I don't really like it.

I agree that the scope of the class doesn't make much sense in this context though, so I'm happy to leave it as-is.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
MattGill98: Oh I see, that sounds like a mistake indeed!

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/SchemaImpl.java
jbee: We can remove the explicit name, but the annotation is actually required to include the property. I think I found the reason back when I added it but can't remember. Anyhow, I remove the name but keep the annotation.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
jbee: You were right, it had a reason. I was wrong, I did not add it, it was already present. Looked a bit into it. When removing the second one the problem is that some paths do `ApplicationProcessor.createSchema` while there already is a schema for the type whereby it gets overridden. Removing the first one however did pass all our tests and the TCK so I am going for that.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
jbee: I agree that this should be in the walker. Moving it there does fail tests though. Reason is that the list of classes is used for the resource mapping which is created before the walker would add inner classes. To me this also looks like it never should have been in the processor but be purely within the walker. The processor has the `addParameter` method which also creates a context wherefore it also recreates a resource mapping again (each time). I think this is kind of a work around. The walking logic of how to process fields annotated with a parameter annotation should be in the walker which would avoid creating the resource mapping many times over and would allow moving the creation of the mapping into the walker which allows the walker to first add the inner classes so these are also considered in the resource mapping.

While this seems a clear resolution to the issue to me this is a more complicated refactoring disentangling processor and walker so the result is still the same. I am note sure we should do this in the context of this issue.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/AnnotationInfo.java
jbee: Added detailed javadoc. It should be clear now.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -395,36 +371,31 @@ public static boolean isAnnotationNull(Annotation annotation) {
...
             return true;
         }
-        boolean allNull = true;
         for (Method m : annotation.annotationType().getDeclaredMethods()) {
             if (m.getParameterCount() == 0) {
                 try {
                     Object value = m.invoke(annotation);
-                    if (value != null) {
```
jbee: FYI: Main point was to change the `.cast(...)` to `(Type)value` style of cast  or `instanceof` where appropriate. Also values from annotation can never be `null` or a `Collection` type so I removed these checks. To be sure this is correct I added tests for this util method.

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/api/visitor/ApiContext.java
```diff
@@ -64,4 +64,10 @@
      * The created operation currently being worked on.
      */
     Operation getWorkingOperation();
+
+    /**
+     * @param type any class, not null
+     * @return true, if the give type is a known type in this context, else false
+     */
+    boolean isApplicationType(Class<?> type);
```
jbee: I added this to allow the `ApplicationProcessor` logic to make this check even though the information on this is within the `OpenApiWalker` that creates the `ApiContext`.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -457,24 +438,9 @@ private void addParameter(AnnotatedElement element, ApiContext context,
         if (context.getWorkingOperation() != null) {
             context.getWorkingOperation().addParameter(newParameter);
         } else {
-            if (element instanceof Field) {
-                Field field = Field.class.cast(element);
-                ApiContext apiContext;
```
jbee: FYI: Below logic moved to the walker so the walker now is the only one creating the context which is both cleaner and was needed to allow moving inner class processing to the walker what needed to delay generation of resource mapping which also moved to the walker.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -907,55 +869,17 @@ public void visitSecurityRequirements(SecurityRequirements securityRequirements,
 
     // PRIVATE METHODS
 
-    /**
-     * Generates a map listing the location each resource class is mapped to.
-     */
-    private static Map<String, Set<Class<?>>> generateResourceMapping(Set<Class<?>> classList) {
```
jbee: FYI: Moved to the walker as is.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -128,13 +137,30 @@ public OpenApiWalker(OpenAPI api, Set<Class<?>> allowedClasses, Map<String, Set<
...
+                addInnerClasses(topLevelClass.getSuperclass());
+            }
+        }
     }
 
     @Override
     public void accept(ApiVisitor visitor) {
         // OpenAPI necessary annotations
-        processAnnotations(OpenAPIDefinition.class, visitor::visitOpenAPI, true);
-        processAnnotations(Schema.class, visitor::visitSchema, true);
```
jbee: FYI: `Schema` was processed twice which was needed previously but now works when keeping the later processing of `Schema` annotation.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -153,113 +179,141 @@ public void accept(ApiVisitor visitor) {
...
-            for (Field field : clazz.getDeclaredFields()) {
-                processAnnotation(field, annotationClass, annotationFunction, altClass, altFunction,
-                        new OpenApiContext(api, null));
+        for (Field field : annotatedClass.getDeclaredFields()) {
+            if (annotations.isAnnotationPresent(annotationClass, field)) {
+                if (   annotationClass == HeaderParam.class
+                    || annotationClass == CookieParam.class
+                    || annotationClass == PathParam.class
+                    || annotationClass == QueryParam.class) {
+                    // NB. if fields are annotated as Param all methods have it
```
jbee: FYI: The field => method logic part was moved here from the application processor

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -153,113 +179,141 @@ public void accept(ApiVisitor visitor) {
...
 
-    private <A extends Annotation, A2 extends Annotation, E extends AnnotatedElement> void processAnnotations(
-            Class<A> annotationClass, VisitorFunction<A, E> annotationFunction, Class<A2> altClass,
-            VisitorFunction<A2, E> altFunction) {
-        processAnnotations(annotationClass, annotationFunction, altClass, altFunction, false);
-    }
+    /**
+     * Generates a map listing the location each resource class is mapped to.
+     */
+    private static Map<String, Set<Class<?>>> generateResourceMapping(Set<Class<?>> classList) {
```
jbee: FYI: Moved here from `ApplicationProcessor`

jbee:
jenkins test please

jbee:
@Pandrex247 I would have liked to split in multiple PRs but its hard to do that for fixes and confirming tests where most of the fixes only in union lead to meaningful and correct results one can test for. Therefore I felt it was still the best option to keep it all in one PR.

MattGill98:
:tada: :tada: :tada:

PR #3853, opened 2019-03-19
Payara-3419: Support deploying timers to deployment group and non-persistent timer service

pdudits:
`asadmin deploy` no longer fails when application with timers are deployed to a deployment group.

EJB Timer Service offers option of persistence type None, where all timers are treated like non-persistent (as in Web Profile / EJB Lite).

### DIFF
--- a/appserver/admingui/ejb/src/main/resources/configuration/ejbContainerTimerService.jsf
+++ b/appserver/admingui/ejb/src/main/resources/configuration/ejbContainerTimerService.jsf
@@ -101,11 +101,7 @@
             <sun:property id="persistenceServiceProp" labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18n_ejb.ejbTimerSettings.persistenceServiceLabel}"  helpText="$resource{i18n_ejb.ejbTimerSettings.persistenceServiceHelp}" visible="#{true}" >
                 <sun:dropDown id="EjbTimerService" selected="#{pageSession.valueMap['ejbTimerService']}" required="#{false}" value="#{pageSession.valueMap['ejbTimerService']}" labels={"Database","DataGrid","None"} values={"Database","DataGrid","None"} >
                 </sun:dropDown>
-            </sun:property>            
-            <sun:property id="clusterFiringModeProp" labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18n_ejb.ejbTimerSettings.clusterFiringModeLabel}"  helpText="$resource{i18n_ejb.ejbTimerSettings.clusterFiringModeHelp}" visible="#{true}" >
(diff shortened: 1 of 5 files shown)


appserver/admingui/ejb/src/main/resources/configuration/ejbContainerTimerService.jsf
```diff
@@ -98,7 +99,7 @@
                 <sun:staticText id="msecs" style="padding: 8pt" text="$resource{i18n.common.Milliseconds}"/>
             </sun:property>
             <sun:property id="persistenceServiceProp" labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18n_ejb.ejbTimerSettings.persistenceServiceLabel}"  helpText="$resource{i18n_ejb.ejbTimerSettings.persistenceServiceHelp}" visible="#{true}" >
-                <sun:dropDown id="EjbTimerService" selected="#{pageSession.valueMap['ejbTimerService']}" required="#{false}" value="#{pageSession.valueMap['ejbTimerService']}" labels={"Database","DataGrid"} values={"Database","DataGrid"} >
+                <sun:dropDown id="EjbTimerService" selected="#{pageSession.valueMap['ejbTimerService']}" required="#{false}" value="#{pageSession.valueMap['ejbTimerService']}" labels={"Database","DataGrid","None"} values={"Database","DataGrid","None"} >
```
pdudits: image

pdudits:
jenkins test please

pdudits:
jenkins test please

appserver/ejb/ejb-connector/src/main/java/org/glassfish/ejb/config/EjbTimerService.java
```diff
@@ -160,11 +162,16 @@
      * Gets the value of the EJB Timer Service  property.
      *
      * The EJB Timer Service type to use in Payara.
+     * Supported values:
+     * <ul>
+     *     <li>{@value #TYPE_DATABASE} for database-backed persistence</li>
+     *     <li>"DataGrid" for DataGrid-backed persistence</li>
```
jGauravGupta: Replace "DataGrid" with {@value #TYPE_DATAGRID}

appserver/ejb/ejb-connector/src/main/java/org/glassfish/ejb/config/EjbTimerService.java
```diff
@@ -68,6 +68,8 @@
 
 @Configured
 public interface EjbTimerService extends ConfigBeanProxy, PropertyBag {
+    String TYPE_NONE = "None";
+    String TYPE_DATABASE = "Database";
```
jGauravGupta: Perhaps TYPE_DATAGRID constant can be added to list all timer types.

pdudits: The reason I included these two was, that these are special -- one is default, the other has special behaviour. There's nothing else in the code that would actually match against value DataGrid in the code, and changing the value would not affect the behavior of the server. Therefore I believe it would be misleading to bake it into API.

PR #3866, opened 2019-03-26
PAYARA-3660: Supply Common Classloader as locator

pdudits:
when CommonModelRegistry is unavailable. Fixes e. g. CDI deployment in Payara Micro, where the definition of annotations are relevant.

The bug workaround mentioned refers to eclipse-ee4j/glassfish-hk2#446


pdudits:
jenkins test please

PR #3757, opened 2019-02-20
Implemented PAYARA-3505 Disable Option for JLine Logging in Asadmin Multimode

fturizo:
In order to prevent unwanted JLine logging in the asadmin multiline mode, a new system property can be used in the respective `asadmin` to disable it:

```
set JAVA=java
:run
%JAVA% -fish.payara.admin.command.jline.log.disable=true -jar "%~dp0..\glassfish\lib\client\appserver-cli.jar" %*
```

PR #3867, opened 2019-03-26
PAYARA-3708 Implemented fixes for JLine Logging

fturizo:
This request is made to complement the changes made by #3757. In particular, the following 2 cases were not considered:

* Running any `asadmin` command outside of multimode
* Running multimode without any domains running.

The gist of the changes was to move the `checkToDisableJLineLogging` method to the `CLICommand` class first and then allow an static initializer to disable the JLine logging before any interaction with a JLine reader is made.


Cousjava:
Jenkins test please

PR #3872, opened 2019-03-28
PAYARA-3468 MP FT Generic types

jbee:
Before I start on 2.0 update I did a quick look through the module addressing warnings.

* Mostly this adds generic types where they where missing,
* removes now unnecessary casts
* removes unnecessary `else` clauses
* makes helper methods `static` where possible

There were a few cases where adding generic type information revealed type inconsistencies. I fixed and marked those with a comment.

Make sure to look at the diff with "ignoring whitespace changes" checked. This makes `else` block changes much less hard to review.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
@@ -108,7 +108,7 @@ public class RetryInterceptor {
             // Attempt to proceed the InvocationContext with Asynchronous semantics if Fault Tolerance is enabled for this
             // method
             if (faultToleranceService.isFaultToleranceEnabled(appName, config)
-                    && ((Boolean) FaultToleranceCdiUtils.getEnabledOverrideValue(
+                    && (FaultToleranceCdiUtils.getEnabledOverrideValue(
                             config, Retry.class, invocationContext)
                             .orElse(Boolean.TRUE))) {
                 // Increment the invocations metric
@@ -127,7 +127,7 @@ public class RetryInterceptor {
             Fallback fallback = FaultToleranceCdiUtils.getAnnotation(beanManager, Fallback.class, invocationContext);
 
             // Only fall back if the annotation hasn't been disabled
-            if (fallback != null && ((Boolean) FaultToleranceCdiUtils.getEnabledOverrideValue(
+            if (fallback != null && (FaultToleranceCdiUtils.getEnabledOverrideValue(
                     config, Fallback.class, invocationContext)
                     .orElse(Boolean.TRUE))) {
                 logger.log(Level.FINE, "Fallback annotation found on method - falling back from Retry");
@@ -182,12 +182,12 @@ public class RetryInterceptor {
         } catch (Exception ex) {
             Class<? extends Throwable>[] retryOn = retry.retryOn();
             try {
-                Optional optionalRetryOn = FaultToleranceCdiUtils.getOverrideValue(
+                Optional<String> optionalRetryOn = FaultToleranceCdiUtils.getOverrideValue(
                         config, Retry.class, "retryOn", invocationContext, String.class);
                 if (optionalRetryOn.isPresent()) {
-                    String retryOnString = ((String) optionalRetryOn.get());
+                    String retryOnString = optionalRetryOn.get();
 
-                    List<Class> classList = new ArrayList<>();
+                    List<Class<?>> classList = new ArrayList<>();
 
                     // Remove any curly or square brackets from the string, as well as any spaces and ".class"es
                     for (String className : retryOnString.replaceAll("[\\{\\[ \\]\\}]", "")
@@ -206,12 +206,12 @@ public class RetryInterceptor {
 
             Class<? extends Throwable>[] abortOn = retry.abortOn();
             try {
-                Optional optionalAbortOn = FaultToleranceCdiUtils.getOverrideValue(
+                Optional<String> optionalAbortOn = FaultToleranceCdiUtils.getOverrideValue(
                         config, Retry.class, "abortOn", invocationContext, String.class);
                 if (optionalAbortOn.isPresent()) {
-                    String abortOnString = (String) optionalAbortOn.get();
+                    String abortOnString = optionalAbortOn.get(); 
 
-                    List<Class> classList = new ArrayList<>();
+                    List<Class<?>> classList = new ArrayList<>();
 
                     // Remove any curly or square brackets from the string, as well as any spaces and ".class"es
                     for (String className : abortOnString.replaceAll("[\\{\\[ \\]\\}]", "")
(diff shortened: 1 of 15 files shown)


appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
```diff
@@ -233,29 +233,29 @@ private Object retry(InvocationContext invocationContext) throws Exception {
...
                 .orElse(retry.maxRetries());
-            long delay = (Long) FaultToleranceCdiUtils.getOverrideValue(
+            long delay = FaultToleranceCdiUtils.getOverrideValue(
                 config, Retry.class, "delay", invocationContext, Long.class)
                 .orElse(retry.delay());
             // Look for a String and cast to ChronoUnit - Use the Common Sense Convertor
-            ChronoUnit delayUnit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
-                config, Retry.class, "delayUnit", invocationContext, String.class)
+            ChronoUnit delayUnit = FaultToleranceCdiUtils.getOverrideValue(
+                config, Retry.class, "delayUnit", invocationContext, ChronoUnit.class)
```
jbee: Note that `String` most likely should have been `ChronoUnit` here and 2x below.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/TimeoutInterceptor.java
```diff
@@ -190,15 +189,15 @@ private Object timeout(InvocationContext invocationContext) throws Exception {
...
         
-        long value = (Long) FaultToleranceCdiUtils.getOverrideValue(
+        long value = FaultToleranceCdiUtils.getOverrideValue(
                 config, Timeout.class, "value", invocationContext, Long.class)
                 .orElse(timeout.value());
         // Look for a String and cast to ChronoUnit - Use the Common Sense Convertor
-        ChronoUnit unit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
-                config, Timeout.class, "unit", invocationContext, String.class)
+        ChronoUnit unit = FaultToleranceCdiUtils.getOverrideValue(
+                config, Timeout.class, "unit", invocationContext, ChronoUnit.class)
```
jbee: Note again, `String` should have been `ChronoUnit`.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
```diff
@@ -233,29 +233,29 @@ private Object retry(InvocationContext invocationContext) throws Exception {
...
                 .orElse(retry.maxRetries());
-            long delay = (Long) FaultToleranceCdiUtils.getOverrideValue(
+            long delay = FaultToleranceCdiUtils.getOverrideValue(
                 config, Retry.class, "delay", invocationContext, Long.class)
                 .orElse(retry.delay());
             // Look for a String and cast to ChronoUnit - Use the Common Sense Convertor
-            ChronoUnit delayUnit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
-                config, Retry.class, "delayUnit", invocationContext, String.class)
+            ChronoUnit delayUnit = FaultToleranceCdiUtils.getOverrideValue(
+                config, Retry.class, "delayUnit", invocationContext, String.class).map(ChronoUnit::valueOf)
```
jbee: Note the `map`. I assume that when loaded from config the `ChronoUnit` is given as `String` and has to be mapped to a `ChronoUnit`. Same two times more below.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/TimeoutInterceptor.java
```diff
@@ -190,15 +189,15 @@ private Object timeout(InvocationContext invocationContext) throws Exception {
...
         
-        long value = (Long) FaultToleranceCdiUtils.getOverrideValue(
+        long value = FaultToleranceCdiUtils.getOverrideValue(
                 config, Timeout.class, "value", invocationContext, Long.class)
                 .orElse(timeout.value());
         // Look for a String and cast to ChronoUnit - Use the Common Sense Convertor
-        ChronoUnit unit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(
-                config, Timeout.class, "unit", invocationContext, String.class)
+        ChronoUnit unit = FaultToleranceCdiUtils.getOverrideValue(
+                config, Timeout.class, "unit", invocationContext, String.class).map(ChronoUnit::valueOf)
```
jbee: Note again the `map` to `ChronoUnit` assuming `config` returns a `String`.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/fallback/FallbackPolicy.java
```diff
@@ -66,18 +67,20 @@
...
-                .orElse(fallback.value().getName()));
-        
-        fallbackMethod = (String) FaultToleranceCdiUtils.getOverrideValue(config, Fallback.class, 
-                "fallbackMethod", invocationContext, String.class)
+        Optional<String> className = FaultToleranceCdiUtils.getOverrideValue(config, Fallback.class, "value", 
+                invocationContext, String.class);
+        fallbackClass = className.isPresent()
+                ? (Class<? extends FallbackHandler<?>>) Thread.currentThread().getContextClassLoader()
+                        .loadClass(className.get())
+                : fallback.value();
```
jbee: Note similar case. Presumably config returns the class name as `String`. If it is present we load it, otherwise use the instance from the annotation. Here `map` cannot be used due to generic capture evaluation by the compiler. Also the conditional avoids calling `fallback.value()` in case override is present which could be relevant in avoiding loading the class.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/validators/FallbackValidator.java
```diff
@@ -65,32 +68,34 @@
...
                 .getOverrideValue(config, Fallback.class, "value", annotatedMethod.getJavaMember().getName(), 
-                annotatedMethod.getJavaMember().getDeclaringClass().getCanonicalName(), String.class)
-                .orElse(fallback.value().getName()));
-        
+                annotatedMethod.getJavaMember().getDeclaringClass().getCanonicalName(), String.class);
+        @SuppressWarnings("unchecked")
+        Class<? extends FallbackHandler<?>> fallbackClass = fallbackClassName.isPresent() 
+            ? (Class<? extends FallbackHandler<?>>) Thread.currentThread().getContextClassLoader()
+                    .loadClass(fallbackClassName.get())
+            : fallback.value();
```
jbee: Note same fix for `Class` loading as above with `String` class name from config.

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
Pandrex247: No I think the original code was simply wrong actually. Should have been:
`ChronoUnit delayUnit = (ChronoUnit) FaultToleranceCdiUtils.getOverrideValue(config, Retry.class, "delayUnit", invocationContext, **ChronoUnit**.class)`

See the this commit that fixes this elsewhere: https://github.com/payara/Payara/commit/8044cf2a5d7265dcb5b2137540a7eb7e41772af6#diff-709d4edc61c9c09ffdaff2ba0f9bc4a7

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/TimeoutInterceptor.java
Pandrex247: See previous comment

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/fallback/FallbackPolicy.java
Pandrex247: For this one I *think* yes it does read it as a String (Afaik anyway).

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/interceptors/RetryInterceptor.java
jbee: Okay, this was actually my first assumption but when I found the case where `Class` was extracted I started to think that the `Config` actually contains the value as `String`. Will change these.

jbee:
jenkins test please

PR #3878, opened 2019-04-01
PAYARA-3464 MicroProfile description does not save

AlanRoth:
The get was using the variable which was null instead of using the string "description" for the key.


cubastanley:
Jenkins test please

AlanRoth:
Jenkins test please

PR #3880, opened 2019-04-03
PAYARA-3674: Preserve Thread Context Class Loader...

pdudits:
in Weld executor service

This fixes the issue with dispatch of CDI asynchronous events. I wonder if this shouldn't be default for Payara Executor Service itself as, API classloader (the default classloader of Payara Executor Service) is rarely useful for anything beyond server startup.


pdudits:
jenkins test please

PR #3887, opened 2019-04-10
PAYARA-3709 Equals sign is added to system properties with empty value even if it's already in the key

AlanRoth:


nucleus/admin/rest/rest-service/pom.xml
```diff
@@ -57,8 +57,8 @@
     
     <description>Rest Interface for GlassFish Management and Monitoring</description>
 
-    <dependencies>
-        <dependency>
+            <dependencies>
```
cubastanley: Weird indentations in this file

AlanRoth: I blame netbeans

MarkWareham:
Alan is making further changes, hence the DO NOT MERGE label

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
```diff
@@ -355,9 +356,24 @@ private String getErrorMessage(Map<String, String> data, ActionReport ar) {
...
+                    if (value.contains("=")) {
+                        value = value.replaceAll("=", "");
+                    }
+
+                }
+
+                if (key != null && !key.trim().isEmpty() && !key.contains("=")) {
+                    if(value != null || !value.trim().isEmpty()){
+                        options.append("=").append(value);
+                    }else{
```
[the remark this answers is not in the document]
AlanRoth: I blame all formatting on netbeans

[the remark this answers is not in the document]
AlanRoth: But should be fixed now

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
```diff
@@ -0,0 +1,94 @@
...
+        data3.put("-Dproduct.name=", "");
+        
+        Map<String, String> data4 = new HashMap<>();
+        data4.put("-Dproduct.name", "XXX");
+        
+        Map<String, String> data5 = new HashMap<>();
+        data5.put("-Dproduct.name", null);
+        
+        Map<String, String> data6 = new HashMap<>();
+        data6.put("-client", "");
```
MarkWareham: two scenarios you missed:
("-Dproduct.name=", null) looks like this would pass
("-client", null)

AlanRoth: Should be added now, pass as expected

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
```diff
@@ -355,8 +356,19 @@ private String getErrorMessage(Map<String, String> data, ActionReport ar) {
                 options.append(sep).append(removeVersioning ? new JvmOption(key).option : key);
 
                 String value = entry.getValue();
-                if (value != null && !value.isEmpty() || key != null && key.startsWith("-D")) {
-                    options.append("=").append(entry.getValue());
+                
+                if (key != null && !key.trim().isEmpty() && key.startsWith("-D")) {    
+                    if (value == null || value.equals("null")) {
```
MarkWareham: Could the text "null" ever be a valid value?

.gitignore
```diff
@@ -27,3 +27,4 @@ appserver/extras/arquillian-containers/payara-common/dependency-reduced-pom.xml
 /nucleus/payara-modules/service-exemplar/nbproject/
 /nucleus/admin/server-mgmt/nbproject/
 appserver/tests/quicklook/quicklook_summary.txt
+/nucleus/admin/rest/rest-service/nbproject/
```
MattGill98: Would you mind removing all lines similar to this and adding a general `nbproject/`? Adding specific directories is generally bad practice, as which is why there's a vast amount of redundancy in this file

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @author asroth
+ */
+public class CollectionLeafResourceTest {
+    
+    public CollectionLeafResourceTest() {
+    }
+
+    @Test
+    public void testProcessData() {
```
MattGill98: I like the use cases, but this is a fairly undescriptive test. Ideally, each test method should contain as close to 1 assertion as possible and should describe exactly what's going wrong so that when an aggregator only lists the method name it's obvious what's gone wrong. Something like `public void when_system_property_has_empty_value_expect_add_to_map()` would fit this criteria.

.gitignore
AlanRoth: Would adding **/nbproject work? I think this got added automatically

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
AlanRoth: Good point

.gitignore
MattGill98: I don't believe it even needs the `**` at the start, although I would check just to be sure!

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
AlanRoth: Changed now

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
AlanRoth: Upon further thought, I think it should be accepted, "null" is a perfectly valid Gaelic name.

AlanRoth:
jenkins test please

PR #3900, opened 2019-04-16
[PAYARA-3417] Flag to use CN as principal-name in glassfish-web.xml

cubastanley:
Implemented a new feature to enable users to optionally use the value of common name as the principal name in an application's glassfish-web.xml for role mapping as opposed to a full string version of the client certificate - will need documentation on passing


nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/certificate/CertificateRealm.java
```diff
@@ -162,11 +168,27 @@ public String getAuthType() {
      * @param principal The Principal object from the user certificate.
      *
      */
-    public void authenticate(Subject subject, Principal principal) {
+    public String authenticate(Subject subject, Principal principal) {
```
arjantijms: At a glance, just wondering if this actually works? The common name is returned here, but not installed in the security context. The returned name is only used for auditing. Am I missing something?

cubastanley: It does work against the written tests in ee7 and against the effective reproducer that I've been working with. It's a change to meet a request made by Ondrej for the sakes of consistency across the authentication logic

arjantijms:
@cubastanley Don't forget that `com.sun.enterprise.security.auth.JaspicToJaasBridge` also uses `CertificateRealm`.

nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/certificate/CertificateRealm.java
arjantijms: @cubastanley I did overlook something indeed, and that's that `name` is used  as the first argument of the security context, my bad, sorry. So that makes it work wrt role mapping. What I was actually looking at however was setting the principal in the Subject to one with the CN as the name". This is pending clarification of the original issue.

arjantijms:
After looking at this a bit better, it seems that the principal name is now inconsistent in the system. Sub-systems that look at the SecurityContext (such as EJB) see the cn only name, but sub-systems that look at the WebPrincipal (such as Servlets) see the full name.

To reproduce this I added this EJB to the existing ee7 client-cert samples test:

```java
@Stateless
public class SecureEJB {

    @Resource
    SessionContext sessionContext;

    public String getEJBPrincipal() {
        return sessionContext.getCallerPrincipal().getName();
    } 
}
```

and updated the existing Servlet in that test to:

```java
@EJB
SecureEJB secureEJB;

@Override
protected void doGet(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {

    response.getWriter().println("\nWeb principal " + request.getUserPrincipal() + " in role g1:" + request.isUserInRole("g1"));

    response.getWriter().println("EJB principal " + secureEJB.getEJBPrincipal());
}
```

The output is:

```
INFO: 

Web principal C=UK, ST=lak, L=zak, O=kaz, OU=bar, CN=lfoo in role g1:true
EJB principal lfoo
```

This is obviously not correct. You should probably look at this bit of code:

com.sun.web.security.RealmAdapter.authenticate:

```java
@Override
    public Principal authenticate(X509Certificate certs[]) {
        if (authenticate(null, null, certs, null)) {
            return new WebPrincipal(certs, SecurityContext.getCurrent());
        }
        return null;
    }
```

And then specifically in the constructor:

```java
  public WebPrincipal(X509Certificate[] certs, SecurityContext context) {
        super(certs[0].getSubjectDN().getName());
        this.certs = certs;
        this.useCertificate = true;
        this.securityContext = context;
    }
```

Maybe adding another constructor with a boolean that instructs it to use the name from the security context instead will fix the issue here.

arjantijms:
Jenkins test

PR #3901, opened 2019-04-16
PAYARA-3754 @Clustered Annotation Doesn't Use Bean Name

MattGill98:


appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/cluster/ClusterScopeContext.java
```diff
@@ -118,8 +118,35 @@ public boolean isActive() {
...
+        if (beanName != null && !beanName.isEmpty()) {
+            return beanName;
+        }
+
+        // Check the bean class name
+        beanName = bean.getBeanClass().getName();
+        if (beanName != null && !beanName.isEmpty()) {
+            return beanName;
+        }
+
```
cubastanley: Is there any way to achieve the same result without using the exact same if statement 3 times? I know it might not be but it would certainly be nicer in my opinion

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/cluster/ClusterScopeContext.java
```diff
@@ -118,8 +118,35 @@ public boolean isActive() {
...
+     * 
+     * @param bean       the bean to reference.
+     * @param annotation the Clustered annotation to reference.
+     * @throws IllegalArgumentException if no name can be found for the bean.
+     */
     static <TT> String getBeanName(Bean<TT> bean, Clustered annotation) {
-        return annotation.keyName().isEmpty()? bean.getName() : annotation.keyName();
+        // Check the annotation keyName
+        String beanName = annotation.keyName();
+        if (beanName != null && !beanName.isEmpty()) {
```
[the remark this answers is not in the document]
AlanRoth: Not sure how important it is but someone might accidentally slap the space bar or something

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/cluster/ClusterScopeContext.java
MattGill98: There is with either something like:

~~~
Stream.of(annotation.keyName(), bean.getName(), bean.getBeanClass().getName())
                .filter(obj -> obj != null && !obj.toString().isEmpty())
                .findFirst()
                .orElseThrow(new IllegalArgumentException("Could not find the name for bean: " + bean.toString()));
~~~

or:

~~~
private static <T> T firstNonNull(T ...items) {
        for(T i : items) if(i != null && !i.toString().isEmpty()) return i;
        throw new IllegalArgumentException("Could not find the name for bean: " + bean.toString());
}
~~~

IMO they both reduce clarity when debugging, although the second is somewhat more palatable. What do you think?

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/cluster/ClusterScopeContext.java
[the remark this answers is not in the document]
MattGill98: Nothing wrong with defensive programming, I like it! I'll incorporate that into the changes I make based on @cubastanley's comments

appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/cluster/ClusterScopeContext.java
cubastanley: I think I'd prefer the latter option to be used. I agree that it may make debugging slightly more involved, although the reduced code repetition will certainly aid in prevention of potential human error in further changes (e.g. someone accidentally changing only one of the if statements) - and I feel this option is still perfectly human readable

AlanRoth: I like the elegance of the stream, and feel like it could be made more readable, maybe by seperating the creation of the list before opening the stream. I like the second solution also, but I see where you're coming from when you mention clarity.

PR #3902, opened 2019-04-16
PAYARA-3757-SPEC-45-Upgrade-jacc-api-to-1.6.1

mulderbaba:


mulderbaba:
jenkins test

arjantijms:
>I think it all looks good. Just one thing to check, does the javax.security.jacc.jar reference in files like appserver/admin/gf_template/src/main/resources/config/default-web.xml have any bearing on this?

The `system-jar-includes` should indeed be updated as well.

mulderbaba:
@arjantijms @MattGill98  `default-web.xml` contains `javax.security.jacc.jar` which is not the artifact I upgraded to. Mine was `javax.security.jacc-api.jar` and updated to `jakarta.security.jacc-api.jar` and yet there is no artifact under Payara with name `javax.security.jacc.jar`, so this is now confusing:)

arjantijms:
@mulderbaba You're right, very confusing that one indeed. Okay, so I think we can leave that one as is, or can remove the invalid entry. Thx!

arjantijms:
Jenkins test please

PR #3905, opened 2019-04-17
PAYARA-3577 Build Payara Server & Payara Micro using JDK8 and get them to pass the Test Suite

jGauravGupta:
Depends on https://github.com/javaee-samples/javaee7-samples/pull/449/

### DIFF
--- a/appserver/admin/gf_template/src/main/resources/config/domain.xml
+++ b/appserver/admin/gf_template/src/main/resources/config/domain.xml
@@ -202,6 +202,12 @@
         <jvm-options>-client</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/jdk.internal.loader=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED</jvm-options>
+        <!-- Hazelcast internal package access requirement to get the best performance results. -->
+        <jvm-options>[9|]--add-exports=java.base/jdk.internal.ref=ALL-UNNAMED</jvm-options>
+        <jvm-options>[9|]--add-opens=java.base/java.lang=ALL-UNNAMED</jvm-options>
+        <jvm-options>[9|]--add-opens=java.base/java.nio=ALL-UNNAMED</jvm-options>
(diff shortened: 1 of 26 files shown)


appserver/admin/gf_template/src/main/resources/config/domain.xml
```diff
@@ -261,6 +257,7 @@
         <jvm-options>[1.8.0u121|1.8.0u160]-Xbootclasspath/p:${com.sun.aas.installRoot}/lib/grizzly-npn-bootstrap-1.7.jar</jvm-options>
         <jvm-options>[1.8.0u161|1.8.0u190]-Xbootclasspath/p:${com.sun.aas.installRoot}/lib/grizzly-npn-bootstrap-1.8.jar</jvm-options>
         <jvm-options>[1.8.0u191|1.8.0u500]-Xbootclasspath/p:${com.sun.aas.installRoot}/lib/grizzly-npn-bootstrap-1.8.1.jar</jvm-options>
+        <jvm-options>[9|]-Xbootclasspath/a:${com.sun.aas.installRoot}/lib/grizzly-npn-bootstrap.jar</jvm-options>
```
smillidge: Why does JDK 11 need the npn-boot jar? Doesn't JDK 11 support NPN negotiation?

jGauravGupta: [Http2AddOn](https://github.com/eclipse-ee4j/grizzly/blob/master/modules/http2/src/main/java/org/glassfish/grizzly/http2/Http2AddOn.java#L121) requires `grizzly-npn-bootstrap.jar#org.glassfish.grizzly.npn.AlpnServerNegotiator` class.

smillidge: OK that may need Grizzly work then as Alpn negotiation I thought was in the SSL stack of the JDK now

smillidge: probably not enough to stop this PR going in but can you raise an issue on Grizzly

jGauravGupta: Added the `grizzly-npn-api.jar` to classpath for JDK 9+

PR #3909, opened 2019-04-18
PAYARA-3784 Fixed OpenIdState equals method

svendiedrichsen:
See sonar finding https://sonarcloud.io/project/issues?id=org.glassfish.main%3Aglassfish-main-aggregator&open=AWoN7ILp9zLXaOU8UM1S&resolved=false&sinceLeakPeriod=true&types=BUG

### DIFF
--- a/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
+++ b/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
@@ -193,7 +193,7 @@ public class OpenIdAuthenticationMechanism implements HttpAuthenticationMechanis
 
             Optional<OpenIdState> expectedState = stateController.get(configuration, httpContext);
             if (expectedState.isPresent()) {
-                if (expectedState.get().equals(receivedState.get())) {
+                if (expectedState.equals(receivedState)) {
                     // (3) Successful Authentication Response : redirect_uri?code=abc&state=123
                     return validateAuthorizationCode(httpContext);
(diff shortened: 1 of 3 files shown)


appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
```diff
@@ -184,18 +186,18 @@ public AuthenticationStatus validateRequest(
...
+        if (receivedState.isPresent()
                 && request.getRequestURL().toString().equals(redirectURI)) {
 
-            OpenIdState expectedState = stateController.get(configuration, httpContext);
-            if (nonNull(expectedState)) {
-                if (expectedState.equals(recievedState)) {
+            Optional<OpenIdState> expectedState = stateController.get(configuration, httpContext);
+            if (expectedState.isPresent()) {
+                if (expectedState.get().equals(receivedState.get())) {
                     // (3) Successful Authentication Response : redirect_uri?code=abc&state=123
```
sgflt: This condition may be simplified to expectedState.equals(recievedState).

If both objects are Optional, equals of Optional delegates call to it's value. As a bonus we will get rid of get call that may throw an exception.

svendiedrichsen: Done. BTW the get() calls in this case couldn't throw an exception because of the isPresent checks.

Cousjava:
Jenkins test please

arjantijms:
Tracked internally by PAYARA-3784

PR #3915, opened 2019-04-24
PAYARA-3781 Use MemberAddressProvider only if --interfaces is not specified

### LINKED ISSUE #3914: Payara Micro --interfaces option does not support wildcard
# Description #
----------

I recently enabled clustering of Payara Micro on Docker Swarm and noticed a peculiar issue when trying to get the containers to work together. When specifying the `--interfaces` value to be used, the command line help suggests wildcards can be used and that the values are comma separated. _I believe the second address below being comma separated is an error._

> --interfaces                                   Specifies the Interfaces that the data grid discovery mode should use to listen can be comma separated e.g. 127.0.0.1,192,168,0,*

The [Hazelcast documentation](https://docs.hazelcast.org/docs/3.12/manual/html-single/index.html#interfaces) indicates that wildcards (192.168.0.*) and ranges (192.168.0.0-15) are supported.

The other issue is that the [MemberAddressProvider](https://docs.hazelcast.org/docs/3.12/manual/html-single/index.html#member-address-provides-spi) is likely unnecessary when `--interfaces` has been specified. I'm not 100% certain of this but my growing Hazelcast knowledge suggests that MemberAddressPicker is in place to identify the public and bind addresses. The public address is already being set directly in the network configuration at [HazelcastCore.java:400](https://github.com/payara/Payara/blob/master/nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/HazelcastCore.java#L400). If `--interfaces` tells Hazelcast what interfaces to use, then the bind address should come from that list of addresses (wildcarded, ranged, or single address). I'm hoping @smillidge has more context.

## Expected Outcome

Specifying `--interfaces` should result in the Hazelcast instance using the interfaces specified and not a random interface selection.

## Current Outcome

If interfaces is specified as a range, `192.168.0.2-15`, the matching interface is **not** selected. The bind address returned from `MemberAddressPicker` is used.

If interfaces is specified as a wildcard, `192.168.0.*`, the matching interface is **not** selected. The bind address returned from `MemberAddressPicker` is used.

To get the proper interface to be selected, every IP address in the range **must** be enumerated in the configuration without whitespace: `192.168.0.2,192.168.0.3,...,192.168.0.15`

## Steps to reproduce 

See the sample `docker-compose.yml` file for a simple deployment example that demonstrates the workaround as well as the expected solution.

### Samples

Deploy the [docker-compose.yml](https://gist.github.com/edthorne/bb3621db9e30b2b4e9eba0db8d018fd4) to a single node Docker Swarm.

## Context

The linked Docker compose file defines a reverse proxy and a set of clustered micro instances. The goal is to have the micro instances communicate with each other only over the cluster network. It uses the stock `payara/micro:5.191` image.

## Environment ##

- **Payara Version**: 5.191
- **Edition**: Micro
- **JDK Version**: OpenJDK 1.8.0_191
- **Operating System**: Linux

edthorne:
Fixes #3914


mulderbaba:
jenkins test

svendiedrichsen:
@edthorne The `MemberAddressPicker` in line 169 uses the provided interfaces list to find a matching one. This list will always be empty. It looks obsolete to me now.

edthorne:
@svendiedrichsen your point about not handling the wildcards is why I'm seeking additional information about the bug. While my solution has solved the problem for me. I'm not 100% certain it's the correct fix. 

My original plan was to add wildcard support to `MemberAddressPicker` and then I realized that the solution provided in the PR _could_ be the correct one. If it is in fact correct, I'll remove the code from `MemberAddressPicker`. The comments at the top of the [findAppropriateInterfaces](https://github.com/payara/Payara/blob/master/nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/MemberAddressPicker.java#L119) method suggest other checks should be taking place but aren't actually implemented.

smillidge:
Thanks for the PR it's appreciated I will review.

To merge we will need a CLA on file see https://github.com/payara/Payara/blob/master/.github/CONTRIBUTING.md

edthorne:
The CLA should be on file for a prior PR submission.

smillidge:
Yes we do have a CLA on file. Apologies

smillidge:
Looks correct to me.

edthorne:
@smillidge do you want me to remove the dead code mentioned earlier?

smillidge:
Yes that would be good thanks

MeroRai:
jenkins test please

PR #3917, opened 2019-04-25
Payara-3482, PAYARA-3742: MP Rest Client 1.2.1 and Jersey 2.29

pdudits:
Two features were added to jaxrs-client-tracing, which is the glue code between opentracing and jersey rest client:

* Rest Client TCK tests that providers are eagerly initialized (before first request). For that, the client decorator needs to support such feature
* Tracing exclusions 

Creating a new module for these looked like a overkill, this way both the initialization and implementation of the features are within single module.

Merging and Jenkins test depend  on payara/Payara_PatchedProjects#243


pdudits:
jenkins test please

PR #3924, opened 2019-05-01
PAYARA-3487 Race condition in ConnectionPool

Cousjava:
When a connectionpool is closed it now logs an error rather than throwing an exception.

A multiexception is printed, showing both the current stack trace and the trace of where
the resource was previously closed.

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ResourceState.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ResourceState.java
@@ -41,11 +41,19 @@
 
 package com.sun.enterprise.resource;
 
+import com.sun.enterprise.resource.pool.ConnectionPool;
+import com.sun.logging.LogDomains;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 public class ResourceState {
     private boolean enlisted;
     private boolean busy;
     private long timestamp;
     private TwiceBusyException busyException;
+    
+    //This is the same logger as in ConnectionPool, used to check the log level
+    private Logger LOGGER = LogDomains.getLogger(ConnectionPool.class,LogDomains.RSR_LOGGER);
 
     public boolean isEnlisted() {
         return enlisted;
@@ -69,7 +77,7 @@ public class ResourceState {
 
     public void setBusy(boolean busy) {
         this.busy = busy;
-        if (!busy) {
+        if (!busy && LOGGER.isLoggable(Level.FINE)) {
             busyException = new TwiceBusyException();
         }
     }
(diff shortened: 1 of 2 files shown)


Cousjava:
Jenkins test please

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ResourceState.java
```diff
@@ -67,6 +69,20 @@ public boolean isBusy() {
 
     public void setBusy(boolean busy) {
         this.busy = busy;
+        if (!busy) {
+            busyException = new TwiceBusyException();
```
pdudits: Can this be controlled by a system property? I believe there is significant overhead in collecting stacktrace at every connection close.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
```diff
@@ -1017,7 +1019,10 @@ public void resourceClosed(ResourceHandle h)
         }
 
         if (!state.isBusy()) {
-            throw new IllegalStateException("state.isBusy() : false");
+            //throw new IllegalStateException("state.isBusy() : false");
+            MultiException noBusyException = new MultiException(state.getBusyStackException());
```
pdudits: And then you will need handle an option of busy stack exception being null. The exception message may suggest to set the system property.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ResourceState.java
```diff
@@ -37,13 +37,15 @@
...
  */
+// Portions Copyright [2019] Payara Foundation and/or affiliates
 
 package com.sun.enterprise.resource;
 
 public class ResourceState {
     private boolean enlisted;
     private boolean busy;
     private long timestamp;
+    private TwiceBusyException busyException;
```
jbee: I don't know much of the context here but race condition sounds like multi-threading which suggests there should thread safe handling somewhere here but I can't see any atomics, `volatile` or `synchronized`. There probably should be?

Cousjava:
Jenkins test please

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/ResourceState.java
jbee: Any comment on my question?

Cousjava: No because the `TwiceBusyException` is used to store details of what has happened to help with tracing where the race condition occurs, there may be a deeper root cause of the issue, but the customer is happy with this.

jbee: So the thread reading and writing the `ResourceState` is always the same?

Cousjava: I think so

PR #3925, opened 2019-05-01
PAYARA-3729 Added missing runtime dependencies

jbee:
Adds the missing runtime dependencies to the client and used the maven flatten plugin that will make the distributed artefact appear as if it is the top most  stand-alone module which makes it cleaner to use the client.

Tested this with both our jersey patched version `2.27.payara-p15`  and `2.28`. 

I attached a maven project with test application to the jira issue. This can be used to deploy (make a jar without the `Client` class) and verify (run `Client` class) the EJB over HTTP.

### DIFF
--- a/appserver/ejb/ejb-http-remoting/client/pom.xml
+++ b/appserver/ejb/ejb-http-remoting/client/pom.xml
@@ -114,10 +114,22 @@
         <dependency>
             <groupId>org.glassfish.jersey.inject</groupId>
             <artifactId>jersey-hk2</artifactId>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.media</groupId>
             <artifactId>jersey-media-json-binding</artifactId>
+            <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>jakarta.json.bind</groupId>
+                    <artifactId>jakarta.json.bind-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/client/pom.xml
```diff
@@ -86,5 +86,13 @@
             <groupId>javax.json</groupId>
             <artifactId>javax.json-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.inject</groupId>
+            <artifactId>jersey-hk2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.media</groupId>
+            <artifactId>jersey-media-json-jackson</artifactId>
```
pdudits: Since the client should depend on JSONB I believe the correct dependency (and probably a runtime one) is `jersey-media-json-binding`

jbee: That worked, thanks for pointing it out.

jbee:
@pdudits I changed the dependency, used the flattened pom and retested.

jbee:
jenkins test please

jbee:
I think in principle this is a valid assumption. However, this would force any client application to override the scope for some of the transitive dependencies, e.g. in case they have annotated entity types what makes using our client less convenient. Therefore we might not set them `runtime`.

jbee:
jenkins test please

jbee:
jenkins test please

PR #3928, opened 2019-05-02
PAYARA-3468 Added a config property to set alternative @Asynchronous annotations

jbee:
@arjantijms pointed out that projects might want to use other `@Asynchronous` annotations. As this is specific a fixed list of supported annotations seemed less feasible. Also some customers might not want to have such a feature enabled by default. Therefore I think the best solution is to allow to configure alternative annotation classes that should have the same effect as FT's `@Asynchronous`.

With this PR user can do so by defining the classes for the added key `MP_Fault_Tolerance_Alternative_Asynchronous_Annotations` as a comma separated list beside other FT configurations.

Implementation wise a bit of duplication cannot be avoided since the marking has to recognise the alternative annotations as well as the `FaultTolceranceConfig` which is not available at that stage.

I added tests that make sure one or two (many) alternatives can be set while FT's own `@Asynchronous` is still be recognised as well.


jbee:
jenkins test please

jbee:
TCK tested and passed locally.

arjantijms:
jenkins test please

PR #3943, opened 2019-05-13
[PAYARA-3812] - System properties unable to be updated

cubastanley:
Issue regarding system properties being unable to be created, updated or deleted through the admin console.


cubastanley:
> Tested and works locally, however, looking at it, it feels like the post method at line 883 is doing a very similar thing to your solution, so the problem might just be the wrong post method is being used somewhere. Any ideas?

It's very similar but not quite the same - firstly the call is made in a separate script dependant on the way the data is provided (it's tricky to explain but understandable if you look at it) Second, the above method takes a parameter for the media type and additionally converts an object of type Map into a MultivalueMap which, from what I've done on this, isn't compatible with the application/json media type that the server is asking for - hence the use of payload instead of formData in my changes

cubastanley:
Jenkins test please

cubastanley:
Further changes relating to this made after a subsequent bug was discovered, see: https://github.com/payara/Payara/pull/3960

PR #3931, opened 2019-05-03
PAYARA-3654 Client-side adapter support for EJB HTTP Client

pdudits:
This implements the proposed solution of adding client side API for creating adapters - application specific proxies, that emulate stateful behavior, and in the end delegate to a stateless remote call.

The PR is best reviewed commit by commit:

38fde3a defines the API, javadoc and some basic tests
967d8b0 is  a clenup of previous api
26fc84e demonstrates how the adapter can be used for emulating JMS api on client side, where remote part implements single method `sendTextMessageToQueue`

PR #3945, opened 2019-05-13
PAYARA-3755: Improve JSON-B support in ejb-http-client

pdudits:
This builds upon #3931 (and the diff will get simpler once that one is merged) and payara/patched-src-yasson#3 in order to handle all of the test cases present in payara/payara-samples#9. Needs payara/Payara_PatchedProjects#244 to build properly.

512fbc7  handles class names especially for primitive types and null
7148154 adds dependency of client to same version of Yasson as in server, because all public Yasson versions at this time have issues with handling primitive types (and null)
177d55b maintains proper generic type of method parameters, so that JSON-B can properly deserialize parameters of types `List<Pojo>`, instead of creating `List<Map<String,Object>>` instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ nb-configuration.xml
 appserver/extras/arquillian-containers/payara-common/dependency-reduced-pom.xml
 appserver/tests/quicklook/quicklook_summary.txt
 **/nbproject
+.flattened-pom.xml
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -24,3 +24,4 @@ nb-configuration.xml
 appserver/extras/arquillian-containers/payara-common/dependency-reduced-pom.xml
 appserver/tests/quicklook/quicklook_summary.txt
 **/nbproject
+.flattened-pom.xml
```
Pandrex247: Thank you

pdudits:
jenkins test please

Issue #3937, opened 2019-05-08
OpenAPI Generation refers to the base class instead of the extending class


deratzmann:
# Description #
----------

## Expected Outcome
We have a REST service with a response object inheriting from a base class. I expect that when I generate the service specification with OpenAPI that in the response section (response code 200) the extending class is referred.

## Current Outcome

In the response section (response code 200) the base class is referred (see txt file).

## Steps to reproduce (Only for bug reports)
We build a simple project to reproduce.

## Environment ##
- **Payara Version**: 5.184
- **Edition**: Full
- **JDK Version**: OpenJDK8
[testopenapi.zip](https://github.com/payara/Payara/files/3157301/testopenapi.zip)
[generated_openapi.txt](https://github.com/payara/Payara/files/3157302/generated_openapi.txt)

jbee:
Thanks for reporting the problem. 

Had a quick look and based on the example and output given I am fairly sure this is a defect. Since we updated OpenAPI after 5.184 we should first see if the problem still exists in 5.192.

jbee:
Made the example into a test. Problem still present in 5.192.

jbee:
Created internal ticket https://payara.atlassian.net/browse/PAYARA-3816

jbee:
@deratzmann I am no longer sure this is a defect. If a schema is referenced using `implementation` the schema created for the annotation uses the `implementation` class as basis of the created schema. This class inherits from `FooResponseBase` wherefore it has a `allOf` reference to that base class. 

To literally refer to  the schema `FooResponse` I think the annotation should use `ref`:
```java
@Schema(ref = "FooResponse")
```

jbee:
@deratzmann Please have a look at my comment above - do you still think this is a defect or does this clarify why you get the results and how to fix it?

deratzmann:
Yes, that fixes our issue. Thank you, very much.

PR #3946, opened 2019-05-13
PAYARA-3816 OpenAPI Generation does not refer to extended base class

jbee:
This is start of a PR where the issue reported by #3937 should be fixed. For this only reproduces the issue in a (now failing) test.


jbee:
This PR is not complete (in progress). I just created the test from the example given in #3937 to confirm if the problem is still present in current version. This is meant to be used as a basis for a fix but so far it just shows the problem through a failing test.

jbee:
No change needed. #3937 was a misunderstanding of the effects of `implementation`.

PR #3950, opened 2019-05-15
[PAYARA-3464] - Microprofile Property Descriptions not saving

cubastanley:
Actually killed two birds with one stone to do this one, now introduced support for creation of properties with dotted names (i.e. DOTTED.NAME.PROPERTY) through the admin console as this was the main reason behind MP property descriptions not being able to be saved.


appserver/ejb/ejb-http-remoting/client/.flattened-pom.xml
```diff
@@ -0,0 +1,183 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pdudits: We don't want this file committed (.gitignore entry is on its way...)

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/PropertiesBagResource.java
```diff
@@ -189,27 +190,33 @@ protected ActionReportResult clearThenSaveProperties(List<Map<String, String>> p
 
                 // the prop name can not contain .
                 // need to remove the . test when http://java.net/jira/browse/GLASSFISH-15418  is fixed
-                boolean canSaveDesc = !((Object)property.get("name")).toString().contains(".");
+                boolean isDottedName = ((Object)property.get("name")).toString().contains(".");
```
pdudits: Can we remove this Object casting madness here? Basically every cast to Object in this method is unnecessary (also when disguised as `Object.class.cast` above).

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/PropertiesBagResource.java
```diff
@@ -189,27 +190,33 @@ protected ActionReportResult clearThenSaveProperties(List<Map<String, String>> p
...
+                    
                     data.put(escapedName, ((Object)property.get("value")).toString());
-                    if (canSaveDesc && (description != null)) {
-                        data.put(escapedName + ".description", ((Object) description).toString());
+                    
+                    if (isDottedName) {
+                        data.put(unescapedName + ".name", unescapedName);
                     }
+                    
+                    if ((description != null)) {
```
pdudits: There are also lots of unnecessary parenthesis in this method, basically no expression needs them (since casts will be removed).

cubastanley:
Jenkins test please

PR #3952, opened 2019-05-15
PAYARA-3314 Add Concept of Docker Nodes (Tech Preview)

Pandrex247:
Adds the concept of a Docker Node to Payara Server.

A Docker Node is a representation of a physical machine onto which docker containers will be spawned, inside of which will reside a Payara Server instance. These instances should function as "standard" remote instances, they simply exist within a docker container. 

Each node is tied to a specific Docker image, so if you would like to run different images on the same machine, you will need to create multiple nodes for it.

As the underlying communication mechanism with Docker is done via the REST API, you will need to expose the docker port to HTTP. A guide on securing this exposed port can be found here: https://docs.docker.com/engine/security/https/.
If following this linked guide, you will then need to: configure the node to use TLS (configurable attribute); add the CA pem to the Payara Server cacerts; create a pkcs bundle of the client cert and key; add the pkcs bundle to the Payara Server keystore.

Several new asadmin commands have been added, but efforts have been made to keep them as similar in functionality to the existing ones (e.g. `start-instance` will start the Docker container and the instance within it)

A default docker image has been created (still needs publishing), but the option is available to provide a custom one - it must however roughly conform to the default image.

Configuration of the Docker containers is also available, and is done using system properties. This allows you to specify them in a config, and override them at the instance level as desired. Configuration options can be found in the Docker Rest API: https://docs.docker.com/engine/api/v1.39/#operation/ContainerCreate. These Json properties match to system properties in the following format:

* All options must be prefixed with `Docker`
* Nested options follow the dotted name format used by system properties (e.g. to configure `{ HostConfig { Memory : 0 } }` the equivalent system property would be `Docker.HostConfig.Memory=0`
* Values of array must be surrounded with `[]`, the split operator is `|`, and the split operator for any objects within the array is `,` (e.g. to configure `{ HostConfig { Mounts [ {Type: Bind, Source: /foo/bar, Target: /opt/foo, ReadOnly: true},  {Type: Bind, Source: /bar/foo, Target: /foo/opt, ReadOnly: true} ] } }` the equivalent system property would be `Docker.HostConfig.Mounts=[Type:Bind,Source:/foo/bar,Target:/opt/foo,ReadOnly:true|Type:Bind,Source:/bar/foo,Target:/foo/opt,ReadOnly:true]`
* Due to a low-level restriction of not having `=` in the value of a system property, it is replaced with `:`, which means configuration of `{ Env: [ foo=bar ] }` would look like so: `Docker.Env=[foo:bar]`


rdebusscher:
The Web Admin console screen info text for "Node Host" doesn't mention that it is required for DOCKER type. But when not filled in, a validation popup is shown.

appserver/docker/src/main/java/fish/payara/docker/instance/admin/CreateDockerContainerCommand.java
```diff
@@ -0,0 +1,578 @@
...
+                propertyComponentObjectBuilders.putIfAbsent(propertyComponent, Json.createObjectBuilder());
+            }
+        }
+
+        // Add lowest level property component to immediate parent builder (second last in list)
+        String immediateParent = propertyComponents.get(propertyComponents.size() - 2);
+
+        // Use the passed in object builder if the immediate parent is the same as the previous property
+        JsonObjectBuilder immediateParentObjectBuilder;
+        if (parent != null && immediateParent.equals(parent)) {
```
rdebusscher: Not null check not needed, equals will handle it fine.

appserver/docker/src/main/java/fish/payara/docker/instance/admin/CreateDockerContainerCommand.java
```diff
@@ -0,0 +1,578 @@
...
+        String dasPort = "";
+        for (Server server : servers.getServer()) {
+            if (server.isDas()) {
+                dasHost = server.getAdminHost();
+                dasPort = Integer.toString(server.getAdminPort());
+                break;
+            }
+        }
+
+        if (dasHost == null || dasHost.equals("") || dasPort == null || dasPort.equals("")) {
```
rdebusscher: `dasPort` can never be null here

appserver/docker/src/main/java/fish/payara/docker/instance/admin/CreateDockerContainerCommand.java
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    /**
+     * Go over all system properties and add them to the Json object.
+     *
+     * @param rootObjectBuilder The top-level object builder that will contain all Docker configuration
+     * @param dasHost The IP address that the DAS is running on
+     * @param dasPort The admin port of the DAS
+     */
+    private void translatePropertyValuesToJson(JsonObjectBuilder rootObjectBuilder, String dasHost,
```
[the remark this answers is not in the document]
Pandrex247: Yeah I've been pondering the best way to test these, as ideally I'd like to ping the resultant JSON request against Docker to validate it, but that requires setup etc.

I suppose the best way is to just do Json String matching.

Pandrex247:
@pdudits Added a few tests for the Json construction.

appserver/extras/docker-images/src/main/resources/payaraserver-node/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG PAYARA_INSTALL
+COPY --chown=payara:payara ${PAYARA_INSTALL} ${PAYARA_HOME}
+RUN rm -rf ${PAYARA_HOME}/glassfish/domains/domain1/ && \
+    rm -rf ${PAYARA_HOME}/glassfish/domains/production
+
+# Install entrypoint script
+ARG ENTRYPOINT_SCRIPT
+COPY --chown=payara:payara ${ENTRYPOINT_SCRIPT} ${WORK_DIR}
+RUN chmod +x ${WORK_DIR}/entrypoint.sh
+
```
rdebusscher: The ARG ENTRYPOINT_SCRIPT takes any name as file name, but RUN command assumes it is `entrypoint.sh`

This is only internal usage, so probably a comment in the pom.xml file will do.

Pandrex247:
Jenkins test please

PR #3957, opened 2019-05-16
PAYARA-3805 Cannot edit, add, or delete JVM options from Admin Console

AlanRoth:
The problem was caused by '-XX:NewRatio=2' as previously it was assumed -X command-line options didn't require a value like in -D command-line options.

### DIFF
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
@@ -352,10 +352,10 @@ public abstract class CollectionLeafResource extends AbstractResource {
                 results.put(key, entry.getValue());
             } else {
                 options.append(sep).append(removeVersioning ? new JvmOption(key).option : key);
-
+                
                 String value = entry.getValue();
                 
-                if (key != null && !key.trim().isEmpty() && key.startsWith("-D")) {    
+                if (key != null && !key.trim().isEmpty() && (key.startsWith("-D") || key.startsWith("-X"))) {    
                     if (value == null) {
                         value = "";
                     } else if(value.contains("=")) {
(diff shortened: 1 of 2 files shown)


nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
```diff
@@ -352,10 +352,10 @@ private String getErrorMessage(Map<String, String> data, ActionReport ar) {
                 results.put(key, entry.getValue());
             } else {
                 options.append(sep).append(removeVersioning ? new JvmOption(key).option : key);
-
+
```
cubastanley: tabz

Pandrex247:
Jenkins test please

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/resources/CollectionLeafResource.java
AlanRoth: Spaces ;)

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
```diff
@@ -122,4 +122,29 @@ public void when_property_is_client_null_value_expect_add_to_map(){
         data8.put("-client", null);
         assertEquals(wrapper.getProcessedData(data8).get("id"), "-client");
     } 
+    
+    @Test
+    public void when_property_is_Xproperty_with_value_expect_add_to_map(){
+        Wrapper wrapper = new Wrapper();
+        Map<String, String> data9 = new HashMap<>();
+        data9.put("-XX:NewRatio=", "2");
+        assertEquals(wrapper.getProcessedData(data9).get("id"), "-XX:NewRatio=2");
```
jbee: I think expected and actual are swapped here (and for the other tests).

Pandrex247:
@jbee It's using TestNg assert equals rather than Junit assert equals - they take _expected_ and _actual_ in reverse of each other (which is just lovely and will never lead to confusion)

nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/resources/CollectionLeafResourceTest.java
AlanRoth: TestNG assert equals just likes to be different - caught me out the first time. 😄

PR #3959, opened 2019-05-16
PAYARA-3798 Create zip file system when processing nested archives

jbee:
**Description**
The issue seems to be that we use the nested jar file without ever creating the `FileSystem` for it as one should according to https://docs.oracle.com/javase/7/docs/technotes/guides/io/fsp/zipfilesystemprovider.html and similar problems reported on SO. 

**Reviewer notes**
I can't say why this would work prior to Java 11. I am also unsure where best to create the `FileSystems.newFileSystem`. In the launcher where I put it it ensures that we do it just for the payara jar  and the method effectively isn't called multiple times but it does feel quite right place.

**Testing performed**
I first ran the reproducer 
```bash
java -jar payara-micro.jar --deploy <some.war> --nested
```
getting the error, added the fix and rebuild the jar and ran the reproducer again without getting the error.

I also tested that the code does not run when not using `--nested`.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExecutableArchiveLauncher.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExecutableArchiveLauncher.java
@@ -17,7 +17,10 @@
 
 package fish.payara.micro.boot.loader;
 
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


jbee:
jenkins test please

jbee:
@pdudits linked the origin to this change https://github.com/payara/Payara/commit/74680274c956bdb57975f4965449a0c1239c4350#diff-5bdbe3594bddef2a0c939928359ecab2 . Using `Paths.get(url.toURI()).toString()` instead of `url.toURI().getPath()` was a fix to irregular parsing of URLs that depending on the way it is constructed does or doesn't accept certain URLs. 

I looked into ways to avoid `Paths.get(...)` (which expects a file system for the jar) but could not find one that would not cause the URL parsing problem.

PR #3960, opened 2019-05-17
[PAYARA-3812] - Fixed issue with checkboxes causing errors due to changes

cubastanley:
Original changes were too generalised and caused issues with checkbox values being saved - created a special case for the saving of system properties to amend this

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/RestUtil.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/RestUtil.java
@@ -893,23 +893,18 @@ public class RestUtil {
 
     public static RestResponse post(String address, Map<String, Object> payload) {
         WebTarget target = getJerseyClient().target(address);
-        Response cr;
-        
-        if(address.endsWith("system-properties")) {
-        cr = target
+        Response cr = target
                 .request(RESPONSE_TYPE)
                 .cookie(new Cookie(REST_TOKEN_COOKIE, getRestToken()))
-                .post(Entity.entity(payload, MediaType.APPLICATION_JSON), Response.class);
-        } else {
-            MultivaluedMap formData = buildMultivalueMap(payload);
-            cr = target
-                .request(RESPONSE_TYPE)
-                .cookie(new Cookie(REST_TOKEN_COOKIE, getRestToken()))
-                .post(Entity.entity(formData, MediaType.APPLICATION_FORM_URLENCODED), Response.class);
-        }
+                .post(getEntityForAddress(address, payload), Response.class);
         
         RestResponse rr = RestResponse.getRestResponse(cr);
         return rr;
+    }
+    
+    private static Entity getEntityForAddress(String address, Map<String, Object> payload) {
+        return address.endsWith("system-properties") ? Entity.entity(payload, MediaType.APPLICATION_JSON)
+                : Entity.entity(buildMultivalueMap(payload), MediaType.APPLICATION_FORM_URLENCODED);
     }
 
     public static RestResponse put(String address, Object payload, String contentType) {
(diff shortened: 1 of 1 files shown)


cubastanley:
Jenkins test please

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/RestUtil.java
```diff
@@ -893,11 +893,21 @@ public static RestResponse post(String address, Object payload, String contentTy
 
     public static RestResponse post(String address, Map<String, Object> payload) {
         WebTarget target = getJerseyClient().target(address);
-        // Map formData = buildMultivalueMap(payload); Commented for sakes of reference in future rather that removal
-        Response cr = target
+        Response cr;
+        
+        if(address.contains("system-properties")) {
```
Pandrex247: Might be better as `endsWith` actually

cubastanley:
Jenkins test please

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/RestUtil.java
```diff
@@ -893,11 +893,21 @@ public static RestResponse post(String address, Object payload, String contentTy
...
                 .request(RESPONSE_TYPE)
                 .cookie(new Cookie(REST_TOKEN_COOKIE, getRestToken()))
                 .post(Entity.entity(payload, MediaType.APPLICATION_JSON), Response.class);
+        } else {
+            MultivaluedMap formData = buildMultivalueMap(payload);
+            cr = target
+                .request(RESPONSE_TYPE)
+                .cookie(new Cookie(REST_TOKEN_COOKIE, getRestToken()))
+                .post(Entity.entity(formData, MediaType.APPLICATION_FORM_URLENCODED), Response.class);
+        }
```
pzygielo: This else branch looks very similar to the first one. Why not inline condition check in `post` method's 1st argument (and then extract it to some method) and have `cr` initialized on definition?

cubastanley: I think it would also be fine to do it that way, although given the if case is true there is no need to create the formData Map instance as it would never be used. Personally, from a readability point of view I feel it's nicer this way as comparatively the line would get much longer using an inline condition

pzygielo: > I think it would also be fine to do it that way, although given the if case is true there is no need to create the formData Map instance as it would never be used.

With `?` `formData` would also be created only when needed.

> Personally, from a readability point of view I feel it's nicer this way as comparatively the line would get much longer using an inline condition

With method extracted to do that it would be more readable IMO.
Pseudo:
```
...
   val cr = target
                .request(RESPONSE_TYPE)
                .cookie(new Cookie(REST_TOKEN_COOKIE, getRestToken()))
                .post(getEntityForAddress(address, payload), Response.class);

   RestResponse rr = RestResponse.getRestResponse(cr);
   return rr;
}

getEntityForAddress(address, payload) {
    return address.condition() ? payload :  buildMultivalueMap(payload);
}
```

Then one could provide test for `getEntityForAddress` (which would be easier to test that decision making).

cubastanley: Made that change now I know what you mean properly, much nicer way to do it, thanks!

Pandrex247:
Jenkins test please

pzygielo:
So [I learnt](https://help.github.com/en/articles/about-required-reviews-for-pull-requests) that my *approval* is meaningless anyway :wink: (for meeting repo's PR requirements). Thus just treat it as confirmation that I find my comment as addressed.

PR #3961, opened 2019-05-17
PAYARA-3521: Only trace JAX-RS Client requests when tracing services are present

pdudits:
Test case in payara-samples#11

Root cause was, that open tracing decorator is present as ServiceLoader-visible service and intercepts creation of JAX-RS client, even if request tracing service (or any other service, as Payara is not running) is not available.


Pandrex247:
Jenkins test please

PR #3964, opened 2019-05-20
PAYARA-3521: Check for request tracing services without side effects

pdudits:
Calling `getService` would start the service, even if the runtime is not ready yet


Pandrex247:
Jenkins test please

PR #3966, opened 2019-05-21
PAYARA-3385 Move JAX-WS tracing into its own module

Pandrex247:
Gets around web dist not having required JAX-WS dependencies


Pandrex247:
Jenkins test please

PR #3974, opened 2019-05-23
Payara-3862: Only return tracing scope when there is active span

pdudits:


pdudits:
jenkins test please

nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/ScopeManager.java
```diff
@@ -55,7 +55,7 @@
 public class ScopeManager implements io.opentracing.ScopeManager {
```
Cousjava: Copyright year need updating

Cousjava:
Jenkins test please

PR #3975, opened 2019-05-23
 PAYARA-3101 MicroProfile endpoints security support

jGauravGupta:
MicroProfile endpoints (/metrics, /health, and /openapi) are configurable through both the asadmin commands and admin console. Here demonstrating the endpoints security configuration via Metrics example.

To enable the security of metrics endpoint, Run the following asadmin command in your terminal:
```
asadmin> set-metrics-configuration --securityenabled=true 
```
When `securityenabled` is set to true, a username and password are required to access Metrics endpoint. By default, the username is set to `mp` and the password is set to `mp` with the role `microprofile`. The user `mp` is defined in the `file` realm (${domain}/config/keyfile).

To provide the access to the Metrics endpoint to any user, User must be defined in the file realm and be part of the microprofile group.

If the application deployed to root context then microprofile endpoints can share the security config of app.

Test PR in payara/payara-samples#12


jGauravGupta:
jenkins test please

arjantijms:
> If the application deployed to root context then microprofile endpoints can share the security config of app.

As a clarification, the endpoints then MUST share the security config with whatever application is deployed to root, as there can be only one security configuration.

arjantijms:
I had rather seen independent security configuration for the endpoints, but for now it does solve the issue so approving.

arjantijms:
@rdebusscher Can you verify that the approach is okay for you? The full review (code and such) can be let to @jbee.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/Constants.java
```diff
@@ -0,0 +1,48 @@
...
+ *     either the CDDL, the GPL Version 2 or to extend the choice of license to
+ *     its licensees as provided above.  However, if you add GPL Version 2 code
+ *     and therefore, elected the GPL Version 2 license, then the option applies
+ *     only if the new code is made subject to such option by the copyright
+ *     holder.
+ */
+
+package fish.payara.microprofile.healthcheck;
+
+public class Constants {
```
jbee: Not sure if we have any convention on this but personally I'd make a constant file an `interface` purely because it is more readable (`public static final` can be skipped) and classes could opt to implement the interface to have constants in scope.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
```diff
@@ -107,13 +114,27 @@
...
     public void execute(AdminCommandContext context) {
         ActionReport actionReport = context.getActionReport();
-
+        Subject subject = context.getSubject();
         Config targetConfig = targetUtil.getConfig(target);
         MetricsHealthCheckConfiguration config = targetConfig.getExtensionByType(MetricsHealthCheckConfiguration.class);
 
+        // Create the default user if it doesn't exist
+        if (!defaultMicroprofileUserExists(actionReport.addSubActionsReport(), subject)) {
+            createDefaultMicroprofileUser(actionReport.addSubActionsReport(), subject);
```
jbee: Using `addSubActionsReport` twice instead of once sounds wrong to me. If it actually adds (modifies state) I believe this should only happen once.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
```diff
@@ -133,6 +157,46 @@ public void execute(AdminCommandContext context) {
...
+
+        ParameterMap parameters = new ParameterMap();
+        parameters.add("authrealmname", "file");
+
+        invocation.parameters(parameters).execute();
+
+        for (ActionReport.MessagePart message : subActionReport.getTopMessagePart().getChildren()) {
+            if (message.getMessage().equals(DEFAULT_USER_NAME)) {
+                exists = true;
+                break;
```
jbee: Not a big fan of `break` in loops as control flow is harder to follow then the alternative of doing a `return true` here and a `return false` at the end of the method instead.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
```diff
@@ -133,6 +157,46 @@ public void execute(AdminCommandContext context) {
...
+        return exists;
+    }
+
+    private void createDefaultMicroprofileUser(ActionReport subActionReport, Subject subject) {
+        CommandRunner.CommandInvocation invocation = commandRunner.getCommandInvocation("create-file-user", subActionReport, subject,
+                false);
+
+        ParameterMap parameters = new ParameterMap();
+        parameters.add("groups", DEFAULT_GROUP_NAME);
+        parameters.add("userpassword", "mp");
```
jbee: Don't know what the "userpassword" is but is sounds like a password of some kind - is it meant to appear in code (safe to be known by everyone)?

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/servlet/HealthCheckServletContainerInitializer.java
```diff
@@ -95,6 +99,10 @@ public void onStartup(Set<Class<?>> c, ServletContext ctx) throws ServletExcepti
             // Register servlet
             ServletRegistration.Dynamic reg = ctx.addServlet("microprofile-healthcheck-servlet", HealthCheckServlet.class);
             reg.addMapping("/" + configuration.getEndpoint());
+            if (Boolean.valueOf(configuration.getSecurityEnabled())) {
```
jbee: Minor improvment: `Boolean.parseBoolean` avoids box/unbox.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/MetricsServiceConfiguration.java
```diff
@@ -98,4 +98,13 @@
     String getVirtualServers();
     void setVirtualServers(String value) throws PropertyVetoException;
 
+    /**
+     * Boolean value determining if the security is enabled or not.
+     *
+     * @return
```
jbee: I also like to keep javadoc short if there is little to say. In would in such case move the above line after the `@return` since its mostly describing the return value. Not sure if we have a guide on it.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/SetMetricsConfigurationCommand.java
```diff
@@ -103,16 +110,31 @@
...
         ActionReport actionReport = context.getActionReport();
-
+        Subject subject = context.getSubject();
         Config targetConfig = targetUtil.getConfig(target);
         MetricsServiceConfiguration metricsConfiguration = targetConfig.getExtensionByType(MetricsServiceConfiguration.class);
         MetricsService metricsService = Globals.getDefaultBaseServiceLocator().getService(MetricsService.class);
+
+        // Create the default user if it doesn't exist
+        if (!defaultMicroprofileUserExists(actionReport.addSubActionsReport(), subject)) {
+            createDefaultMicroprofileUser(actionReport.addSubActionsReport(), subject);
```
jbee: `addSubActionsReport` twice again. Please check.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/SetMetricsConfigurationCommand.java
```diff
@@ -154,6 +180,46 @@ public void execute(AdminCommandContext context) {
...
             actionReport.failure(LOGGER, "Failed to update Metrics configuration", ex);
         }
+
+        // If everything has passed, scrap the subaction reports as we don't want to print them out
+        if (!actionReport.hasFailures() || !actionReport.hasWarnings()) {
+            actionReport.getSubActionsReport().clear();
+        }
+    }
+
+    private boolean defaultMicroprofileUserExists(ActionReport subActionReport, Subject subject) {
```
jbee: Both `defaultMicroprofileUserExists` and `createDefaultMicroprofileUser` did occur before. Can and should be shared for them?

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
jGauravGupta: Sub ActionReport new instance is required for both function `defaultMicroprofileUserExists` and `createDefaultMicroprofileUser` to invoke `list-file-users` and `create-file-user` commands

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
jGauravGupta: When `securityenabled` is set to true, a username and password are required to access MicroProfile endpoints. By default, the username & password is set to `mp` and defined in the `file` realm.

jbee: I understand. I just wonder how good a security is when everyone can read the password in our public sources. So do we expect the default to practically never be used? Wouldn't it be more secure to force configuration of a password in case security is enabled?

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
```diff
@@ -621,7 +620,8 @@ public WebModuleConfig createSystemDefaultWebModuleIfNecessary(WebArchivist webA
             WebBundleDescriptorImpl webBundleDescriptor = webArchivist.getDefaultWebXmlBundleDescriptor();
             webBundleDescriptor.setModuleID(DEFAULT_WEB_MODULE_NAME);
             webBundleDescriptor.setContextRoot("");
-            
+            webBundleDescriptor.getLoginConfiguration().setRealmName(DEFAULT_WEB_MODULE_REALM);
+
```
pdudits: `file` is not always a good default. I think the correct default is whatever is specified in `SecurityService#getDefaultRealm` of current server's config.

appserver/payara-appserver-modules/microprofile/microprofile-common/src/main/java/fish/payara/microprofile/SetSecureMicroprofileConfigurationCommand.java
```diff
@@ -0,0 +1,108 @@
...
+                );
+
+        ParameterMap parameters = new ParameterMap();
+        parameters.add("groups", DEFAULT_GROUP_NAME);
+        parameters.add("userpassword", "mp");
+        parameters.add("target", target);
+        parameters.add("authrealmname", "file");
+        parameters.add("DEFAULT", DEFAULT_USER_NAME);
+
+        invocation.parameters(parameters).execute();
```
pdudits: If the command fails, does it fail the parent command as well?

The user might have chosen not to use default realm, and even remove `file` from the realms altogether.

rdebusscher:
> @rdebusscher Can you verify that the approach is okay for you? 

If it is by default off, then I assume it is OK when they use the maven/gradle plugins (since they deploy app on root)

PAYARA-3101 is not only about /metrics, so I assume other PRs will be created for the other MP endpoints? @jGauravGupta 

@OndroMih You worked with BMW on that ticket, what is you opinion of this solution?

@arjantijms  I'll try to have a more detailed look this evening ...

jGauravGupta:
Hi @rdebusscher, 

This PR covers all 3 endpoints /metrics, /health, and /openapi.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/GetMPHealthCheckConfiguration.java
```diff
@@ -79,7 +79,7 @@
 })
 public class GetMPHealthCheckConfiguration implements AdminCommand {
 
-    private final String OUTPUT_HEADERS[] = {"Enabled", "EndPoint", "VirtualServers"};
+    private final String OUTPUT_HEADERS[] = {"Enabled", "EndPoint", "VirtualServers", "Security Enabled"};
```
rdebusscher: C-Style Array declaration

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/GetMetricsConfigurationCommand.java
```diff
@@ -82,7 +82,7 @@
 })
 public class GetMetricsConfigurationCommand implements AdminCommand {
 
-    private final String OUTPUT_HEADERS[] = {"Enabled", "SecureMetrics", "Dynamic", "EndPoint", "VirtualServers"};
+    private final String OUTPUT_HEADERS[] = {"Enabled", "SecureMetrics", "Dynamic", "EndPoint", "VirtualServers", "Security Enabled"};
```
rdebusscher: C-Style Array declaration

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/admin/GetOpenApiConfigurationCommand.java
```diff
@@ -76,7 +76,7 @@
 })
 public class GetOpenApiConfigurationCommand implements AdminCommand {
 
-    private final String OUTPUT_HEADERS[] = {"Enabled", "VirtualServers", "CorsHeaders"};
+    private final String OUTPUT_HEADERS[] = {"Enabled", "VirtualServers", "CorsHeaders", "Security Enabled"};
```
rdebusscher: C-Style array declaration

arjantijms:
Jenkins test please

jGauravGupta:
Jenkins test please

jGauravGupta:
Jenkins test please

OndroMih:
The behavior described by @jGauravGupta looks good to me and it's according to what the customer expects.

I would rename the new `--securityenabled` argument to `--authorizationenabled`. The current name is confusing. There's already a `--securemetrics` argument, which triggers SSL encryption.  We already use the word "security" in various cases, and very often it means encryption (secure admin, secure metrics, secure HTTP listener). I suggest using terms like "restricted access",  "authorized access" or "authorization" for things related to allowing access only for privileged users and avoid using the word "security" in these cases.

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
```diff
@@ -69,13 +69,17 @@ faultTolerance.configuration.managedScheduledExecutorServiceNameHelp=The JNDI na
 
 healthCheck.configuration.enabled=Enabled
 healthCheck.configuration.enabledHelp=Enables or Disables the Health Check.
+healthCheck.configuration.securityEnabled=Enable Secure Access
```
OndroMih: Change the text to "Require authorization"

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
```diff
@@ -69,13 +69,17 @@ faultTolerance.configuration.managedScheduledExecutorServiceNameHelp=The JNDI na
...
 healthCheck.configuration.virtualServers=Virtual Servers:
 healthCheck.configuration.virtualServersHelp=Associates an Internet domain name with a physical server.
 
 metrics.configuration.enabled=Enabled
 metrics.configuration.enabledHelp=Enables or Disables the Metrics.
+metrics.configuration.securityEnabled=Enable Secure Access
+metrics.configuration.securityEnabledHelp=If checked, you will be required to log in before accessing the Metrics endpoint.
 metrics.configuration.dynamic=Dynamic
 metrics.configuration.dynamicHelp=If set to true, applies the changes instantly without a restart. Otherwise a restart is required.
 metrics.configuration.secureMetrics=Secure Metrics
```
OndroMih: Change the text from "Secure Metrics" to "Require encrypted access over HTTPS"

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
```diff
@@ -87,6 +91,8 @@ metrics.configuration.virtualServersHelp=Associates an Internet domain name with
 
 openAPI.configuration.enabled=Enabled
 openAPI.configuration.enabledHelp=Enables or Disables the OpenAPI.
+openAPI.configuration.securityEnabled=Enable Secure Access
```
OndroMih: Change the text to "Require authorization"

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
```diff
@@ -69,13 +69,17 @@ faultTolerance.configuration.managedScheduledExecutorServiceNameHelp=The JNDI na
...
+healthCheck.configuration.securityEnabled=Enable Secure Access
+healthCheck.configuration.securityEnabledHelp=If checked, you will be required to log in before accessing the Health Check endpoint.
 healthCheck.configuration.endpoint=EndPoint
 healthCheck.configuration.endpointHelp=The context root used to expose the health checks.
 healthCheck.configuration.virtualServers=Virtual Servers:
 healthCheck.configuration.virtualServersHelp=Associates an Internet domain name with a physical server.
 
 metrics.configuration.enabled=Enabled
 metrics.configuration.enabledHelp=Enables or Disables the Metrics.
+metrics.configuration.securityEnabled=Enable Secure Access
```
OndroMih: Change the text to "Require authorization"

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/config/MetricsHealthCheckConfiguration.java
```diff
@@ -57,33 +57,36 @@
...
 
     void setVirtualServers(String value) throws PropertyVetoException;
     
+    /**
+     * @return a Boolean value determining if the security is enabled or not.
+     */
+    @Attribute(defaultValue = "false", dataType = Boolean.class)
+    String getSecurityEnabled();
+
+    void setSecurityEnabled(String value) throws PropertyVetoException;
```
OndroMih: Rename get and set methods to `setAuthorizationRequired` and `getAuthorizationRequired`

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/MetricsServiceConfiguration.java
```diff
@@ -54,48 +54,47 @@
...
     @Attribute(dataType = String.class)
     String getVirtualServers();
     void setVirtualServers(String value) throws PropertyVetoException;
 
+    /**
+     * @return a Boolean value determining if the security is enabled or not.
+     */
+    @Attribute(defaultValue = "false", dataType = Boolean.class)
+    String getSecurityEnabled();
+    void setSecurityEnabled(String value) throws PropertyVetoException;
```
OndroMih: Rename get and set methods to `setAuthorizationRequired` and `getAuthorizationRequired`

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/admin/OpenApiServiceConfiguration.java
```diff
@@ -60,22 +60,28 @@
...
 
     void setVirtualServers(String value) throws PropertyVetoException;
 
+    /**
+     * @return a Boolean value determining if the security is enabled or not.
+     */
+    @Attribute(defaultValue = "false", dataType = Boolean.class)
+    String getSecurityEnabled();
+
+    void setSecurityEnabled(String value) throws PropertyVetoException;
```
OndroMih: Rename get and set methods to `setAuthorizationRequired` and `getAuthorizationRequired`

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/admin/SetMetricsConfigurationCommand.java
```diff
@@ -103,16 +105,25 @@
     @Param(name = "virtualServers", optional = true)
     private String virtualServers;
 
-    @Param(optional = true, defaultValue = "server-config")
-    private String target;
-    
+    @Param(optional = true, alias = "securityenabled")
+    private Boolean securityEnabled;
```
OndroMih: Rename to `authorizationRequired`, alias `authorizationrequired`

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/admin/SetOpenApiConfigurationCommand.java
```diff
@@ -73,28 +75,32 @@
...
     private String virtualServers;
 
     @Param(name = "corsHeaders", optional = true, defaultValue = "false")
     private Boolean corsHeaders;
-    
-    @Param(optional = true, defaultValue = "server-config")
-    private String target;
+
+    @Param(optional = true, alias = "securityenabled")
+    private Boolean securityEnabled;
```
OndroMih: Rename to `authorizationRequired`, alias `authorizationrequired`

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
```diff
@@ -107,13 +106,24 @@
     @Inject
     UnprocessedConfigListener unprocessedListener;
 
+    @Param(optional = true, alias = "securityenabled")
+    private Boolean securityEnabled;
```
OndroMih: Rename to `authorizationRequired`, alias `authorizationrequired`

jGauravGupta:
Hi @OndroMih,

I agree that with existing `--securemetrics` argument & the new `--securityenabled` argument is confusing and `--authorizationRequired` makes more sense. But to make consistency with other existing asadmin command like [set-rest-monitoring-configuration](https://docs.payara.fish/documentation/payara-server/rest-monitoring/configuring-rest-monitoring.html) introduced `--securityenabled` argument. 
// @Pandrex247

Pandrex247:
Well for one is `--securemetrics` actually any different to `--securityenabled`? They look suspiciously similar...

And yes I agree @jGauravGupta ideally we'd keep things consistent.

Hmm, not convinced about `--authorizationrequired`. As you said @OndroMih `securityEnabled` is pretty well established in Payara as turning on HTTPS, but I'd also argue that it's fairly well established that it **also** makes auth required - are there any cases where you can enable HTTPS and auth separately?

Specifically making it _auth required_ also sounds a bit "developer-y" to me: _technically_ correct vs. common usage.

jGauravGupta:
Thanks, @OndroMih and @Pandrex247 for your feedback.
IMO `--securemetrics` argument can be marked as obsolete and `--securityenabled` argument can be used to turn on HTTPS and make auth required. Wdyt?

appserver/payara-appserver-modules/microprofile/microprofile-common/src/main/java/fish/payara/microprofile/SetSecureMicroprofileConfigurationCommand.java
jGauravGupta: Good catch, fixed it.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/admin/SetMPHealthCheckConfiguration.java
jGauravGupta: > Do we expect the default to practically never be used?

Yes

> Wouldn't it be more secure to force configuration of a password in case security is enabled?

I agree that if default user not exist and security is enabled then prompting for default username and password would be good. But prompting for password functionality and `com.sun.enterprise.admin.cli.ProgramOptions` injected instance is available for commands which inherits `com.sun.enterprise.admin.cli.CLICommand` class.

jGauravGupta:
jenkins test please

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
jGauravGupta: "Enable Secure Access" label is added here to maintain consistency across other endpoints like REST Monitoring.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/Constants.java
```diff
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2019] [Payara Foundation and/or its affiliates]
```
pdudits: There's not change in this file, so copyright can be also removed

jGauravGupta:
jenkins test please

OndroMih:
@Pandrex247 
> Well for one is --securemetrics actually any different to --securityenabled?

Well, your question exactly proves that there's a lot of confution in how the word "secure" is used. There's a remarkably huge difference between `--securemetrics` and `--securityenabled`, at least to how `--securemetrics` used to work before 5.192. While the first only required a connection over SSL, the second requires an SSL connection AND authentication.

If we agree that "secure" always means both SSL and authentication, then I'm fine with leaving the wording as is. I understand that we're deprecating the `--securemetrics` option. So "securing" in different contexts means:

* secured MP metrics/health/openapi means only authorized users are allowed and only over SSL :ballot_box_with_check:
* secured JMX connector means it works only over SSL. The Admin console hints that client authentication needs to be enabled separately but I'm not sure how this works.  :question:
* Secured DAS means access is allowed only over SSL and it requires authentication :ballot_box_with_check:
* Secure access for REST Monitoring means authentication is required. It is bound to the admin listener so SSL is configured separately. :ballot_box_with_check:

It seems that all contexts I found are more or less in line with this definition of "secure". However, I still recommend to use the word "restrict access" instead of just "secure" internally for new features to ensure that it's also possible to require authentication and not only SSL encryption.

appserver/admingui/microprofile-console-plugin/src/main/resources/fish/payara/admingui/microprofile/Strings.properties
OndroMih: The wording in REST Monitoring is can be wrong so I wouldn't blindly follow it for keeping consistency. Both should rather say "Require Secure Access" or "Require secure authorization". But I'm not going to argue, we can leave it as it is.

OndroMih: The wording in REST Monitoring is can be wrong so I wouldn't blindly follow it for keeping consistency. Both should rather say "Require Secure Access" or "Require secure authorization". But I'm not going to argue, we can leave it as it is.

PR #3976, opened 2019-05-24
PAYARA-3736: Let HazelcastConfigSpecificConfiguration register itself

pdudits:
After HK2 upgrade, the `@Named` annotation prevents the config object to be
added into config automatically.

To test, backup a Payara4 domain, and restore it into Payara5. Before it failed because of this injection not being satisfied, now it will only compain about missing Grizzly ALPN classes (but will start).


pdudits:
Jenkins test please

arjantijms:
> now it will only compain about missing Grizzly ALPN classes (but will start).

This may need further investigation?

pdudits:
> This may need further investigation?

No, it is known issue (next step in migration guide is adding relevant JVM options). However internal issue was created ti possibly automate that step via `asadmin start-domain --upgrade`.

PR #3978, opened 2019-05-27
Cleanup sonar issues 1 - Resources + Misc

jbee:
Cleanup of blocker and critical sonar issues (bugs) hand picked from our sonar issues trying to fix more severe first. As the issues do not have a common theme except their severity I annotated the changes with some context or thoughts.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/WeightedSnapshot.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/WeightedSnapshot.java
@@ -94,7 +94,7 @@ public class WeightedSnapshot extends Snapshot {
 
         for (int i = 0; i < copy.length; i++) {
             this.values[i] = copy[i].value;
-            this.normWeights[i] = copy[i].weight / sumWeight;
+            this.normWeights[i] = sumWeight == 0d ? 0d : copy[i].weight / sumWeight;
         }
 
         for (int i = 1; i < copy.length; i++) {
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/WeightedSnapshot.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/WeightedSnapshot.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  * 
- *    Copyright (c) [2018] Payara Foundation and/or its affiliates. All rights reserved.
+ *    Copyright (c) [2018-2019] Payara Foundation and/or its affiliates. All rights reserved.
  * 
  *     The contents of this file are subject to the terms of either the GNU
  *     General Public License Version 2 only ("GPL") or the Common Development
--- a/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
+++ b/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
@@ -158,8 +158,8 @@ public class HASingleSignOnEntry extends SingleSignOnEntry {
 
     /** convert a principal into byte array */
     private byte[] convertToByteArray(Principal obj) {
-        ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        try (ObjectOutputStream oos = ioUtils.createObjectOutputStream(new BufferedOutputStream(baos), true)) {
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                ObjectOutputStream oos = ioUtils.createObjectOutputStream(baos, true)) {
             oos.writeObject(obj);
             oos.flush();
             return baos.toByteArray();
--- a/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
+++ b/appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
@@ -159,50 +159,23 @@ public class HASingleSignOnEntry extends SingleSignOnEntry {
     /** convert a principal into byte array */
     private byte[] convertToByteArray(Principal obj) {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        BufferedOutputStream bos = null;
-        ObjectOutputStream oos = null;
-        try {
-            bos = new BufferedOutputStream(baos);
-            oos = ioUtils.createObjectOutputStream(bos, true);
+        try (ObjectOutputStream oos = ioUtils.createObjectOutputStream(new BufferedOutputStream(baos), true)) {
             oos.writeObject(obj);
             oos.flush();
             return baos.toByteArray();
         } catch(Exception ex) {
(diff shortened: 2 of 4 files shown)


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/WeightedSnapshot.java
```diff
@@ -94,7 +94,7 @@ public WeightedSnapshot(Collection<WeightedSample> values) {
 
         for (int i = 0; i < copy.length; i++) {
             this.values[i] = copy[i].value;
-            this.normWeights[i] = copy[i].weight / sumWeight;
+            this.normWeights[i] = sumWeight == 0d ? 0d : copy[i].weight / sumWeight;
```
jbee: from the context it should not be happening that `sumWeight` is zero but theoretical it could in case of bad test data. I figured that a normalised value of zero is the right value in case sum is zero.

appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
```diff
@@ -159,50 +159,23 @@ public long incrementAndGetVersion() {
...
-            bos = new BufferedOutputStream(baos);
-            oos = ioUtils.createObjectOutputStream(bos, true);
+        try (ObjectOutputStream oos = ioUtils.createObjectOutputStream(new BufferedOutputStream(baos), true)) {
             oos.writeObject(obj);
             oos.flush();
             return baos.toByteArray();
         } catch(Exception ex) {
             throw new IllegalStateException("Could not convert principal to byte array", ex);
-        } finally {
-          closeSilently(baos);
```
jbee: A `ByteArrayOutputStream` does not have to be closed. Its close is a NOOP anyway. Therefore there does not need to be another level of try for the `baos`.

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/composite/CompositeUtil.java
```diff
@@ -637,50 +637,41 @@ public Locale getLocale(MultivaluedMap<String,String> requestHeaders) {
...
+                            list.add(ext);
                         }
-                        list.add(ext);
                     }
                 }
-
             }
         } catch (IOException ex) {
             RestLogging.restLogger.log(Level.SEVERE, null, ex);
-        } finally {
```
jbee: I think this block wasn't just "old style" it was also buggy because looking closely the resource is created in the outer `while` loop but this will only close the last resource from the loop.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/provider/FlashlightProbe.java
```diff
@@ -264,7 +264,7 @@ public String getModuleProviderName() {
 	}
 
     public String getModuleName() {
-        return moduleProviderName;
```
jbee: Looking at the class I am pretty sure this was actually a case of maybe auto-completing to the wrong field. There is another getter for the `moduleProviderName` and this should use the `moduleName`.

jbee:
jenkins test please

appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
jbee: It cannot be within the block as it is used both in the resource construction expression and within the try block. If we want it covered it would need its own level of try-with-resouce.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/provider/FlashlightProbe.java
jbee: I don't know any details on flashlight but I was thinking that maybe the `getProbeDesc` is used as or for the key when I was looking why such a field confusion would go unnoticed for years. At least a method called `createProbe` uses it to do:
```java
FlashlightProbe fp = genericProvider.getProbe(probeDescriptor);
```

pdudits: Yes, I wasn't able to trigger code that would involve this method both over HTTP or JMX interfaces, so it's likely safe to remove it.

appserver/web/web-ha/src/main/java/org/glassfish/web/ha/authenticator/HASingleSignOnEntry.java
pdudits: try can accept multiple assignment statements.

jbee: Learned about multi-variable try-with resources ;) Thanks @pdudits . Whet ahead and used it. As discussed I also removed the buffer wrapping as well that in this case should just cause overhead for no benefit since a `ByteArrayOutputStream` already is a "buffer".

jbee:
jenkins test please

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/composite/CompositeUtil.java
```diff
@@ -637,50 +637,41 @@ public Locale getLocale(MultivaluedMap<String,String> requestHeaders) {
...
+                            continue;
                         }
-                        String[] entry = line.split(":");
-                        String base = entry[0];
-                        String ext = entry[1];
-                        List<String> list = modelExtensions.get(base);
-                        if (list == null) {
-                            list = new ArrayList<String>();
-                            modelExtensions.put(base, list);
+                        if (line.charAt(0) != '#') {
```
cubastanley: Just out of curiosity, why use line.charAt(0) != '#' over line.startsWith('#')?

jbee: Don't know the context here but I'd assume `#` is used for line comments so any line starting with `#` is effectivly ignored by this check.

jbee:
jenkins test please

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/composite/CompositeUtil.java
jbee: @cubastanley is your questions answered?

sgflt: I think the question is still the same. If line **starting with** "#" is supposed to be ignored, why not use **startsWith** method? :)

jbee: Opsi, somehow missed the _over_ part when reading it. 

`charAt` is _cheaper_, maybe that is why it was chosen. Or just personal style. I would not see `startsWith` as _more readable_ and maybe the author felt the same so he was opting for the option that execution-wise does exactly what he wants to check, no less, no more, which means less surprises of some odd corner case possibly existing with all you could do with a `String`. Using `charAt` you could skip thinking if such a corner case exist as it is a very basic operation. 
We will never know - I think both are just fine for such a check in general. Sometimes context can be a reason to chose one over the other. Within a character by character parser `charAt` is more clear, within string processing `startsWith`.

PR #3979, opened 2019-05-28
Cleanup sonar issues 2 - Resources + Misc

jbee:
Cleanup of blocker and critical sonar issues (bugs) hand picked from our sonar issues trying to fix more severe first. As the issues do not have a common theme except their severity I annotated the changes with some context or thoughts.

### DIFF
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
@@ -83,25 +83,17 @@ public class TextResourcesGenerator extends ResourcesGeneratorBase {
     public String endGeneration() {
         //generate date info in 1 single file
         File file = new File(generationDir+ "/codegeneration.properties");
-        BufferedWriter out = null;
         try {
             if (file.createNewFile()) {
-                FileWriter fstream = new FileWriter(file);
-                out = new BufferedWriter(fstream);
-                out.write("generation_date=" + new Date() + "\n");
+                try (BufferedWriter out = new BufferedWriter(new FileWriter(file))) {
+                    out.write("generation_date=" + new Date() + "\n");
+                }
             } else {
                 RestLogging.restLogger.log(Level.SEVERE, RestLogging.FILE_CREATION_FAILED, "codegeneration.properties");
             }
         } catch (Exception e) {
             e.printStackTrace();
-        } finally {
-            if (out != null) {
-                try {
-                    out.close();
-                } catch (IOException ex) {
-                    RestLogging.restLogger.log(Level.SEVERE, null, ex);
-                }
-            }
+            RestLogging.restLogger.log(Level.SEVERE, null, e);
         }
 
         return  "Code Generation done at : " + generationDir;
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2019] Payara Foundation and/or affiliates
 
 package org.glassfish.admin.rest.generator;
 
--- a/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
+++ b/nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
@@ -92,7 +92,6 @@ public class TextResourcesGenerator extends ResourcesGeneratorBase {
                 RestLogging.restLogger.log(Level.SEVERE, RestLogging.FILE_CREATION_FAILED, "codegeneration.properties");
             }
         } catch (Exception e) {
-            e.printStackTrace();
             RestLogging.restLogger.log(Level.SEVERE, null, e);
         }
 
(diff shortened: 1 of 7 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
```diff
@@ -577,39 +577,37 @@ public Types getDeployableTypes(DeploymentContext context) throws IOException {
             Types types = context.getTransientAppMetaData(Types.class.getName(), Types.class);
             if (types!=null) {
                 return types;
-            } else {
```
jbee: Wanted to address the resource issue but its unclear to me if that is correct since the parser is passed to something outliving the context of this method so all this change is is the removal or unnecessary `else` clause I did when looking at it.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/services/impl/ContainerMapper.java
```diff
@@ -470,7 +470,7 @@ public void unregister(String contextRoot) {
 
     public void register(final Endpoint endpoint) {
         if (LOGGER.isLoggable(Level.FINE)) {
-            LOGGER.log(Level.FINE, "MAPPER({0}) REGISTER endpoint: {1}", endpoint);
+            LOGGER.log(Level.FINE, "MAPPER({0}) REGISTER endpoint: {1}", new Object[] { this, endpoint });
```
jbee: Found a similar `log` call in the class which was using `this` as the "MAPPER" so it thought it should be same here.

nucleus/osgi-platforms/osgi-cli-interactive/src/main/java/org/glassfish/osgi/cli/interactive/LocalOSGiShellCommand.java
```diff
@@ -247,43 +266,23 @@ protected int executeCommand()
...
-                        return;
-                    }
-                };
-
-                reader = new ConsoleReader(REMOTE_COMMAND,
-                        new FileInputStream(file), out,
-                        null);
+                reader.setBellEnabled(false);
+                reader.addCompleter(getCommandCompleter());
+                return executeCommands(reader); // NB: wrapper on system.in/out does not need try-with-resource
```
jbee: sonar should complain about this line now but we then can mark it as false positive as we just wrap `System.in` and `System.out` which should not be closed. Hopefully the comment make recognising this in sonar easier.

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/osgi-platforms/osgi-cli-interactive/src/main/java/org/glassfish/osgi/cli/interactive/LocalOSGiShellCommand.java
pdudits: Actually we will be redoing this a bit, as with 192, this `System.out` no longer reaches the console.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
jbee: Added a suppression.

nucleus/osgi-platforms/osgi-cli-interactive/src/main/java/org/glassfish/osgi/cli/interactive/LocalOSGiShellCommand.java
jbee: Made the comment more general. I assume it still is the case that the general in/out should not be closed by a try-with-resources.

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/generator/TextResourcesGenerator.java
```diff
@@ -83,25 +83,17 @@ public ClassWriter getClassWriter(String className, String baseClassName, String
...
-                out = new BufferedWriter(fstream);
-                out.write("generation_date=" + new Date() + "\n");
+                try (BufferedWriter out = new BufferedWriter(new FileWriter(file))) {
+                    out.write("generation_date=" + new Date() + "\n");
+                }
             } else {
                 RestLogging.restLogger.log(Level.SEVERE, RestLogging.FILE_CREATION_FAILED, "codegeneration.properties");
             }
         } catch (Exception e) {
             e.printStackTrace();
```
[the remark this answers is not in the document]
jbee: Removed it.

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
```diff
@@ -571,45 +571,43 @@ public ApplicationInfo deploy(Collection<? extends Sniffer> sniffers, final Exte
...
-                                libAdapter.close();
-                            }
+            }
+            try {
+                ResourceLocator locator = determineLocator();
+                // scan the jar and store the result in the deployment context.
+                ParsingContext.Builder parsingContextBuilder = new ParsingContext.Builder().logger(context.getLogger())
+                        .executorService(executorService.getUnderlyingExecutorService());
+                // workaround bug in Builder
+                parsingContextBuilder.locator(locator);
```
Pandrex247: Wonder if this bug is still here...

jbee: Would have been nice if the author had left a clue about the bug.
Since my change only removed the unnecessary `else` clause I would think that this had no effect on it.

pdudits: Author here, it's a bug in the api, where locator is void method rather than fluent setter.

jbee: :D I guess it becomes apparent once you try to "fix" it so comment almost makes it more confusing.

PR #3980, opened 2019-05-28
Cleanup sonar issues 3 - potential NPEs

jbee:
Moved to Major sonar issues (bugs). This PR is almost exclusivity preventing NPEs from happening at locations identified by sonar. I annotated more complicated places.

### DIFF
--- a/appserver/admingui/core/src/main/java/org/glassfish/admingui/util/SunOptionUtil.java
+++ b/appserver/admingui/core/src/main/java/org/glassfish/admingui/util/SunOptionUtil.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2019] Payara Foundation and/or affiliates
 
 /**
  *
--- a/appserver/admingui/core/src/main/java/org/glassfish/admingui/util/SunOptionUtil.java
+++ b/appserver/admingui/core/src/main/java/org/glassfish/admingui/util/SunOptionUtil.java
@@ -110,9 +110,11 @@ public class SunOptionUtil {
         SelectItem[] options =
                 (SelectItem[]) Array.newInstance(SUN_OPTION_CLASS, size);
         options[0] = getSunOption("", "");
-        for (int i = 0; i < size-1; i++) {
-            SelectItem option = getSunOption(values[i], values[i]);
-            options[i + 1] = option;
+        if (values != null) {
+            for (int i = 0; i < size-1; i++) {
+                SelectItem option = getSunOption(values[i], values[i]);
+                options[i + 1] = option;
+            }
         }
         return options;
     }
--- a/appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/NestedAppClientDeployerHelper.java
+++ b/appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/NestedAppClientDeployerHelper.java
@@ -352,14 +352,14 @@ public class NestedAppClientDeployerHelper extends AppClientDeployerHelper {
 
         URI appClientURIWithinEAR = URI.create(appClientDesc().getModuleDescriptor().getArchiveUri());
         final Artifact appClientJARArtifact = newArtifact(appClientURIWithinEAR);
-
-        /*
-         * Processing the client artifact will recursively process any artifacts
-         * on which it depends plus the transitive closure thereof.
-         */
-        appClientJARArtifact.processArtifact(dependencyURIsProcessed,
-                clientLevelDownloads(), earLevelDownloads());
-
+        if (appClientJARArtifact != null) {
+            /*
+             * Processing the client artifact will recursively process any artifacts
+             * on which it depends plus the transitive closure thereof.
+             */
+            appClientJARArtifact.processArtifact(dependencyURIsProcessed,
+                    clientLevelDownloads(), earLevelDownloads());
+        }
         /*
          * Now incorporate the library JARs and, if v2 compatibility is chosen,
          * EJB JARs and top level JARs.
--- a/appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/NestedAppClientDeployerHelper.java
+++ b/appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/NestedAppClientDeployerHelper.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2019] Payara Foundation and/or affiliates
 
 package org.glassfish.appclient.server.core;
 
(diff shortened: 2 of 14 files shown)


appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/NestedAppClientDeployerHelper.java
```diff
@@ -352,14 +352,14 @@ private void processDependencies() throws IOException {
...
         final Artifact appClientJARArtifact = newArtifact(appClientURIWithinEAR);
-
-        /*
-         * Processing the client artifact will recursively process any artifacts
-         * on which it depends plus the transitive closure thereof.
-         */
-        appClientJARArtifact.processArtifact(dependencyURIsProcessed,
-                clientLevelDownloads(), earLevelDownloads());
-
+        if (appClientJARArtifact != null) {
```
jbee: In other places `newArtifact` was used (which can return null) a null check was performed so I opted for the same.

appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/jws/JWSAdapterManager.java
```diff
@@ -187,7 +187,10 @@ synchronized File rootForSignedFilesInDomain() {
      * @param content
      */
     void addStaticSystemContent(final String lookupURI, StaticContent newContent) throws IOException {
-        systemAdapter().addContentIfAbsent(lookupURI, newContent);
+        AppClientHTTPAdapter adapter = systemAdapter();
```
jbee: If starting the `systemAdapter` fails `systemAdapter()` logs an error and returns `null` so I figured that this should be handled without throwing an exception to the outside. Error is already logged so we just don't add the entry.

appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/jws/XPathToDeveloperProvidedContentRefs.java
```diff
@@ -221,12 +221,14 @@ void addToContentIfInApp(
                         new Object[] {referringDocument, pathToContent});
             } else {
                 final ApplicationSignedJARManager signedJARManager = helper.signedJARManager();
-                if (signedJARManager == null && helper instanceof NestedAppClientDeployerHelper) {
```
jbee: Problem is the `&&` that makes null be possible in other branches but its not expected to be. I believe what was intended here is to only log an error for null `signedJARManager` in case of a `NestedAppClientDeployerHelper`. Should any other helper occur with null this goes unnoticed but from the comments it sounds like we expect this to be a nested client helper. We might want to do some other logging or we just be satisfied as it is. What previously was ending with a NPE now has no log entry which seems reasonable since nobody had reported a NPE for this location so I would guess its not possible or common.

appserver/batch/glassfish-batch-connector/src/main/java/fish/payara/jbatch/persistence/rdbms/JBatchJDBCPersistenceManager.java
```diff
@@ -213,7 +213,7 @@ public CheckpointData getCheckpointData(CheckpointDataKey key) {
 
         tryObtainTableLock();
 
-        CheckpointData checkpointData = queryCheckpointData(key.getCommaSeparatedKey());
+        CheckpointData checkpointData = key == null ? null : queryCheckpointData(key.getCommaSeparatedKey());
```
jbee: While the logging in the method suggest that `key` can be null I believe it is not really expected that callers use null keys. The query performed is a "select where id =" which does not work for null so I concluded that we simply don't even try and finish with `null` data for `null` key which is logged and handled fine by the rest of the method.

appserver/common/amx-javaee/src/main/java/org/glassfish/admin/amx/impl/j2ee/RegistrationSupport.java
```diff
@@ -311,25 +311,27 @@ else if (ejbType.equals("Message-driven"))
...
                 intf = SingletonSessionBean.class;
                 impl = SingletonSessionBeanImpl.class;
             }
             else
             {
-                throw new IllegalArgumentException( "Unknown ejbSessionType: " + ejbSessionType + ", expected Stateless or Stateful");
+                throw new IllegalArgumentException("Unknown ejbSessionType: " + ejbSessionType + ", expected Stateless or Stateful");
             }
+        } else {
+            throw new IllegalArgumentException("Unknown ejbType: " + ejbType + ", expected Entity, Message-driven or Session");
```
jbee: This case needs to be added in order to prevent that `impl` and `intf` cannot be null when calling `registerJ2EEChild` below which requires them to be non-null. Therefore I opted to also throw an `IllegalArgumentException` for ùnexpected `ejbType` similar to the one present for unexpected `ejbSessionType`.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/EntityManagerWrapper.java
```diff
@@ -227,7 +227,9 @@ private EntityManager _getDelegate() {
                     }
                 }
             }
-
+            if (extendedEntityManager == null) {
+                throw new IllegalStateException("No EntityManager available in current extended persistence context.");
```
jbee: The `delegate` set to the `extendedEntityManager` in this branch is expected to never return null (inferred from users of the method) but currently the `extendedEntityManager` could be null at this point which would lead to calling methods fail with a NPE because of the `delegate` being null. In such cases we not get the `IllegalStateException` instead. So we still get an exception for those cases where this is true but now its a more descriptive one at the point of the root cause and not a NPE later one.

appserver/admingui/core/src/main/java/org/glassfish/admingui/util/SunOptionUtil.java
```diff
@@ -110,9 +110,11 @@ public static Option getOption(String value, String label) {
         SelectItem[] options =
                 (SelectItem[]) Array.newInstance(SUN_OPTION_CLASS, size);
         options[0] = getSunOption("", "");
-        for (int i = 0; i < size-1; i++) {
-            SelectItem option = getSunOption(values[i], values[i]);
```
jbee: NPE for `values` would never happen because of indexes and loop condition but sonar can't figure this out so I helped with the extra null check.

appserver/appclient/server/core/src/main/java/org/glassfish/appclient/server/core/jws/RestrictedContentAdapter.java
```diff
@@ -379,9 +379,6 @@ protected void finishErrorResponse(final Response gResp, final int status) {
     }
 
     protected String dumpContent() {
-        if (content == null) {
```
jbee: Content is a final field initialised in the constructor. So unless there is some initialisation madness going on sonar should be correct that this cannot be null.

jbee:
jenkins test please

jbee:
jenkins test please

PR #3983, opened 2019-05-29
Cleanup sonar issues 4 - potential NPEs

jbee:
Major sonar issues (bugs). Focussed on potential NPEs. Sometimes I also cleaned other problems found at the same location. As before I annotate the changes that are less trivial and require bit of context.

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -236,7 +236,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             transactionManager.clearThreadTx();
         }
         
-        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (invocation != null && requestTracing != null && requestTracing.isRequestTracingEnabled()) {
             startConcurrentContextSpan(invocation, handle);
         }
         
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -236,7 +236,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             transactionManager.clearThreadTx();
         }
         
-        if (invocation != null && requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
             startConcurrentContextSpan(invocation, handle);
         }
         
@@ -250,24 +250,26 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
     private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
         Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(
                 Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)));
-        
+
         SpanContext spanContext = tracer.extract(Format.Builtin.TEXT_MAP, new MapToTextMap(handle.getSpanContextMap()));
-        
+
         if (spanContext != null) {
             // Start a trace in the request tracing system
             SpanBuilder builder = tracer.buildSpan("executeConcurrentContext")
-                    .asChildOf(spanContext)
-                    .withTag("App Name", invocation.getAppName())
-                    .withTag("Component ID", invocation.getComponentId())
-                    .withTag("Module Name", invocation.getModuleName());
+                    .asChildOf(spanContext);
+            if (invocation != null) {
+                builder = builder.withTag("App Name", invocation.getAppName())
+                        .withTag("Component ID", invocation.getComponentId())
+                        .withTag("Module Name", invocation.getModuleName());
 
-            Object instance = invocation.getInstance();
-            if (instance != null) {
-                builder.withTag("Class Name", instance.getClass().getName());
+                Object instance = invocation.getInstance();
+                if (instance != null) {
+                    builder.withTag("Class Name", instance.getClass().getName());
+                }
             }
-            
+
             builder.withTag("Thread Name", Thread.currentThread().getName());
-            
+
             // Check for the presence of a propagated parent operation name
             try {
                 String operationName = ((RequestTraceSpanContext) spanContext).getBaggageItems().get("operation.name");
@@ -277,7 +279,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             } catch (ClassCastException cce) {
                 logger.log(Level.FINE, "ClassCastException caught converting Span Context", cce);
             }
-            
+
             builder.startActive(true);
         }
     }
--- a/appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsUtil.java
+++ b/appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsUtil.java
@@ -111,7 +111,7 @@ public class ConnectorsUtil {
 
     public static boolean getPingDuringPoolCreation(PoolInfo poolInfo, Resources allResources) {
         ResourcePool pool = getConnectionPoolConfig(poolInfo, allResources);
-        return Boolean.parseBoolean(pool.getPing());
+        return pool != null && Boolean.parseBoolean(pool.getPing());
     }
 
     /**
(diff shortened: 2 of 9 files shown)


appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -236,7 +236,7 @@ else if(backupClassLoader != null) {
             transactionManager.clearThreadTx();
         }
         
-        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (invocation != null && requestTracing != null && requestTracing.isRequestTracingEnabled()) {
```
jbee: As identified by sonar `invocation` could be null according to other logic in the same method. Only option I saw is to skip calling `startConcurrentContextSpan` in case `invocation` s null as this only works for non-null invocations. So NPE becomes NOOP.

appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsUtil.java
```diff
@@ -111,7 +111,7 @@ public static boolean belongsToSystemRA(String raName) {
 
     public static boolean getPingDuringPoolCreation(PoolInfo poolInfo, Resources allResources) {
         ResourcePool pool = getConnectionPoolConfig(poolInfo, allResources);
-        return Boolean.parseBoolean(pool.getPing());
+        return pool != null && Boolean.parseBoolean(pool.getPing());
```
jbee: Theoretically `getConnectionPoolConfig` can return null which I thought defaults to `false`. Didn't wanted to make this more complicated with logging or exceptions. So NPE becomes `false`.

appserver/core/javaee-kernel/src/main/java/org/glassfish/kernel/javaee/WebContainerStarter.java
```diff
@@ -162,23 +162,17 @@
...
-                isStartNeeded = true;
-            }
-        }
+        boolean isStartNeeded = serverConfig != null
+                && (isStartNeeded(serverConfig.getHttpService()) || isStartNeeded(serverConfig.getNetworkConfig()));
 
         if (isStartNeeded) {
             startWebContainer();
-        } else {
+        } else if (serverConfig != null){
```
jbee: It seems logical that if the `serverConfig` is null (as suggested to be possible by `isStartNeeded`) we simply cannot do the initialisation. Behaviour change: NPE becomes NOOP

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/EnvironmentProperty.java
```diff
@@ -386,50 +386,49 @@ public void print(StringBuilder toStringBuilder) {
...
-	    } else if (Float.class.equals(type)) {
-		return new Float(string);
-	    } else if (Short.class.equals(type)) {
-		return Short.valueOf(string);
-	    } else if (Byte.class.equals(type)) {
-		return Byte.valueOf(string);
-	    } else if (Long.class.equals(type)) {
-		return Long.valueOf(string);
-	    } else if (Character.class.equals(type)) {
+        if (string == null || ("".equals(string) && type != String.class)) {
```
jbee: Using the proper way of comparing `Class`es this also removes the possible NPE for `type` and maps any `null` input for `type` to `null` output which seems reasonable. 

Below I corrected funny indentation, comparison and unnecessary `else` nesting.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/DeploymentDescriptorNode.java
```diff
@@ -1120,15 +1120,16 @@ public void appendQNameChild(String elementName, Node parent, String namespaceUr
...
         // Always set prefix mapping on leaf node. If the DOL was
         // populated from an existing deployment descriptor it does
         // not preserve the original node structure of the XML document,
         // so we can't reliably know what level to place mapping.
         // Alternatively, if we're writing out a descriptor that was created
         // by the deploytool, there is no prefix->namespace information in
         // the first place.
-        setAttributeNS(element, prefix, namespaceUri);
+        if (element != null) {
+            setAttributeNS(element, prefix, namespaceUri);
```
jbee: Should the `element` be null (empty text) we also don't want to add the attribute. Behaviour change: NPE becomes NOOP and now allows adding empty text nodes.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/ResourcePropertyNode.java
```diff
@@ -84,15 +84,14 @@ public Node writeDescriptor(Node node, Descriptor desc) {
...
-
-        Set keys = properties.keySet();
-
-        for (Object key : keys) {
-            String name = (String) key;
-            String value = (String) properties.get(name);
-            Node propertyNode = appendChild(node, TagNames.RESOURCE_PROPERTY);
-            appendTextChild(propertyNode, TagNames.RESOURCE_PROPERTY_NAME, name);
-            appendTextChild(propertyNode, TagNames.RESOURCE_PROPERTY_VALUE, value);
+        if (properties != null) {
```
jbee: There seems to be a set of classes we are able to extract `properties` from. Should this be called with another implementation class for `desc` we don't know how to access properties so we have to skip adding text nodes. Behaviour change: NPE for certain - maybe plainly theoretical - descriptor types becomes NOOP.

I did not add any logging or throwing of an exception as this seems to do an extra thing for a known set of classes. In such a case I would expect unknown classes just not to have this extra functionality.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/ComponentValidator.java
```diff
@@ -468,70 +468,65 @@ protected void accept(EjbReference ejbRef) {
...
+            BundleDescriptor referringBundle = 
+                ejbRef.getReferringBundleDescriptor();
+
+            // If we can verify that the current ejb 3.0 reference is defined
+            // in any Application Client module or in a stand-alone web module
+            // it must be remote business.
+            if (((referringBundle == null) && (getEjbBundleDescriptor() == null))
+                    || ((referringBundle != null) && (referringBundle.getModuleType() == DOLUtils.carType()))
+                    || ((getApplication() == null) && (referringBundle != null)
+                            && referringBundle.getModuleType() != null
```
jbee: The above line is the fix for a potential NPE testing the condition. I assume this was simply a miss.
Rest of the change is caused by removing unnecessary `else` clauses. Behaviour change: NPE during condition check now does no longer occur.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsUtil.java
pdudits: `ResourcePool#getPing` has a default value, so I believe it cannot return null.

appserver/core/javaee-kernel/src/main/java/org/glassfish/kernel/javaee/WebContainerStarter.java
pdudits: But I cannot imagine a situation where serverConfig is not provided. I think this is worth an archeological expedition on why server config could be optional.

appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/node/ResourcePropertyNode.java
pdudits: I believe there is set of xml schema checks before these nodes are filled, so it is quite possible that this is pure theory (but you can try creating a descriptor that would trigger this ;)

appserver/connectors/connectors-internal-api/src/main/java/com/sun/appserv/connectors/internal/api/ConnectorsUtil.java
jbee: I think you are - also `parseBoolean` could handle a null value.

appserver/core/javaee-kernel/src/main/java/org/glassfish/kernel/javaee/WebContainerStarter.java
jbee: Can't give much insight into when a `serverConfig` would be null. This is purely based on preexisting code suggesting that it could be.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
Pandrex247: This seems like a valid change, but I worry how this will affect the call to `endTrace` in the `reset` method.
As far as I can tell this is an "entry point" for tracing spans on async threads, and as such it's not immediately apparent to me how this might tracing further down the line (assuming execution actually somehow proceeds without having an invocation...).

It might be safer to simply move the null check to the `startConcurrentContextSpan` method and move the creation of the Span tags etc. to inside of that check, so that when the invocation proceeds there's a least a "skeleton" span.

jbee:
jenkins test please

PR #3984, opened 2019-05-29
Cleanup sonar issues 5

jbee:


appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/ResourceValidator.java
```diff
@@ -833,15 +833,15 @@ private boolean isEmbedded(Application application, String raname) {
      */
     private void validateJNDIRefs(DeploymentContext deploymentContext, Application application, AppResource resource, JNDINamespace namespace) {
         // In case lookup is not present, check if another resource with the same name exists
-        if (!resource.hasLookup() && !namespace.find(resource.getName(), resource.getEnv())) {
+        String jndiName = resource.getJndiName();
+        if (jndiName == null || !resource.hasLookup() && !namespace.find(resource.getName(), resource.getEnv())) {
```
jbee: I believe the below `throws` also was meant to occur in case `jndiName` is `null`. This is suggested both by the method comment and the parameters to the message passing `null` for the "Lookup:" value (which is another name of the jndiName).

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/CallbackInvocationContext.java
```diff
@@ -227,42 +225,43 @@ public void invokeSpecial() throws Throwable {
...
-                }
-                index++;
-            }
-        } else {
-            throw new IllegalStateException("Internal Error: Got null constructor");
-        }
+       if( ctor == null) {
+           throw new IllegalStateException("Internal Error: Got null constructor");
+       }
+       if (params == null) {
```
jbee: `params` needed proper null handling. All cases were practically present just not nested and formulated in a way that makes sure the intended `IllegalArgumentException` is thrown. Moved the `else` from end of method to top to get rid of 1 nesting level for rest of the method body. Behaviour change: NPE becomes IllegalArgumentException (as intended)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorUtil.java
```diff
@@ -37,120 +37,82 @@
...
+        }
+        return set;
+    }
 
-    public static boolean hasCompatiblePrimitiveWrapper(Class type, Class typeTo) {
-        Set<Class> compatibles = compatiblePrimitiveWrapper.get(type);
-        return compatibles.contains(typeTo);
+    public static boolean hasCompatiblePrimitiveWrapper(Class<?> type, Class<?> typeTo) {
+        Set<Class<?>> compatibles = compatiblePrimitiveWrapper.get(type);
+        return compatibles != null && compatibles.contains(typeTo);
```
jbee: Another theoretical NPE here in case method is called with a non-primitive `type` value.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorUtil.java
```diff
@@ -37,120 +37,82 @@
...
-         smallerPrimitiveWrappers.add(Integer.class);
-         compatiblePrimitiveWrapper.put(int.class, smallerPrimitiveWrappers);
-
-         smallerPrimitiveWrappers = new HashSet<>();
-         smallerPrimitiveWrappers.add(Byte.class);
-         smallerPrimitiveWrappers.add(Short.class);
-         smallerPrimitiveWrappers.add(Integer.class);
-         smallerPrimitiveWrappers.add(Long.class);
-         compatiblePrimitiveWrapper.put(long.class, smallerPrimitiveWrappers);
+    private static final Map<Class<?>, Set<Class<?>>> compatiblePrimitiveWrapper = new ConcurrentHashMap<>();
```
jbee: I am quite sure this class is used by multiple threads so `HashMap` is not adequate. Practically this does work since there is only initial writing that eventually is visible for all threads by chance but formally there are no guarantees. 

Also cleaned up the duplication and added missing generics for the file.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorUtil.java
```diff
@@ -37,120 +37,82 @@
...
-
-        if( method != null) {
-
-            Class[] paramTypes = method.getParameterTypes();
-            if ((params == null) && (paramTypes.length != 0)) {
+        if( method == null) {
+            throw new IllegalStateException("Internal Error: Got null method");
+        }
+        Class<?>[] paramTypes = method.getParameterTypes();
+        if (params == null) {
```
jbee: Same problem as before for constructors the code had all cases but wasn't written so that it makes sure `IllegalArgumentException` is thrown in case of `null` `params` when more then zero parameters were expected and accepted otherwise. Again moved the `else` block from end of method to the top to get rid of nesting.

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/deployment/node/runtime/EjbNode.java
```diff
@@ -136,80 +138,81 @@ public void setElementValue(XMLElement element, String value) {
...
                 }
             }
             return;
-        } else if (descriptor==null && ! RuntimeTagNames.AVAILABILITY_ENABLED.equals(element.getQName())) {
+        } else if (descriptor==null && !isAvailabilityEnabled) {
             DOLUtils.getDefaultLogger().log(Level.SEVERE, "enterprise.deployment.backend.addDescriptorFailure",
-                            new Object[] {element.getQName() , value });
+                    new Object[] {qName , value });
+            super.setElementValue(element, value);
+            return;
```
jbee: The issue was that `descriptor` can be null what is not handled by below if-else-if block. If `qName` is set to do `isAvailabilityEnabled` no descriptor is involved so no warning was logged and no NPE can occur. But if it is some other `qName` we assume a `descriptor` is needed and log the warning. To void the NPE in case one of the `else if` matches we do return. As the default for no match (last `else`) is to call `super` we do this before returning assuming it could be any other `qName` that does not need a `descriptor`.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/deployment/node/runtime/EjbNode.java
```diff
@@ -390,7 +393,7 @@ public Node writeDescriptor(Node parent, String nodeName, EjbDescriptor ejbDescr
      * @param parent node for the information
      * @param descriptor the descriptor containing the generated info
      */
-    private void writeGenClasses(Node parent, EjbDescriptor ejbDescriptor) {
+    private static void writeGenClasses(Node parent, EjbDescriptor ejbDescriptor) {
```
jGauravGupta: Why did `writeGenClasses` instance method change to a static method?

jbee: Because it can be a `static` method - something we can reason about as a function. 

This helps a reader to see quickly (and with IDE also on call site) that this is side effect free (except from doing static side-effect which hopefully should be rare). It also is a hint that whatever is going on in this method is independent of the class defining it so it might either be a utility that could be in a general utility class or it should maybe be a member of one of its argument classes instead).

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorUtil.java
pdudits: HashMap only has issues with concurrent access while writing. Writing only in static initializer prevents that from happening (as the field is not accessible for reading until the static block finishes). I'd rather wrap it in `Collections.unmodifiableMap` just to make the limitation clear.

jbee: I wasn't concerned about concurrent access but visibility across threads but as constant this should not be an issue. Will change to `unmodifiableMap`.

jbee:
jenkins test please

PR #3986, opened 2019-05-29
Cleanup sonar issues 6 - potential NPEs

jbee:
Major sonar issues (bugs). Focussed on potential NPEs. Sometimes I also cleaned other problems found at the same location. As before I annotate the changes that are less trivial and require bit of context.

### DIFF
--- a/appserver/ha/ha-file-store/src/main/java/org/glassfish/ha/store/adapter/file/FileBackingStore.java
+++ b/appserver/ha/ha-file-store/src/main/java/org/glassfish/ha/store/adapter/file/FileBackingStore.java
@@ -259,7 +259,7 @@ public class FileBackingStore<K extends Serializable, V extends Serializable>
         if (logger.isLoggable(TRACE_LEVEL)) {
             logger.log(TRACE_LEVEL, debugStr + "Entered save(" + sessionKey + ")");
         }
-        writetoFile(sessionKey, fileName, getSerializedState(sessionKey, value));
+        writetoFile(sessionKey, fileName, getSerializedState(value));
         if (logger.isLoggable(TRACE_LEVEL)) {
             logger.log(TRACE_LEVEL, debugStr + "Done save(" + sessionKey + ")");
         }
@@ -326,33 +326,16 @@ public class FileBackingStore<K extends Serializable, V extends Serializable>
         return success;
     }
 
-    private byte[] getSerializedState(K key, V value)
+    private byte[] getSerializedState(V value)
             throws BackingStoreException {
-
-        byte[] data = null;
-        ByteArrayOutputStream bos = new ByteArrayOutputStream();
-        ObjectOutputStream oos = null;
-        try {
-            oos = new ObjectOutputStream(bos);
+        try (ByteArrayOutputStream bos = new ByteArrayOutputStream();
+                ObjectOutputStream oos = new ObjectOutputStream(bos)) {
             oos.writeObject(value);
             oos.flush();
-            data = bos.toByteArray();
+            return bos.toByteArray();
         } catch (IOException ioEx) {
             throw new BackingStoreException("Error during getSerializedState", ioEx);
-        } finally {
-            try {
-		if (oos != null) {
-                    oos.close();
-                }
-            } catch (IOException ioEx) {/* Noop */}
-            try {
-                if (bos != null) {
-                    bos.close();
-                }
-            } catch (IOException ioEx) {/* Noop */}
         }
-
-        return data;
     }
 
     private byte[] readFromfile(String fileName) {
@@ -366,11 +349,7 @@ public class FileBackingStore<K extends Serializable, V extends Serializable>
         if (file.exists()) {
             int dataSize = (int) file.length();
             data = new byte[dataSize];
-            BufferedInputStream bis = null;
-            FileInputStream fis = null;
-            try {
-                fis = new FileInputStream(file);
-                bis = new BufferedInputStream(fis);
+            try (BufferedInputStream bis = new BufferedInputStream(new FileInputStream(file))) {
                 int offset = 0;
                 for (int toRead = dataSize; toRead > 0;) {
                     int count = bis.read(data, offset, toRead);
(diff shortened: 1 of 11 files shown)


appserver/ha/ha-file-store/src/main/java/org/glassfish/ha/store/adapter/file/FileBackingStore.java
```diff
@@ -326,33 +326,16 @@ private boolean removeFile(final File file) {
...
-        ByteArrayOutputStream bos = new ByteArrayOutputStream();
-        ObjectOutputStream oos = null;
-        try {
-            oos = new ObjectOutputStream(bos);
+        try (ByteArrayOutputStream bos = new ByteArrayOutputStream();
+                ObjectOutputStream oos = new ObjectOutputStream(bos)) {
             oos.writeObject(value);
             oos.flush();
-            data = bos.toByteArray();
+            return bos.toByteArray();
```
jbee: static analysis now should be able to work out that result is never null. Also removed the key that wasn't used and that made this somewhat confusing. The caller for this is the like change above.

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/DMManagedConnectionFactory.java
```diff
@@ -191,34 +191,13 @@
      * @param key
      * @return
      */
-    private String getParsedKey(String key) throws ResourceException {
```
jbee: This could return `null` for any key not starting with `"set"` as well as the `null` key but the caller did not expect to get back null ever and fixing it there didn't make much sense. Also this method already had a test checking for null or empty key in case it was removing the "set" prefix so it seem logical that passing `null` or the empty string to this should equally cause the `ResourceException`.

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/ManagedConnectionFactoryImpl.java
```diff
@@ -600,33 +600,35 @@ private void detectSqlTraceListeners() {
...
-                            try {
-
-                                constructors = listenerClass.getConstructors();
-                                for (Constructor constructor : constructors) {
-                                    parameterTypes = constructor.getParameterTypes();
-                                    //For now only the no argument constructors are allowed.
-                                    //TODO should this be documented?
-                                    if (parameterTypes != null && parameterTypes.length == 0) {
-                                        listener = (SQLTraceListener) constructor.newInstance(initargs);
+                    if (listenerClass != null) {
```
jbee: `listenerClass` can be null which now will not try to register any listener (which would cause a NPE)

appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/data/InstanceDescriptorImpl.java
```diff
@@ -176,18 +176,10 @@ public void removeApplication(ApplicationDescriptor applicationInfo) {
...
     /**
      * Overrides equals purely based on the UUID value
-     *
-     * @param obj
-     * @return
      */
     @Override
     public boolean equals(Object obj) {
-        boolean result = false;
-        if (InstanceDescriptorImpl.class.isInstance(obj)) {
```
jbee: sonar didn't like use of `isInstance` over `instanceof`. I agree :)

appserver/persistence/cmp/ejb-mapping/src/main/java/com/sun/jdo/api/persistence/mapping/ejb/SunCmpMappingsUtils.java
```diff
@@ -176,24 +176,23 @@ else if (l.size() > 1)
...
-        if (len > 0)
-            retVal = new ArrayList(); 
-        for (int i = 0; i < len; i++) {
+        if (props == null) {
+            return null;
+        }
+        List<BaseBean> retVal = new ArrayList<>(); 
+        for (int i = 0; i < props.length; i++) {
             // get the bean that is the property's parent.
-            BaseBean candidate = gm.getPropertyParent(props[i]);
```
jbee: sonar can't do index based exclusion so I rewrote the body to both make it more readable for humans and tools.

appserver/persistence/cmp/enhancer/src/main/java/com/sun/jdo/api/persistence/enhancer/PersistenceLauncher.java
```diff
@@ -192,8 +192,10 @@ public static void main(String[] args)
             out.println("    }");//NOI18N
             out.println("    targetClassname = " + targetClassname);//NOI18N
             out.print("    targetClassArgs = { ");//NOI18N
-            for (int i = 0; i < targetClassArgs.length; i++) {
-                out.print(targetClassArgs[i] + " ");//NOI18N
```
jbee: Again index based exclusion needs some help so I added a null check.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/ha/ha-file-store/src/main/java/org/glassfish/ha/store/adapter/file/FileBackingStore.java
```diff
@@ -427,17 +388,6 @@ private void writetoFile(K sessionKey, String fileName, byte[] data)
             }
             String errMsg = "Could not save session: " + sessionKey;
             throw new BackingStoreException(errMsg, ex);
-        } finally {
```
MattGill98: Should we keep the failure logging here with a catch?

jbee: I wouldn't say so. Its logging of exceptions occurring in something that is backed into try-with-resource. I would think that these were only added because the author didn't want to just ignore a cached exception.

jbee: Maybe a further clarification: Using try-with-resources is a change of behaviour since exceptions occurring while closing will no longer be caught. My assumptions here was that the author was using the replaced code as the pattern to use for IO and I replaced it with the new pattern, the try-with-resouce.

PR #3987, opened 2019-05-29
PAYARA-3876 Added profiles for adding and embedding source in build

arjantijms:
PAYARA-3876

Signed-off-by: arjantijms <arjan.tijms@gmail.com>


arjantijms:
Jenkins test please

arjantijms:
Jenkins test please

appserver/packager/pom.xml
```diff
@@ -298,80 +471,51 @@
...
+    <profiles>
+        <profile>
+            <id>embed-source</id>
+            <activation><activeByDefault>false</activeByDefault></activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-dependency-plugin</artifactId>
+                        <executions>
```
pdudits: Given the packaging process is complicated already, wouldn't be better to define those source-only steps in profile only?

nucleus/pom.xml
```diff
@@ -1482,6 +1482,27 @@ Parent is ${project.parent}</echo>
                 </pluginManagement>
             </build>
         </profile>
+        
+        <profile>
+            <id>add-source</id>
```
pdudits: Why different from `embed-source` in packager?

arjantijms: One profile just adds the source jars, but leaves them where they are.  

The packager profile embeds them into the existing jars, which is a somewhat distinct operation.

appserver/packager/pom.xml
arjantijms: Lines 495 to 498 are in the embed-source profile, so the source-only steps are already in profile. Or did you mean something else?

arjantijms:
Jenkins test please

Cousjava:
How is this different from `mvn pre-site -Psource`?

appserver/packager/pom.xml
pdudits: Yes, whether you considered configuration of this execution to be put here as well. The `<phase>none</phase>` is easy to miss when reasoning about what the actual process is.

nucleus/pom.xml
pdudits: Ok, so you don't need both of them to work together. @Cousjava seem to suggest there's a way to achieve that already, though.

arjantijms:
> How is this different from mvn pre-site -Psource?

This embeds the source inside the java jar archives, so the IDE always finds it. `mvn pre-site -Psource` does nothing like that.

appserver/packager/pom.xml
arjantijms: I put it there since that's how the IPS profile works too. All the config in one place, and then the profiles activate it.

arjantijms:
Jenkins test please

arjantijms:
Jenkins test please

PR #3988, opened 2019-05-29
PAYARA-3733 Change Appclient Namespace

MattGill98:
Renamed all groupIds and some artifactIds.
Renamed folders to match.


MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/appclient/server/pom.xml
```diff
@@ -39,24 +39,28 @@
...
-    <name>GlassFish App Client Server-side Modules</name>  
+
+    <name>Payara App Client - Server-Side Modules</name>  
+
     <modules>
-       <module>connector</module>
-       <module>core</module>
-       <module>core-l10n</module>
-<!--       <module>jws</module> -->
+       <module>appclient-connector</module>
```
jGauravGupta: Considering the naming consistency of other refactored appclient client and server modules. IMHO this module should be renamed to `appclient-server-connector`.

MattGill98: I thought about that, but I decided it sounded weird since `server-connector` implies it's for connecting to the server. What do you think?

MattGill98: After speaking to @Pandrex247 I've changed my mind and changed it as you said!

MattGill98:
jenkins test please

MattGill98:
jenkins test please for the 10000000000th time

MattGill98:
jenkins test please

MattGill98:
jenkins test please

Pandrex247:
Upon further investigation, do we even need to change the namespace so as to publish these artefacts? These are _build_ artefacts, the appclient is actually bundled in with the release, and has extra Jars added to it after this point.

pdudits:
I was investigating changing the groupIds globally and I'd suggest replacing `org.glassfish.main` with `fish.payara.server` to keep [top level group](https://repo.maven.apache.org/maven2/fish/payara/) lean. Otherwise we'd end up adding tens of subgroups there.

Pandrex247:
I think we'd need to be careful throwing the _server_ word around - Payara Server is specifically the name of the full distribution (in comparison to Payara Micro)

Pandrex247:
Closing as this isn't the solution to the problem

PR #3989, opened 2019-05-30
Cleanup sonar issues 7 - Locks not released

jbee:
Complexity of locks deserves a PR focused on the `AdminCommandLock` entirely. 

~~There are two cases of not releasing the lock on all paths that were corrected. I annotate them in context.~~

@pdudits provided a PR-PR that removes the full suspension feature as it is unused whereby the inconsistent lock handling fixed originally no longer exists.

### DIFF
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
- * Portions Copyright [2017] Payara Foundation and/or affiliates
+ * Portions Copyright [2017-2019] Payara Foundation and/or affiliates
  */
 
 package org.glassfish.api.admin;
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
@@ -77,7 +77,7 @@ public class AdminCommandLock {
      * and thus there being exactly one such lock object, shared by all
      * users of this class.
      */
-    private static ReentrantReadWriteLock rwlock = new ReentrantReadWriteLock();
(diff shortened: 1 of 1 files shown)


nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
```diff
@@ -590,27 +590,23 @@ private void semaphoreWait(Semaphore s, String logMsg) {
...
-                    lock.unlock();
-
+                lock.unlock();
                 // Run the caller's commands without a lock in place.
                 r.run();
             } finally {
                 // Relock before returning.  This may block if someone else
                 // already grabbed the lock.
-                if (lock != null)
-                    lock.lock();
```
jbee: This is best understood looking at the method name. The method should execute the `Runnable` in a context where the lock is not locked. So if no lock is active we can run `r` plain, otherwise we unlock, run and lock again (in a way that static analysis can follow).

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/admin/AdminCommandLock.java
```diff
@@ -507,55 +505,57 @@ public void run() {
...
                         suspendCommandsTimedOut = true;
                 } catch (java.lang.InterruptedException e) {
-                    logger.log(Level.FINE, 
-                               "Interrupted acquiring command lock. ",
-                               e); 
-               }
+                    logger.log(Level.FINE, "Interrupted acquiring command lock. ", e);
+                }
             }
+            try {
```
jbee: The body is doing various things left and right and it has two flags saying the same thing in a different way.
There is  the `lockAcquired` variable and the `suspendCommandsTimedOut` field (which means lock was not acquired because of a timeout trying). Because it is a field its unclear if or how this can be simplified. Therefore, I think, a `try` block has to put around the body as soon as the lock is acquired that does `unlock` in its `finally` at the end only in case `lockAcquired` is true.

jbee:
jenkins test please

jbee:
jenkins test please

jbee:
jenkins test please

PR #3990, opened 2019-05-30
Fixed  typo in path.

johnmanko:


smillidge:
Thanks good catch

johnmanko:
I seems like there are a lot of places where META-INF, WEB-INF, *.xml strings are defined.  I wouldn't mind consolidating all of those to a single definition some place.  If you're ok with that, I'll work on it.  Which class/interface should define all of those?  There is ` Payara/appserver/resources/resources-connector/src/main/java/org/glassfish/resources/api/ResourceConstants.java`, but I think the current ones defined in there will also have to be renamed to account for `WEB-INF`.  Perhaps move the definitions from ` Payara/appserver/resources/resources-connector/src/main/java/org/glassfish/resources/util/ResourceUtil.java` to `ResourceConstants.java`?

cubastanley:
Jenkins test please

pdudits:
@johnmanko, please check our [contribution guide](/payara/Payara/tree/master/.github/CONTRIBUTING.md), and fill out the CLA. 

Thanks for your contribution!

PR #3991, opened 2019-05-30
Cleanup sonar issues 8 - Extracting constants and Adding generics

jbee:
Started extracted constants as indicated by sonar and got side tracked filling in generics in classes full of raw types. 

Reviewers should look out for generics added to something that would be considered public API and therefore should not change. As everything is part of the `admingui` to my knowledge it should be fine and not be problematic for 3rd party users.


jbee:
jenkins test please

jbee:
jenkins test please

PR #4006, opened 2019-06-04
PAYARA-3815: Preserve configured properties when looking up administe…

pdudits:
…red object

Ignoring env resulted in properties defined in jndi.properties taking over the values
configured in the naming context the lookup was done in.


pdudits:
jenkins test please

PR #4015, opened 2019-06-05
PAYARA-3895 Fix check for if name is in use by deployment group

Pandrex247:
_Since Jan 2018_
Whoopsies!


jGauravGupta:
jenkins test please

Pandrex247:
@jGauravGupta re-requesting since actually on second look at this the if logic is just wrong and weird.

Pandrex247:
Jenkins test please

PR #4016, opened 2019-06-05
PAYARA-3822 Ensure AMX Bean is created for metrics after server start

Cousjava:


Cousjava:
Jenkins test please

nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/MonitoringBootstrap.java
```diff
@@ -215,6 +215,9 @@ public void event(Event event) {
             if (LOGGER.isLoggable(Level.FINE))
                 LOGGER.log(Level.FINE, "Discovering the XML ProbeProviders from lib/monitor");
             discoverXMLProviders();
+            if (spmd != null) {
+                spmd.updateAllStatsProviders();
```
jbee: From what I understood the issue is that at restart a problem occurs. If making this `null` check helps prevent it it must mean that `spmd` is null which makes me wonder if `updateAllStatsProviders` needs to be called later or will `spmd` be initialised at some later stage and this does include the effects of `updateAllStatsProviders`?

PR #4020, opened 2019-06-07
PAYARA-3821 Add asadmin command to reset AMX metrics

AlanRoth:
The feature has been requested by Rakuten to turn all monitoring items to OFF once, and then set it back to the previous value on the modules the user is watching to perform a 'reset'

The new command
```
restart-monitoring
```
Parameters
```
--target 
--verbose or -v
```
The target parameter specifies which config to reset the monitoring levels for.


nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
+/*
```
jGauravGupta: license header missing

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+
+        //Setting all modules to "OFF"
+        for (String module : validModuleList) {
+            try {
+                ConfigSupport.apply((final MonitoringService monitoringServiceProxy) -> {
+                    monitoringServiceProxy.setMonitoringLevel(module, "OFF");
+                    return monitoringServiceProxy;
+                }, monitoringService);
+            } catch (TransactionFailure ex) {
+                logger.log(Level.WARNING, "Failed to execute the command set-monitoring-level, while setting modules to 'OFF': {0}", ex.getCause().getMessage());
```
jGauravGupta: change command name `set-monitoring-level` to `reset-monitoring-levels`

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+            actionReport.setMessage("Cound not find target: " + target);
+            actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
+            return;
+        }
+
+        Map<String, String> enabledModules = getPreviouslyEnabledModules(validModuleList);
+
+        //Setting all modules to "OFF"
+        for (String module : validModuleList) {
```
Pandrex247: Do you need to set the ones that are already off to off? Does that actually do anything (e.g. poke it)?
Might be a slight performance improvement if you used your enabledModules map like you do when re-enabling them if it doesn't actually do anything.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+@PerLookup
+@ExecuteOn({RuntimeType.DAS, RuntimeType.INSTANCE})
+@TargetType({CommandTarget.DAS, CommandTarget.STANDALONE_INSTANCE, CommandTarget.CLUSTER, CommandTarget.CONFIG})
+public class ResetMonitoringLevels implements AdminCommand {
+
+    @Param(name = "target", optional = true, defaultValue = "server-config")
+    String target;
+
+    @Param(name = "verbose", optional = true, shortName = "v")
+    private boolean verbose;
```
jbee: So lets see if I remember... what I sure of is that using `boolean` (instead of `Boolean`) for `optional` turned out to be a bad idea. I had the same when I did my first command thinking that `false` for "not provided" would be fine. I think the issue was that the injection mechanism would act unexpected when not stating the parameter on command line.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+        for (String module : validModuleList) {
+            try {
+                ConfigSupport.apply((final MonitoringService monitoringServiceProxy) -> {
+                    monitoringServiceProxy.setMonitoringLevel(module, "OFF");
+                    return monitoringServiceProxy;
+                }, monitoringService);
+            } catch (TransactionFailure ex) {
+                logger.log(Level.WARNING, "Failed to execute the command set-monitoring-level, while setting modules to 'OFF': {0}", ex.getCause().getMessage());
+                actionReport.setMessage(ex.getCause().getMessage());
+                actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
```
Pandrex247: If you're going to carry on execution, might it be better to set the exit code to warning so that the output is "command completed with warnings" (or w/e the exact string is)?

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: 🤦‍♂

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+            monitoringService = config.getMonitoringService();
+        } else {
+            actionReport.setMessage("Cound not find target: " + target);
+            actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
+            return;
+        }
+
+        Map<String, String> enabledModules = getPreviouslyEnabledModules(validModuleList);
+
+        //Setting all modules to "OFF"
```
jbee: Might want to extract below block into a method with a name that makes the comment unnecessary.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+                    return monitoringServiceProxy;
+                }, monitoringService);
+            } catch (TransactionFailure ex) {
+                logger.log(Level.WARNING, "Failed to execute the command set-monitoring-level, while setting modules to 'OFF': {0}", ex.getCause().getMessage());
+                actionReport.setMessage(ex.getCause().getMessage());
+                actionReport.setActionExitCode(ActionReport.ExitCode.FAILURE);
+            }
+        }
+
+        //Setting modules back to what they were previously
```
jbee: Again might want to extract a method.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+        if (verbose) {
+            actionReport.setMessage(getFormattedColumns(enabledModules).toString());
+        } else {
+            actionReport.setMessage("Reset " + enabledModules.size() + " modules");
+        }
+    }
+
+    private Map<String, String> getPreviouslyEnabledModules(List<String> validModules) {
+        Map<String, String> enabledModules = new HashMap<>();
+        if (!validModules.isEmpty()) {
```
jbee: Usually such a check before a for loop isn't needed unless you have a suspicion that creating an iterator is a surprisingly expensive operation for the list while `isEmpty` wouldn't be.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+            for (String module : validModules) {
+                String level = monitoringService.getMonitoringLevel(module);
+                if (!level.equals("OFF")) {
```
jbee: Unless you are sure about level never being `null` it is better to write such tests in reverse `!"OFF".equals(level)` as that is null safe.

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,136 @@
...
+import org.jvnet.hk2.config.TransactionFailure;
+
+/**
+ *
+ * @author Alan Roth
+ */
+@Service(name = "reset-monitoring-levels")
+@PerLookup
+@ExecuteOn({RuntimeType.DAS, RuntimeType.INSTANCE})
+@TargetType({CommandTarget.DAS, CommandTarget.STANDALONE_INSTANCE, CommandTarget.CLUSTER, CommandTarget.CONFIG})
```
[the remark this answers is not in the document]
AlanRoth: I'd assume so, since cluster is deprecated? And I believe targetting a specific config was a requested feature

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: I had the same thought and assumed it did, since the jira asks to turn all monitoring items to OFF once. Since the reset command would be used when "AMX metrics are broken due to some reason" there may be factors I don't know about affecting them that setting to OFF could maybe fix, so I played it safe.

AlanRoth: However I'm happy to change it if you believe its not neccesery

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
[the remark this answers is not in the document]
Pandrex247: Actually, yes, there is a case for targetting a config - ignore that bit :)

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,170 @@
...
+import org.glassfish.internal.api.Target;
+import org.jvnet.hk2.annotations.Service;
+import org.jvnet.hk2.config.ConfigSupport;
+import org.jvnet.hk2.config.TransactionFailure;
+
+/**
+ *
+ * @author Alan Roth
+ */
+@Service(name = "reset-monitoring-levels")
```
Pandrex247: I'm not sold on the name of this command.
At first glance without reading any docs I'd assume a command with this name would reset the monitoring levels to their defaults.

It might be better simply as "reset-monitoring" or "restart-monitoring"

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: Good idea

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: I had the same feeling, restart-monitoring has a nice ring to it

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: Didn't know that, pretty handy.

AlanRoth:
> 
> 
> Looks good. Just some code style comments.
> 
> I see you made it work changing the configuration directly. Maybe you can provide some insight (e.g. in the class javadoc) why this is enough or how the reset becomes effective.

I'm going to have to find out more about it; my knowledge on how changing the config directly works is limited

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: I'll keep that in mind in the future, should be changed now

AlanRoth:
Jenkins test please

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
```diff
@@ -0,0 +1,185 @@
...
+ * 
+ * This was Feature request due to AMX metrics "breaking" sometimes, where restarting it manually was somewhat tedious and prone to error.
+ * 
+ * @author Alan Roth
+ */
+@Service(name = "restart-monitoring")
+@PerLookup
+@ExecuteOn({RuntimeType.DAS, RuntimeType.INSTANCE})
+@TargetType({CommandTarget.DAS, CommandTarget.STANDALONE_INSTANCE, CommandTarget.DEPLOYMENT_GROUP, CommandTarget.CONFIG})
+public class ResetMonitoringLevels implements AdminCommand {
```
Pandrex247: Can you make the class name reflect the command name please? :)

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
Pandrex247: I don't think it's necessary - it's hard for an "off" metric to be wrong :P

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/ResetMonitoringLevels.java
AlanRoth: Oops

AlanRoth:
Jenkins test please

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/RestartMonitoring.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        
+        if (verbose) {
+            actionReport.setMessage(getFormattedColumns(previouslyEnabledModules).toString());
+        } else {
+            actionReport.setMessage("Restarted " + previouslyEnabledModules.size() + " modules");
+        }
+    }
+
+    private void setModulesToOff(Map<String, String> modules){
```
jbee: This should maybe be implemented with the help of `setModules`.

AlanRoth:
Jenkins test please

AlanRoth:
Jenkins test please

nucleus/admin/monitor/src/main/java/fish/payara/admin/monitor/cli/RestartMonitoring.java
AlanRoth: Yep

PR #4021, opened 2019-06-07
[PAYARA-3860] - Added synchronization block double check to avoid race condition

cubastanley:
Added a synchronized double check in PayaraInstanceImpl to avoid race condition - tested through sheer brute forcing of Thread.sleep() and spamming of functionality and problem no longer appears to be present

### DIFF
--- a/appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraInstanceImpl.java
+++ b/appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraInstanceImpl.java
@@ -152,7 +152,7 @@ public class PayaraInstanceImpl implements EventListener, MessageReceiver, Payar
     private PayaraExecutorService executor;
     
     private synchronized InstanceDescriptorImpl getMyInstanceDescriptor() {
-        if(me == null) {
+        if (me == null) {
             initialiseInstanceDescriptor();
         }
         return me;
@@ -243,7 +243,6 @@ public class PayaraInstanceImpl implements EventListener, MessageReceiver, Payar
     @SuppressWarnings({"unchecked"})
     public void event(Event event) {
         if (event.is(EventTypes.SERVER_READY)) {
-            //initialiseInstanceDescriptor();
             PayaraInternalEvent pie = new PayaraInternalEvent(PayaraInternalEvent.MESSAGE.ADDED, getMyInstanceDescriptor());
             ClusterMessage<PayaraInternalEvent> message = new ClusterMessage<>(pie);
             this.cluster.getEventBus().publish(INTERNAL_EVENTS_NAME, message);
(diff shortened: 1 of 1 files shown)


cubastanley:
Jenkins test please

appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraInstanceImpl.java
```diff
@@ -243,25 +240,29 @@ public void event(Event event) {
...
-        } 
-        // Adds the application to the clustered register of deployed applications
+        } // Adds the application to the clustered register of deployed applications
         else if (event.is(Deployment.APPLICATION_STARTED)) {
             if (event.hook() != null && event.hook() instanceof ApplicationInfo) {
                 ApplicationInfo applicationInfo = (ApplicationInfo) event.hook();
-                if (me == null) {//race check
-                    initialiseInstanceDescriptor();
+                //Race condition double check
+                if (me == null) {
```
Pandrex247: https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java
https://www.javaworld.com/article/2074979/double-checked-locking--clever--but-broken.html
Page 348 of the book I'm about to plonk on your desk :)

pdudits: Oh, why did I miss that article? :)

Likely the easiest solution then is to replace any access to `me` with synchronized getter that can also initialize the descriptor when null.

cubastanley: Is something like
`private synchronized InstanceDescriptorImpl getMyInstanceDescriptor() {
        if(me == null) {
            initialiseInstanceDescriptor();
        }
        return me;
    }`
the kind of thing you mean?

cubastanley:
Jenkins test please

appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraInstanceImpl.java
```diff
@@ -139,19 +138,26 @@
...
 
     @Inject
     private ApplicationRegistry appRegistry;
-    
+
     @Inject
     private PayaraExecutorService executor;
     
+    private synchronized InstanceDescriptorImpl getMyInstanceDescriptor() {
+        if(me == null) {
```
pdudits: space after if, please

appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraInstanceImpl.java
```diff
@@ -239,50 +243,46 @@ void postConstruct() {
     @SuppressWarnings({"unchecked"})
     public void event(Event event) {
         if (event.is(EventTypes.SERVER_READY)) {
-            initialiseInstanceDescriptor();
-            PayaraInternalEvent pie = new PayaraInternalEvent(PayaraInternalEvent.MESSAGE.ADDED, me);
+            //initialiseInstanceDescriptor();
```
pdudits: remove this one as well

cubastanley:
Jenkins Test Please

PR #4024, opened 2019-06-09
Fixes issue-3746

cubastanley:


cubastanley:
Was created as example for marketing task

PR #4026, opened 2019-06-10
SonarQube - Optimizing String.lastIndexOf() for single char in appserver

dvmarcilio:
This pull request contains fixes only for module _appserver_.

Replacing `String.lastIndexOf("/")` with `String.lastIndexOf('/')` which can be more performant.

This fixes SonarQube violations of the rule: [String function use should be optimized for single characters](https://rules.sonarsource.com/java/RSPEC-3027)


jbee:
jenkins test please

PR #4030, opened 2019-06-10
SonarQube - Using Wrapper.parseWrapper() when converting String to primitive

dvmarcilio:
parseWrapper() avoids creating a boxed primitive, and this is more performant when just the primitive is desired.

This fixes SonarQube violations of the rule: [Parsing should be used to convert "Strings" to primitives](https://rules.sonarsource.com/java/RSPEC-2130)


jbee:
jenkins test please

PR #4032, opened 2019-06-10
SonarQube - Replacing raw Collections EMPTY_... fields with generic methods empty…()

dvmarcilio:
This PR solves some cases of the SonarQube rule: ["Collections.EMPTY_LIST", "EMPTY_MAP", and "EMPTY_SET" should not be used](https://rules.sonarsource.com/java/RSPEC-1596).

I only replaced cases where a method returns with a typed Collection return type.

### DIFF
--- a/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigModel.java
+++ b/nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigModel.java
@@ -982,7 +982,7 @@ public final class ConfigModel {
         return ( dv );        
     }
     private static List<String> getMetadataFromDescription(Map<String, List<String>> map, String key) {
-        return map.get(key) == null ? Collections.emptyList() : map.get(key);
+        return map.getOrDefault(key, Collections.emptyList());
     }
     
     private static class SafeHk2Loader implements HK2Loader {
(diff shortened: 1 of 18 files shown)


jbee:
jenkins test please

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigModel.java
```diff
@@ -982,7 +982,7 @@ private static String getMetadataFieldKeyedAs(List<String> strings, String name)
         return ( dv );        
     }
     private static List<String> getMetadataFromDescription(Map<String, List<String>> map, String key) {
-        return map.get(key) == null ? Collections.EMPTY_LIST : map.get(key);
+        return map.get(key) == null ? Collections.emptyList() : map.get(key);
```
svendiedrichsen: This ternary expression calls get twice. Could be refactored to a temp variable or the use of getOrDefault method.

dvmarcilio: I changed it to use `getOrDefault()` because I couldn't think of a descriptive name for the temp variable.

svendiedrichsen: Thank you.

jbee:
jenkins test please

PR #4033, opened 2019-06-10
SonarQube - Instantiating wrappers - from 'new Wrapper(arg)' to 'Wrapper.valueOf(arg)'

dvmarcilio:
This PR fixes multiple SonarQube violations of the rule: [Constructors should not be used to instantiate "String", "BigInteger", "BigDecimal" and primitive-wrapper classes](https://rules.sonarsource.com/java/RSPEC-2129)

The rule says: 

> Constructors for String, BigInteger, BigDecimal and the objects used to wrap primitives should never be used. Doing so is less clear and uses more memory than simply using the desired value in the case of strings, and using valueOf for everything else.


jbee:
jenkins test please

dvmarcilio:
> Thanks @dvmarcilio , much appreciated. I think for the primitives in array initialisers you can drop the wrapper altogether and let the compiler do that, e.g. `Object[] foo = { 1, 2 }` works just fine.

Good point! I think I covered all the cases for this situation. While I was at it, I took the liberty to add blank spaces between the arguments.

jbee:
jenkins test please

PR #4034, opened 2019-06-11
PAYARA-3911 Fix domain.xml java-home is not honored in version-specific jvm-options (#4025)

### LINKED ISSUE #4025: The domain.xml java-home is not honored in version-specific jvm-options PAYARA-3911
# Description #
----------

On domain.xml, all the jvm-options are filtered according to the JDK used to run asadmin -- not the one configured to be used with the domain. This will result in errors if the arguments are not valid for the JDK set for the domain, especially using JDK before and after 9.

## Expected Outcome

Using JDK 1.8 to run asadmin, and JDK 11 configured in domains.xml, the domain should start normally, running on JDK 11.

## Current Outcome

The execution will fail with the following output:

    Waiting for domain1 to start .Error starting domain domain1.
    The server exited prematurely with exit code 1.
    Before it died, it produced the following output:
    
    -Xbootclasspath/p is no longer a supported option.
    Error: Could not create the Java Virtual Machine.
    Error: A fatal exception has occurred. Program will exit.
    
    Command start-domain failed.

1. Update the domain.xml, pointing the java-home to a JDK 11 installation, e.g.

    <java-config classpath-suffix=""
        debug-options="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=9009"
        java-home="/Library/Java/JavaVirtualMachines/jdk-11.0.1.jdk/Contents/Home"
        system-classpath="">

2. Make sure to use JDK 1.8 for asadmin

    $ java -version
    java version "1.8.0_211"
    Java(TM) SE Runtime Environment (build 1.8.0_211-b12)
    Java HotSpot(TM) 64-Bit Server VM (build 25.211-b12, mixed mode)

3. Leave the AS_JAVA unset (default value on glassfish/config/asenv.conf)

4. Use JDK 1.8 to start the domain

    ./asadmin start-domain domain1

## Context (Optional)

This issue came up when trying to start Payara 5.192 with JDK 11 from IntelliJ IDEA, with the system default JDK still being 1.8 (set on .bash_properties).

## Environment ##

- **Payara Version**: 5.192
- **Edition**: Full
- **JDK Version**: 1.8.0_211 and 11.0.1
- **Operating System**: Mac

vlumi:
Fix GFLauncher to use the correct JDK when determining the JVM options.

Using the javaExe already set for launching the domain, run it first with the -version parameter, and use the version parsed from its output (if available and matching) when setting the JVM options instead of the JVM currently running.

In case of non-fatal errors, the functionality falls back on the current JVM version (like before). On fatal errors a GFLauncherException will be thrown.

Fixes #4025

### DIFF
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.admin.launcher;
 
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -73,7 +73,7 @@ import java.util.stream.Collectors;
  */
 public abstract class GFLauncher {
 
-    private static final Pattern JAVA_VERSION_PATTERN = Pattern.compile(".* version \"([^\"]+)\".*");
+    private static final Pattern JAVA_VERSION_PATTERN = Pattern.compile(".* version \"([^\"\\-]+)(-.*)?\".*");
     private final List<String> commandLine = new ArrayList<String>();
     private final List<String> jvmOptionsList = new ArrayList<String>();
     private final GFLauncherInfo info;
--- a/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
+++ b/nucleus/admin/launcher/src/main/java/com/sun/enterprise/admin/launcher/GFLauncher.java
@@ -60,6 +60,8 @@ import static com.sun.enterprise.util.SystemPropertyConstants.*;
 import static com.sun.enterprise.admin.launcher.GFLauncherConstants.*;
 import com.sun.enterprise.util.JDK;
 import fish.payara.admin.launcher.PayaraDefaultJvmOptions;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


jbee:
jenkins test please

vlumi:
>  I am unsure about the pattern though.

I'll try to find some specification for the output of `java -version` and update the PR.

In addition to the pattern, an assumption is also made that it will be on the first line -- hopefully some specification will give these assumptions a more solid foundation.

vlumi:
> extract via a small program run with the target JRE that uses System.getProperty("java.specification.version") (as done for the JRE_VERSION constant).

I could not find this pattern in the Payara repository -- did I overlook some place? The same version as output by `java -version` is in the property `java.version`, and an even more detailed one on `java.runtime.version`.

The only document I found describing the version output format is a pretty old one: https://www.oracle.com/technetwork/java/javase/versioning-naming-139433.html

The output appears to be stable, with the version number always on the first line, with the expected format. Also, according to the specification, the output always matches the system property, so either approach should give the same result.

vlumi:
Found one more edge case that was not working correctly -- pre-release (-ea) versions, e.g. "9-ea". According to [JEP 223](https://openjdk.java.net/jeps/223), the version string may be added by pre-release and/or build information, which starts with a hyphen.

This is handled in JDK.initialize() by simply ignoring everything followed by the first hyphen -- I have updated this PR to take the same approach, only considering the pre-hyphen part as the version proper.

jbee:
@vlumi Thanks for looking into the JEPs and adopting the expression.

jbee:
jenkins test please

arjantijms:
Note that a modified version of the same `JDK` class that was updated here also exists in this repo:

https://github.com/payara/ecosystem-eclipse-plugin/blob/master/plugins/org.eclipse.payara.tools/src/org/eclipse/payara/tools/sdk/server/JDK.java

See specifically line [292](https://github.com/payara/ecosystem-eclipse-plugin/blob/master/plugins/org.eclipse.payara.tools/src/org/eclipse/payara/tools/sdk/server/JDK.java#L292) and onwards.

vlumi:
Another one is in ecosystem-netbeans-plugin, too:

https://github.com/payara/ecosystem-netbeans-plugin/blob/master/payara.tooling/src/org/netbeans/modules/payara/tooling/server/JDK.java

This one is almost identical to the one in this PR.

PR #3799, opened 2019-02-26
PAYARA-3824 Use KeyID from JWT header to find public key in JSON Web Key Set

lreimer:
This patch will use the KeyID `kid` header field (if present) from the JSON web token to find the matching public key in a JSON Web Key Set (JWKS) array. If there is no KeyID, then the first key entry will be used just like before.

Following changes:
- I had to split parsing and verification of the JWT in `JwtTokenParser.java` so that I can access the JWT header without having to parse the `SignedJWT` twice.
- I had to pass the obtained `keyID` down to the `createPublicKeyFromJWKS` method in order to be able to find the matching JWK.

PR #4036, opened 2019-06-11
TEST-198 Refactor JsonWebToken API in MP TCK Runner JWT Auth

AlanRoth:
**Note: Slightly misleading title**

The original refactors blocked PR PAYARA-3824 (#3799) however, it was suggested that the TokenParser class be removed from MicroProfile TCK runner and the code added to Payara itself, so I tried my best to make it compatible with #3799 

Tested the changes locally with #3799 and seems to work


appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jwt/JwtTokenParser.java
```diff
@@ -81,9 +84,28 @@ public JwtTokenParser() {
...
+            throw new IllegalStateException("Not JWT");
+        }
+    }
+    
+     public JsonWebTokenImpl parse(String bearerToken, String issuer, PublicKey signedBy) throws Exception {
+         try{
+             parse(bearerToken);
+             return verify(issuer, signedBy);
+         }catch(Exception e){
+             throw new IllegalStateException("", e);
```
cubastanley: Out of curiosity, why throw a new exception here rather than just printing the stack trace of the caught exception with the same type? Could be a good reason for this, I'd be curious to know

AlanRoth: I thought it was somewhat strange, but the code comes from TokenParser class in MP TCK

AlanRoth:
Closing due to getting clarification on the best way of doing this - this is not it

PR #4037, opened 2019-06-11
PAYARA-3864 EJB Java serialisation alongside with JSONB

jbee:
### Description
Adds implementation for https://payara.atlassian.net/wiki/spaces/PAYAR/pages/851836929/EJB+HTTP+Protocol+version+1 
including both Java serialisation and JSONB as well as automatic detection of served protocol by the client. In case both versions are supported the client chooses the newer version.

To explicitly select a serialisation type or protocol version two environment variables were added to the `RemoteEJBContextFactory`:
* `fish.payara.serialization` given as "java" or "json" (default) - not case sensitive
* `fish.payara.version` given as "0" or "1" (default)

### Testing performed
Component tests were added to both the endpoint and the client.
The existing integration test were adopted https://github.com/payara/payara-samples/pull/17 so they also cover the new protocol version for both Java Serialisation and JSONB.


jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/endpoint/pom.xml
```diff
@@ -86,6 +86,36 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
```
smillidge: Anyway this can be done without adding the commons dependency?

jbee: I use if for the `ClassLoaderObjectInputStream`. Its Apache licensed so I guess we should be allowed to only grab that class?

We already do use `commons-io` elsewhere so if we don't want to use it we should probably remove it everywhere? Maybe you can help me understand why this dependency is unwanted while we have lots of upstream dependencies. Is it because its a general utility that might cause trouble in case applications use it in a different version?

smillidge: If we are using it anyway then I'm fine with it.

smillidge: In general there are a number of reasons we don't want to bring in new 3rd party dependencies lightly;

1. It bloats the server
2. Applications may use a different version and therefore it can cause problems
3. It could increase the attack surface of the server. 
4. They need a license review and update to all the 3rd party license files

jbee: As discussed on Teams we avoid adding `commons.io` by repackaging the needed class.

jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/endpoint/src/test/java/fish/payara/ejb/endpoint/EjbOverHttpResourceTest.java
```diff
@@ -203,6 +204,18 @@ public static void stopServer() {
         }
     }
 
+    @Test
+    public void discover() {
+        try (Response response = target.path("/").request().build("HEAD").invoke()) {
+            assertEquals(Status.OK.getStatusCode(), response.getStatus());
+            assertEquals(2, response.getLinks().size());
+            assertEquals("http://localhost:" + port + "/jndi/lookup",
```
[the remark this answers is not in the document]
jbee: In the test there is only the root context since we testing on a HTTP server level.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/LookupV1.java
```diff
@@ -0,0 +1,178 @@
...
+                return equals(args[0]);
+            }
+            return invokeRemote(method, args);
+        }
+
+        private Object invokeRemote(Method method, Object[] args) throws Exception {
+            InvokeMethodRequest request = createRequest(method, args);
+            try (Response response = invoke.request(mediaType).buildPost(Entity.entity(request, mediaType)).invoke()) {
+                if (response.getStatus() == Status.OK.getStatusCode()) {
+                    return response.readEntity(InvokeMethodResponse.class).result;
```
pdudits: This will not he that easy, unfortunately. As noted in comment to JSON-B deserializer for this class, it wouldn't work for container generic return types (`List<Pojo>`). 
Using that deserializer (and making use of the provided return type) makes only sense if method's return type is an abstract class that is not a `Map` or `Collection`.

jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/endpoint/pom.xml
jbee: Backed adding as repackaged class since removing `commons.io` from nucleus results in errors. It is needed. Created PAYARA-3923 to investigate where it is used and if we could remove it.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContext.java
```diff
@@ -133,130 +132,151 @@ public Object lookup(String name) throws NamingException {
...
+    }
+
+    /**
+     * Determine supported protocol on server side and construct corresponding Lookup instance.
+     */
+    private Lookup determineLookup(URI root) throws NamingException {
+        ClientBuilder clientBuilder = null;
+        try {
+            clientBuilder = getClientBuilder();
+            Client client = clientBuilder.build();
```
Pandrex247: Could you not have just put this on one line?

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContext.java
```diff
@@ -133,130 +132,151 @@ public Object lookup(String name) throws NamingException {
...
+
+            LookupDiscoveryServiceImpl lookupDiscovery = new LookupDiscoveryServiceImpl();
+            LookupDiscoveryResponse discovery = lookupDiscovery.discover(client, root);
```
Pandrex247: Again, could have been on one line.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContext.java
```diff
@@ -133,130 +132,151 @@ public Object lookup(String name) throws NamingException {
...
+
+            Integer version = getVersion(discovery);
+            if (discovery.isV1target() && (version == null || version.intValue() == 1)) {
```
Pandrex247: Haven't you just done this check in the `getVersion` method?
`discovery.isV1target() && (version == null || version.intValue() == 1)`
vs.
`version == 1 && discovery.isV1target()`

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContextFactory.java
```diff
@@ -79,7 +79,8 @@
     public static final String JAXRS_CLIENT_HOSTNAME_VERIFIER = "fish.payara.hostnameVerifier";
     public static final String JAXRS_CLIENT_EXECUTOR_SERVICE = "fish.payara.executorService";
     public static final String JAXRS_CLIENT_CONNECT_TIMEOUT = "fish.payara.connectTimeout";
-
+    public static final String JAXRS_CLIENT_SERIALIZATION = "fish.payara.serialization";
+    public static final String JAXRS_CLIENT_VERSION = "fish.payara.version";
```
Pandrex247: I think both of these probably need to be a bit more specific.
fish.payara.version particularly sounds more like a property to get the version of Payara itself rather than the version of the JAX-RS client.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/EjbHttpProxyHandlerV1.java
```diff
@@ -0,0 +1,144 @@
...
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.UndeclaredThrowableException;
+import java.util.Map;
+
+import static java.util.Arrays.asList;
+import static javax.naming.Context.SECURITY_CREDENTIALS;
+import static javax.naming.Context.SECURITY_PRINCIPAL;
+
+final class EjbHttpProxyHandlerV1 implements InvocationHandler {
```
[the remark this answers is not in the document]
pdudits: Because the previous version was so informal, that it didn't deserve to be called one.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/LookupDiscoveryServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Default implementation for the {@link LookupDiscoveryService} that uses HEAD request at the root {@link URI} to
+ * discover options and falls back to v0 at {@code /ejb} in case no options found.
+ * 
+ * @author Patrik Duditš
+ * @author Jan Bernitt
+ */
+class LookupDiscoveryServiceImpl implements LookupDiscoveryService {
+
+    static final String INVOKER_V1_REL = "https://payara.fish/ejb-http-invoker/v1";
```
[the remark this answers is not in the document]
pdudits: As in "link rel" https://tools.ietf.org/html/rfc5988#section-4

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContextFactory.java
pdudits: I agree on that one, also JAXRS_CLIENT_VERSION seems to be telling different story. Probably it should  mention protocol version

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContext.java
jbee: Not exactly. The version selection is a bit delicate. The goal is to use the version specified if specified *and* available otherwise fallback on the newest available version. This part is the fallback while the `getVersion` method returns the selected version if specified and available.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContextFactory.java
jbee: I thought the same myself but the existing ones didn't use some namespacing so I did not add one. In hindsight it might have been nicer to have a common namespace like `fish.payara.ejb.http` for all of them. Any particular suggestions on the two names?

Pandrex247: Actually I completely overlooked those other properties - yeah it probably would be better to have a proper namespace for them.

TBH I like the namespace you gave, `fish.payara.ejb.http`, but I recognise that it's perhaps a bit verbose. I think the names are fine if you use that namespace though (e.g. `fish.payara.ejb.http.version`).

I kind of agree with Patrik in that the variable name for _JAXRS_CLIENT_VERSION_ is a bit misleading as it seems to be more to do with the EJB HTTP protocol version (unless I'm misreading it).

jbee:
jenkins test please

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/RemoteEJBContextFactory.java
jbee: Changed the name and properties.

jbee:
Thanks @Pandrex247

PR #4041, opened 2019-06-12
WIP: HTTP Client protocol negotiation

pdudits:
This decouples lookup mechanism from rest of the remote context, allowing to use different request and response processing based on protocol supported on server side.

To discover supported protocol a HEAD request to the root of invoker is made, and expects Link header pointing it to V1 lookup endpoint. 

Tested against 192 server endpoint, that one is discovered correctly, and (finally) handles 404 correctly ;)


appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/LookupDiscovery.java
```diff
@@ -0,0 +1,164 @@
...
+import java.util.Set;
+
+import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
+import static javax.ws.rs.core.Response.Status.Family.REDIRECTION;
+import static javax.ws.rs.core.Response.Status.Family.SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.Family.SUCCESSFUL;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.OK;
+
+class LookupDiscovery {
```
jbee: This class seems a mix of data structure and service that I personally find confusing. I'd like to see a more clear separation of state passed to a service with a interface to help mentally separate a concept and its invocation from implementation details.

appserver/ejb/ejb-http-remoting/client/src/main/java/fish/payara/ejb/http/client/LookupDiscovery.java
```diff
@@ -0,0 +1,164 @@
...
+    private WebTarget v1lookup;
+
+    LookupDiscovery(Map<String, Object> environment, Client client, URI root) {
+        this.environment = environment;
+        this.root = root;
+        this.resolvedRoot = root;
+        this.client = client;
+    }
+
+    void discover() throws NamingException {
```
jbee: To me this should return the result of the discovery as data which allows the caller to inspect.

pdudits:
Merged with main PR

PR #4043, opened 2019-06-13
PAYARA-3928 Fix NPE when notification via CDI eventbus comes in

arjantijms:
Signed-off-by: arjantijms <arjan.tijms@gmail.com>

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -236,7 +236,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             transactionManager.clearThreadTx();
         }
         
-        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (requestTracing != null && requestTracing.isRequestTracingEnabled() && handle.getSpanContextMap() != null) {
             startConcurrentContextSpan(invocation, handle);
         }
         


arjantijms:
Jenkins test please

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -236,7 +236,7 @@ else if(backupClassLoader != null) {
             transactionManager.clearThreadTx();
         }
         
-        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (requestTracing != null && requestTracing.isRequestTracingEnabled() && handle.getSpanContextMap() != null) {
```
svendiedrichsen: Shouldn't this NULL check be in the startConcurrentContextSpan method.

arjantijms: Would that make a lot of difference?

The check would have to be here then:

```java
private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
    if (handle.getSpanContextMap() == null) {
        return;
    }
    Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(
            Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)));

    SpanContext spanContext = tracer.extract(Format.Builtin.TEXT_MAP, new MapToTextMap(handle.getSpanContextMap()));
```

To my mind the check is at a similar level to request tracing being enabled (globally). If there's no span context map, then request tracing is not active for the current context/thread. In both cases we shouldn't even attempt to start a concurrent context span.

svendiedrichsen: I just thought that the reason for the check would be clearer if it is close to the usage. And you wouldn't need to check it again if the method would be called anywhere else.

arjantijms: It's not a wrong point indeed, although what the check is *actually* asking for is

"isTracingLocallyActive()"

From a semantic point of view it would be:

```java
if (isTracingGloballyEnabled() && isTracingLocallyActive()) {
   startConcurrentContextSpan();
}
```

vs

```java
startConcurrentContextSpan();
```

And then inside `startConcurrentContextSpan`:

```java
private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
    if (!isTracingGloballyEnabled() || !isTracingLocallyActive()) {
        return;
    }
    ...
}
```

In other words, always (or rather, often) call methods unconditionally, then check inside the methods right away if they should be called.

Might be best to discuss this as a base practice for the entire code convention. I'll merge this as is now, but certainly take the consideration on board.

thx!

Pandrex247: I'm not 100% convinced on this change - I also think it should be moved inside the `startConcurrentContextSpan` method.

We very recently had a similar change being suggested by Jan (via Sonar) about the ComponentInvocation being null, and I think I want to echo the same response: we still want to start a span, just exclude the stuff we can't get.
Reason being is that this is the entrypoint for essentially all async spans - the ramifications of not starting a span here will echo outwards.

So in this case, I'd suggest just editing the method so that it works a bit more like the `JaxrsClientRequestTracingFilter` class (see snippit below):
```

SpanContext parentSpanContext = (SpanContext) requestContext.getProperty(
                        PropagationHeaders.OPENTRACING_PROPAGATED_SPANCONTEXT);
                if (parentSpanContext != null) {
                    spanBuilder.asChildOf(parentSpanContext);
}
```
So _try_ and get the context from the InvocationContext, and create a span as a child of it if there is one, otherwise just create a span with no parent.

PR #4044, opened 2019-06-14
PAYARA-3933 Inconsistent metrics time units

### LINKED ISSUE #3883: Inconsistent metrics time units
Description 
When using injected Timer Metric, the chosen Metric units are not respected.

----------
When using:

@Inject
@Metric(name= "TestTimer", unit = MetricUnits.MILLISECONDS, description="test")
private Timer testTimer;
...
testTimer.update(1000, TimeUnit.MILLISECONDS);

## Expected Outcome

When using the metrics url: .../metrics/application

Should show:
  TYPE application:test_test_test_timer_max_seconds gauge
application:test_test_test_timer_max_seconds 1.0

## Current Outcome

  TYPE application:test_test_test_timer_max_seconds gauge
application:test_test_test_timer_max_seconds 1000000.0

## Environment ##

- **Payara Version**: payara-micro-5.184
- **Edition**: Micro
- **JDK Version**: 8 - OpenJDK
- **Operating System**: alpine Linux

vlumi:
Github Issue: #3883

Align the MicroService Metrics timer conversion factor to the actual type (ns) and the expected output (s) type on the plain-text output.


smillidge:
Thanks for the contribution

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/writer/PrometheusExporter.java
```diff
@@ -323,25 +323,13 @@ private double getConversionFactor(String unit) {
...
-                    break;
                 case MINUTES:
-                    conversionFactor = MINUTE_CONVERSION;
-                    break;
                 case HOURS:
-                    conversionFactor = HOUR_CONVERSION;
-                    break;
                 case DAYS:
-                    conversionFactor = DAY_CONVERSION;
+                    conversionFactor = NANOSECOND_CONVERSION;
```
jGauravGupta: `Second` is the base time unit in the Prometheus format, instead of returning `NANOSECOND_CONVERSION` for all, May you fix the conversion factor like `NANOSECOND_CONVERSION = 1d`, `MICROSECOND_CONVERSION = 1d / 1_000d`, `MILLISECOND_CONVERSION = 1d / 1_000_000d`, ... etc in `PrometheusExporter`.

vlumi: The conversion here is from nanoseconds to seconds, both of which are currently fixed. The other time conversion constants would be correct if the time would be stored internally in the units given (e.g. the unit parameter to @Metric).

As it stands, `fish.payara.microprofile.metrics.impl.TimerImpl` always stores its final measured values in the histogram as nanoseconds, and converts the value given to the public method `update(long, TimeUnit)` to nanoseconds as well:
```
@Override
public void update(long duration, TimeUnit unit) {
    update(unit.toNanos(duration));
}
```

jGauravGupta: Makes sense but histogram conversion factor should not be changed. I have updated the PR, please review it!

jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/writer/PrometheusExporter.java
vlumi: I have verified that it works correctly for both timer and histogram, with Prometheus and JSON output.

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

PR #4050, opened 2019-06-14
PAYARA-3950 Process all properties files

### LINKED ISSUE #4049: Microprofile config implementation only loads one properties file
# Description

Looking at the implementation of `ConfigProviderResolverImpl.initialiseApplicationConfig(...)` and it's change history I believe it was broken with commit 99370d7a433f9e86c0a54c9a00a21042e9d98f77.

## Expected Outcome

The [microprofile configuration spec](https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc) for default `ConfigSources` indicates that **all** `META-INF/microprofile-config.properties` files on the classpath should be loaded.

> A ConfigSource for each property file META-INF/microprofile-config.properties found on the classpath. (default ordinal = 100).

## Current Outcome

The configuration provider loads only the first properties file returned by the classloader. This results in default configuration values in any other files not being loaded into the config.

## Steps to reproduce

I experienced this bug during test execution in my project. I've been using Arquillian to test features and injecting a config file into the test deployment with defaults for the test case.

When I recently started exploring usage of the MP REST client, I added another config file to the project in the source classpath for default locations of the services. Adding this config hides the configuration included in the test deployment and causes the tests to fail because the values used by the test are no longer present in the configuration.

While Arquillian **should** only be using the config file inside the deployment during testing, I suspect that using the embedded Payara container has the source/test classpath available and the MP config is picking up both.

## Environment

- **Payara Version**: 5.192
- **Edition**: Micro 
- **JDK Version**: OpenJDK 8 
- **Operating System**: Linux

edthorne:
Fixes #4049

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
@@ -44,6 +44,7 @@ import java.io.InputStream;
 import java.lang.reflect.Type;
 import java.net.InetAddress;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
```diff
@@ -426,9 +428,9 @@ private Properties getPropertiesFromFile(ClassLoader appClassLoader, String file
             try (InputStream is = url.openStream()) {
                 p.load(url.openStream());
```
svendiedrichsen: Is this call twice to openStream intentional or should this variable _is_ be used for the second call?

edthorne: Clearly I was so focused on fixing my problem I didn't review the code for additional issues!

smillidge:
Jenkins test please

smillidge:
This needs an internal Jira reference.

arjantijms:
Internally tracked by PAYARA-3950

PR #3317, opened 2018-10-25
PAYARA-3072 PAYARA-3073 remove support for java.ext.dirs and java.endorsed.dirs

Cousjava:

PR #4051, opened 2019-06-14
PAYARA-3931

Pandrex247:
Was removed when we merged in the JDK 11 branch.

There are checks in place which should prevent it being used when running Payara Server with JDK11, as these properties are "hard" deprecated (JDK11 refuses to run if they're present).

This is essentially a revert of https://github.com/payara/Payara/pull/3317 with a couple of modifications


Pandrex247:
Closing for now as testing has revealed an underlying issue.

PR #2978, opened 2018-07-23
Add .gitlab-ci.yml

krichter722:
I'm aware that this might be closed pretty much immediately because the script requires maintenance and a CI is already set up, however I want to share my finding during the long process of enable GitLab CI for this project which might be useful to someone who wishes to develop on GitLab. [The build passes](https://gitlab.com/krichter/Payara/-/jobs/83585365).

PR #4053, opened 2019-06-16
Add .gitlab-ci.yml

krichter722:
see #2978


cubastanley:
Jenkins test please

MarkWareham:
Thank you for your contribution. Unfortunately the team here is not using Gitlab. We do not support it so not going to include in our codebase.

PR #4054, opened 2019-06-17
TEST-198 Refactor JsonWebToken API in MP TCK Runner JWT Auth 

AlanRoth:
Moved TokenParser class to Payara Server from MP TCK, plus changes to make it compatible with #3799 which is mostly the same changes #3799 does to the Parse method.

Relies on #44 in MicroProfile-TCK-Runners
https://github.com/payara/MicroProfile-TCK-Runners/pull/44


AlanRoth:
jenkins test please

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jwt/JwtTokenParser.java
```diff
@@ -81,8 +84,17 @@ public JwtTokenParser() {
...
+        rawToken = bearerToken;
+        signedJWT = SignedJWT.parse(rawToken);
+        
+        if(!checkIsJWT(signedJWT.getHeader())){
+            throw new IllegalStateException("Not JWT");
+        }
+    }
+    
+    public JsonWebTokenImpl verify(String issuer, PublicKey publicKey) throws Exception {
+        SignedJWT signedJWT = SignedJWT.parse(rawToken);
```
cubastanley: Don't feel great about the ambiguity of this local variable having the same name as the signedJWT variable at line 76 - could it have a slightly different name?

AlanRoth: signedJWT2? I don't think its worth the change as #3799 is refactoring this class, I just did the minimum to make it function for now with the least amount of changes; #3799 is due for a merge shortly after this.

AlanRoth: That line can be removed and a null check added - as you should call parse before calling verify

AlanRoth:
jenkins test please

AlanRoth:
jenkins test please

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jwt/JwtTokenParser.java
```diff
@@ -81,8 +84,19 @@ public JwtTokenParser() {
...
+        rawToken = bearerToken;
+        signedJWT = SignedJWT.parse(rawToken);
+        
+        if(!checkIsJWT(signedJWT.getHeader())){
+            throw new IllegalStateException("Not JWT");
+        }
+    }
+    
+    public JsonWebTokenImpl verify(String issuer, PublicKey publicKey) throws Exception {
+        if(signedJWT == null){
```
MattGill98: This needs formatting
```suggestion
        if (signedJWT == null) {
```

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jwt/JwtTokenParser.java
```diff
@@ -81,8 +84,19 @@ public JwtTokenParser() {
         this(Optional.empty(), Optional.empty());
     }
     
-    public JsonWebTokenImpl parse(String bearerToken, String issuer, PublicKey publicKey) throws Exception {
-        SignedJWT signedJWT = SignedJWT.parse(bearerToken);
+    public void parse(String bearerToken) throws Exception {
+        rawToken = bearerToken;
+        signedJWT = SignedJWT.parse(rawToken);
+        
+        if(!checkIsJWT(signedJWT.getHeader())){
```
MattGill98: Needs formatting
```suggestion
        if (!checkIsJWT(signedJWT.getHeader())) {
```

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/jwt/JwtTokenParser.java
```diff
@@ -72,6 +72,9 @@
     private final boolean enableNamespacedClaims;
     private final Optional<String> customNamespace;
 
+    private String rawToken;
+    private SignedJWT signedJWT;
```
MattGill98: It's typically bad practice to have two class variables depending directly on each other. Is it possible to make the signed JWT a local variable without breaking the TCK or introducing another problem?

AlanRoth: Shouldn't be too much trouble, I'll cook something up.

AlanRoth:
> 
> 
> Agreed with Matt's requests, otherwise I'm happy

I agree too, and I had a chat with Matt and #3799 will revert any changes back to using those two variables, it wouldn't be too bad of a shout on pointing it out on #3799

AlanRoth:
jenkins test please

PR #4075, opened 2019-07-05
[PAYARA-3909] - Can't Use Any of the Variable Types in JDBC Connection Pool Settings

cubastanley:
It is desired that variable types can be used in place of set values in JDBC connection pool settings (i.e. set maxPoolSize = ${MAX_POOL_SIZE_PROPERTY}) - this was previously not possible due to the validation process of JDBC connection pool properties (results in a NumberFormatException trying to parse an incorrect format string to an Integer) - the changes in this PR get around that

### DIFF
--- a/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
+++ b/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.jdbc.config;
 
--- a/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
+++ b/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
@@ -187,7 +187,7 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
     /**
      * Gets the value of the maxPoolSize property.
      *
-     * Maximum number of conections that can be created
+     * Maximum number of connections that can be created
      * 
      * @return possible object is
      *         {@link String }
@@ -251,7 +251,7 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      * maximum time in seconds, that a connection can remain idle in the pool.
      * After this time, the pool implementation can close this connection.
      * Note that this does not control connection timeouts enforced at the
-     * database server side. Adminsitrators are advised to keep this timeout
+     * database server side. Administrators are advised to keep this timeout
      * shorter than the database server side timeout (if such timeouts are
      * configured on the specific vendor's database), to prevent accumulation of
      * unusable connection in Application Server.
@@ -798,7 +798,7 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *
      * Init sql is executed whenever a connection created from the pool. 
      * This is mostly useful when the state of a
-     * connection is to be initialized
+     * connection is to be initialised
      * 
      * @return possible object is
      *         {@link String }
--- a/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
+++ b/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
@@ -58,7 +58,6 @@ import org.jvnet.hk2.config.*;
 import org.jvnet.hk2.config.types.Property;
 import org.jvnet.hk2.config.types.PropertyBag;
 
-import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.Pattern;
 import java.beans.PropertyVetoException;
@@ -175,8 +174,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="8")
-    @Min(value=0)
-    @Max(value=Integer.MAX_VALUE)
     String getSteadyPoolSize();
 
     /**
@@ -196,8 +193,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="32")
-    @Min(value=1)
-    @Max(value=Integer.MAX_VALUE)
     String getMaxPoolSize();
 
     /**
@@ -218,8 +213,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="60000")
-    @Min(value=0)
-    @Max(value=Integer.MAX_VALUE)
     String getMaxWaitTimeInMillis();
 
     /**
@@ -242,8 +235,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="2")
-    @Min(value=1)
-    @Max(value=Integer.MAX_VALUE)
     String getPoolResizeQuantity();
 
     /**
@@ -269,8 +260,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="300")
-    @Min(value=0)
-    @Max(value=Integer.MAX_VALUE)
     String getIdleTimeoutInSeconds();
 
     /**
@@ -507,8 +496,6 @@ public interface JdbcConnectionPool extends ConfigBeanProxy, Resource, ResourceP
      *         {@link String }
      */
     @Attribute (defaultValue="0")
-    @Min(value=0)
-    @Max(value=Integer.MAX_VALUE)
     String getValidateAtmostOncePeriodInSeconds();
 
     /**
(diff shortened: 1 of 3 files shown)


cubastanley:
Jenkins test please

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -71,17 +72,58 @@ public boolean isValid(final ResourcePool pool,
 
         if(poolFaults == ConnectionPoolErrorMessages.MAX_STEADY_INVALID) {
             if(pool instanceof JdbcConnectionPool) {
+                
+                Properties systemProps = System.getProperties();
+                
                 JdbcConnectionPool jdbcPool = (JdbcConnectionPool) pool;
                 String maxPoolSize = jdbcPool.getMaxPoolSize();
                 String steadyPoolSize = jdbcPool.getSteadyPoolSize();
```
AlanRoth: Whats an example string for steadyPoolSize? For context

cubastanley: It would be an integer value representing the minimum size of the connection pool (default value is "8")

AlanRoth: I meant the raw value, sorry, as I see you do some string manipulation, 'twas trying to figure out what you're trying to extract in that case.

cubastanley: So in the event that a system variable has been passed in, ${MAX_POOL_SIZE}, for example - that will cause an error when running the Integer.parseInt method (for obvious reasons) so in that case the method will retrieve the supposedly integer value stored in that system variable rather than trying to parse the variable name string instead

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -71,17 +72,58 @@ public boolean isValid(final ResourcePool pool,
...
+                        }
+                        steadyPoolSize = System.getenv(steadyPoolSize);
+                    } else {
+                        steadyPoolSize = systemProps.getProperty(steadyPoolSize);
+                    }
                 }
+                
                 if (maxPoolSize == null) {
                     maxPoolSize = Constants.DEFAULT_MAX_POOL_SIZE;
+                } else if(maxPoolSize.startsWith("$")) {
```
AlanRoth: These blocks seem pretty similar, maybe it's possible to extract it into a method to make it a bit more generic? Or perhaps only partly, some of the environment and system variable checks could be

cubastanley:
Jenkins test please

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
```diff
@@ -190,14 +187,12 @@
...
-     * Maximum number of conections that can be created
+     * Maximum number of connections that can be created
      * 
      * @return possible object is
      *         {@link String }
      */
     @Attribute (defaultValue="32")
-    @Min(value=1)
-    @Max(value=Integer.MAX_VALUE)
     String getMaxPoolSize();
```
svendiedrichsen: Isn't removing the min/max value begging for illegal values to be configured? Shouldn't they be at least checked for in code somewhere?

svendiedrichsen:
How about supporting the existing variables and their different types here: https://docs.payara.fish/documentation/payara-server/server-configuration/var-substitution/

cubastanley:
Jenkins test please

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/JdbcConnectionPool.java
cubastanley: I agree it is a rather risky move, although as far as I'm aware these annotations will always cause a validation exception in the event that a user tries to input a string that cannot be parsed to an Integer (as would be the case with a system variable name) I've altered the basic check in the implementation of JdbcConnectionPool so as to prevent the value from being below 1 but I do agree that it isn't the safest situation

jbee: I think @svendiedrichsen has a good point here. Surely extending the mechanism isn't part of the task as originally requested but if we don't include this or make it a precondition of this PR we introduce room for error that might never be closed again even when we would like to see this happen in the near future.

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,71 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
+                    maxPoolSize = getParsedVariable(maxPoolSize.substring(2, maxPoolSize.length() - 1));
+                    if(maxPoolSize == null) return false;
                 }
-                if (Integer.parseInt(maxPoolSize) <
-                        (Integer.parseInt(steadyPoolSize))) {
+                
+                //By this point it should be the case that the value is always castable
+                //to an integer
+                maxPoolSizeValue = Integer.parseInt(maxPoolSize);
```
MeroRai: This will cause `NumberFormatException` if the string does not have the appropriate numeric value.

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,71 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
+                //Check microprofile config for requested value
+                Config config = ConfigProvider.getConfig();
+                if (config.getValue(variableReference[1], String.class) != null && !config.getValue(variableReference[1], String.class).isEmpty()) {
+                    return config.getValue(variableReference[1], String.class);
+                }
+                break;
+            default:
+                //We got a system variable as no split occured
+                if (systemProps.getProperty(variableReference[0]) != null && !systemProps.getProperty(variableReference[0]).isEmpty()) {
+                    return systemProps.getProperty(variableReference[0]);
```
MeroRai: Can't you just simply use `System.getProperty(variableToRetrieve)` here.

cubastanley:
Jenkins Test Please

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,75 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
     }
 
+    public String getParsedVariable(String variableToRetrieve) {
+
+        String[] variableReference = variableToRetrieve.split("=");
+        
+        switch(variableReference[0]) {
+            case "ENV":
+                //Check environment variables for requested value
+                if (System.getenv(variableReference[1]) != null && !System.getenv(variableReference[1]).isEmpty()) {
```
jbee: `System.getenv(variableReference[1])` should be extracted to local var

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,75 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
+                    return System.getenv(variableReference[1]);
+                }
+                break;
+            case "MPCONFIG":
+                //Check microprofile config for requested value
+                Config config = ConfigProvider.getConfig();
+                if (config.getValue(variableReference[1], String.class) != null && !config.getValue(variableReference[1], String.class).isEmpty()) {
```
jbee: `config.getValue(variableReference[1], String.class)` should be extracted to local var

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,75 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
+                    return config.getValue(variableReference[1], String.class);
+                }
+                break;
+            default:
+                //We got a system variable as no split occured
+                if (System.getProperty(variableReference[0]) != null && !System.getProperty(variableReference[0]).isEmpty()) {
```
jbee: local var

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -65,23 +68,75 @@ public void initialize(final JdbcConnectionPoolConstraint constraint) {
...
         if(poolFaults == ConnectionPoolErrorMessages.MAX_STEADY_INVALID) {
             if(pool instanceof JdbcConnectionPool) {
+                
                 JdbcConnectionPool jdbcPool = (JdbcConnectionPool) pool;
                 String maxPoolSize = jdbcPool.getMaxPoolSize();
                 String steadyPoolSize = jdbcPool.getSteadyPoolSize();
-                if(steadyPoolSize == null) {
+                
+                int maxPoolSizeValue = 0;
+                int steadyPoolSizeValue = 0;
```
jbee: It should be possible to add a helper function where the min/max/default and actual value go in and the successfully parsed or resolved value comes out that can be used for both of these. Essentially making a method that does the extended version of that the annotations did semantically.

cubastanley:
Jenkins Test Please

PR #4087, opened 2019-07-12
PAYARA-3922 Updated jersey.version to 2.29.payara-p4

dmatej:
- fixes PAYARA-3922:
  - support for Stateless(name = "xxx")
  - support for implicit Local interface

### DIFF
--- a/nucleus/pom.xml
+++ b/nucleus/pom.xml
@@ -273,7 +273,7 @@
         <!-- JAX-RS -->
         <jax-rs-api.spec.version>2.1.5</jax-rs-api.spec.version>
         <jax-rs-api.impl.version>2.1.5</jax-rs-api.impl.version>
-        <jersey.version>2.29.payara-p3</jersey.version>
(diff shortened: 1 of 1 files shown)


nucleus/pom.xml
```diff
@@ -273,7 +273,7 @@
         <!-- JAX-RS -->
         <jax-rs-api.spec.version>2.1.5</jax-rs-api.spec.version>
         <jax-rs-api.impl.version>2.1.5</jax-rs-api.impl.version>
-        <jersey.version>2.29.payara-p3</jersey.version>
+        <jersey.version>2.29.payara-p4-SNAPSHOT</jersey.version>
```
Pandrex247: Snapshot? 
Not simply p4?

dmatej: You're right, it should be waiting yet for a review of the jersey and after that I shoud run into this PR with the final version.

dmatej:
Jenkins test please

PR #4096, opened 2019-07-18
PAYARA-4015 Module admingui/common-console replaced synchronized classes...

svendiedrichsen:
... with unsynchronized. extracted a method for code copy.

Apologies for the reformatting of one of the classes. The formatting was way off.

The replacement of the used Stack class in the inner class JsonChars is thread-safe because every call creates a new instance of the JsonChars class. Thus only a single thread will enter it.

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
@@ -55,6 +55,8 @@ import com.sun.jsftemplating.layout.descriptors.handler.HandlerContext;
 import org.glassfish.admingui.common.util.GuiUtil;
 
 import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
 
 /**
  *
@@ -91,10 +93,9 @@ public class NewSSLHandlers {
         Object ciphers = handlerCtx.getInputValue("ciphers");
         if (ciphers != null) {
             if (ciphers instanceof String) {
-                String[] ary = getSelectedCiphersList((String) ciphers);
-                ciphersList = getCiphers(ary);
+                ciphersList = Arrays.asList(getSelectedCiphersList((String) ciphers));
             } else {
-                ciphersList = getCiphers((String[]) ciphers);
+                ciphersList = Arrays.asList((String[]) ciphers);
             }
         }
         handlerCtx.setOutputValue("CommonCiphersList", getCommonCiphers(ciphersList));
@@ -128,19 +129,14 @@ public class NewSSLHandlers {
         handlerCtx.setOutputValue("ciphers", ciphers);
     }
 
-
-
     private static String[] getSelectedCiphersList(String selectedCiphers){
-        List<String> selItems = new ArrayList<>();
         if(selectedCiphers != null){
-            String[] sel = selectedCiphers.split(","); //NOI18N
-            for (String cName : sel) {
-                if (cName.startsWith("+")) { //NOI18N
-                    selItems.add(cName.substring(1));
-                }
-            }
+            return filterCiphers(
+                    Arrays.asList(selectedCiphers.split(",")),
+                    cipher -> cipher.startsWith("+"),
+                    cipher -> cipher.substring(1));
         }
-        return selItems.toArray(new String[0]);
+        return new String[0];
     }
 
     private static String processSelectedCiphers(String[] selectedCiphers, String ciphers){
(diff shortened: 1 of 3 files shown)


appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
```diff
@@ -126,100 +129,85 @@ public static void convertCiphersItemsToStr(HandlerContext handlerCtx) {
...
 
-        private static Vector getCiphersVector(String[] allCiphers){
-            Vector ciphers = new Vector();
-            if (allCiphers != null){
-                for(int i=0; i<allCiphers.length; i++){
-                    ciphers.add(allCiphers[i]);
+    private static String[] getSelectedCiphersList(String selectedCiphers){
+        List<String> selItems = new ArrayList<>();
+        if(selectedCiphers != null){
+            String[] sel = selectedCiphers.split(","); //NOI18N
```
jbee: We try to avoid short names or abbreviations for variable names. Maybe this can just be inlined in the `for` loop?

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
```diff
@@ -47,57 +48,59 @@
...
     )
     public static void convertToDifferentCiphersGroup(HandlerContext handlerCtx) {
-
-        Vector ciphersVector = null;
-        Object ciphers = (Object)handlerCtx.getInputValue("ciphers");
+        List<String> ciphersList = null;
+        Object ciphers = handlerCtx.getInputValue("ciphers");
         if (ciphers != null) {
             if (ciphers instanceof String) {
                 String[] ary = getSelectedCiphersList((String) ciphers);
```
jbee: Inline to avoid short variable name?

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
```diff
@@ -126,100 +129,85 @@ public static void convertCiphersItemsToStr(HandlerContext handlerCtx) {
...
         }
-        
-        private static String[] getOtherCiphers(Vector ciphers){
-            Vector bitCiphers = filterCiphers(ciphers, BIT_CIPHERS);
-            String[] ciphersList = (String[])bitCiphers.toArray(new String[bitCiphers.size()]);
-            return ciphersList;
+        return sb.toString();
+    }
+
+    private static List<String> getCiphers(String[] allCiphers){
```
jbee: Couldn't this method be replaced with `Arrays.asList(allCiphers)` or does the resulting `List` have to be modifiable?

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
```diff
@@ -126,100 +129,85 @@ public static void convertCiphersItemsToStr(HandlerContext handlerCtx) {
...
+        return filterCiphers(ciphers, BIT_CIPHERS).toArray(new String[0]);
+    }
+
+    private static List<String> filterCiphers(List<String> ciphers, Set<String> filterList){
+        List<String> listCiphers = new ArrayList<>();
+        if (ciphers != null) {
+            listCiphers.addAll(ciphers);
+            listCiphers.retainAll(filterList);
         }
+        return listCiphers;
```
jbee: If all methods using `filterCiphers` convert the returned list to an array directly maybe that common part could be done here as well?

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/NewSSLHandlers.java
```diff
@@ -126,100 +129,85 @@ public static void convertCiphersItemsToStr(HandlerContext handlerCtx) {
...
+        List<String> ephmCiphers = breakUpCiphers(new ArrayList<>(), ciphers, "_DHE_RSA_"); //NOI18N
+        ephmCiphers = breakUpCiphers(ephmCiphers, ciphers, "_DHE_DSS_"); //NOI18N
+        return ephmCiphers.toArray(new String[0]);
+    }
+
+    private static String[] getOtherCiphers(List<String> ciphers){
+        return filterCiphers(ciphers, BIT_CIPHERS).toArray(new String[0]);
+    }
+
+    private static List<String> filterCiphers(List<String> ciphers, Set<String> filterList){
```
jbee: If `filterCiphers` would take a `Predicate<String>` evaluating if a cipher is accepted all the filters could be done with the same function (`breakUpCiphers` could also use filter) and in one pass (`breakUpCiphers` would not needed to be called multiple times) and the code would be more readable formulating the condition of the ciphers to keep as a lambda. WDYT?

svendiedrichsen:
@jbee I have made the requested changes.

Pandrex247:
Jenkins test please

PR #4101, opened 2019-07-24
PAYARA-3775 Disabled failing domain creation in payara-minimal module

dmatej:
- currently fails with NPE


dmatej:
Jenkins test please

PR #4102, opened 2019-07-24
[WIP] PAYARA-3850 Monitoring Console Data Collection and trivial App

jbee:
### Synopsis
Introduces a new API to collect monitoring data consisting of two interface:

* `MonitoringDataSource` a `@Contract` implemented by `@Service` classes that have references to monitoring data in some form
* `MonitoringDataCollector` is passed to the sources so they can add any number of data points.

Data points are added in in a collection context that is implicit for the `MonitoringDataCollector` and can be extended using its `tag` method to create a new extended collection context.

The majority of actual code are the implementations of the `MonitoringDataSource`s where the different services add their information to the collection. 

The Web-app at this point simply resolves all `MonitoringDataSource`s and collects their data using some utilities. The data is made available as JSON in a REST API. A simple GUI has been created using the REST API to populate a dropdown of all metrics. When choosing one of them the diagram of the sliding window is shown and will update.

### Usage
Some URLs:
* "monitoring app": http://localhost:8080/monitoring-console-webapp/
* all points of all metrics in the sliding window: http://localhost:8080/monitoring-console-webapp/api/points
* all series (for dropdown): http://localhost:8080/monitoring-console-webapp/api/points/series/
* points of a specific metric (series following after `/series/`): http://localhost:8080/monitoring-console-webapp/api/points/series/ns=collect,duration

The data of points is not presented as arrays of numbers (times and values) because JS graph libraries do expect an array of objects from which they extract X and Y axis values using names of the fields holding the values.

### Testing
To test the app build `appserver/monitoring-console/webapp` and deploy it as `monitoring-console-webapp` (automatically suggested name). Since the data collection is partly in the app and partly modification to the server the server has to be deployed from this feature branch as well. 

### Review
Note that the webapp uses jQuery and Chart.js which uses MIT License
https://github.com/chartjs/Chart.js/blob/master/LICENSE.md


jbee:
jenkins test please

jbee:
Closed in favour of #4121 - Another PR will be issues later for the webapp part.

PR #4106, opened 2019-07-25
Fixed random errors identified by IDE

jbee:
Quickly fixed two occurrences of errors that IDE reports as such which gets kind of annoying.

### DIFF
--- a/nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/mbean/LoggingImpl.java
+++ b/nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/mbean/LoggingImpl.java
@@ -411,7 +411,7 @@ public final class LoggingImpl extends AMXImplBase //implements /*Logging,*/ Log
     public Map<String, Number>[] getErrorInfo() {
         unimplemented();
 
-        final List<Map<String, Object>> infos = null; //getLogMBean().getErrorInformation();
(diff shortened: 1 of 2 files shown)


nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/impl/mbean/LoggingImpl.java
```diff
@@ -411,7 +411,7 @@ private Properties attributesToProps(List<Attribute> attrs) {
     public Map<String, Number>[] getErrorInfo() {
         unimplemented();
 
-        final List<Map<String, Object>> infos = null; //getLogMBean().getErrorInformation();
```
jbee: `null` would cause NPE in line 416 - except we never get there because `unimplemented()` above already throws an exception.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
```diff
@@ -63,7 +60,6 @@
         this.store = new BoundedTreeSet<>(0);
         this.maxStoreSize = store.size();
         this.strategy = strategy;
-        this.events = events;
```
jbee: This does not have any effect, with it deleted it becomes apparent that the `events` field wasn't used at all so I removed it.

jbee:
jenkins test please

PR #4107, opened 2019-07-25
PAYARA-3830 Cleanup sonar warnings as a result of MP Metrics 2.0

Cousjava:


appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/BindableFaultToleranceMetrics.java
```diff
@@ -128,7 +130,13 @@ public void addToHistogram(String keyPattern, long duration) {
     @Override
     public void linkGauge(String keyPattern, LongSupplier gauge) {
         String metricName = metricName(keyPattern);
-        Gauge<?> existingGauge = metricRegistry.getGauges().get(metricName);
+        Gauge<?> existingGauge = null;
+        for (Map.Entry<MetricID, Gauge> gaugeEntry : metricRegistry.getGauges().entrySet()) {
+            if (gaugeEntry.getKey().getName().equals(metricName)) {
+                existingGauge = gaugeEntry.getValue();
+                break;
```
jbee: To avoid `break` I'd prefer to extract a function that then could use `return gaugeEntry.getValue()` instead. Alternatively: shouldn't it be possible to create an instance of `MetricID` from the `metricName` `String`?

Cousjava: You can create an instance of `MetricID` from a `String`, but that will be tagless and not match any entries that do have tags/

Cousjava:
Jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/BindableFaultToleranceMetrics.java
jbee: Hm... that implies that multiple metric IDs could have the named looked for why would it be correct to use the first one found? Or maybe the ID is not a good value to use as key of the map? Something is odd here.

PR #4109, opened 2019-07-26
PAYARA-4016 Mp Metrics does not work with multiple MBeans with the sa…

Cousjava:
…me name but different tags

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadataHelper.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadataHelper.java
@@ -54,6 +54,7 @@ import org.eclipse.microprofile.metrics.Metadata;
 import org.eclipse.microprofile.metrics.MetadataBuilder;
 import org.eclipse.microprofile.metrics.Metric;
 import org.eclipse.microprofile.metrics.MetricFilter;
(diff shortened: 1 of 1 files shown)


Cousjava:
Jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadataHelper.java
```diff
@@ -93,8 +94,13 @@
 
         List<MBeanMetadata> unresolvedMetadataList = resolveDynamicMetadata(metadataList);
         for (MBeanMetadata beanMetadata : metadataList) {
+            List<Tag> tags = new ArrayList<>();
+            for (fish.payara.microprofile.metrics.jmx.XmlTag tag : beanMetadata.getTags()) {
```
Pandrex247: I'm aware it's moved from slightly further down, but is there a reason this class isn't imported? I don't see a conflict

svendiedrichsen:
There are some Sonar-Bugs introduced just a few days ago to the class MBeanMetadata. You might want address these as well.

https://sonarcloud.io/project/issues?id=org.glassfish.main%3Aglassfish-main-aggregator&open=AWwWEaokTNTtgv3YW2fk&resolved=false&sinceLeakPeriod=true&types=BUG

Cousjava:
Jenkins test please

PR #4115, opened 2019-07-31
PAYARA-4026 Fault Tolerance 2.0.1 fix for failing TCK

Cousjava:
Circuit Breaker should be considered success if exception thrown is not
marked as failOn


Cousjava:
Jenkins test please

PR #4117, opened 2019-07-31
SPEC-259: Correctly identify if timer service relates to managed entity

pdudits:
In some previous refactorings the invocation of the second constructor
disappeared, replacing with runtime instance check.

Fixes all 35 failures in `ejb/ee/timer`, and thus entire `ejb` suite


pdudits:
jenkins test please

PR #4121, opened 2019-08-02
PAYARA-3850 Data collection API and sources

jbee:
### Synopsis
Introduces a new API to collect monitoring data consisting of two interface:

* `MonitoringDataSource` a `@Contract` implemented by `@Service` classes that have references to monitoring data in some form
* `MonitoringDataCollector` is passed to the sources so they can add any number of data points.

Data points are added in in a collection context that is implicit for the `MonitoringDataCollector` and can be extended using its `tag` method to create a new extended collection context.

The majority of actual code are the implementations of the `MonitoringDataSource`s where the different services add their information to the collection. 

More details can be found in the design document https://payara.atlassian.net/wiki/spaces/PAYAR/pages/870055939/Payara+Monitoring+Console+Design

### DIFF
--- a/appserver/payara-appserver-modules/healthcheck-checker/src/main/java/fish/payara/healthcheck/mphealth/MicroProfileHealthChecker.java
+++ b/appserver/payara-appserver-modules/healthcheck-checker/src/main/java/fish/payara/healthcheck/mphealth/MicroProfileHealthChecker.java
@@ -1,7 +1,7 @@
 /*
  *  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  * 
- *  Copyright (c) [2018] Payara Foundation and/or its affiliates. All rights reserved.
+ *  Copyright (c) [2018-2019] Payara Foundation and/or its affiliates. All rights reserved.
  * 
  *  The contents of this file are subject to the terms of either the GNU
  *  General Public License Version 2 only ("GPL") or the Common Development
@@ -43,9 +43,12 @@
 package fish.payara.healthcheck.mphealth;
 
 import com.sun.enterprise.config.serverbeans.Domain;
+
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import javax.inject.Inject;
+import javax.validation.constraints.Pattern;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.client.WebTarget;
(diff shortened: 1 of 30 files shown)


jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -126,6 +140,80 @@ public void init() {
         }
     }
 
+    @Override
+    public void collect(MonitoringDataCollector rootCollector) {
+        MonitoringDataCollector metricsCollector = rootCollector.in("metric");
+        if (metricsEnabled != null) {
+            metricsCollector.collect("enabled", metricsEnabled);
+        }
+        if (metricsSecure != null) {
```
jGauravGupta: `getSecureMetrics()` is marked as deprecated, please use the `isSecurityEnabled()`:
`metricsCollector.collect("securityEnabled", isSecurityEnabled());`

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -126,6 +140,80 @@ public void init() {
         }
     }
 
+    @Override
+    public void collect(MonitoringDataCollector rootCollector) {
+        MonitoringDataCollector metricsCollector = rootCollector.in("metric");
```
jGauravGupta: Minor but may you name it: `metrics`

appserver/payara-appserver-modules/healthcheck-checker/src/main/java/fish/payara/healthcheck/mphealth/MicroProfileHealthChecker.java
```diff
@@ -115,44 +120,58 @@
...
                     try {
-                        String endpoint = metricsConfig.getEndpoint();
-
-                        URI remote = buildURI(server, endpoint);
-
-                        result.add(pingHealthEndpoint(remote));
-
+                        result.add(pingHealthEndpoint(instanceName,  buildURI(server, metricsConfig.getEndpoint())));
                     } catch (URISyntaxException ex) {
+                        serverHttpStatus.put(instanceName, 420);
```
jGauravGupta: Constant may be defined for `420` status code as it is not part of the HTTP standard.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
jbee: So far I did use singular for the namespace and type tags. Overall this reads better but this is all up to discussion. I should just end up consistent.

appserver/payara-appserver-modules/healthcheck-checker/src/main/java/fish/payara/healthcheck/mphealth/MicroProfileHealthChecker.java
jbee: Actually I think I should change it to a 5xx code. Both cases I used it it is a server error more then a client error.

jbee:
jenkins test please

appserver/payara-appserver-modules/healthcheck-checker/src/main/java/fish/payara/healthcheck/mphealth/MicroProfileHealthChecker.java
jbee: Only one really applying is plain 500 so I changed to that.

PR #4124, opened 2019-08-02
PAYARA-4034 Cleanup sonar warnings for nucleus/common/amx-core

Cousjava:


Cousjava:
Jenkins test please

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/util/jmx/JMXUtil.java
```diff
@@ -1071,7 +979,7 @@ public static boolean isSetter(Method method)
         return (method.getName().startsWith(SET) &&
                 method.getParameterTypes().length == 1 &&
                 method.getParameterTypes()[ 0] != Attribute.class &&
-                method.getReturnType().getName().equals("void"));
+                method.getReturnType().equals(Void.TYPE));
```
jbee: `method.getReturnType() == void.class` is fine as well

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/util/jmx/ReadWriteAttributeFilter.java
```diff
@@ -37,83 +37,67 @@
...
+    }
 
+    /**
+     */
+    @Override
+    public boolean filterAttribute(final MBeanAttributeInfo info) {
+        throw new RuntimeException("Can't get here");
+    }
 
+    public static final ReadWriteAttributeFilter READ_ONLY_FILTER = new ReadWriteAttributeFilter() {
```
jbee: I think these all could be lambdas since it is not important that they implement the `ReadWriteAttributeFilter` but the `AttributeFilter` interface. If another type distinction is wanted the `ReadWriteAttributeFilter` could also be made an interface I think.

Pandrex247:
a) Conflicts
b) Although the build has been deleted due to it being too old, this actually simply failed to compile

jbee:
@Pandrex247 , @Cousjava compilation failure is likely to be caused by unfinished merge in one file.

Cousjava:
Jenkins test please

nucleus/common/amx-core/src/main/java/org/glassfish/admin/amx/util/ArrayConversion.java
```diff
@@ -158,24 +158,20 @@ private ArrayConversion()
...
-                theSet = Collections.emptySet();
-                break;
+                return Collections.emptySet();
             case 1:
-                theSet = Collections.singleton(array[ 0]);
-                break;
+                return Collections.singleton(array[ 0]);
             default:
-                theSet = new HashSet<T>();
+                Set<T> theSet = new HashSet<T>();
```
jbee: You could use `new HashSet<T>(Arrays.asList(array));`

PR #4129, opened 2019-08-05
PAYARA 3792 asadmin print certificate

dmatej:
- prints informations about a certificate saved in a file
- supported file types: der, pem, p12, jks, jceks
- if using additional provider, it's jar must be in payara5/glassfish/lib/asadmin/ directory (f.e. BC)

### DIFF
--- a/nucleus/admin/server-mgmt/pom.xml
+++ b/nucleus/admin/server-mgmt/pom.xml
@@ -51,7 +51,7 @@
     <modelVersion>4.0.0</modelVersion>
     <artifactId>server-mgmt</artifactId>
     <packaging>glassfish-jar</packaging>
-    
+
     <name>admin-server-management</name>
     <description>Server Management</description>
 
@@ -77,7 +77,7 @@
                     <include>**/*.xsd</include>
                 </includes>
             </resource>
-        </resources>        
+        </resources>
         <plugins>
             <plugin>
                 <groupId>org.codehaus.mojo</groupId>
@@ -129,7 +129,7 @@
             </plugin>
         </plugins>
     </build>
-    
+
     <dependencies>
         <dependency>
             <groupId>org.glassfish.main.common</groupId>
@@ -161,6 +161,14 @@
         <dependency>
             <groupId>org.glassfish.annotations</groupId>
             <artifactId>logging-annotation-processor</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcpkix-jdk15on</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
@@ -1,0 +1,254 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) 2019 Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ *     The contents of this file are subject to the terms of either the GNU
+ *     General Public License Version 2 only ("GPL") or the Common Development
+ *     and Distribution License("CDDL") (collectively, the "License").  You
+ *     may not use this file except in compliance with the License.  You can
+ *     obtain a copy of the License at
+ *     https://github.com/payara/Payara/blob/master/LICENSE.txt
+ *     See the License for the specific
+ *     language governing permissions and limitations under the License.
+ *
+ *     When distributing the software, include this License Header Notice in each
+ *     file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ *     GPL Classpath Exception:
+ *     The Payara Foundation designates this particular file as subject to the "Classpath"
+ *     exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ *     file that accompanied this code.
+ *
+ *     Modifications:
+ *     If applicable, add the following below the License Header, with the fields
+ *     enclosed by brackets [] replaced by your own identifying information:
+ *     "Portions Copyright [year] [name of copyright owner]"
+ *
+ *     Contributor(s):
+ *     If you wish your version of this file to be governed by only the CDDL or
+ *     only the GPL Version 2, indicate your decision by adding "[Contributor]
+ *     elects to include this software in this distribution under the [CDDL or GPL
+ *     Version 2] license."  If you don't indicate a single choice of license, a
+ *     recipient has the option to distribute your version of this file under
+ *     either the CDDL, the GPL Version 2 or to extend the choice of license to
+ *     its licensees as provided above.  However, if you add GPL Version 2 code
+ *     and therefore, elected the GPL Version 2 license, then the option applies
+ *     only if the new code is made subject to such option by the copyright
+ *     holder.
+ */
+package fish.payara.admin.servermgmt.cli;
+
+import com.sun.enterprise.admin.cli.CLICommand;
+import com.sun.enterprise.admin.servermgmt.KeystoreManager;
+import com.sun.enterprise.security.auth.realm.certificate.CertificateRealm;
+
+import java.io.File;
+import java.security.KeyStore;
+import java.security.Provider;
+import java.security.PublicKey;
+import java.security.Security;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.security.interfaces.DSAKey;
+import java.security.interfaces.RSAKey;
+import java.util.Collection;
+import java.util.concurrent.Callable;
+
+import javax.security.auth.x500.X500Principal;
+
+import org.glassfish.api.Param;
+import org.glassfish.api.admin.CommandException;
+import org.glassfish.api.admin.CommandValidationException;
+import org.glassfish.hk2.api.PerLookup;
+import org.jvnet.hk2.annotations.Service;
+
+/**
+ * Prints information about a certificate given in a file.
+ * <p>
+ * Uses RFC-2253 and the {@link CertificateRealm#OID_MAP} for the principal distinguished name.
+ * The Subject from the output can be directly used for a role mapping when using client
+ * certificate authentication, f.e.<br>
+ * <code>&lt;principal-name&gt;CN=My Client,OU=Payara,O=Payara Foundation,L=Great Malvern,ST=Worcestershire,C=UK&lt;/principal-name&gt;</code>
+ * <p>
+ * <b>WARNING:</b> JKS and JCEKS may be removed from some JDKs, then this command may fail.
+ *
+ * @author David Matejcek
+ */
+@Service(name = "print-certificate")
+@PerLookup
+public class PrintCertificateCommand extends CLICommand {
+
+    @Param(name = "file", primary = true)
+    String file;
+
+    @Param(name = "certificatealias", optional = true, defaultValue = "")
+    String certificateAlias;
+
+    @Param(name = "providerclass", optional = true)
+    String providerClass;
+
+    private Provider provider;
+
+    private File derFile;
+
+    private File keystoreFile;
+
+    private char[] keystorePassword;
+
+    @Override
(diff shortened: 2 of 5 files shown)


dmatej:
Jenkins test please

nucleus/admin/server-mgmt/pom.xml
```diff
@@ -162,5 +162,17 @@
             <groupId>org.glassfish.annotations</groupId>
             <artifactId>logging-annotation-processor</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.62</version>
```
jbee: I'm not sure about this particular artifact but `org.bouncycastle` stuff usually occurs in multiple modules so it maybe should be manged using dependency management?

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,255 @@
...
+            } catch (final ReflectiveOperationException e) {
+                throw new CommandValidationException("The provider class was not found on classpath.", e);
+            }
+        }
+
+        if (isDerEncodedFile()) {
+            this.derFile = sourceFile;
+        } else if (isKeystoreFile()) {
+            this.keystoreFile = sourceFile;
+            if (!StringUtils.ok(this.certificateAlias)) {
```
jbee: Either use `ok` with import or without consistently. :D

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,255 @@
...
+
+    private char[] keystorePassword;
+
+    @Override
+    protected void validate() throws CommandException, CommandValidationException {
+        super.validate();
+        if (!ok(this.file)) {
+            throw new CommandValidationException("The file with the certificate must be specified.");
+        }
+        final File sourceFile = new File(this.file);
```
jbee: Not 100% sure about our style guide but from what I did I'd assume we do use `this` when we assign to a field but do not use it when reading a field unless the field is shadowed by a local variable.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,255 @@
...
+import javax.security.auth.x500.X500Principal;
+
+import org.glassfish.api.Param;
+import org.glassfish.api.admin.CommandException;
+import org.glassfish.api.admin.CommandValidationException;
+import org.glassfish.hk2.api.PerLookup;
+import org.jvnet.hk2.annotations.Service;
+
+/**
+ * Prints information about a certificate given in a file.
```
jbee: Maybe you can give a bit of background why we would want to print it to system out?

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,255 @@
...
+            return "PKCS12";
+        }
+        throw new IllegalStateException("Reached unreachable code, validation is incomplete!");
+    }
+
+
+    private X509Certificate getCertificate() throws CommandException {
+        if (this.derFile != null) {
+            return getCertificateFromDerFile();
+        } if (this.keystoreFile != null) {
```
jbee: `if` should be on a new line.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,255 @@
...
+    private static X509Certificate getX509Certificate(final CertificateSupplier supplier) throws KeyStoreException {
+        final Certificate certificate  = supplier.get();
+        if (certificate instanceof X509Certificate) {
+            return (X509Certificate) certificate;
+        }
+        throw new KeyStoreException("The certificate was found but it is not supported X509 certificate.");
+    }
+
+
+    private interface CertificateSupplier {
```
jbee: You could use `Callable` but it might be better readability you are after.

nucleus/admin/server-mgmt/src/test/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommandTest.java
```diff
@@ -0,0 +1,287 @@
...
+        this.command = new PrintCertificateCommandMock();
+        this.command.certificateAlias = alias;
+        this.command.file = file.getAbsolutePath();
+    }
+
+
+    @Before
+    public void init() {
+        this.stdout = new ByteArrayOutputStream();
+        System.setOut(new PrintStream(this.stdout));
```
jbee: I don't think we want to change output stream but not restore it after the test. Maybe that is done by JUnit? Don't know.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
pzygielo: Or - drop this interface definition and use `java.util.function.Supplier`.

jbee: @pzygielo `java.util.function.Supplier` lacks the checked exception and cannot be used.

pzygielo: > @pzygielo `java.util.function.Supplier` lacks the checked exception and cannot be used.

:+1: right, missed that!

nucleus/admin/server-mgmt/pom.xml
dmatej: It is used in javaee7-samples, but not in payara, but it could be (as a test dependency).
But generally I agree, it is good idea to put it to dependency-management at least to have versions in one place.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
dmatej: I think it is good to always have some visual difference between fields and local variables. It forces to think about the context of the variable on the first sight. Payara code conventions page tells nothing about this...

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
dmatej: Supplier does not allow to throw exceptions.

dmatej: Callable throws java.lang.Exception. But yes, because currently I'm not interested in the exception type, there is no need for the interface.

nucleus/admin/server-mgmt/src/test/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommandTest.java
dmatej: It is restored via @BeforeClass - @AfterClass mechanism.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
jbee: Sure. Most IDEs will already make it visually very different right out of the box and if not you could configure it. Its basically the same argument people had to prefix every field with underscore or `m_`. At least I am happy we have IDEs now that don't make us prefix stuff.

nucleus/admin/server-mgmt/src/test/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommandTest.java
jbee: Funny, I was looking for that and didn't see it then but had no problem now.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
dmatej: I updated javadocs. The original requirement was to print "standardized" DN useful for principal-name in role mapping. I went more far, now it prints all basic informations about the certificate and also allows to use different provider (but it's jar must be on asadmin classpath).
Because usually it is a client certificate, it probably is not present in any domain keystore, so that's another parameter - the input file (standard types supported). 
This asadmin command has no dependency on domains, it is more standalone utility than some server tool.

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
dmatej: The problem in IDEs is that most of the code in the world has so many issues that one more warning is simply ignored.
I never used prefixes, because they change too many things, finally the readibility would be worse than better.
The `this` does not change anything in compiled class, it is only about syntax. It is so obvious for me, because it helped to cleanup classes with 5000 lines, hundred variables, one method ... yes, that's not an issue in Payara, so here you don't see any advantage of `this`.
I think more similar discussion is about using `final` ...

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
```diff
@@ -0,0 +1,254 @@
...
+ * <p>
+ * <b>WARNING:</b> JKS and JCEKS may be removed from some JDKs, then this command may fail.
+ *
+ * @author David Matejcek
+ */
+@Service(name = "print-certificate")
+@PerLookup
+public class PrintCertificateCommand extends CLICommand {
+
+    @Param
```
Pandrex247: We normally name all parameters to make finding them easier, even if the name is the default.

Also as the only non-optional parameter this could be made primary so that the user doesn't have to explicitly enter the parameter name each time.

dmatej: 1) Ok, I found that there are no conventions, but in most cases its lowercased name of the attribute

2) In fact I hated it here, because then I even cannot use the --file and this parameter must be last ... if I want to fix alias, I have to move cursor before the path ... but naturally the file would be the first thing I would write. So here I would defend current implementation - especially because this command is somehow special (no domain, only reads file).

dmatej: Ad 1) Fixed.

Pandrex247:
Not sure if I'm being silly, but this doesn't seem to have any auto-generated help?
`asadmin print-certificate --help` gives `Cannot find man page for command: print-certificate`

Is this because it's a CLI command rather than an Admin command?

dmatej:
Then I am silly too :-/
I found more commands with this problem, but I don't know how it is processed.

dmatej:
Ok, finally I will mark it as primary, because it is "standard" and I hope users use certificate files more often than keystores, because then it is incomfortable as the primary argument must be the last ... (that was always pain).

nucleus/admin/server-mgmt/src/main/java/fish/payara/admin/servermgmt/cli/PrintCertificateCommand.java
dmatej: ad 2) Finally accepted, because

- for der/pem file there is only the file argument, no need to prefix it with --file
- for keystores yes, I don't like to move cursor before long path for every execution (typos in password), but ... yes, it is standard. 

It would be nice if it would be possible to use the --file too, but ... nobody requests this feature except me :) ... and I can live with that.

dmatej:
Jenkins test please

PR #4131, opened 2019-08-07
SPEC-267: Fail operations on expired nonpersistent timers

pdudits:
Spec mandates throwing an exception on accessing nonpersistent
expired timers when accessed over public API.
Since NonPersistentEJBTimerService is base class for persistent
implementations, which pass their non-persistent functionality to it,
it needs internal method for verifying it would handle that timer.

This fixed ejb32/lite/timer tests, while not breaking ejb30/timer/tests at same time.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/NonPersistentEJBTimerService.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/NonPersistentEJBTimerService.java
@@ -41,7 +41,6 @@
 
 package com.sun.ejb.containers;
 
-import static com.sun.ejb.containers.EJBTimerService.logger;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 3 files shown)


pdudits:
jenkins test please

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/NonPersistentEJBTimerService.java
```diff
@@ -106,17 +127,20 @@ protected boolean cancelNonPersistentTimer(TimerPrimaryKey timerId)
...
-        if (rt != null) {
-            if (rt.isCancelled()) {
-                // already cancelled or removed
-                return true;
-            }
-
-            cancelTimerSynchronization(null, timerId,
-                    rt.getContainerId(), ownerIdOfThisServer_, false);
+        if (rt == null) {
+            return false;
```
jGauravGupta: In case of cancellation, If a timer is null or already canceled then the container must throw `javax.ejb.NoSuchObjectLocalException` instead of `FinderException` (EJB Spec 3.2 - Section 13.4.4 Timer Cancellation).

pdudits: Yes. This exception is wrapped into `NoSuchObjectLocalException` in `TimerWrapper`

PR #4132, opened 2019-08-08
SPEC-282: Introduce JVM option to turn off variable support outside config

pdudits:
Performing variable expansion in descriptors is not spec compliant, therefore it needs to be switched off for TCK tests. The system property is called `fish.payara.substitution.disable` and needs to be set to true in order to disable the feature.

Some of the expansions need to be done regardless of the setting, therefore the original method was split into two.

### DIFF
--- a/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdUtil.java
+++ b/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdUtil.java
@@ -66,7 +66,9 @@ public final class OpenIdUtil {
         if (configResult.isPresent()) {
             return configResult.get();
         }
-        result = (T) TranslatedConfigView.getTranslatedValue(result);
+        if (type == String.class) {
+            result = (T) TranslatedConfigView.expandValue((String) result);
+        }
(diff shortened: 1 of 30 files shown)


appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdUtil.java
```diff
@@ -66,7 +66,9 @@ private OpenIdUtil() {
         if (configResult.isPresent()) {
             return configResult.get();
         }
-        result = (T) TranslatedConfigView.getTranslatedValue(result);
+        if (type == String.class) {
+            result = (T) TranslatedConfigView.expandValue((String) result);
+        }
```
jbee: Is the _else_ case handled elsewhere?

pdudits: `getTranslatedValue` does transformation only on Strings. I think the story is, that it inherited too general signature, when it was extracted from the interceptor-style method `invoke` before 2011.

jGauravGupta:
jenkins test please

PR #4133, opened 2019-08-08
PAYARA-4031 Master Password Synchronisation Inconsistencies Across Nodes

MattGill98:
For anyone reviewing, a brief explanation of the behaviour change. The following functionality causes a problem:

1. Create a node
2. Create an instance on that node
3. Change master password for domain
4. Attempt to start instance (node still has the old master password saved, which doesn't unlock the keystore as the keystore has been synchronised from the DAS to the instance, so must be unlocked using the DAS master password)
5. Attempt to change master password for node (can't as the `change-master-password` command for the node will check the password against the keystore of each instance, which has now been updated with the DAS keystore)

This PR attempts to remedy this by causing the `change-master-password` command, when run against a node, to only change the saved password for the node, and not touch existing keystores. This will mean that the keystores will only have their password changed when changing the master password for the DAS, and the master password should then be updated on each node to reflect this change.

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -576,8 +576,12 @@ public abstract class LocalServerCommand extends CLICommand {
         for (int i = 0; i < times; i++) {
             // XXX - I18N
             String prompt = STRINGS.get("mp.prompt", (times - i));
-            char[] mpvArr = super.readPassword(prompt);
-            mpv = mpvArr != null ? new String(mpvArr) : null;
+            try {
+                char[] mpvArr = super.readPassword(prompt);
+                mpv = mpvArr != null ? new String(mpvArr) : null;
+            } catch (Throwable t) {
+                throw new CommandException(STRINGS.get("no.console"), t);
+            }
             if (mpv == null)
                 throw new CommandException(STRINGS.get("no.console"));
             // ignore retries :)
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -83,7 +83,11 @@ public abstract class LocalServerCommand extends CLICommand {
     ////////////////////////////////////////////////////////////////
     private ServerDirs serverDirs;
     private static final LocalStringsImpl STRINGS = new LocalStringsImpl(LocalDomainCommand.class);
-    private static final String DEFAULT_MASTER_PASSWORD = "changeit";
+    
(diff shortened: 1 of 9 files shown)


MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
```diff
@@ -171,89 +164,88 @@ public int execute(String... argv) throws CommandException {
...
-            FileProtectionUtility.chmod0600(pwdFile);
-        } catch (Exception ex) {
-            throw new CommandException(strings.get("masterPasswordFileNotCreated", pwdFile),ex);
+    protected String findOldPassword() throws CommandException {
+        // Fetch from password file
+        String oldPassword = passwords.get(OLD_PASSWORD_ALIAS);
+
+        // Prompt user
+        if (oldPassword == null) {
+            char[] opArr = super.readPassword(STRINGS.get("old.mp"));
```
Pandrex247: This (or possibly the whole command) probably needs to be hidden behind an `isInteractive()` check.
Currently if you run `asadmin -I false change-master-password Noddy1` it will still prompt you for a password.

nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
```diff
@@ -566,8 +576,12 @@ private String retry(int times) throws CommandException {
         for (int i = 0; i < times; i++) {
             // XXX - I18N
             String prompt = STRINGS.get("mp.prompt", (times - i));
-            char[] mpvArr = super.readPassword(prompt);
-            mpv = mpvArr != null ? new String(mpvArr) : null;
+            try {
+                char[] mpvArr = super.readPassword(prompt);
+                mpv = mpvArr != null ? new String(mpvArr) : null;
+            } catch (Throwable t) {
```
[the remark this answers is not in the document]
MattGill98: I'd done that as a easy way to catch the sneaky thrown IOException from JLine, but turns out I can catch it as an exception without the compiler kicking off! I've added the check anyway just to be safe

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
MattGill98: Good shout, the `getPassword` method did this but the `readPassword` method didn't. I've added the check

MattGill98:
jenkins test please

MarkWareham:
@dmatej changes requested are now resolved?

MattGill98:
jenkins test please

MattGill98:
jenkins test please

dmatej:
@MarkWareham  Now they are :-)

dmatej:
Jenkins test please

PR #4134, opened 2019-08-08
SPEC-320: Fix PostConstruct import

pdudits:
`ejbCreate` wasn't considered as postconstruct method, because wrong postconstruct class was passed down the stream.


Pandrex247:
Jenkins test please

PR #4135, opened 2019-08-09
PAYARA-4065 Ensure that the threadpools are correctly initialized even when Embedded server is restarted

realityforge:
While there was a previous attempt to do this in place, it failed if the shutdown of the thread pool took any amount of time. And thus calling `stop()` and then `start()` in rapid succession would fail. This change attempts to eliminate this scenario

### DIFF
--- a/nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
+++ b/nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
@@ -131,15 +131,36 @@ public class PayaraExecutorService implements ConfigListener, EventListener {
     @Override
     public void event(Event event) {
         if (event.is(Deployment.ALL_APPLICATIONS_LOADED)) {
-            // there is awkward point in embedded EJBContainer initialization, where same server
-            // instance is started twice. In such case, we need to reinitialize.
-            if (threadPoolExecutor.isShutdown()) {
+            // Embedded containers can be started and stopped multiple times.
+            // Thus we need to initialize anytime the server instance is started.
+            if (null == threadPoolExecutor) {
                 initialiseThreadPools();
             }
         } else if (event.is(EventTypes.SERVER_SHUTDOWN)) {
-            threadPoolExecutor.shutdown();
-            scheduledThreadPoolExecutor.shutdown();
+            terminateThreadPools();
         }
+    }
+
+    private void terminateThreadPools() {
+        threadPoolExecutor.shutdown();
+        scheduledThreadPoolExecutor.shutdown();
+
+        // Wait until the schedulers actually terminate
+        try {
+          threadPoolExecutor.awaitTermination(5, TimeUnit.SECONDS);
+          scheduledThreadPoolExecutor.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (InterruptedException ignored) {
+        }
+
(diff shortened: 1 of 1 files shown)


nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
```diff
@@ -131,14 +129,29 @@ public void postConstruct() {
...
-            // there is awkward point in embedded EJBContainer initialization, where same server
-            // instance is started twice. In such case, we need to reinitialize.
-            if (threadPoolExecutor.isShutdown()) {
-                initialiseThreadPools();
-            }
+            // Embedded containers can be started and stopped multiple times.
+            // Thus we need to initialize anytime the server instance is started.
+            initialiseThreadPools();
         } else if (event.is(EventTypes.SERVER_SHUTDOWN)) {
             threadPoolExecutor.shutdown();
```
svendiedrichsen: Maybe have a method like `terminateThreadPools` for this?

realityforge: Seems reasonable ... and done.

pdudits:
jenkins test please

pdudits:
jenkins test please

nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
```diff
@@ -120,8 +120,6 @@ public void postConstruct() {
         if (transactions != null) {
             transactions.addListenerForType(PayaraExecutorServiceConfiguration.class, this);
         }
-
-        initialiseThreadPools();
```
pdudits: I believe it is this point that makes the tests fail. Reinitializing at ALL_APPLICATIONS_LOADED was a heuristic I found to work for multiple embedded starts, but the threadpool is already needed during load.

nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
```diff
@@ -131,15 +129,34 @@ public void postConstruct() {
...
     public void event(Event event) {
         if (event.is(Deployment.ALL_APPLICATIONS_LOADED)) {
-            // there is awkward point in embedded EJBContainer initialization, where same server
-            // instance is started twice. In such case, we need to reinitialize.
-            if (threadPoolExecutor.isShutdown()) {
-                initialiseThreadPools();
-            }
+            // Embedded containers can be started and stopped multiple times.
+            // Thus we need to initialize anytime the server instance is started.
+            initialiseThreadPools();
```
pdudits: And in such case the test whether he executor is alive is needed.

nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
realityforge: Seems reasonable. I have reworked it slightly to follow the same strategy but use null value of pool as the trigger to recreate

pdudits:
jenkins test please

PR #4137, opened 2019-08-09
 SPEC-321: Create instance at the and of AroundConstruct chain

pdudits:
SPEC-321: Create instance at the and of AroundConstruct chain

Producing the instance eagerly triggered interceptor chain too early with incomplete context, and instance was created after interceptors completed, not when end of chain was reached (i. e. in the middle of interceptors processing).

Also clean up unused imports and methods.

And trying to fix embedded shell, which sometimes build with incorrect MANIFEST.MF, causing tests with embedded EjbContainer to fail.


pdudits:
Jenkins test please

PR #4138, opened 2019-08-09
Change parent POM of nucleus to top-level parent

Pandrex247:
This helps prevent IntelliJ from getting stuck trying to resolve the project

PR #4139, opened 2019-08-10
PAYARA-4043 Removed Jvnet parent

dmatej:
- new parent relation introduced problems with it's configuration poisoning
  all it's grandchildren - profiles source and javadoc did not work.
  This is now fixed.
- source and javadoc profiles now produce same artifacts as before
- it is possible to combine profiles and build phases
- removed duplicated and redundant configuration parts.
- modules moved from profiles to the main pom.
- aggregator also changed the id

### DIFF
--- a/nucleus/pom.xml
+++ b/nucleus/pom.xml
@@ -46,30 +46,15 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>net.java</groupId>
-        <artifactId>jvnet-parent</artifactId>
-        <version>5</version>
-        <relativePath />
+        <groupId>fish.payara</groupId>
+        <artifactId>payara-aggregator</artifactId>
+        <version>5.193-SNAPSHOT</version>
     </parent>
 
-    <groupId>fish.payara</groupId>
     <artifactId>payara-nucleus-parent</artifactId>
-    <version>5.193-SNAPSHOT</version>
     <packaging>pom</packaging>
     <name>Payara Nucleus Parent Project</name>
     <url>https://github.com/payara/Payara</url>
-
-    <scm>
-        <connection>scm:git:git@github.com:payara/payara.git</connection>
-        <url>https://github.com/payara/Payara/tree/master</url>
-        <developerConnection>scm:git:git@github.com:payara/payara.git</developerConnection>
-        <tag>payara-server-5.193-SNAPSHOT</tag>
-    </scm>
-
-    <issueManagement>
-        <system>Github</system>
-        <url>https://github.com/payara/Payara/issues</url>
-    </issueManagement>
 
     <licenses>
       <license>
@@ -79,16 +64,6 @@
         <comments>A business-friendly OSS license</comments>
       </license>
     </licenses>
-
-
-    <developers>
-        <developer>
-            <name>Payara Team</name>
-            <email>info@payara.fish</email>
-            <organization>Payara Foundation</organization>
-            <organizationUrl>http://www.payara.fish</organizationUrl>
-        </developer>
-    </developers>
 
     <repositories>
         <!-- Try Maven central first, not last, which happens when omitted here -->
@@ -162,7 +137,7 @@
             </snapshots>
         </repository>
 
-        <!--  
+        <!--
         <repository>
            <id>payara-patched-externals-local</id>
            <name>Payara Patched Externals local</name>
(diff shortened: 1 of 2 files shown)


dmatej:
Jenkins test please

dmatej:
This is an alternative PR to #4138

nucleus/pom.xml
```diff
@@ -162,7 +137,7 @@
             </snapshots>
         </repository>
 
-        <!--  
+        <!--
```
pdudits: Since we're cleaning up, can we also get rid of the commented out parts of pom?

nucleus/pom.xml
```diff
@@ -180,15 +155,15 @@
...
-	   <!--<url>file:///opt/PayaraDev/Payara_PatchedProjects</url>-->
-           	<releases>
-           		<enabled>true</enabled>
-           	</releases>
-            	<snapshots>
-                	<enabled>false</enabled>
-            	</snapshots>
+             <name>Payara Patched Externals</name>
+             <url>https://raw.github.com/payara/Payara_PatchedProjects/master</url>
+             <!--<url>file:///opt/PayaraDev/Payara_PatchedProjects</url>-->
```
pdudits: And also remove these alternatives....

pom.xml
```diff
@@ -39,181 +39,145 @@
...
-                <module>api</module>
-            </modules>
-            <properties>
-                <maven.test.skip>true</maven.test.skip>
-            </properties>
             <build>
+                <finalName>payara-${project.version}</finalName>
                 <plugins>
                     <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
```
pdudits: I'm personally not fan of dropping groupId, there are enough rules to remember about maven already :)

pom.xml
```diff
@@ -39,181 +39,145 @@
...
-    Portions Copyright [2017-2018] Payara Foundation and/or affiliates
+    Portions Copyright [2017-2019] Payara Foundation and/or affiliates
 -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <groupId>org.glassfish.main</groupId>
-    <artifactId>glassfish-main-aggregator</artifactId>
+    <groupId>fish.payara</groupId>
+    <artifactId>payara-main-aggregator</artifactId>
```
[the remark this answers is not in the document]
dmatej: Hmm. I think I will merge those parents into one soon. But I need there are no pull requests that would be in conflict, so the is only the one step of the proces. But ok, I can rename it.

Pandrex247:
@pdudits  @dmatej  Discussion from previous PR about naming of groupId:
https://github.com/payara/Payara/pull/4062#discussion_r300003137

dmatej:
> When I did this in Payara BOM PR, I had to change few versions of dependencies that were indeed pulled from javanet. Didn't you have issues building with empty local repo?

I think JVnet did not declare dependencies, only plugin dependencies, some of them were updated, but nothing big (I think it was only jar plugin and maybe source plugin).
Build with an empty local repo was successful.

nucleus/pom.xml
dmatej: Yes and no, only if we create documentation how to use fake local repo for patched projects, this is somethink like howto for beginers :-)
And no for another reason - target of this issue is not cleanup, I only moved common things to a new parent to avoid new duplications.
I have to look how much conflicts I created to your "BOM" issue, but I want to minimize that.

nucleus/pom.xml
dmatej: Same ...

pom.xml
dmatej: It's default, I think there is no need to remember any "new rule". You don't even use the groupId when you are looking for a documentation and you can even run build without that.
https://maven.apache.org/guides/introduction/introduction-to-plugin-prefix-mapping.html

On the command line the maven- and -plugins are defaults too, so usually you run commands simply like this:
`mvn dependency:build-classpath -pl :payara`

nucleus/pom.xml
pdudits: I will redo BOM in next sprint, so no worries about conflicts.

dmatej:
Jenkins test please

nucleus/pom.xml
```diff
@@ -46,31 +46,16 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>net.java</groupId>
-        <artifactId>jvnet-parent</artifactId>
-        <version>5</version>
-        <relativePath />
+        <groupId>fish.payara</groupId>
```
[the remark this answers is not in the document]
dmatej: Then I would have to update all childrens. And I even don't believe the "server" is useful here, it is even controversial, because "micro" is "serverless".

pdudits:
jenkins test please

nucleus/pom.xml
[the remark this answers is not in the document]
pdudits: Ok let's leave nucleus as is for now, we'll move it with next great renaming.

dmatej:
Jenkins test please

PR #4140, opened 2019-08-10
PAYARA-4012 Removed javaee modules, not used for years

dmatej:
- also javaee.version was currently resolved to 8.193-SNAPSHOT, because
  minor_version is Payara's minor version
- ips profile was empty
- activating the javaee-api module would cause failure


pdudits:
Jenkins test please

PR #4150, opened 2019-08-14
PAYARA-4064 Fixed dependency on operating system's java

dmatej:
- maven plugin should prioritize maven environment over
  the user's environment
- honor ~/.mavenrc settings when executing asadmin via Ant exec.


dmatej:
Jenkins test please

PR #4154, opened 2019-08-15
PAYARA-4067: Put JSON-B api on AppClient classpath

pdudits:
This will make `javax.json.bind-api` to appear in the `Class-Path` manifest of `acc-standalone` (also known as `gf-client.jar`), and therefore makes classes available to application client runtime.

Makes Jakarta EE 8 signature tests pass for these packages.


pdudits:
jenkins test please

PR #4161, opened 2019-08-20
PAYARA-4033 RemoteCommand notify now use false as default when callin…

Cousjava:
…g super constructor


Cousjava:
Jenkins test please

PR #4164, opened 2019-08-21
PAYARA-4075 Initial branding for Monitoring Console

jbee:
PR extracts JS and CSS to separate files. Branding is added to CSS. JS is doing more proper scoping providing the functionality as the `MonitoringConsole` global object.

There is a config for each chart. UI manipulates configs and triggers rendering.

I don't think it is useful to look too closely on code level. Its early stage of development and the CSS/JS will change a lot. I know I have to encapsulate it better and make a more clear distinction between internal and external state and organise the JS code more. These things will happen. This should just give us a branded version of the app.

For **testing** it:

* build the app
* deploy it (manually)
* open localhost `/monitoring-console-webapp/` context
* dropdown should contain a number of entries, select one
* press "add" button (multiple charts can be added)

Once added a chart can be removed with `[ - ]` button in its top right corner.
Config is stored in local storage so refreshing the page brings up last state. There is some odd behaviour when playing with adding and subtracting and reloading. I am aware of that and plan to change the way it works so it is more consistent.


appserver/monitoring-console/webapp/src/main/webapp/monitoring-console.js
```diff
@@ -0,0 +1,213 @@
+Chart.defaults.global.defaultFontColor = "#fff";
```
Pandrex247: Copyright

appserver/monitoring-console/webapp/src/main/webapp/monitoring-console.css
```diff
@@ -0,0 +1,45 @@
+body {
```
Pandrex247: Copyright

jbee:
Done.

jbee:
jenkins test please

MattGill98:
jenkins test please

PR #4165, opened 2019-08-21
PAYARA-4079: Fix NPE in DeployCommand

pdudits:
When deployment fails, prepare returns null and action report contains details of the failure. Line 542 caused NullPointerException in such case, silently erasing that diagnostic information.

### DIFF
--- a/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
+++ b/nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
@@ -537,11 +537,11 @@ public class DeployCommand extends DeployCommandParameters implements AdminComma
             Transaction t = deployment.prepareAppConfigChanges(deploymentContext);
             span.finish(); // next phase is launched by prepare
             Deployment.ApplicationDeployment deplResult = deployment.prepare(null, deploymentContext);
-            if(!loadOnly) {
+            if(deplResult != null && !loadOnly) {
                 // initialize makes its own phase as well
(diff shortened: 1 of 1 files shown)


pdudits:
jenkins test please

nucleus/deployment/admin/src/main/java/org/glassfish/deployment/admin/DeployCommand.java
```diff
@@ -537,11 +537,11 @@ public void execute(AdminCommandContext context) {
             Transaction t = deployment.prepareAppConfigChanges(deploymentContext);
             span.finish(); // next phase is launched by prepare
             Deployment.ApplicationDeployment deplResult = deployment.prepare(null, deploymentContext);
-            if(!loadOnly) {
+            if(deplResult != null && !loadOnly) {
                 // initialize makes its own phase as well
                 deployment.initialize(deplResult.appInfo, deplResult.appInfo.getSniffers(), deplResult.context);
             }
-            ApplicationInfo appInfo = deplResult.appInfo;
+            ApplicationInfo appInfo = deplResult != null ? deplResult.appInfo : null;
```
jbee: `appInfo` is later used in `deployment.registerAppInDomainXML(appInfo, deploymentContext, t);`. From the looks of that statement I'd assume this should not be called with `null`.

pdudits: Yes, but when `appInfo` is null, action exit code is not success so the branch that uses it is not invoked (yes there are too many signalling paths in deployment process).

PR #4171, opened 2019-08-22
PAYARA-4087 Allow the use of a single char as the operand for Micro

Cousjava:
For commands used in pre- or post-boot file allow the operand to be a single
char


Cousjava:
Jenkins test please

PR #4181, opened 2019-08-27
TEST-367 Import Payara Sample Suite

MattGill98:
See individual commits for change overview


MattGill98:
jenkins test please

MattGill98:
Can a reviewer run the tests locally to check?

dmatej:
Fix the parent as Maven says:
```
dmatej@dmatej-G5-5587:~/work/repo/git/Payara/appserver/tests/payara-samples$ mvn clean install
[INFO] Scanning for projects...
Downloading from ossrh: https://oss.sonatype.org/content/repositories/snapshots/fish/payara/payara-aggregator/5.193-SNAPSHOT/maven-metadata.xml
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for fish.payara.samples:parent-aggregator:pom:5.193-SNAPSHOT
[WARNING] 'parent.relativePath' of POM fish.payara.samples:parent-aggregator:[unknown-version] (/home/dmatej/work/repo/git/Payara/appserver/tests/payara-samples/pom.xml) points at org.glassfish.main.tests:tests instead of fish.payara:payara-aggregator, please verify your project structure @ fish.payara.samples:parent-aggregator:[unknown-version], /home/dmatej/work/repo/git/Payara/appserver/tests/payara-samples/pom.xml, line 9, column 13
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
...
```

MattGill98:
Now waiting on https://github.com/payara/payara-samples/pull/30/files to remove the embedded dependencies.

MattGill98:
I believe it's good to go now!

MattGill98:
Strictly speaking as it is this PR should be able to run against any Payara version still (I've been deliberately unobtrusive in that respect). Does anyone reviewing have any strong opinions on whether I should strip out the irrelevant stuff in each PR like the wrong `ee-client` dependency etc?

Pandrex247:
IMO any version specific stuff can be cleaned up separately

dmatej:
Only a single note: I can run payara-samples tests only against any compatible preinstalled Payara version with the payara-remote profile. But I would expect here would be used payara5 profile instead - this crashes with well known problem with injections and incompatible JDK from system that I fixed in glassfishbuild maven plugin.I tried to fix it here in the same way, but it had no effect, I don't know why ...
But the same result I have in both repositories.

MattGill98:
@dmatej This is because when you activate the `payara-server-remote` profile, the `payara5` profile disables itself. I've drafted up a change to use [our maven extension](https://github.com/payara/ecosystem-maven-regex-profile-activation-extension) to fix this, but annoyingly there's a bug in it because the `payara.version` is set to `project.version`, which is tested literally. I've made a PR to fix it here: https://github.com/payara/ecosystem-maven-regex-profile-activation-extension/pull/5. Once it's approved and merged I'll push the inclusion of the plugin to this branch

MattGill98:
@dmatej I think it should be fixed now, new regex profile activator release woo!

dmatej:
First, **sorry** for confusion, I can't see any remote profile in payara-samples, it was javaee7/8 repository.

Hmm. Still the same. The problem is not with profles, but as it is described in the SO link - with the environment. When build starts built payara.zip itself, it uses asadmin executed directly - but then it does not honor maven environment, but runs system process with system default environment - and uses default java command (which is from OpenJDK11 on Kubuntu), because generally there is no AS_JAVA or JAVA_HOME set.

**But ok, this is not related to this PR. I will try to fix it later.**

Btw, I would rename the module from:
```
    <groupId>fish.payara.samples</groupId>
    <artifactId>parent-aggregator</artifactId>

``` 
to 
```
    <groupId>fish.payara.samples</groupId>
    <artifactId>payara-samples-aggregator</artifactId>
``` 
It affects -rf or -pl arguments for maven, for example. Current state would conflict with the payara aggregator so we would have to use full module id.

Pandrex247:
Jenkins test please

MarkWareham:
Jenkins test please

PR #4184, opened 2019-08-29
PAYARA-3819: Remove java.net repositories

pdudits:
The repositories do not contain anything relevant outside one unreleased version of fighterfish osgi-cdi, which already has more recent version on central.

Also for local patched_projects it's just better to place that as a mirror in settings.xml, no need to pollute pom.xml with that.


pdudits:
Jenkins test please

Issue #4186, opened 2019-08-29
PAYARA-4104 Disabling Hazelcast for 5.193 Server causes startup failure


mwnorman:
# Description #

Server fails to start when hazelcast is disabled.

## Current Outcome
On macOS, the port number for (VNC) screen-sharing is 5900 which is used by hazelcast.
Disabling it via:
`asadmin set-hazelcast-configuration --enabled=false`
results in a NPE:

```
MultiException stack 1 of 1
MultiException stack 1 of 2
java.lang.NullPointerException
        at fish.payara.monitoring.store.InMemoryMonitoringDataRepository.init(InMemoryMonitoringDataRepository.java:116)
        
```
## Steps to reproduce
Starting with un-modified "domain1" config ...
Temporarily disable Screen Sharing
`asadmin start-domain`
`asadmin set-hazelcast-configuration --enabled=false`
`asadmin stop-domain`
Re-enable Screen Sharing
`asadmin start-domain`

## Environment ##

- **Payara Version**: 5.193
- **Edition**: Server (Full)
- **JDK Version**:
`openjdk version "1.8.0_222"`
`OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_222-b10)`
`OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.222-b10, mixed mode)`
- **Operating System**: macOS 10.14.6

mwnorman:
work-around:
`<hazelcast-runtime-configuration start-port="59000"></hazelcast-runtime-configuration>`

since it is nigh-impossible to change the Screen Sharing port on the Mac

jbee:
Thanks @mwnorman for fast report and workaround. We unfortunately missed that new monitoring console at the moment requires hazelcast turned on. 

Another workaround I want to verify is to remove `monitoring-console-core.jar` from `glassfish/modules` folder.

jbee:
Tracked by internal issue PAYARA-4104.

jbee:
Fixed by #4188 . As a workaround remove the `monitoring-console-core.jar` from `glassfish/modules`.

PR #4188, opened 2019-08-30
PAYARA-4104 Fixes NPE in monitoring console startup in case hazelcast is disabled

jbee:
PR fixes error reported in #4186 .

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
@@ -112,16 +112,22 @@ public class InMemoryMonitoringDataRepository implements MonitoringDataRepositor
     public void init() {
         serviceLocator = Globals.getDefaultBaseServiceLocator();
         serverEnv = serviceLocator.getService(ServerEnvironment.class);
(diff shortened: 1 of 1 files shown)


jbee:
jenkins test please

jbee:
@AlanRoth Does not look related to me but we can wait with the merge and look for more insight.

AlanRoth:
@jbee If you think it's unrelated, you're free to merge. :)

svendiedrichsen:
@jbee Would this fix lead to a release of Payara 5.193.1?

dmatej:
@AlanRoth  Usually this is caused by one of these:

- JDK11 as default java program (no AS_HOME or JAVA_HOME set)
- some instance is already running

jbee:
@svendiedrichsen I have a chat internal on a 5.193.1 release.

jbee:
@svendiedrichsen We feel that turning of hazelcast on DAS is so uncommon that a release isn't justified. There is an easy enough workaround in removing the `monitoring-console-core.jar` from `glassfish/modules` that can be applied by those that use the setting for some reason.

svendiedrichsen:
> @svendiedrichsen We feel that turning of hazelcast on DAS is so uncommon that a release isn't justified. There is an easy enough workaround in removing the `monitoring-console-core.jar` from `glassfish/modules` that can be applied by those that use the setting.

Thanks for the info, @jbee .

PR #4193, opened 2019-09-03
PAYARA-4113 Log Boot Command Exceptions

MattGill98:
Try starting a domain with the following post boot command file to test:

~~~
start-local-instance --node localhost-domain1 instance1
~~~

Make sure you've created the instance first, although it shouldn't matter in principle


MattGill98:
jenkins test please

PR #4194, opened 2019-09-03
PAYARA-4115 Fix invalid number of columns message when running list-nodes-docker

Pandrex247:
Title.

Previously when you run the command you'd just get an error, due to the number of columns in the formatter not matching the number of actual parameters.


MattGill98:
jenkins test please

PR #4195, opened 2019-09-03
PAYARA-4083: Do not use priority annotations for classes without public API

pdudits:
Marking CDI bean with `@Priority` makes it an alternative. Since these classes doesn't implement any public API, application have no means of overriding these classes. And this also breaks one of CDI TCK tests that expect only application alternatives to be present.

36969e1  is refinement of 792efdd, removing unnecessary interface/impl split and injection, since there is only one possible implementation.


pdudits:
jenkins test please

jbee:
> Do we also need to make changes to the FaultToleranceInterceptor class?

I don't quite follow the reasoning that `@Priority` makes things an alternative. At least for FT I can say it is used because it gives order to the interception which is part of the spec so it cannot be removed without breaking the TCK that actually checks it.

pdudits:
Meaning of `@Priority` is different amongst an interceptor and regular CDI bean.

For bean it is governed by [5.1.1.1. Declaring selected alternatives for an application](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#declaring_selected_alternatives_application), while for interceptors its [9.4. Interceptor enablement and ordering](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#enabled_interceptors).

jbee:
Thanks @pdudits , now this makes sense to me and it should also answer @Pandrex247 question. As FT uses it on an interceptor this should stay as it is.

PR #4198, opened 2019-09-04
PAYARA-3501 Configuration changes in MP health and metrics in Admin console don't warn about restart needed

AlanRoth:
The `MetricsService` was not subscribed to the `UnprocessedConfigListener`

I removed the `restart = true` after every config change as `UnprocessedConfigListener` when not dynamic, will ask for server restart even if there are no changes when pressing save, which is expected behaviour, so `restart` only determined if the `ActionReport` message asked for restart or not, which I believe can be assumed that the message should be printed if `dynamic` is false, as if true, restart is not required.


AlanRoth:
jenkins test please

jGauravGupta:
**Command 1:**
`asadmin set-metrics-configuration --enabled=true --dynamic=false`
```
Restart server for change to take effect
Command set-metrics-configuration executed successfully.
```

**Command 2:**
`asadmin get-metrics-configuration`
```
Enabled  Dynamic  EndPoint  VirtualServers  Security Enabled  Roles
true         false        metrics                             false                     microprofile
Command get-metrics-configuration executed successfully.
```

**Command 3:**
`asadmin set-metrics-configuration --enabled=true --dynamic=true`
```
Command set-metrics-configuration executed successfully.
```

**Command 4:**
`asadmin get-metrics-configuration`
```
Enabled  Dynamic  EndPoint  VirtualServers  Security Enabled  Roles
true         true         metrics                             false                     microprofile
Command get-metrics-configuration executed successfully.
```

**Command 5:**
**If by default `dynamic` is true then on executing command stil shows restart server warning.**
`asadmin set-metrics-configuration --enabled=true`
```
Restart server for change to take effect
Command set-metrics-configuration executed successfully.
```

AlanRoth:
@jGauravGupta I see the problem, I've changed it now so it will only ask for restart if `getDynamic` from the config is false, which means it also "remembers" the last state of `dynamic`. Testing it locally I'm getting the expected output now with your commands, thanks for spotting that :)

AlanRoth:
jenkins test please

AlanRoth:
@Pandrex247 Found the issue, seems to work now

AlanRoth:
jenkins test please

Pandrex247:
I still get the error - only on the first run though.
It seems to come from the "old" value for virtual server being null (it's trying to change null -> "")

AlanRoth:
jenkins test please

PR #4201, opened 2019-09-05
PAYARA-4143 Replace usage of File.deleteOnExit() with manual deletion on exit

### LINKED ISSUE #4199: Resource leak (temporary files left) after payara micro run.
# Description #

After payara micro run there are files left in temporary location (in directory `payaramicro-rt*`).

## Expected Outcome
Temporary files are removed upon payara shutdown.

## Current Outcome
Temporary files left after payara shutdown.

1. Start payara-micro.
1. Stop it.
1. Examine temporary location: `payaramicro-rt*` and its content were not removed.

## Environment ##

- **Payara Version**: 5.193
- **Edition**: Micro
- **JDK Version**: OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1ubuntu1~16.04.1-b10)
- **Operating System**: Linux
- **Database**: none

---
Calling `deleteOnExit` like in:
https://github.com/payara/Payara/blob/5b1ff95c8e01be695be8cddb482b43270dc37957/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExplodedURLClassloader.java#L80-L85
or https://github.com/payara/Payara/blob/5b1ff95c8e01be695be8cddb482b43270dc37957/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/RuntimeDirectory.java#L150-L160
will not result in directory deletion if given directory is not empty, as [DeleteOnExitHook](https://github.com/openjdk/jdk/blob/0a92dc786d1e5f3162067154beec8f079437f4f7/src/java.base/share/classes/java/io/DeleteOnExitHook.java#L81) calls [File.delete()](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/io/File.html#delete()):
>  If this pathname denotes a directory, then the directory must be empty in order to be deleted.

svendiedrichsen:
Fixes #4199 

I have replaced the usage of File.deleteOnExit with common-util FileUtils.deleteOnExit() to allow non-empty directories being deleted.


nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -551,8 +551,52 @@ else if (!deleteFile(f) && undeletedFiles != null) {
...
+     * Java Runtime. Non-empty directories will be deleted recursively.
+     * @param f the file to delete on JVM shutdown
+     */
+    public static void deleteOnExitRecursively(File f) {
+        if (f != null && f.exists()) {
+            final SecurityManager securityManager = System.getSecurityManager();
+            if (securityManager != null) {
+                securityManager.checkDelete(f.getAbsolutePath());
+            }
+            FILES_TO_DELETE_ON_SHUTDOWN.add(f);
```
sgflt: I wonder how much memory this will allocate and make unavailable for application. In some angle of view it could be considered as a memory leak.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -551,8 +551,52 @@ else if (!deleteFile(f) && undeletedFiles != null) {
...
+            deleteOnExitRecursively(f);
+        }
+    }
+
+    /**
+     * Deletes the provided file by registering a shutdown hook with the
+     * Java Runtime. Non-empty directories will be deleted recursively.
+     * @param f the file to delete on JVM shutdown
+     */
+    public static void deleteOnExitRecursively(File f) {
```
sgflt: Is it ok to delete recursively files that are not directories?

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -551,8 +551,52 @@ else if (!deleteFile(f) && undeletedFiles != null) {
...
+                    }
+                    FILES_TO_DELETE_ON_SHUTDOWN.clear();
+                }
+                private void delete(File file) {
+                    if (file.exists()) {
+                        if (file.isDirectory()) {
+                            final File[] files = file.listFiles();
+                            if (files != null) {
+                                for (File f : files) {
+                                    delete(f);
```
sgflt: I cant see the actual deletion, this does just recursion?

dmatej:
Both ways are unreliable - SIGKILL prevents execution.
I would prefer variant 6 with NIO2: https://www.baeldung.com/java-delete-directory 
This is interesting too: https://stackoverflow.com/questions/1424736/making-sure-file-gets-deleted-on-jvm-exit

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExplodedURLClassloader.java
```diff
@@ -73,14 +75,13 @@ public ExplodedURLClassloader(File explodeTo) throws IOException {
 
     public ExplodedURLClassloader() throws IOException {
         super(new URL[0]);
-        File directory;
-        directory = File.createTempFile("payaramicro-rt", "tmp");
+        File directory = File.createTempFile("payaramicro-rt", "tmp");
```
dmatej: This is why we need recursive deletion of files and directories ... ok

dmatej: But ... the directory variable will be a regular file, not a directory. See Files.createTempDirectory

pzygielo: > But ... the directory variable will be a regular file, not a directory. See Files.createTempDirectory

Note, that only generated name is used, as in next line https://github.com/payara/Payara/blob/5b1ff95c8e01be695be8cddb482b43270dc37957/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExplodedURLClassloader.java#L79 there is attemp to _convert_ it to directory.

dmatej: You're right ... I forgot. Just another argument to use NIO instead of old io... it simplifies many things.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
svendiedrichsen: This mimics how to JDK does it. Except for the fact that the JDK holds a static list of Strings.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
svendiedrichsen: Sure. The same way it is ok to iterate a list with one item. But I see the wording problem.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
svendiedrichsen: A few lines below there is the file.delete() call.

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/ExplodedURLClassloader.java
svendiedrichsen: Sure, I can switch it to NIO. Recursion is a bit tricky to read.

svendiedrichsen:
@sgflt @dmatej I have made the requested changes.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -561,43 +548,54 @@ public static void deleteFileNowOrLater(File f) {
...
-                        }
-                        file.delete();
-                    }
+        Runtime.getRuntime().addShutdownHook(FILE_DELETION_ON_EXIT);
+    }
+
+    private static final class FileDeletionThread extends Thread {
+
+        private static final Logger LOG = Logger.getLogger(FileDeletionThread.class.getName());
+        private final Set<String> filesToDelete = new LinkedHashSet<>();
```
sgflt: Is it sure only one thread at a time will access field filesToDelete? If not, there is a race condition that may cause ConcurrentModificationException during clean up (line 580).

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -561,43 +548,54 @@ public static void deleteFileNowOrLater(File f) {
...
+            for (String path : filesToDelete) {
+                delete(new File(path));
+            }
+            filesToDelete.clear();
+        }
+
+        private void delete(File file) {
+            if (file.exists()) {
+                try {
+                    Files.walk(file.toPath())
```
sgflt: > The stream is <i>weakly consistent</i>. It does not freeze the
>      * file tree while iterating, so it may (or may not) reflect updates to
>      * the file tree that occur after returned from this method.

Temporary directory may still exist after clean up if a new file appeared during clean up phase.

svendiedrichsen: How would you like to address this?

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
svendiedrichsen: Good point. I will address this.

svendiedrichsen: I have addressed it.

svendiedrichsen:
@sgflt I have fixed the concurrency problem. How do you think one could fix the _weakly consistency_ problem. IMO there is no way to handle it within the payara process.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
```diff
@@ -551,8 +538,68 @@ else if (!deleteFile(f) && undeletedFiles != null) {
...
+            }
+        }
+
+        @Override
+        public void run() {
+            while (!filesToDelete.isEmpty()) {
+                final Set<String> filesToIterate = filesToDelete;
+                filesToDelete = new LinkedHashSet<>();
+                for (String path : filesToIterate) {
+                    delete(new File(path));
```
pzygielo: Could `File`s (instead of `String`s) be stored at 570 and this construction skipped here?

svendiedrichsen: Sure. That was my first approach. I have changed it to String from File when @sgflt was concerend about the amount of memory this would consume and that this was a memory leak. Are you sure you want to go back to storing File instances?

pzygielo: > Sure. That was my first approach...

I missed [that ...when @sgflt was concerned..](https://github.com/payara/Payara/pull/4201#discussion_r321582378) already about this case.

IMO keeping `File`s would be more clear.
I think you acted way too fast with https://github.com/payara/Payara/pull/4201/commits/9e8cccf2715f8b93460f43f57c60fbfe1c048cd7, as I'm 3rd party here :grin:

svendiedrichsen: We will see. I'm also for using Files.

sgflt: Files are OK for me. I was initially thinking about static array that is now gone. :) (moved to class as a field wich looks better to me)


```
final Set<String> filesToIterate = filesToDelete;
filesToDelete = new LinkedHashSet<>();
```
I was able to reproduce ConcurrentModificationException.
Without sync there is possible following sequence of steps:

T1 loads S1
T2 loads S1
T2 creates S2
T2 iterates over S1
T1 adds element to iterated S1

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
sgflt: I agree that every attempt may fail with possible race conditions and deleteOnExit does job as "best effort". (Windows has open descriptor to deleted file, new files appearing during hookup processing, etc.)

I don't know ho much robust have this utility be. Maybe it is ok just admit the fact in some cases there will be left some temporary directory. Or maybe not and we should spend more time with thinking about solution.

svendiedrichsen:
@sgflt Using a ConcurrentSkipListSet to hold the file references now to prevent ConcurrentModificationExceptions.

nucleus/common/common-util/src/main/java/com/sun/enterprise/util/io/FileUtils.java
svendiedrichsen: We could still improve by having the files where deletion failed retried several times.

svendiedrichsen:
@sgflt I have added code to retry failed file deletion. Extracted FileDeletionThread into separate class file.

pdudits:
Jenkins test please

appserver/extras/payara-micro/payara-micro-boot/pom.xml
```diff
@@ -53,13 +53,18 @@
     <name>Payara Micro Boot</name>
     <description>Payara Micro Edition Boot</description>
     <dependencies>
+        <dependency>
+            <groupId>org.glassfish.main.common</groupId>
```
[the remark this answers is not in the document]
svendiedrichsen: I don't think there is a difference. The size comparing the payara-micro to a different previous build is 75.51 MB (before on a different build) to 75.52 MB (on this branch).

svendiedrichsen:
@pdudits Are there any tests included in the automatic build for the Payara Micro server?

pdudits:
No, but I'll inspect the resulting .jar now and will run some.

pdudits:
This doesn't clean up the directories for me on Windows (but that could be Windows thing). However I wonder why I'm not getting any classloading exceptions, as my understanding is, that `ExplodedURLClassLoader` is the one that loads `FileUtils`, and therefore should blow up at startup.

I'll look into this more once my pile of tasks thickens.

svendiedrichsen:
If you look at the maven payara-micro dependency graph you will see that common-util is pulled in as transitive dependency via `glassfish -> config-api -> common-util`.

pdudits:
It's not about dependency graph, it's about runtime classpath. `payara-micro-boot` prepares classloader of all modules, inclusive `common-util`, by expanding contents of `MICRO-INF` from the jar into that temp directory -- that's what `ExplodedURLClassLoader` does. 

Therefore that classloader has no way of accessing `FileUtils`. At least that's what theory says, I'm looking forward to figuring out what practice looks like ;)

svendiedrichsen:
Another question that came to my mind: Should we delete files that couldn't be deleted on exit delete on startup? Meaning when FileUtil class is loaded. I'd only need to store the list of undeleteable files in a text file in the temp folder on exit and pick it up when the class is loaded. This should work around the Windows problem with deleting files on exit.

pdudits:
Oh, I was testing with wrong binary! 🤣 The right one indeed fails along my expectations with:

```
java.lang.NoClassDefFoundError: com/sun/enterprise/util/io/FileUtils
        at fish.payara.micro.boot.loader.ExplodedURLClassloader.<init>(ExplodedURLClassloader.java:84)
```
There are no nice solutions to this one, but after it initializes, it could use reflection to call `FileUtils`. I'll try that.

svendiedrichsen:
> Oh, I was testing with wrong binary! 🤣 The right one indeed fails along my expectations with:
> 
> ```
> java.lang.NoClassDefFoundError: com/sun/enterprise/util/io/FileUtils
>         at fish.payara.micro.boot.loader.ExplodedURLClassloader.<init>(ExplodedURLClassloader.java:84)
> ```
> 
> There are no nice solutions to this one, but after it initializes, it could use reflection to call `FileUtils`. I'll try that.

Couldn't you just delay calls to FileUtils after the Jars have been extracted and added?

svendiedrichsen:
Now I got the problem. Took some time.

pdudits:
jenkins test please

pdudits:
I pushed the fix for classloader.

As for deleting files on startup, I wouldn't take the feature that far. It might fix the the issue with windows locking `.jar` files, but it would cause troubles during concurrent start of multiple instances, e. g. when system with multiple micro instances boots. And loose files to delete when multiple instances shutdown.

PR #4202, opened 2019-09-06
PAYARA-4125 MP HealthCheck NPE from when first saving values in admin console

AlanRoth:
Fixed for MP metrics in PAYARA-3105


AlanRoth:
jenkins test please

PR #4203, opened 2019-09-06
PAYARA-4097 Presets for Monitoring Console

jbee:
**Summary**

Changes in context of to the task goal of having a preset starting page for the Monitoring Console:

* direct collection of web and jvm statistics from providers (still needs configuration to level HIGH but no JMX/AMX)
* added metric _CPU Usage_ (%) to provider
* added metric _Heap Usage_ (%) to provider
* added _per second_ option to display values as delta per second instead of absolute values
* added preset with requested starting page used in case no UI configuration is found in local storage
* added page reset to preset (if available)
* added manual import/export of UI configuration
* removed unnecessary house keeping from `CpuUsageHealthCheck`
* fix for inconsistent behaviour of web statistics provider after removal of application

This PR also contains numerous other changes and improvements (done between sprints):

* more readable series name format, shorter names (now 2-3 parts: `ns:<ns> [@:<group>] <metric>`)
* removed many uninteresting series from collection
* allow filtering by instances (basic)
* fetch all data in one request for any number of charts on the page
* fixed NPE(s) on server to assemble result sets
* less noisy REST API URLs

**Review Notes**
On Javascript code I don't think there is a point in reviewing the exact details. At this point we should focus on the general code style. Please add your suggestions on how to improve.
I split the sources into multiple files which are merged as part of the build so keep it at more manageable sizes. So `monitoring-console.js` is generated and does not need a review. 

**Testing**
1. build server.
2. deploy webapp in `appserver/monitoring-console/webapp`
3. Open `localhost:8080/monitoring-console-webapp/`

Browser should now show the 5 metrics requested by the task. (see Jira)
_Request Count_ and _Active Sessions_ will be zero since monitoring level has to switched to HIGH first. To do that:

1. goto _Configurations_ => _server-config_ => _Monitoring_
2. select all checkboxes in the list
3. select level HIGH from dropdown and press _Change Level_
4. click _Save_

_Request Count_ should now start to show data. _Active Sessions_ might still be zero because no session exist.

To test the reset move a chart left or right or remove it (x) and then press the rest button (most left on top icon toolbar).
The full UI configuration can also be exported and imported using the buttons right to the reset button in the top icon toolbar.


appserver/web/admin/src/main/java/org/glassfish/web/admin/monitor/WebStatsProviderBootstrap.java
```diff
@@ -65,50 +67,37 @@
...
     }
 
+    @Override
     public void postConstruct(){
         //Register the Web stats providers
         registerWebStatsProviders();
     }
 
     private synchronized void registerWebStatsProviders() {
-        if (isWebStatsProvidersRegistered.get()) {
```
jbee: FYI: I found an inconsistency in behaviour when investigating why undeloy + deploy would cause configuration level not have the expected effect. The `registerWebStatsProviders` would be called in `postConstruct` and when new application was registered since unregistering an application would remove the providers in case this was the last application. This seems inconsistent since on server startup those provider do exist even with no application. Since there might be something depending on this I decided that it was the best fix to never remove them to be consistent over time.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/CpuUsageHealthCheck.java
```diff
@@ -110,50 +92,29 @@ protected HealthCheckResult doCheckInternal() {
...
-                times.setId(id);
-                times.setStartCpuTime(c);
-                times.setEndCpuTime(c);
-                times.setStartUserTime(u);
-                times.setEndUserTime(u);
-                threadTimes.put(id, times);
-            }
-            else {
-                times.setEndCpuTime(c);
-                times.setEndUserTime(u);
```
jbee: FYI: Having to duplicate the computation of CPU usage for the statistic provider where such a value was added I needed to understand what is computed. Doing this I discovered that the temporary map for the statistics per thread does not serve any purpose other than totalling the time of all threads which simply could be done by looping and adding their times directly.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/monitoring-console/webapp/pom.xml
```diff
@@ -94,4 +94,35 @@
             <scope>provided</scope>
         </dependency>
     </dependencies>
+    
+    <build>
+        <plugins>
+            <plugin>
+              <groupId>com.bekioui.maven.plugin</groupId>
+              <artifactId>merge-maven-plugin</artifactId>
```
Pandrex247: We try to avoid bringing in third-party dependencies and plugins when they aren't strictly necessary.
Is there specifically a reason that you need this plugin to "merge" these files?

License seems to be fine as its Apache, just want to hear your reasoning :)

Pandrex247:
Jenkins test please

appserver/monitoring-console/webapp/pom.xml
jbee: There is no particular reason other than finding a platform independent way of merging the JS files into one. Having everything in one file got out of hand length wise so I split it to multiple files that are merged. But I also didn't want to pull in the full JS build stack madness so I found the merge plugin that does everything I need for now. Maybe there is a better way that I just have not thought about?

jbee: @Pandrex247 Are you happy with the use of the plugin or should we look into how to else solve the issue? I'd imagine a web-server might have some feature to serve a virtual file composed of a list of resource files? :)

PR #4210, opened 2019-09-10
PAYARA-4077: Implement Weld SecurityContext

pdudits:
This enable security context propagation to asynchronous event observers.

This is build on top of PAYARA-4078 to make CDI TCK pass.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
@@ -110,13 +110,7 @@ import org.glassfish.web.deployment.descriptor.AppListenerDescriptorImpl;
 import org.glassfish.web.deployment.descriptor.ServletFilterDescriptor;
 import org.glassfish.web.deployment.descriptor.ServletFilterMappingDescriptor;
 import org.glassfish.weld.connector.WeldUtils;
-import org.glassfish.weld.services.EjbServicesImpl;
-import org.glassfish.weld.services.ExternalConfigurationImpl;
-import org.glassfish.weld.services.InjectionServicesImpl;
-import org.glassfish.weld.services.NonModuleInjectionServices;
-import org.glassfish.weld.services.ProxyServicesImpl;
-import org.glassfish.weld.services.SecurityServicesImpl;
-import org.glassfish.weld.services.TransactionServicesImpl;
+import org.glassfish.weld.services.*;
 import org.jboss.weld.bootstrap.WeldBootstrap;
 import org.jboss.weld.bootstrap.spi.BeanDeploymentArchive;
 import org.jboss.weld.bootstrap.spi.EEModuleDescriptor;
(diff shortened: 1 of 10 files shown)


pdudits:
Jenkins test please

pdudits:
jenkins test please

pdudits:
Jenkins test please

pdudits:
Jenkins test please

pdudits:
jenkins test please

pdudits:
jenkins test please

pdudits:
Jenkins test please

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
```diff
@@ -110,13 +110,7 @@
...
 import org.glassfish.web.deployment.descriptor.ServletFilterMappingDescriptor;
 import org.glassfish.weld.connector.WeldUtils;
-import org.glassfish.weld.services.EjbServicesImpl;
-import org.glassfish.weld.services.ExternalConfigurationImpl;
-import org.glassfish.weld.services.InjectionServicesImpl;
-import org.glassfish.weld.services.NonModuleInjectionServices;
-import org.glassfish.weld.services.ProxyServicesImpl;
-import org.glassfish.weld.services.SecurityServicesImpl;
-import org.glassfish.weld.services.TransactionServicesImpl;
+import org.glassfish.weld.services.*;
```
dmatej: Whe should agree on some rule how many imports should be replaced with asterisk ... I have 20 as default ... and I would not like it under 10. Only minor comment here, does not block PR.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ResourceInjectionServicesImpl.java
```diff
@@ -0,0 +1,173 @@
...
+    private static final Logger LOG = Logger.getLogger(ResourceInjectionServicesImpl.class.getName());
+
+
+    private final InitialContext namingContext;
+
+    public ResourceInjectionServicesImpl() {
+        try {
+            this.namingContext = new InitialContext();
+        } catch (NamingException e) {
+            throw new RuntimeException("Initial context not available at deployment time", e);
```
dmatej: Use IllegalStateException, don't use RTE directly ...

appserver/web/weld-integration/src/main/java/org/glassfish/weld/WeldDeployer.java
Pandrex247: Yeah IntelliJ defaults to something like 5 which I think is way too low.
We don't currently have one set, but I agree we should agree one.

Personally I hate wildcard imports in all situations.
You pretty much **never** need to import everything, and IDEs can auto-hide imports anyway so the "readability" argument is moot.

dmatej: Yep, SonarQube also has "too many imports" rule ... so usually asterisk should be unreachable :)

pdudits: It was set to 9 :). I'll turn it up a bit.

PR #4222, opened 2019-09-17
PAYARA-4140: Ignore unresolvable JNDI references

pdudits:
that are likely resolved by other definition


pdudits:
jenkins test please

PR #4224, opened 2019-09-17
PAYARA-4131: Fix race conditions in resource class generation

pdudits:
The generation is recursive and therefore we need to mark class as generated as soon as generation starts


pdudits:
Jenkins test please

PR #4225, opened 2019-09-18
PAYARA-4097 Fixed Grid Layout, Dynamic Settings 

jbee:
### Summary

Pure JS/HTML/CSS update for MC webapp that does:

* fixed grid layout algorithm
* made settings dynamic or responsive so they reflect the actual chart selected and its current values - for this the settings are now purely generated by JS
* organised and cleaned the JS code (main separation into model view and chart specific code all under the global object `MonitoringConsole`)

### Reviewer Notes
Remember that `monitoring-console.js` is merged from the other JS files and does not need to be reviewed.

### DIFF
--- a/appserver/monitoring-console/webapp/pom.xml
+++ b/appserver/monitoring-console/webapp/pom.xml
@@ -113,9 +113,10 @@
                         <override>true</override>
                         <target>src/main/webapp/monitoring-console.js</target>
                         <sources>
-                            <source>src/main/webapp/mc-ui.js</source>
-                            <source>src/main/webapp/mc-render.js</source>
-                            <source>src/main/webapp/mc-page.js</source>
+                            <source>src/main/webapp/mc-main.js</source>
+                            <source>src/main/webapp/mc-model.js</source>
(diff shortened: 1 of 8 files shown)


jbee:
jenkins test please

jbee:
jenkins test please

jbee:
@AlanRoth Removing charts is implemented but I discovered that I broke it recently (and fixed it later). Other instances can be monitored. Just start one and if it has a series with non zero value it should occur in the chart and also in the list of instances. I might also have broken something there recently since the list of instances actually has an effect of the shown data. Have to check.

jbee:
@AlanRoth Other instances still works. Just create them and the core tab should start showing multiple lines.

jbee:
@AlanRoth Since I already fixed the issue with chart removal in another branch I hope it is ok for you if we keep it in this one as it is not related to the feature of this PR and was caused by general code reorganisation I have to do every now and then in the JS code to keep it readable. Slowly I learn how to do that without it growing to big or confusing.

PR #4228, opened 2019-09-19
PAYARA-3501 Configuration changes in MP health and metrics in Admin console don't warn about restart needed

AlanRoth:
Reopening due to small change that removed Dynamic config, thanks to @jGauravGupta for spotting it!


AlanRoth:
jenkins test please

PR #1081, opened 2016-09-03
PAYARA-917 Fixes #927 store forwarded path in the request and return …

smillidge:
…if not null

PR #4229, opened 2019-09-19
PAYARA-4056 HttpServletRequest.getServletPath() not returning original path

Pandrex247:
Revert fix made for [PAYARA-917](https://github.com/payara/Payara/pull/1081) as it currently breaks the Servlet TCK.
Requires the following PR into the EE7 samples repo to be merged to pass Jenkins tests - as the fix was originally put it to fix a couple of tests in `jaspic/dispatching-jsf-cdi` around something that doesn't seem to be mentioned in any specs: https://github.com/payara/patched-src-javaee7-samples/pull/44

What should be the outcome of calling `getServletPath()` on a HttpServletRequest injected into a Bean, which is injected into a Servlet which has been forwarded to? The original servlet (before the forward), or the forwarded-to servlet (as is what you'd get if calling the method on the HttpServletRequest parameter of the forwarded to Servlet's `doGet` method)

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/RequestFacade.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/RequestFacade.java
@@ -55,7 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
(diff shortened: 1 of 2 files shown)


Pandrex247:
Also contains a tiiiiiny refactor of removing a redundant else

cubastanley:
Jenkins test please

pdudits:
jenkins test please

PR #4234, opened 2019-09-20
PAYARA-4055: Minimize the need for decoration of a ClientBuilder

pdudits:
Decoration is only needed for Microprofile tracing scenarios:

* When using opentracing directly, the client must call and use the builder from `ClientRegistrar.configure`
* MP Rest client provides its own async invocation interception facility, that can be used to propagate context

Both opentracing and MP Rest client supporting classes were moved into `microprofile-opentracing`. Additional context propagation facility was added, `SpanPropagator` uses threadlocal to store parent span context, and is called from async invocation interceptor for MP Rest Client calls.

Passes OpenTracing TCK after fixing the test (pdudits/microprofile-opentracing@5c44919e342c4a applied)
Passes `jaxrs/client/api` TCK.


pdudits:
jenkins test please

PR #4239, opened 2019-09-25
PAYARA-3263 Support Alternate Opentracing Implementation

### LINKED ISSUE #3405: Add possibility to override provided MicroProfile io.opentracing.Tracer instance
# Description #
----------

While trying the MicroProfile OpenTracing functionality provided by Payara, I realized that currently there is no way of plugging in a different OpenTracing tracer implementation, like Jaeger for example.

The current implementation in combination with the log or slack notifier is nice, but I really want to send the trace info to one of the currently supported tracers (see https://opentracing.io/docs/supported-tracers/) like Jaeger.

## Expected Outcome

Trace and span info can be send to any supported tracer. Alternatively, have the possibility to override the Payara internal `Tracer` implementation using an application supplied CDI `@Alternative` or `@Spezializes` tracer instance.

## Current Outcome

The internal `Tracer` instance can not be overridden via CDI , because most OpenTracing code does not use CDI but the internal lookup mechanism. The trace info can only be send to log or slack.

None. This is an enhancement request.

### Samples

A small tracing demo can be found here: https://github.com/lreimer/mp-tracing-payara-demo

## Context (Optional)

I was trying to override the internal `io.opentracing.Tracer` instance using a CDI alternative bean.
Creating a Jaeger tracer using CDI works, but the internal OpenTracing implementation basically ignores the `@Alternative` instance and always uses the internal instance.

## Environment ##

- **Payara Version**: 5.183
- **Edition**: Micro
- **JDK Version**: Zulu 8u181
- **Operating System**: CentOS base image

Cousjava:
fixes #3405 

This requires RequestTracing in Payara to be enabled, but it does not send any traces from the request tracing service to the tracer, only those traced as part of OpenTracing are.


Cousjava:
Jenkins test please

appserver/payara-appserver-modules/microprofile/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
```diff
@@ -94,16 +100,27 @@ public void event(Event<?> event) {
     public synchronized Tracer getTracer(String applicationName) {
```
jbee: I'd like if we take the change and remove the `synchronized` and use `tracers.computeIfAbsent` instead extracting the initialisation function to a static function in the class. This would both make more clear what the sections of the method do and get rid of some thread contention.

appserver/payara-appserver-modules/microprofile/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
```diff
@@ -94,16 +100,27 @@ public void event(Event<?> event) {
...
+
         // If there isn't a tracer for the application, create one
         if (tracer == null) {
             // Check which type of Tracer to create
+
+            try {//See if an alternate implementation of Tracer is available in a library.
+                ServiceLoader<Tracer> tracerLoader = ServiceLoader.load(Tracer.class);
+                Iterator<Tracer> loadedTracer = tracerLoader.iterator();
+                if (loadedTracer.hasNext()) {
+                    tracer = tracerLoader.iterator().next();
```
rdebusscher: We already have the iterator in a variable.

Cousjava:
Jenkins test please

PR #4240, opened 2019-09-26
PAYARA-3832 Fixed request tracing maxSize was ineffective

jbee:
### Summary
Both request tracing store and strategy implementations had defects causing the store to grow beyond the `maxSize`. Both have two alternatives:

* `LocalRequestTraceStore` and `ClusteredRequestTraceStore` for the store
* `LongestTraceStorageStrategy` and `ReservoirTraceStorageStrategy` for the strategy
  
**Cluster Store**:
* would not be a shared store in the cluster but per instance because `instanceId` was used as key for a multi-map where the values set would be the traces in the store
* would not `remove(Object)` items since it uses the value as argument where the key was expected

**Local Store**:
* would not remove items since `size()` would return wrong values due to incorrect implementation of `RequestTrace.compareTo` (more later)
* `getStoreSize()` would return actual size not max size (while cluster does other way what I believe was meant)

**Reservoir Sampling Strategy**
* would not return traces for removal over max size due wrongly implemented algorithm
* would potentially use lots of memory to pick a random element should it decide to

**Keep Longest Strategy**:
* worked correctly but had potential to use lots of memory to do so

I also made the store `setSize` method not throw away all traces but only those that do not exceed the new max size limit. The strategy is used to pick the traces to remove.

Due to the changes the `TraceStorageStrategy` could be simplified to a single method.

### Testing
As fixing the issues required quite some changes I added tests for both stores and both strategies showing their behaviour as I would expect them to be intended to work.

Noticed that one of the existing tests was basically a empty shell where everything was commented out. It seemed to be based on an no longer existing API so I deleted it.

### Other Reviewer Notes
Part of the issue was an implementation of `RequestTrace.compareTo` that does not satisfy the semantics required by the method as documented in its javadoc. Since it is only comparing `elapsedTime` and `startTime` two traces of same length and same starting time are falsely considered the same. Since a change could have far reaching consequences in other code relying on the way it is implemented now I opted to not address this in the PR. Suggestion to fix it: Traces are equal by their ID if present, all traces without an ID yet (no span yet) are equal ("empty instance"). `compareTo` then should compare the ID in case `elapsedTime` and `startTime` are identical.


nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/RequestTracingService.java
```diff
@@ -274,13 +286,6 @@ public void bootstrapRequestTracingService() {
                         0, period, TimeUnit.SECONDS);
             }
             
-            if (executionOptions.getAdaptiveSamplingEnabled()) {
```
jbee: This just moved up. I had a initialisation order issue causing a NPE that was caused by an exception earlier on and this fixed it but since that earlier exception should not be thrown in the first place it would not be strictly needed. Anyhow, I forgot to change it back and being earlier makes the code a bit more robust in case of failure so why not.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/ClusteredRequestTraceStore.java
```diff
@@ -54,46 +54,30 @@
...
-    private final String instanceId;
-    private final MultiMap<String, RequestTrace> store;
+    private final Map<UUID, RequestTrace> store;
     private int maxStoreSize;
-    
+
     private final TraceStorageStrategy strategy;
 
-    ClusteredRequestTraceStore(MultiMap<String, RequestTrace> store, String instanceId, TraceStorageStrategy strategy) {
+    ClusteredRequestTraceStore(Map<UUID, RequestTrace> store, TraceStorageStrategy strategy) {
```
jbee: I have no clue why the `instanceId` came in but it does not make sense as it would effectively give each instance its own store just that it is stored distributed instead of  in local memory. I believe the goal here was to have all instances share a common distributed store. This is best done by a simpler distributed data structure. I opted for a `Map` where the keys are the trace IDs and the values are the traces.  The instances will all use the same map as it has a constant name within the cluster. A small side benefit is that monitoring will have statistics available on the map.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/ClusteredRequestTraceStore.java
```diff
@@ -109,6 +93,9 @@ public RequestTrace addTrace(RequestTrace trace, RequestTrace traceToRemove) {
 
     @Override
     public void setSize(int maxSize) {
+        while (store.size() > maxSize) {
+            store.remove(strategy.getTraceForRemoval(getTraces(), maxSize, null).getTraceId());
+        }
```
jbee: The semantics of changing size seemed unclear and contradictory to the goal of never having more items then the `maxSize` so I decided to remove items using the strategy as long as the store has too many items for the new size.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
```diff
@@ -51,29 +54,21 @@
  */
 public class LocalRequestTraceStore implements RequestTraceStoreInterface {
 
-    private BoundedTreeSet<RequestTrace> store;
+    private final Set<RequestTrace> store = ConcurrentHashMap.newKeySet();
```
jbee: Since the strategy will already prevent the store from growing to large there is no need to also have a bound data structure. More importantly the concurrent hash set will have the expected behaviour. `BoundedTreeSet` will not work. Try this test:

```
@Test
public void test() {
  Set<RequestTrace> set = new ConcurrentSkipListSet<>(); // base class of BoundedTreeSet already show the problem
  for (int i = 0; i < 10; i++) {
    set.add(createRandomRequestTraceWithSingleSpanAndSameTime());
  }
  assertEquals(10, set.size());
}
```
The set should have size 10 but the `ConcurrentSkipListSet` actually can return any number between 1 and 10 at this point.

> Beware that, unlike in most collections, the size method is not a constant-time operation. Because of the asynchronous nature of these sets, determining the current number of elements requires a traversal of the elements, and so may report inaccurate results if this collection is modified during traversal.

This is just a hint the javadoc gives us on how the class works. Together with the improper implementation of `RequestTrace.compareTo` the size does come out wrong most of the times. 
I here opted to simply switch the collection as hash set is more adequate anyway. Still we do need to fix the `RequestTrace.compareTo` method but since that might have far reaching consequences I didn't want to bite this bullet in this PR.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/ClusteredRequestTraceStore.java
```diff
@@ -54,46 +54,30 @@
...
     }
 
     @Override
     public RequestTrace addTrace(RequestTrace trace) {
-        store.put(instanceId, trace);
-        RequestTrace traceToRemove = strategy.getTraceForRemoval(getTraces(), maxStoreSize);
-        if (traceToRemove == null) {
-            return null;
-        }
-        store.remove(traceToRemove);
```
jbee: This was another bug since `remove` expects the _key_ not the _value_.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
```diff
@@ -98,20 +93,20 @@ public RequestTrace addTrace(RequestTrace trace, RequestTrace traceToRemove) {
 
     @Override
     public void setSize(int maxSize) {
-        // Set the store to be one bigger than the specified value. Allows the store to transparently select which item to remove
-        this.store = new BoundedTreeSet<>(maxSize + 1);
+        while (store.size() > maxSize) {
+            store.remove(strategy.getTraceForRemoval(getTraces(), maxSize, null));
+        }
```
jbee: Again changing size updates the store if needed (same as above in other strategy)

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
```diff
@@ -98,20 +93,20 @@ public RequestTrace addTrace(RequestTrace trace, RequestTrace traceToRemove) {
...
         this.maxStoreSize = maxSize;
     }
 
     @Override
     public int getStoreSize() {
-        return this.store.size();
+        return maxStoreSize;
```
jbee: Its unclear what `getStoreSize()` should return since one strategy would return the `maxSize` while this was returning the actual number of values. I felt that it was meant to return the `maxSize` as the value size could be obtained by `getTraces().size()` while there was no other way to obtain the `maxSize`.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/RequestTraceStoreFactory.java
```diff
@@ -86,8 +89,8 @@ public static RequestTraceStoreInterface getStore(Events events, boolean reservo
         // Get a clustered store if possible
         ClusteredStore clusteredStore = Globals.getDefaultHabitat().getService(ClusteredStore.class);
         if (clusteredStore != null && clusteredStore.isEnabled()) {
-            MultiMap<String, RequestTrace> store = (MultiMap) clusteredStore.getMultiMap(storeName);
-            return new ClusteredRequestTraceStore(store, clusteredStore.getInstanceId(), strategy);
+            Map<UUID, RequestTrace> store = (Map) clusteredStore.getMap(storeName);
+            return new ClusteredRequestTraceStore(store, strategy);
```
jbee: See comment on `ClusteredRequestTraceStore` on the change from `MultiMap` to `Map`

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/strategy/LongestTraceStorageStrategy.java
```diff
@@ -40,62 +40,40 @@
...
     @Override
-    public RequestTrace getTraceForRemoval(Collection<RequestTrace> traces, int maxSize) {
+    public RequestTrace getTraceForRemoval(Collection<RequestTrace> traces, int maxSize, RequestTrace traceToRemove) {
         if (traces == null || traces.isEmpty() || traces.size() <= maxSize) {
             return null;
         }
-        // Sort and return shortest request
-        List<RequestTrace> tracesCopy = new ArrayList<>(traces);
-        Collections.sort(tracesCopy);
-        return traces.toArray(new RequestTrace[]{})[traces.size() - 1];
```
jbee: While this works this was just too painful performance wise to let it be - we do not know how large users might make the store - the target trace can be found by iterating instead of copying, sorting it (which copies again) and once again copy it to an array.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/strategy/ReservoirTraceStorageStrategy.java
```diff
@@ -51,66 +52,42 @@
 public class ReservoirTraceStorageStrategy implements TraceStorageStrategy {
 
     private final Random random;
-    private long counter;
```
jbee: I believe this `counter` is the code translation of the wiki article text talking about an _unknown large number_. I think the algorithm got misunderstood here. If there was an ever increasing divider for the `probability` (below) the probability would ever decrease over time (with the number of invocations) making it less and less likely an item is kept. Also inverting this does not make sense. That would increase the likelihood to 1 over time. Both are wrong and must be based in an misunderstanding of the algorithm.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/strategy/ReservoirTraceStorageStrategy.java
```diff
@@ -51,66 +52,42 @@
...
-        if (counter < Long.MAX_VALUE) {
-            counter++;
-        }
-
-        // Probability of keeping the new item
-        double probability = (double) maxSize / counter;
-        boolean keepItem = random.nextDouble() < probability;
-
-        // If the probability fails or the list isn't full, nothing should be removed
-        if (!keepItem || traces.size() <= maxSize) {
```
jbee: Here it should have been `keepItem` instead of `!keepItem` except that other errors in the algorithm partly cancelled this out or shadowed it. Now it does not play a role as we see below.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/strategy/ReservoirTraceStorageStrategy.java
```diff
@@ -51,66 +52,42 @@
...
+                : findRandomTrace(traces);
+    }
+
+    /**
+     * Since this implementation is only dealing with a special case of the Reservoir sampling scenario where there is a
+     * single element more than the maximum size a selection of equal probability is simply to select any of the
+     * elements at random whereby we are left with the elements to keep.
+     */
+    private RequestTrace findRandomTrace(Collection<RequestTrace> traces) {
+        int itemToReplace = random.nextInt(traces.size());
```
jbee: The "reservoir sampling" for our case where we always have `n < maxSize` or `n == maxSize+1` indeed is just a random number in range 0  to (n-1) as all we need is to decide which element to throw out to get to `maxSize` again. Since all element should have same probability the random number does just that. Good thing: you do not have to take my word for it, there is a test that proves it.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/strategy/TraceStorageStrategy.java
```diff
@@ -49,14 +49,13 @@
...
-     * removed.
+     * Get the trace that needs removing suggesting a candidate that should be chosen in case it is contained in the
+     * given list of traces.
+     * 
+     * @param traces        the list of traces to test.
+     * @param maxSize       the maximum size of the list.
+     * @param traceToRemove the trace to remove if present or null if no particular candidate if favoured
+     * @return the trace that needs removing, or null if no traces need removing.
      */
-    RequestTrace getTraceForRemoval(Collection<RequestTrace> traces, int maxSize);
```
jbee: As both implementations would just call the other `getTraceForRemoval` with `null` for the `traceToRemove` I felt it could very well be removed and called with `null` directly.

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
Pandrex247: Is the functionality the same though?
The bounded tree set was used specifically so that the last (read: shortest) trace would be replaced if the store was full (assuming you're using the longest traces strategy)

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/ClusteredRequestTraceStore.java
jbee: This can be a topic for Monday meeting since it also has consequences for the UI. If there is a store per instance and given we make all of them accessible in the UI you get instances times max-size traces which sounds like could be quite a number at times.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/LocalRequestTraceStore.java
jbee: Yes and no. 

No: The property of the tree set to remove the shortest on "overflow" after adding is practically superfluous. If the strategy works correctly the strategy already will determine the element to remove before it ever comes to an overflow as the overflow occurs at max-size + 2 but strategy will remove max-size +1 to max-size after an add.  

Yes: If the strategy or other code involved to remove elements on behalf of the strategy does not work correctly (as it was) the "overflow" logic might have kick in and removed the shortest trace. I say might because of the `size()` bug this wasn't guaranteed either. Also this might be wrong in case the other strategy is used. Since there now are tests that verify correct behaviour I think we don't need to rely on the collection rescuing the failing strategy. I guess this behaviour has contributed to hiding the issues in the rest of the involved code.

jbee:
jenkins test please

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/store/ClusteredRequestTraceStore.java
jbee: I was thinking about this a bit more and to me it appears less and less desirable to have a non-shared  store. While instances theoretically do have access to the data of all instances the reality is that asking the store only makes the local data visible. As a consequence when using the admin GUI you can never look into anything not deployed on the DAS which doesn't feel that useful. Say we would allow to get data from other instances there would still be the problem that any tool using request tracing data had to make this decision or forward it and there is the risk of configuration not being what the user wanted for something were it is quite hard to verify that a certain area is covered. I'd think for something that should warn you about something that you might not be able to trigger to verify correct configuration there should not be such a configuration so you can be sure that the tool will warn you

PR #4241, opened 2019-09-26
PAYARA-4164 Robust null-safe code

dmatej:
- invocationManager is null for default admin context
- there is no guarantee to have all needed services accessible
- application name can be null for context of server modules

# Testing

### Testing Performed

Fixes TCK/CDI

### Test suites executed
- [x ] Quicklook
- [x] Payara Samples
- [ ] Java EE7 Samples
- [ ] Java EE8 Samples
- [ ] Payara Private Tests
- [x] Payara Microprofile TCKs Runner
- [x] Jakarta TCKs
- [ ] Mojarra
- [ ] Cargo Tracker

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/OpenTracingApplicationEventListener.java
+++ b/appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/OpenTracingApplicationEventListener.java
@@ -88,9 +88,14 @@ public class OpenTracingApplicationEventListener implements ApplicationEventList
             LOG.config("Default base service locator is null, JAX-RS server tracing is disabled.");
             return;
         }
+        final InvocationManager invocationManager = serviceLocator.getService(InvocationManager.class);
         this.requestTracing = serviceLocator.getService(RequestTracingService.class);
         this.openTracing = serviceLocator.getService(OpenTracingService.class);
-        this.applicationName = this.openTracing.getApplicationName(serviceLocator.getService(InvocationManager.class));
+        if (invocationManager == null || this.openTracing == null) {
+            this.applicationName = null;
+        } else {
+            this.applicationName = this.openTracing.getApplicationName(invocationManager);
+        }
     }
 
 
--- a/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
+++ b/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
@@ -51,6 +51,7 @@ import javax.annotation.PostConstruct;
 import javax.interceptor.InvocationContext;
 import org.glassfish.api.event.EventListener;
 import org.glassfish.api.event.Events;
+import org.glassfish.api.invocation.ComponentInvocation;
 import org.glassfish.api.invocation.InvocationManager;
 import org.glassfish.internal.api.Globals;
 import org.glassfish.internal.data.ApplicationInfo;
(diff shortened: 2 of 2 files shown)


svendiedrichsen:
@dmatej You could use `java.util.Optional` to clean up code from all the `NULL` checks.

dmatej:
@svendiedrichsen Then I would need to combine two optionals, Java 9 has Optional.or not sure how to do that in Java 8, but I'll try to google it tomorrow :)
How would you do that?

svendiedrichsen:
@dmatej I have tried on the OpenTracingService.getApplicationName method. I doesn't look too pretty. You probably shouldn't do it.
```
    public String getApplicationName(final InvocationManager invocationManager) {
        return Optional.ofNullable(invocationManager.getCurrentInvocation().getAppName())
                .orElseGet(() ->
                    Optional.ofNullable(invocationManager.getCurrentInvocation().getModuleName())
                        .orElseGet(() ->
                                Optional.ofNullable(invocationManager.getCurrentInvocation().getComponentId())
                                    .filter(name -> Globals.getDefaultBaseServiceLocator().getService(ApplicationRegistry.class).get(name) == null)
                                    .map(name -> name.split("_/")[0])
                                    .orElse(null)
                        )
                );
    }
```

sgflt:
I would use this:

```
final String appName = Stream.of(
  invocation.getAppName(),
  invocation.getModuleName()
).filter(Objects::nonNull)
.findFirst()
.orElseGet(() -> buildAppNameFromComponentId(invocation.getComponentId()))

```

dmatej:
I thought we are talking about OpenTracingApplicationEventListener, not OpenTracingService. Then I would rewrite the whole file in the same style ;)

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/OpenTracingApplicationEventListener.java
```diff
@@ -88,9 +88,14 @@ public void postConstruct() {
...
             return;
         }
+        final InvocationManager invocationManager = serviceLocator.getService(InvocationManager.class);
         this.requestTracing = serviceLocator.getService(RequestTracingService.class);
         this.openTracing = serviceLocator.getService(OpenTracingService.class);
-        this.applicationName = this.openTracing.getApplicationName(serviceLocator.getService(InvocationManager.class));
+        if (invocationManager == null || this.openTracing == null) {
+            this.applicationName = null;
+        } else {
+            this.applicationName = this.openTracing.getApplicationName(invocationManager);
```
jbee: I don't know if this is a personal preference but I'd use a conditional if I had two cases to initialise the same variable. 
```java
this.applicationName = <test>
  ? <a>
  : <b>
```
Makes it more clear that there is a section of code and after the variable is initialised to a value.

jbee:
Please be careful with `Optional`. In the particular situation I don't think it improves the situation at all. It makes the code just appear much more complex than it actually is. I'd rather stick the candidates in a list and flatten the list to the first non null value or something like that. Or just do the cascading null check. Not the nicest code but it is very clear what it does. Takes a second to understand if you see it first time. Can't say the optional construct has the same quality.

dmatej:
Jenkins test please

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/OpenTracingApplicationEventListener.java
dmatej: Hmm. It is very similar, but I looked at it after another week and the if/else was optically a bit more straightforward. I use conditionals most time only if it is single true/false condition. More complicated logic => method/function. Two lines ok, three lines warning sign :-)

```
 this.applicationName = invocationManager == null || this.openTracing == null ? null
            : this.openTracing.getApplicationName(invocationManager);
```

PR #3437, opened 2018-11-19
Logging: Replacing usage of Thread/Timer/TimerTask with PayaraExecuto…

svendiedrichsen:
…rService in GFFileHandler

PR #4246, opened 2019-10-02
PAYARA-4118 Speed up PayaraExecutorService Shutdown 

Cousjava:
# Description
This is a performance imporvement

reverts #3437 
Due to the behaviour of GFFileHandler logpump it is better for it to be in it's own thread rather than use the executor service, having it as part of the executor service causes it to carry on running until killed.

# Testing

### New tests
None

### Testing Performed
Ran the command `asadmin start-domain --verbose --debug && date +"%Y-%m-%d %H:%M:%S.%3N` to print out time with milliseconds after domain is stopped, and compared that time with time that server shutdown was started.

Time before PR - 6.398 seconds
Time with 3c57d71 - 6.099 seconds
Time with c98a042 - 0.984 seconds

### Test suites executed

- [ ] Quicklook
- [ ] Payara Samples
- [ ] Java EE7 Samples
- [ ] Java EE8 Samples
- [ ] Payara Private Tests
- [ ] Payara Microprofile TCKs Runner
- [ ] Jakarta TCKs
- [ ] Mojarra
- [ ] Cargo Tracker

### Testing Environment
"Zulu JDK 1.8_222 on Ubuntu 19.04 DIsco Dingo with Maven 3.6.0"-->

### DIFF
--- a/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
+++ b/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
@@ -52,16 +52,28 @@ import com.sun.enterprise.util.io.FileUtils;
 import com.sun.enterprise.v3.logging.AgentFormatterDelegate;
 import fish.payara.enterprise.server.logging.JSONLogFormatter;
 import fish.payara.enterprise.server.logging.PayaraNotificationLogRotationTimer;
-import fish.payara.nucleus.executorservice.PayaraExecutorService;
-import java.io.*;
+import static java.security.AccessController.doPrivileged;
+
+import java.io.BufferedOutputStream;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
 import java.security.PrivilegedAction;
 import java.text.FieldPosition;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+import java.util.ResourceBundle;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.ErrorManager;
 import java.util.logging.Formatter;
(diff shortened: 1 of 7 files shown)


Cousjava:
Jenkins test please

nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
```diff
@@ -136,14 +136,19 @@ public void event(Event event) {
             if (null == threadPoolExecutor) {
                 initialiseThreadPools();
             }
+        } else if (event.is(EventTypes.PREPARE_SHUTDOWN)) {
+            stopThreadPools();
```
pdudits: I'm not sure if this is not too early -- prepare shutdown is broadcasted before applications are stopped, as initlevel is decreased after broadcasting this event.
Therefore applications might be failing during their stop.

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
```diff
@@ -88,53 +101,46 @@
 /**
  * GFFileHandler publishes formatted log Messages to a FILE.
  *
- * @author Jerome Dochez
- * @author Carla Mott
+ * @AUTHOR: Jerome Dochez
+ * @AUTHOR: Carla Mott
```
pdudits: nitpick: could you retain the cleanups from reverted commit, as these invalid javadocs, or typos further down the source?

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
```diff
@@ -551,68 +567,65 @@ private void configureUniformLogFormatter(String excludeFields, boolean multiLin
...
-                        try {
-                            log();
-                        } catch (Exception e) {
-                            // GLASSFISH-19125
-                            // Continue the loop without exiting
-                        }
+        pump = new Thread() {
+            @Override
+            public void run() {
+                while (!done.isSignalled()) {
```
pdudits: There also was another branch in the newer version, flushing records if `logToFile` is set to false in course of run. I believe that one should be retained

nucleus/payara-modules/payara-executor-service/src/main/java/fish/payara/nucleus/executorservice/PayaraExecutorService.java
smillidge: agree this is probably too early

smillidge:
Complete change of subject. Did you look to see if GFFileHandler log pump could be refactored to not need its own thread?

Cousjava:
replaced by #4250

PR #4249, opened 2019-10-03
PAYARA-4067: Add JSONB implementation into appclient runtime

pdudits:
# Description
This is a bug fix. 

When using appclient feature both API and implementation of JSOB need to be on available on application client classpath. Adding the dependency into acc magically changes classpath of `gf-client.jar` -- the entry point to application client.

# Testing

### Test suites executed

- Jakarta EE TCKs - jsonb in appclient suite.

### Testing Environment

Zulu JDK 1.8_222 on Ubuntu 18.04


pdudits:
jenkins test please

PR #4250, opened 2019-10-04
PAYARA-4118 Speedup Payara shutdown by having GFFileHandler use its o…

Cousjava:
# Description
This is a performance imporvement

reverts #3437 
Due to the behaviour of GFFileHandler logpump it is better for it to be in it's own thread rather than use the executor service, having it as part of the executor service causes it to carry on running until killed.

# Testing

### New tests
None

### Testing Performed
Ran the command `asadmin start-domain --verbose --debug && date +"%Y-%m-%d %H:%M:%S.%3N` to print out time with milliseconds after domain is stopped, and compared that time with time that server shutdown was started.

Time before PR - 6.398 seconds
Time with 4ddcf7f - 1.235 seconds

### Test suites executed

- [ ] Quicklook
- [ ] Payara Samples
- [ ] Java EE7 Samples
- [ ] Java EE8 Samples
- [ ] Payara Private Tests
- [ ] Payara Microprofile TCKs Runner
- [ ] Jakarta TCKs
- [ ] Mojarra
- [ ] Cargo Tracker

### Testing Environment
"Zulu JDK 1.8_222 on Ubuntu 19.04 DIsco Dingo with Maven 3.6.0"-->

### DIFF
--- a/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
+++ b/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
@@ -51,8 +51,7 @@ import com.sun.enterprise.util.LocalStringManagerImpl;
 import com.sun.enterprise.util.io.FileUtils;
 import com.sun.enterprise.v3.logging.AgentFormatterDelegate;
 import fish.payara.enterprise.server.logging.JSONLogFormatter;
-import fish.payara.enterprise.server.logging.PayaraNotificationLogRotationTimer;
-import fish.payara.nucleus.executorservice.PayaraExecutorService;
+import fish.payara.enterprise.server.logging.PayaraNotificationLogRotationTimer;;
 import java.io.*;
 import java.security.PrivilegedAction;
 import java.text.FieldPosition;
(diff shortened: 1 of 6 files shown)


Cousjava:
Jenkins test please

Cousjava:
Jenkins test please

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/GFFileHandler.java
```diff
@@ -585,14 +573,10 @@ private void configureUniformLogFormatter(String excludeFields, boolean multiLin
...
     void initializePump() {
-        if (pumpFuture != null) {
-            pumpFuture.cancel(true);
-            pumpFuture = null;
-        }
-        if (logToFile) {
-            pumpFuture = payaraExecutorService.submit(
-                () -> {
-                    while (!done.isSignalled() && logToFile) {
+        pump = new Thread() {
```
Pandrex247: It might be worth putting in a comment explaining why we're not using a the managed executor service here so that we don't come back to this in a year's time and go "why is that not using a managed executor?"

Cousjava:
Jenkins test please

PR #4252, opened 2019-10-04
PAYARA-4144 Add support for deployment groups in commands that don't list targets

Cousjava:
# Description
This is a bug fix

Error occurs with some asadmin commands, i.e. _list-app-refs when targeting a deployment group, leading to messages like:
    Target local-group is not a supported type. Command _list-app-refs supports these types of targets only: Default server, Stand alone instance, Cluster, Config, 

# Important Info

# Testing

### Testing Performed
 create-instance --node=localhost-domain1 local-instance
create-deployment-group local-group
add-instance-to-deployment-group --instance=local-instance --deploymentGroup=local-group
start-deployment-group local-group
Then ran test app (See Jira)(which uses cargo plugin) with `mvn integration-test`

### Test suites executed

- Quicklook
- Payara Samples
- Java EE7 Samples
- Java EE8 Samples
- Payara Private Tests
- Jakarta TCKs
- Mojarra
- Cargo Tracker

### Testing Environment
Zulu JDK 1.8_222 on Ubuntu 18.04 with Maven 3.6.0


Cousjava:
Jenkins test please

PR #4254, opened 2019-10-06
PAYARA-3829 MicroProfile Healthcheck 2.1 implementation

jGauravGupta:
# Description
MicroProfile Healthcheck 2.0 and 2.1 implementation:
- Message body of Health check response was modified, `outcome` and `state` replaced by `status` [Breaking Change]
- Introduction of Health checks for `@Liveness` and `@Readiness` on `/health/ready` and `/health/live` endpoints
- To enable the backward compatibility (`outcome` and `state` not replaced by `status` in JSON payload) of `/health` endpoint set the value of `mp.health.enable-backward-compatibility` MP config property to true.
- `/health` is deprecated and return combines Health checks data of `@Liveness` , `@Readiness` and `@Health`.

# Important Info

### Dependant PRs 
https://github.com/payara/MicroProfile-TCK-Runners/pull/59

# Testing

### New tests
Microprofile HealthCheck TCK v2.1

### Test suites executed
- Payara Microprofile TCKs Runner - HealthCheck TCK v1.0 where `mp.health.enable-backward-compatibility=true`
- Payara Microprofile TCKs Runner - HealthCheck TCK v2.1

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
+++ b/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
@@ -108,6 +108,8 @@ public class HealthCheckService implements EventListener, ConfigListener {
     private final Map<String, ClassLoader> applicationClassLoaders = new ConcurrentHashMap<>();
     private final List<String> applicationsLoaded = new CopyOnWriteArrayList<>();
 
+    private static final String BACKWARD_COMP_ENABLED_PROPERTY = "MP_HEALTH_BACKWARD_COMPATIBILITY_ENABLED";
+
     @PostConstruct
     public void postConstruct() {
         if (events == null) {
@@ -115,7 +117,7 @@ public class HealthCheckService implements EventListener, ConfigListener {
         }
         events.register(this);
         this.backwardCompEnabled = ConfigProvider.getConfig()
-                .getOptionalValue("mp.health.enable-backward-compatibility", Boolean.class)
+                .getOptionalValue(BACKWARD_COMP_ENABLED_PROPERTY, Boolean.class)
                 .orElse(false);
     }
 
@@ -209,7 +211,7 @@ public class HealthCheckService implements EventListener, ConfigListener {
         return healthChecks;
     }
     
-        private Map<String, Set<HealthCheck>> getCollectiveHealthChecks(HealthCheckType type){
+    private Map<String, Set<HealthCheck>> getCollectiveHealthChecks(HealthCheckType type) {
         final Map<String, Set<HealthCheck>> healthChecks;
         if (type == READINESS) {
             healthChecks = readiness;
--- a/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
+++ b/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
@@ -55,10 +55,10 @@ import javax.json.Json;
 import javax.json.JsonArrayBuilder;
 import javax.json.JsonObjectBuilder;
 import javax.servlet.http.HttpServletResponse;
-
 import fish.payara.microprofile.healthcheck.config.MetricsHealthCheckConfiguration;
-
+import java.util.HashMap;
 import static java.util.logging.Level.WARNING;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.health.HealthCheck;
 import org.eclipse.microprofile.health.HealthCheckResponse;
 import org.glassfish.api.StartupRunLevel;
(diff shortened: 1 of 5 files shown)


jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -104,6 +109,9 @@ public void postConstruct() {
             events = Globals.getDefaultBaseServiceLocator().getService(Events.class);
         }
         events.register(this);
+        this.backwardCompatibilityEnabled = ConfigProvider.getConfig()
+                .getOptionalValue("mp.health.enable-backward-compatibility", Boolean.class)
```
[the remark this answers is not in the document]
jGauravGupta: This flag is specifically for the Payara not part of MicroProfile Health spec.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -220,35 +196,50 @@ public void registerClassLoader(String appName, ClassLoader classloader) {
...
+    }
+    
+        private Map<String, Set<HealthCheck>> getCollectiveHealthChecks(HealthCheckType type){
+        final Map<String, Set<HealthCheck>> healthChecks;
+        if (type == READINESS) {
+            healthChecks = readiness;
+        } else if (type == LIVENESS) {
+            healthChecks = liveness;
+        } else {
+            healthChecks = new HashMap<>(health);
```
jbee: What I understand from this now is that when running health checks when doing type `READINESS` or `LIVENESS` we run just those but for `HEALTH` we run any kind of check? Is that it?

jbee: What I understand from this now is that when running health checks when doing type `READINESS` or `LIVENESS` we run just those but for `HEALTH` we run any kind of check? Is that it?

jGauravGupta: Yes @jbee, `HEALTH` is deprecated hence merging all other health checks in output.

> The MicroProfile Health Check architecture consists of two /health/ready and /health/live endpoints in a MicroProfile runtime that respectively represent the readiness and the liveness of the entire runtime.
For backward compatibility, a 3rd endpoint /health may also be used to provide a combination of previous endpoints and Health Check procedures annotated with the deprecated @Health annotation.

https://github.com/eclipse/microprofile-health/blob/master/spec/src/main/asciidoc/overview.adoc

jbee: Thanks @jGauravGupta for the explanation. Couldn't we keep all checks in a common data structure and process them by using a filter? I'd think that this gets much easier to manage a single collection and to send a filter predicate to the processing - something like `(check) -> check.getType() == READINESS` and for backwards compatibility it is just `(check) -> true`.

jGauravGupta: One health check could also be part of more than one type e.g both `@Readiness` and `@Liveness`. 
Also, Health Check endpoints are invoked more frequently and intended for the machine to machine communication, Removing type-based filter (and iteration) optimized the request further.

jbee: I don't see how constructing a new collection each time is more optimised then iterating the same one and applying a filter. Especially when a check can be of more then one type the filter seems a good approach to me. Still, it is just a suggestion.

jGauravGupta: > how constructing a new collection each time is more optimised
But this case is specifically for deprecated `/health` endpoint, which is not recommended to use.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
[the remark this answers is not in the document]
jbee: We might want to use another name since the `mp.<spec>.*` usually is used for constants in the spec. In FT they also don't use the canonical names for configuration values but simple names like `MP_Fault_Tolerance_Metrics_Enabled`. Can't say for the other MP parts but it would be nice if there was some consistency in the naming scheme.

jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -92,18 +97,28 @@
...
 
+    private static final String BACKWARD_COMP_ENABLED_PROPERTY = "MP_HEALTH_BACKWARD_COMPATIBILITY_ENABLED";
+
     @PostConstruct
     public void postConstruct() {
         if (events == null) {
             events = Globals.getDefaultBaseServiceLocator().getService(Events.class);
         }
         events.register(this);
+        this.backwardCompEnabled = ConfigProvider.getConfig()
```
Pandrex247: Should this also be checked on each invocation?
As it is, it's not dynamic which slightly defeats one of the purposes of MP Config.

jGauravGupta: I purposely added this snippet to PostConstruct method instead of each invocation for very minor optimization. Do you think, end-user will need to change `MP_HEALTH_BACKWARD_COMPATIBILITY_ENABLED` property value dynamically?

Pandrex247: In 95% of cases, probably not.
But it's the fact that by being only checked in the postconstruct it essentially means that a restart is required if you want to change this. We can't however actually indicate that, because since it's MP config the option can be defined outside of Payara, so we have no way to check if it has changed except by, well, running this exact check :P

It's _probably_ fine as is, we'd just need to make sure to explicitly document this case.

PR #4257, opened 2019-10-07
PAYARA-4196 Use unit separator to split tags, rather than space

Cousjava:
# Description
This is a bug fix

As tag values can be any valid UTF-8 character, and spaces are used frequently
i.e. in the name of the garbage collector PS Marksweep

Picked a control character to split tags as that is least likely to be split on

# Testing

### New tests
New unit test added (see PR)

### Testing Performed
Testing also performed by starting up a DAS and instance, no errors occur in logs

### Test suites executed

- Quicklook
- Payara Samples
- Java EE7 Samples
- Java EE8 Samples
- Payara Private Tests
- Jakarta TCKs
- Mojarra
- Cargo Tracker

### Testing Environment
"Zulu JDK 1.8_222 on Ubuntu 18.04 with Maven 3.6.0

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/Series.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/Series.java
@@ -52,8 +52,8 @@ import java.util.regex.Pattern;
 public final class Series implements Comparable<Series>, Serializable {
 
     public static final char TAG_ASSIGN = ':';
-    public static final char TAG_SEPARATOR = ' ';
-    private static final char[] TAG_SEPARATORS = { ' ', ',', ';' };
+    public static final char TAG_SEPARATOR = '\u001F'; //As tag values can be any valid UTF-8 character, use a control character to separate tags
+    private static final char[] TAG_SEPARATORS = { TAG_SEPARATOR, TAG_ASSIGN };
 
     private static final String SPLIT_PATTERN = "[" + Pattern.quote(new String(TAG_SEPARATORS)) + "]+";
 
@@ -130,11 +130,6 @@ public final class Series implements Comparable<Series>, Serializable {
         if (c == TAG_ASSIGN) {
             return true;
         }
-        for (char sep : TAG_SEPARATORS) {
-            if (sep == c) {
-                return true;
-            }
-        }
-        return false;
+        return TAG_SEPARATOR == c;
     }
 }
--- a/appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/collect/DataSinkCollectorTest.java
+++ b/appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/collect/DataSinkCollectorTest.java
@@ -57,6 +57,7 @@ import org.junit.Test;
 
 import fish.payara.monitoring.store.MonitoringDataSink;
 import fish.payara.monitoring.store.SinkDataCollector;
+import fish.payara.monitoring.model.Series;
 
 /**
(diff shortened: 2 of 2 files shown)


Cousjava:
Jenkins test please

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/Series.java
```diff
@@ -52,8 +52,8 @@
 public final class Series implements Comparable<Series>, Serializable {
 
     public static final char TAG_ASSIGN = ':';
-    public static final char TAG_SEPARATOR = ' ';
-    private static final char[] TAG_SEPARATORS = { ' ', ',', ';' };
+    public static final char TAG_SEPARATOR = '\u001F'; //As tag values can be any valid UTF-8 character, use a control character to separate tags
```
jbee: The design that series are human readable and writable is intentional - using a control character compromises this property which I don't like as it messes up tooling and ease of testing things manually. The benefit of using a control character would only fully apply if the `TAG_ASSIGN` is made some control character as well. Besides that it is superficial because there would just be other characters that mess up the parsing even if we can expect them to be less frequent. 

We have no requirement to support unchanged arbitrary names so I think the solution I suggested to simply append the tag name with special character replacement is the most balanced. It yields the advantages of a clear human readable and writable format which are useful in many regards while its disadvantages of character replacement only apply to edge cases where a replacement does not alter the name or value in a way that would make it hard to recognise.

appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/collect/DataSinkCollectorTest.java
```diff
@@ -168,37 +170,42 @@ public void plainAfterSubHasNoTags() {
...
     @Test
     public void sameTagDoesReplaceExistingTagFromThatTagOn() {
-        assertDataPoint("sub:five reset", 11L);
+        assertDataPoint("sub:five" + Series.TAG_SEPARATOR + "reset", 11L);
+    }
+    
+    @Test
+    public void tagwithUnusualCharacters() {
+        assertDataPoint("complex:sp aced; str,\u1F408ange" + Series.TAG_SEPARATOR +"sub", 1);
     }
```
Cousjava: New test

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/Series.java
Cousjava: If this is meant to be human-readable, that is what MetricWriter and its sub-classes are for. The PR includes a test which ensures that odd values are written correctly, so TAG_ASSIGN does not need to be a control character.

jbee: I don't say that the conversion of metric data is how it should be. Still, the collection API should protect against values being passed that later on fail to parse to a `Series` object that expects certain characters to have special meaning. I opted to simply replace those characters with `_` as all of them are kind of a separator character which `_` can replace quite ok.
So if the collection API has this mechanism anyway there is little point to use custom `MetricWriter`s unless we really want to have them map to other replacement characters. 

Anyhow, the pain that would be inflicted by a non human readable and writable format would be far far bigger then to little itch of having 5 characters replaced with `_`. You have to remember that the series name is THE identifier concept that humans do read and write, copy & paste, send in mails and might even put in URLs. Compare the headache a control character in that would cause in comparison to the minor inconvenience some replaced characters in odd names have.

Cousjava:
replaced by #4261

PR #4260, opened 2019-10-08
PAYARA-4099 Allow Creation of Temp Nodes

Pandrex247:
# Description
This is a new feature intended for use with Docker/Kubernetes to more easily allow autoscaling. 

Previously, autoscaling could only be done in a limited capacity, and very messily to boot, due to the restriction of requiring a node to be mapped to each instance for the DAS to talk to. This PR introduces the concept of Temp (temporary) nodes, which are similar to config nodes, but only persist as long as they have instances assigned to them. As demonstrated in the updated Docker Node image, this allows containers to create their own nodes and instances and register themselves to the DAS (autogenerating names as required).

Temp nodes and instances created upon them have some behaviour that differentiate themselves from normal Config nodes:
* When an instance created on a temp node is stopped, it is automatically deleted.
* When Payara Server detects that a temp node has no more instances assigned to it, the node is deleted.
  * Payara Server performs this check upon instance stop/deletion, and when shutting down. This is to avoid unnecessary locking or race conditions caused by a periodic reaper service: creation of nodes and instances are done separately and can be interrupted.
* Temp nodes do not show up on the admin console, and are read only.
  * They do however still show up if you look in the REST management console, or inspect the dotted name tree / domain.xml - I'm not aware of any way around this.

# Testing

### New tests
`JsonRequestConstructorTest.java` updated with custom config name (rather than using default).
Otherwise none - any test requires environment setup (Docker, multiple VMs etc.).

### Testing Performed
All testing performed with DAS on a Windows machine, and Docker containers on a Linux VM (so I can use _host_ network mode).

**Creating an instance using Docker REST API**:
* Setup DAS on Windows Host:
  * Start clean domain
  * `change-admin-password`
  * `enable-secure-admin`
  * `restart-domain`
* Create Docker container:
  * `curl -X POST -H 'Accept: application/json' -H 'Content-Type: application/json' -i 'http://192.168.150.141:2376/containers/create' --data '{
  "Image": "payara/server-node:latest",
  "HostConfig": {
    "Mounts": [
      {
        "Type": "bind",
        "Source": "/home/andrew/DockerTest/passwordfile.txt",
        "Target": "/opt/payara/passwords/passwordfile.txt",
        "ReadOnly": true
      }
    ],
    "NetworkMode": "host"
  },
  "Env": [
    "PAYARA_DAS_HOST=192.168.253.1"
  ]
}'`
* _Repeated with various bits of info added or missing (e.g. instance name provided, config name provided, node name provided_)

**Restarting a container that was stopped manually**: 
If given an instance name, will attempt to start existing instance, if given no info, will create new instance and node from scratch. 

**Temp Nodes Deletion**
* Stop a running instance on a temp node from DAS > instance stopped and deleted, plus node deleted (checked via REST Management interface)
* Delete a stopped instance on a temp node from DAS (instance stopped without deletion by stopping _container_) > instance and node deleted.
* Delete a stopped instance on a temp node with more than one instance > instance deleted, node left alone.
* Temp nodes deleted on DAS restart:
  * Stop containers to stop instances without deletion
  * Restart DAS > instances and node should be gone upon restart.

### Test suites executed

- Quicklook
- MicroProfile TCK Runners

### Testing Environment
Zulu JDK8u222, OpenSUSE Tumbleweed Linux, Maven 3.6.0.
Zulu JDK8u222, Windows 10, Maven 3.6.0.
Docker 19.03.1, API 1.40

# Documentation
https://github.com/payara/Payara-Server-Documentation/pull/632

# Notes for Reviewers
You'll need to build the Docker image and then specify that it's the image you want to use in your containers. Depending on your environment you can build the image as a part of the normal Maven build by adding the profile `-PBuildDockerImages`. Otherwise build the image manually using Docker CLI.


Pandrex247:
Jenkins test please

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
```diff
@@ -148,9 +150,14 @@ public void addRemoteDatasets(Message<SeriesDatasetsSnapshot> message) {
...
-            long value = snapshot.values[i];
-            remoteInstanceDatasets.compute(series, (key, seriesByInstance) -> addRemotePoint(seriesByInstance, instance, key, time, value));
+            try {
+                Series series = new Series(snapshot.series[i]);
+                long value = snapshot.values[i];
+                remoteInstanceDatasets.compute(series, (key, seriesByInstance) -> addRemotePoint(seriesByInstance, instance, key, time, value));
+            } catch (IndexOutOfBoundsException ioobe) {
+                Logger.getLogger(InMemoryMonitoringDataRepository.class.getName()).log(Level.FINE,
+                        "Could not create new series object for snapshot", ioobe);
+            }
```
Cousjava: Please revert this, the bug this is trying to fix is fixed properly by #4261

appserver/admingui/cluster/src/main/java/fish/payara/cluster/admingui/handlers/PayaraRestApiHandlers.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.Map;
+
+import static org.glassfish.admingui.common.util.RestUtil.buildChildEntityList;
+
+/**
+ * Handler methods for the admingui cluster module.
+ *
+ * @author AndrewPielage <andrew.pielage@payara.fish>
+ */
+public class PayaraRestApiHandlers {
```
dmatej: I will play the SonarQube rules: Minor: "undocumented public methods".

appserver/admingui/cluster/src/main/resources/node/nodes.jsf
```diff
@@ -79,7 +79,7 @@
     <!beforeCreate
         setPageSessionAttribute(key="parentUrl" value="#{sessionScope.REST_URL}/nodes");
         setPageSessionAttribute(key="childType" value="node");
-        gf.getChildList(parentEndpoint="#{pageSession.parentUrl}", childType="#{pageSession.childType}", result="#{requestScope.listOfRows}");
+        py.getNodesList(parentEndpoint="#{pageSession.parentUrl}", childType="#{pageSession.childType}", result="#{requestScope.listOfRows}");
```
dmatej: Copyrights ...

appserver/docker/src/main/java/fish/payara/docker/node/admin/DeleteNodeDockerCommand.java
```diff
@@ -99,7 +99,6 @@
     @Override
     public void execute(AdminCommandContext context) {
         ActionReport actionReport = context.getActionReport();
-        Logger logger = context.getLogger();
```
dmatej: Remove also the unused import

appserver/extras/docker-images/src/main/resources/payaraserver-node/entrypoint.sh
```diff
@@ -4,15 +4,105 @@ set -e
...
+    echo "No Docker container IP override given, setting to first result from 'hostname -I'"
+    DOCKER_CONTAINER_IP="$(hostname -I | cut -f1 -d ' ')"
+    echo "Hostname is ${DOCKER_CONTAINER_IP}"
+fi
+
+echo "Docker Container IP is: ${DOCKER_CONTAINER_IP}"
+
 ### Functions ###
+function checkAndCreateNewNodeIfRequired {
+    if [ -z ${PAYARA_NODE_NAME} ]; then
```
dmatej: Missing quotation marks

appserver/extras/docker-images/src/main/resources/payaraserver-node/entrypoint.sh
```diff
@@ -4,15 +4,105 @@ set -e
...
+                PAYARA_INSTANCE_NAME="$(./payara5/bin/asadmin -I false -T -a -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} create-local-instance --node ${PAYARA_NODE_NAME} --config ${PAYARA_CONFIG_NAME} --dockernode true --ip ${DOCKER_CONTAINER_IP})"
+            fi
+        fi
     else
-        echo "./payara5/bin/asadmin -I false -T -a -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} create-local-instance --node ${PAYARA_NODE_NAME} --dockernode true ${PAYARA_INSTANCE_NAME}"
-        PAYARA_INSTANCE_NAME="$(./payara5/bin/asadmin -I false -T -a -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} create-local-instance --node ${PAYARA_NODE_NAME} --dockernode true ${PAYARA_INSTANCE_NAME})"
+        if [ -z ${PAYARA_CONFIG_NAME} ]; then
+            echo "./payara5/bin/asadmin -I false -T -a -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} create-local-instance --node ${PAYARA_NODE_NAME} --dockernode true --ip ${DOCKER_CONTAINER_IP} ${PAYARA_INSTANCE_NAME} "
```
dmatej: To debug commands you can also put this after the first line of the bash script:
`set -x`
Or put the command to a variable and then use echo and eval.

appserver/extras/docker-images/src/main/resources/payaraserver-node/entrypoint.sh
```diff
@@ -34,9 +125,9 @@ else
         # Check if an instance with this name is actually registered
         echo "Checking if an instance with name ${PAYARA_INSTANCE_NAME} has been registered with the DAS"
         echo "./payara5/bin/asadmin -I false -t -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} list-instances --nostatus ${PAYARA_INSTANCE_NAME}"
-        EXISTS="$(./payara5/bin/asadmin -I false -t -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} list-instances --nostatus ${PAYARA_INSTANCE_NAME})" || true
+        INSTANCE_EXISTS="$(./payara5/bin/asadmin -I false -t -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} list-instances --nostatus ${PAYARA_INSTANCE_NAME})" || true
 
-        if [ ! -z ${EXISTS} ]; then
+        if [ ! -z ${INSTANCE_EXISTS} ]; then
```
dmatej: Missing quotation marks

appserver/extras/docker-images/src/main/resources/payaraserver-node/entrypoint.sh
Pandrex247: Are they necessary? I haven't got them anywhere else and it works fine from my testing.

dmatej: https://stackoverflow.com/questions/51112152/how-to-accept-a-quoted-sentence-as-argument-in-shell-script - I didn't try, but I expect that if I would set the property with spaces inside, instead of an error reponse the script would crash.

dmatej:
Jenkins test please

PR #4261, opened 2019-10-08
PAYARA-4196 Replace spaces in tag values with an underscore

Cousjava:
# Description

This is a bug fix

As tag values can be any valid UTF-8 character, and spaces are used frequently
i.e. in the name of the garbage collector PS Marksweep

Picked a control character to split tags as that is least likely to be split on
 # Testing
### New tests

New unit test added (see PR)
### Testing Performed

Testing also performed by starting up a DAS and instance, no errors occur in logs

# Testing Environment

Zulu JDK 1.8_222 on Ubuntu 18.04 with Maven 3.6.0


Cousjava:
Jenkins test please

Cousjava:
Jenkins test please

PR #4262, opened 2019-10-08
PAYARA-4202 PAYARA-4206 Support for optional arrays and char type in config

pdudits:
# Description
This is a bug fix. 

Obtaining an array over `Config.getOptionalValue(String, Class)` previously threw an exception
```
Caused by: java.lang.IllegalArgumentException: Unable to convert value to type java.lang.String[]
        at fish.payara.nucleus.microprofile.config.spi.PayaraConfig.getConverter(PayaraConfig.java:243)
        at fish.payara.nucleus.microprofile.config.spi.PayaraConfig.getOptionalValue(PayaraConfig.java:99)
```
First commit of the PR fixes this.

This use of API was not covered by TCK test, therefore tests have been added upstream in https://github.com/eclipse/microprofile-config/pull/452.

Running against these tests showed, that we miss converters for `char` and `Character` types, which is added by the second commit.

Last commit cleans up logic around selecting fitting converter for a type, and moves discovery of automatic converters into `CommonSenseConverter.forClass(Class)`

# Important Info

# Testing

### New tests
https://github.com/eclipse/microprofile-config/pull/452

### Test suites executed

- Payara Microprofile TCKs Runner - Config

### Testing Environment

Zulu JDK 1.8_221 on Windows 10 with Maven 3.6.1

# Notes for Reviewers

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/AutomaticConverter.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/AutomaticConverter.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2018 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018-2019 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/AutomaticConverter.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/AutomaticConverter.java
@@ -46,6 +46,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
@@ -54,17 +55,21 @@ import java.util.stream.Stream;
  *
  * @author steve
  */
-public class CommonSenseConverter implements Converter<Object> {
+public class AutomaticConverter implements Converter<Object> {
 
     /**
      * Return implicit converter for a class following section "Automatic Converters" of the spec, if class has
      * matching construct
      *
-     * @param type target type of property
+     * @param generalType target type of property
      * @param <T> target type of property
      * @return Optional of converter using a method found in the class, empty if none were found
      */
-    public static <T> Optional<Converter<T>> forClass(Class<T> type) {
+    public static <T> Optional<Converter<T>> forType(Type generalType) {
+        if (!(generalType instanceof Class)) {
+            return Optional.empty();
+        }
+        Class<T> type = (Class<T>)generalType;
         return Stream.<Supplier<Converter<T>>>of(
                 () -> forMethod(type, "of", String.class),
                 () -> forMethod(type, "valueOf", String.class),
@@ -78,11 +83,11 @@ public class CommonSenseConverter implements Converter<Object> {
     private Method conversionMethod;
     private Constructor constructor;
     
-    private CommonSenseConverter(Method method) {
+    private AutomaticConverter(Method method) {
         conversionMethod = method;
     }
     
-    private CommonSenseConverter(Constructor method) {
+    private AutomaticConverter(Constructor method) {
         constructor = method;
     }
 
@@ -110,7 +115,7 @@ public class CommonSenseConverter implements Converter<Object> {
         try {
             Method factoryMethod = type.getMethod(method, argumentTypes);
             if (Modifier.isStatic(factoryMethod.getModifiers()) && Modifier.isPublic(factoryMethod.getModifiers())) {
-                return (Converter<T>) new CommonSenseConverter(factoryMethod);
+                return (Converter<T>) new AutomaticConverter(factoryMethod);
             } else {
                 return null;
             }
@@ -123,7 +128,7 @@ public class CommonSenseConverter implements Converter<Object> {
         try {
             Constructor<T> constructor = type.getConstructor(argumentTypes);
             if (Modifier.isPublic(constructor.getModifiers())) {
-                return (Converter<T>) new CommonSenseConverter(constructor);
+                return (Converter<T>) new AutomaticConverter(constructor);
             } else {
                 return null;
             }
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/CharacterConverter.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/CharacterConverter.java
@@ -1,0 +1,59 @@
+/*
+ *    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) [2019] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ *    The contents of this file are subject to the terms of either the GNU
+ *    General Public License Version 2 only ("GPL") or the Common Development
+ *    and Distribution License("CDDL") (collectively, the "License").  You
+ *    may not use this file except in compliance with the License.  You can
+ *    obtain a copy of the License at
+ *    https://github.com/payara/Payara/blob/master/LICENSE.txt
+ *    See the License for the specific
+ *    language governing permissions and limitations under the License.
+ *
+ *    When distributing the software, include this License Header Notice in each
+ *    file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ *    GPL Classpath Exception:
+ *    The Payara Foundation designates this particular file as subject to the "Classpath"
+ *    exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ *    file that accompanied this code.
+ *
+ *    Modifications:
+ *    If applicable, add the following below the License Header, with the fields
+ *    enclosed by brackets [] replaced by your own identifying information:
+ *    "Portions Copyright [year] [name of copyright owner]"
+ *
+ *    Contributor(s):
+ *    If you wish your version of this file to be governed by only the CDDL or
+ *    only the GPL Version 2, indicate your decision by adding "[Contributor]
+ *    elects to include this software in this distribution under the [CDDL or GPL
+ *    Version 2] license."  If you don't indicate a single choice of license, a
+ *    recipient has the option to distribute your version of this file under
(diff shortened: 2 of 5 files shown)


pdudits:
jenkins test please

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -177,74 +174,60 @@ private String getValue(String propertyName) {
...
+
+        throw new IllegalArgumentException("Unable to convert value to type " + propertyType.getCanonicalName());
+    }
+
+
+
+    private static <T> Class<T> boxedTypeOf(Class<T> type) {
+        if (!type.isPrimitive()) {
+            return type;
+        } else if (type.equals(int.class)) {
```
jbee: `else` can be spared - I know it makes the method longer with our formatting rules but it is simpler without the `else`. `equals` can be `==`.

I benchmarked this years ago - using a map is surprisingly equally fast but I also ended up keeping the if's since those can be ordered checking for the most likely first.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -177,74 +174,60 @@ private String getValue(String propertyName) {
...
+            return (Class<T>) Float.class;
+        } else if (type.equals(byte.class)) {
+            return (Class<T>) Byte.class;
+        } else if (type.equals(char.class)) {
+            return (Class<T>) Character.class;
+        } else if (type.equals(short.class)) {
+            return (Class<T>) Short.class;
         }
-        return converter;
+        return type;
```
jbee: We already established that `type` is a primitive type so should we ever get here this should mean a primitive had been added to the JVM and java that we have not covered so its more of a `throw new` situation if you ask me.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -177,74 +174,60 @@ private String getValue(String propertyName) {
...
-            for (Entry<Type, Converter> entry : converters.entrySet()) {
-                Type type1 = entry.getKey();
-                if (type1 instanceof ParameterizedType) {
-                    ParameterizedType ptype = (ParameterizedType)type1;
-                    if (ptype.getRawType().equals(propertyType)) {
-                        converter = entry.getValue();
-                        break;
-                    }
+        // search for a matching raw type
+        for (Entry<Type, Converter> entry : converters.entrySet()) {
```
jbee: Maybe the converters should be organized by raw type first, then in a map of parameterized types as second level.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -177,74 +174,60 @@ private String getValue(String propertyName) {
...
+            Type type1 = entry.getKey();
+            if (type1 instanceof ParameterizedType) {
+                ParameterizedType ptype = (ParameterizedType)type1;
+                if (ptype.getRawType().equals(propertyType)) {
+                    return entry.getValue();
```
jbee: Returning the first converter with matching raw types must be a bug. Say I have a converter `List<String>` and `List<? extends Number>` this could respond with `List<String` when I ask for `List<Integer>` even though there was a proper match.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/CommonSenseConverter.java
```diff
@@ -39,51 +39,97 @@
...
+import java.util.stream.Stream;
 
 /**
  *
  * @author steve
  */
 public class CommonSenseConverter implements Converter<Object> {
 
+    /**
+     * Return implicit converter for a class following section "Automatic Converters" of the spec, if class has
```
jbee: Just for the record: I'd find the name `AutomaticConverter` as suggested by this comment much more explanatory.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/CharacterConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+@Priority(1)
+public class CharacterConverter implements Converter<Character> {
+    @Override
+    public Character convert(String value) {
+        if (value == null || value.equals(ConfigProperty.UNCONFIGURED_VALUE) || value.length() == 0) return null;
+        if (value.length() == 1) {
+            return value.charAt(0);
+        } else {
+            throw new IllegalArgumentException("Value is more than one character: ["+value+"]");
```
jbee: Is there a spec section on this scenario? _Getting more than needed or wanted_

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -177,74 +174,60 @@ private String getValue(String propertyName) {
...
         }
-        if (converter == null) {
-            throw new IllegalArgumentException("Unable to convert value to type " + propertyType.getCanonicalName());
+
+        throw new IllegalArgumentException("Unable to convert value to type " + propertyType.getCanonicalName());
+    }
+
+
+
+    private static <T> Class<T> boxedTypeOf(Class<T> type) {
```
jbee: The spec says (https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/converters.asciidoc): 
> The following Converters are provided by MicroProfile-Config by default:

Then naming primitive and their wrapper types. Does this mean this code is actually build into some shared code or does it want to tell us how we have to implement this? If this is already implemented there should not be a need to have this method since primitives and their wrappers are already covered.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
pdudits: Yes, but do we want such surprise few years later? :D But yes, initially these branches didn't list `short` and `char`.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/CharacterConverter.java
pdudits: It doesn't. But other converters fail with `IllegalArgumentException` when they cannot parse the value. E. g. `IntegerConverter` doesn't trim decimal point when there is one....

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
jbee: We could also change the name to `boxedTypeOrNot` to make the caller aware of this behaviour :D

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
pdudits: The latter. We need to implement this, which is exactly why this conversion happens. Other option would be to populate default converters map with both options, which is likely even simpler.

jbee: Unless... overrides. And of course there can be.

MarkWareham:
jenkins test please

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
pdudits: That's right, I think the support for generic types work only for Lists and Sets, and it has specific code in CDI extension for that. I'll remove this part

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
pdudits: Since this ends up just being key in the map, I think I'll just initialize default converters map with both types.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
pdudits: Actually, the only type reaching that return statement is... `void`, which is quite improbable for config property, but makes the code completely correct :D

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/converters/AutomaticConverter.java
```diff
@@ -39,51 +39,102 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 /**
  *
  * @author steve
  */
-public class CommonSenseConverter implements Converter<Object> {
+public class AutomaticConverter implements Converter<Object> {
```
jbee: Needs copyright update ;(

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -39,29 +39,24 @@
  */
 package fish.payara.nucleus.microprofile.config.spi;
 
-import fish.payara.nucleus.microprofile.config.converters.CommonSenseConverter;
+import fish.payara.nucleus.microprofile.config.converters.AutomaticConverter;
```
jbee: Needs copyright update.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
```diff
@@ -62,6 +62,7 @@
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import fish.payara.nucleus.microprofile.config.converters.CharacterConverter;
```
jbee: Needs copyright update.

pdudits:
jenkins test please

PR #4263, opened 2019-10-08
Payara-4176 TestContainers+Arquillian+JUnit5 Prototype

dmatej:
Prototype of tests with the usage of TestContainers, Arquillian and JUnit5

# Status

**STILL IN PROGRESS!**

* remote tests OK
* in container tests - injections does not work
* DockerNodesITest - "docker in docker" 
  - management of docker payara instances
* DatabaseITest 
  - real database server and real payara server
  - incomplete, will be refactored
* CertificateRealmITest
  - complete
  - note that we don't need Arquillian, we can deploy the physical application using asadmin commands or client, and find the url using it too. But this is simplier.
* VerboseLoggingITest 
  - server hangs with full logging (known issue, czclic dependency with HK2)
* JtaTimeoutLoggingITest 
  - tests logging of the Payara Server DAS, asynchronous
```
git  checkout PAYARA-4176-TestContainers-prototype
mvn clean install -PQuickBuild -DskipTests=true 
mvn clean verify -Ptest-containers  -pl :test-containers
```
- compiles Payara fastest as possible, skips broken tests
- runs tests; first run is slow, downloads docker image to cache. Next run takes around minute.

# Dependencies
https://github.com/dmatej/arquillian-junit5-extension (fixed master) (local mvn clean install required)
https://github.com/zforgo/arquillian-junit5-extension (original master)

# See Also
https://github.com/arquillian/arquillian-core/issues/137


appserver/tests/pom.xml
```diff
@@ -50,17 +50,44 @@
     </parent>
 
     <groupId>fish.payara.server.internal.tests</groupId>
-    <artifactId>tests</artifactId>
+    <artifactId>payara-tests</artifactId>
```
MarkWareham: affects payara-samples module

dmatej: I think it does not, payara-samples inherit from another module, payara-aggregator. And it is not compilable, I wanted to fix it, but I had to postpone it, because there is much more issues ... but they are not related to this (but this can make them visible, because it integrates samples to the reactor).

Samples still use old group ids before Patrik's changes.

MarkWareham: So it is!. I assumed similarity :pensive:

dmatej:
I'm closing this PR, but not the branch. The branch is and will be used as an incubator for other issues, and it will in some future it will be split to standalone pull requests.
Also TestContainers usage will be refactored, because with new docker images it can build tests on them, not on own (expensive) image.

PR #4273, opened 2019-10-16
Print the class name of the original exception in DeploymentException

OndroMih:
# Description
This is a small enhancement to log the class name of the original exception thrown during app deployment. The current solution transfers the stack trace and the message from the original exception to the DeploymentException but the name of the original exception is lost.

This helps in investigation of deployment exceptions which don't give enough detail in the message and in the stack trace, especially if the exception comes from the native code which is the case I have now with one of our customers.

# Important Info

# Testing

### Testing Performed

I haven't tested this as it's a minor enhancement and doesn't really impact anything besides the message in the exception. It should be enough if the code compiles.


dmatej:
Why it does not use standard exception mechanism and replaces original exception with DeploymentException with the original stacktrace? Is it some specification requirement or ... why?

svendiedrichsen:
But you are still losing the stacktrace and message of the original exception. Wouldn't it be helpful to have this and go with usual way by passing the exception to the DeploymentException?

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/DeploymentException.java
```diff
@@ -67,7 +67,7 @@ public DeploymentException(Throwable t)
             // the DeploymentException travels between processes and needs
```
MattGill98: I know that this comment suggests that they don't want to use the normal `super(Throwable)` mechanism. I would suggest that this is faulty logic. All exceptions should be serializable (as `Throwable` implements `Serializable`. Perhaps it could use this method, and then perform some logic to make sure every chained exception is serializable?

dmatej: Aha, they throw away all internal causes, create new exception and only use stacktrace of the last exception, so deployer would never see internals. But also does not know what caused the failure.
Nevermind, I would vote for standard mechanism. Usually the person deploying the application is admin user.
The problem might be some old RMI client and deserialization, but I don't know noone who would use it today. Maybe deployment-client, JSR-88? We can try it ;)

PR #4277, opened 2019-10-17
PAYARA-3811 improve deployment times

Cousjava:
# Description
This is a bug fix

Contains some small improvements to deployment, reducing deployment times by ~9%.

# Testing

### Testing Performed
Tested by using netinium's test domain + app and see how long deployment takes
Before: 77 seconds
After:    70 seconds

### Testing Environment
Zulu JDK 1.8_181 on Windows 10 with Maven 3.6.0

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -114,7 +114,7 @@ import java.util.logging.Logger;
  * Specialized web application class loader.
  * <p>
  * This class loader is a full reimplementation of the
- * <code>URLClassLoader</code> from the JDK. It is desinged to be fully
+ * <code>URLClassLoader</code> from the JDK. It is designed to be fully
  * compatible with a normal <code>URLClassLoader</code>, although its internal
  * behavior may be completely different.
  * <p>
@@ -188,8 +188,7 @@ public class WebappClassLoader
      * Use this variable to invoke the security manager when a resource is
      * loaded by this classloader.
      */
-    private boolean packageDefinitionEnabled =
-         System.getProperty("package.definition") == null ? false : true;
+    private final boolean packageDefinitionEnabled = System.getProperty("package.definition") != null;
     // END OF PE 4989455
 
     /**
@@ -293,8 +292,7 @@ public class WebappClassLoader
      * A list of read File and Jndi Permission's required if this loader
      * is for a web application context.
      */
-    private ConcurrentLinkedQueue<Permission> permissionList =
-        new ConcurrentLinkedQueue<Permission>();
+    private final ConcurrentLinkedQueue<Permission> permissionList = new ConcurrentLinkedQueue<Permission>();
     
     //holder for declared and ee permissions
     private PermsHolder permissionsHolder;
@@ -310,8 +308,7 @@ public class WebappClassLoader
      * The PermissionCollection for each CodeSource for a web
      * application context.
      */
-    private ConcurrentHashMap<String, PermissionCollection> loaderPC =
-        new ConcurrentHashMap<String, PermissionCollection>();
+    private final ConcurrentHashMap<String, PermissionCollection> loaderPC = new ConcurrentHashMap<String, PermissionCollection>();
 
     /**
      * Instance of the SecurityManager installed.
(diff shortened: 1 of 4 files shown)


Cousjava:
Jenkins test please

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -2147,11 +2147,10 @@ protected void nullInstance(Object instance) {
             return;
         }
         Field[] fields = instance.getClass().getDeclaredFields();
-        for (int i = 0; i < fields.length; i++) {
-            Field field = fields[i];
+        for (Field field : fields) {
             int mods = field.getModifiers();
             if (field.getType().isPrimitive()
-                    || (field.getName().indexOf("$") != -1)) {
+                    || (field.getName().indexOf('$') != -1)) {
```
svendiedrichsen: This could be achieved using _contains_

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -2100,11 +2102,10 @@ private final void clearReferencesStaticFinal() {
             if (clazz != null) {
                 try {
                     Field[] fields = clazz.getDeclaredFields();
-                    for (int i = 0; i < fields.length; i++) {
-                        Field field = fields[i];
+                    for (Field field : fields) {
                         int mods = field.getModifiers();
                         if (field.getType().isPrimitive()
-                                || (field.getName().indexOf("$") != -1)) {
+                                || (field.getName().indexOf('$') != -1)) {
```
svendiedrichsen: This could be achieved using _contains_

nucleus/deployment/common/src/main/java/com/sun/enterprise/deployment/deploy/shared/InputJarArchive.java
```diff
@@ -311,20 +310,13 @@ protected static JarFile getJarFile(URI uri) throws IOException {
         JarFile jf = null;
         try {
             File file = new File(uri);
-            if (file.exists()) {
-                jf = new JarFile(file);
-            }
+            jf = new JarFile(file);
```
svendiedrichsen: Does the removal of the existance check save the time?

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
```diff
@@ -571,7 +574,9 @@ private void checkManifest(JarFile jar, File file) throws IOException {
 
                 // add to class path of this class loader
                 try {
-                    appendURL(newFile);
+                    if (newFile.exists()) {
```
svendiedrichsen: Funny. Here you are adding the existance check which you have removed below.

svendiedrichsen:
What's your opinion about why this is speeding up 9%. I couldn't spot any _major_ differences.

nucleus/deployment/common/src/main/java/com/sun/enterprise/deployment/deploy/shared/InputJarArchive.java
Cousjava: Yes, since the jar will exist, unless deleted by the user while deployment is ongoing, this is unlikely and so if this happens then the exception is better.

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
Cousjava: This prevents adding jars to the classpath that don't exist, so the above check isn't needed.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
Cousjava: indexOf(char) is faster than contains, I did some benchmarking to see what was the faster method.
```
Benchmark                                          Mode  Cnt     Score      Error  Uni
IndexOfAndContainsPerfTest.containsString          avgt   30    17.515 ±    0.249  ns/op
IndexOfAndContainsPerfTest.containsStringIndirect  avgt   30    19.722 ±    0.187  ns/op
IndexOfAndContainsPerfTest.indexOfChar             avgt   30    15.080 ±    0.199  ns/op
IndexOfAndContainsPerfTest.indexOfCharIndirect     avgt   30    16.278 ±    0.591  ns/op
IndexOfAndContainsPerfTest.indexOfString           avgt   30    17.334 ±    0.326  ns/op
IndexOfAndContainsPerfTest.indexOfStringIndirect   avgt   30    20.242 ±    0.340  ns/op
IndexOfAndContainsPerfTest.indexSplit              avgt   30   149.581 ±    7.285  ns/
```

PR #4274, opened 2019-10-16
PAYARA-4193 Monitoring Console Various Improvments

jbee:
### Summary
Improvements include:

* Widget have dedicated unit
* Axis labelling with units
* Possibility to set axis min/max value
* Presets for Core (with 6th HTTP Threadpool count), HTTP and Request Tracing
* Possibility to set waterline vale (baseline, limit)
* Possibility to set alarming and critical thresholds
* Request statistics for individual web-applications is collected
* Possibility to set status texts (used in presets to help user to understand that some series require configuration)
* reset pages are always added when missing
* web resources organised on subfolders (`/js/`, `/css/`, ...)

### Note for Reviewers
Note that `monitoring-console.js` is a generated merged file of the other JS files and does not need individual review.

### Testing
1. To test the server needs to be build, installed and MC webapp in `appserver/monitoring-console/webapp/` be deployed on the installation.
2. When started open the app in the browser at http://localhost:8080/monitoring-console-webapp/

You should see a view similar to the below:
PAYARA-4193-mc-improvments

Some points to note comparing to the unimproved version before:
* web based metrics need to be enabled as they point out by a message
* HTTP based metric is available but most likely isn't yet enabled for you (similar message)
* _server_ is now given as _DAS_ in legend (also note tooltip)
* main (current value) line is now thicker for better visibility in monitor screen scenario
* x-axis has labels with unit, %, ms, or count /s
* y-axis has labels from _now_ (right) to _x s ago_ (x usually 60) on left
* legend now has value with unit
* legend show values _per second_ as an average of last few values (otherwise it could show zero at point of update even though in between values were non-zero which seams wrong)
* legend now distinguishes between _No Data_ and _Lost Connection_ (undeploy webapp to see)
* _CPU Usage_ and _Heap Usage_ have additional red and yellow lines indicating the set alarming and critical level (see _Decorations_ in image below)

When selecting a widget by clicking on the metric name in the caption the side panel shows its configuration similar to seen below:
PAYARA-4193-mc-improvments-side-panel

Points to note about the side panel in general:
* configuration is now organised more structured having multiple inputs for a row where it makes sense
* no stacked groupings are used any more

Points to note about the _Data_ box:
* a _Unit_ can be chosen controlling what the data means
* a _min_ and _max_ value for _Y-Axis_ can be set
* _X-Axis_ labelling can be turned on/off

Points to note about the _Decorations_ box:
* Decorations are client side visuals added to a chart entered in the charts unit, e.g. % or ms
[...]

PR #4278, opened 2019-10-18
PAYARA-4121 Fixed Updating of Flashlight Class Instrumentation

jbee:
### Symthom
When un-deploying and re-deploying an application the web and HTTP statistics would no longer update.

### Finding 
The reason for updates no longer reaching the statistics counters was that the connection done via class instrumentation was cut when un-deploying an app. This occurred because un-deploying an application does disable statistics providers specifically responsible for the application which wrongly triggered a removal of the class instrumentation for the probe provider class connecting the events to the statistics counters. So instead of just disconnecting the application specific statistics all statistics based on a particular class were disabled. Since the overall statistics for HTTP and web use the same class as the application specific ones this also disabled the overall statistics.

### Solution
Because the trigger for changes of class instrumentation was a static call in the innermost part of the logic there was no clear way to only trigger updates to the class instrumentation after the logic is done updating the changed probes and invokers. Another mistake was that disabling a invoker caused disabling of the class even though there might be other invokers and other probes for the same provider class. This is why both `transform` and `untransform` became just `update`. If updating would require to update the instrumentation or remove it is now evaluated based on the probes and their active invokers. If there are active invokers and the probe is enabled the instrumentation is kept and updated. Methods with no invokers are no longer instrumented.

To avoid running the update for each trigger call, which is each method times the statistics provider instances, the trigger only flags for update and a asynchronous daemon thread is used to run the actual updates with a small delay. By granting a delay the logic causing multiple triggers is likely to have completed and flagged all needed updates before the thread actually performs them. 

During shutdown the thread ends from interruption and does not perform (most of) the triggered transformation that a shutdown normally causes.

### Note to the Reviewer
Main change is in `ProbeProviderClassFileTransformer`. Besides that I only added some missing generics and made some hash maps into concurrent ones while looking for the problem.

Note that `synchronized` was largely removed where present. Synchronisation is done by now using a concurrent map and atomics. To allow reclaim of key class key was changed to `String` which alled use of concurrent map. Other methods no longer need synchronisation because only the updater thread will call them.

This also fixes PAYARA-1285 which should be checked with the reproducer given in the ticket.

### Testing
Easiest is to test this after #4274 is merged since it will be easy to verify in the monitoring console that un-deploying and re-deploying an application does no longer cause the request/sec to zero out.

With MC started enable web and HTTP monitoring until you see some request/sec in the _Core_ view of MC. Redeploy MC and check that the requests/sec are still above 0, most likely 0.5 due to data polling every 2secs.

Testing should also check that enabling and disabling the web or HTTP monitoring has the expected effect. Again this is verified easiest after #4274 already being in master. 

To check that instrumentation no longer is added or removed multiple times from provider classes I put a log output on INFO level in both cases of this if-statement https://github.com/payara/Payara/pull/4278/files#diff-16bec2e5206edbe4e95b050af4617992R193. Note that occasionally it is still possible that same class is handled twice due to unlucky async timing but not more often then that.

### DIFF
--- a/nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
+++ b/nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
@@ -51,13 +51,13 @@ import org.glassfish.external.probe.provider.PluginPoint;
 import org.glassfish.external.probe.provider.StatsProviderInfo;
 
 public class StatsProviderRegistry {
-    private final Map<String, List<StatsProviderRegistryElement>> configToRegistryElementMap = new HashMap();
-    private final Map<Object, StatsProviderRegistryElement> statsProviderToRegistryElementMap = new HashMap();
+    private final Map<String, List<StatsProviderRegistryElement>> configToRegistryElementMap = new ConcurrentHashMap<>();
+    private final Map<Object, StatsProviderRegistryElement> statsProviderToRegistryElementMap = new ConcurrentHashMap<>();
     private boolean isAMXReady = false;
     private boolean isMBeanEnabled = true;
     
     static final String[] DEFAULT_CONFIG_LEVELS = new String[] {"LOW","HIGH"};
-    protected static final Map<String, Integer> configLevelsMap = new ConcurrentHashMap();
+    protected static final Map<String, Integer> configLevelsMap = new ConcurrentHashMap<>();
 
     public StatsProviderRegistry(MonitoringRuntimeDataRegistry mrdr) {
         for (int i = 0; i < DEFAULT_CONFIG_LEVELS.length; i++) {
@@ -84,7 +84,7 @@ public class StatsProviderRegistry {
             List<StatsProviderRegistryElement> spreList = configToRegistryElementMap.get(configStr);
             spreList.add(spre);
         } else {
-            List<StatsProviderRegistryElement> spreList = new ArrayList();
+            List<StatsProviderRegistryElement> spreList = new ArrayList<>();
             spreList.add(spre);
             configToRegistryElementMap.put(configStr, spreList);
         }
--- a/nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
+++ b/nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
@@ -118,7 +118,8 @@ public class StatsProviderRegistry {
     }
 
     List<StatsProviderRegistryElement> getStatsProviderRegistryElement(String configElement) {
-        return (this.configToRegistryElementMap.get(configElement));
+        List<StatsProviderRegistryElement> plain = configToRegistryElementMap.get(configElement);
+        return plain == null ? Collections.emptyList() : new ArrayList<>(plain);
     }
 
     Collection<StatsProviderRegistryElement> getSpreList() {
--- a/nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
+++ b/nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
@@ -52,6 +52,7 @@ import java.lang.reflect.Method;
 import java.security.ProtectionDomain;
 import java.util.logging.Logger;
 import java.util.logging.Level;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -115,9 +116,13 @@ public class ProbeProviderClassFileTransformer implements ClassFileTransformer {
         Thread t = new Thread(() -> {
             try {
                 while (true) {
-                    Thread.sleep(200); // wait first so triggering calls flag first before the update runs 
-                    for (ProbeProviderClassFileTransformer transformer : instances.values()) {
-                        if (transformer.flaggedForUpdate.get()) {
+                    Thread.sleep(200); // wait first so triggering calls flag first before the update runs
+                    Iterator<ProbeProviderClassFileTransformer> iter = instances.values().iterator();
+                    while (iter.hasNext()) {
+                        ProbeProviderClassFileTransformer transformer = iter.next();
+                        if (transformer.providerClassRef.get() == null) {
+                            iter.remove();
+                        } else if (transformer.flaggedForUpdate.get()) {
                             transformer.runUpdate();
                         }
                     }
--- a/nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
+++ b/nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
@@ -53,9 +53,9 @@ import java.security.ProtectionDomain;
 import java.util.logging.Logger;
 import java.util.logging.Level;
 import java.util.Map;
-import java.util.Set;
-import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.glassfish.flashlight.FlashlightLoggerInfo;
 import static org.glassfish.flashlight.FlashlightLoggerInfo.*;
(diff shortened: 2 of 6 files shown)


nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
```diff
@@ -51,13 +51,13 @@
 import org.glassfish.external.probe.provider.StatsProviderInfo;
 
 public class StatsProviderRegistry {
-    private final Map<String, List<StatsProviderRegistryElement>> configToRegistryElementMap = new HashMap();
-    private final Map<Object, StatsProviderRegistryElement> statsProviderToRegistryElementMap = new HashMap();
+    private final Map<String, List<StatsProviderRegistryElement>> configToRegistryElementMap = new ConcurrentHashMap<>();
+    private final Map<Object, StatsProviderRegistryElement> statsProviderToRegistryElementMap = new ConcurrentHashMap<>();
```
jbee: Just a concurrency fix that most likely is needed.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -79,141 +78,94 @@
...
-        try {
-            instrumentation.retransformClasses(providerClass);
-        }
-        catch (UnmodifiableClassException e) {
-            logger.log(Level.WARNING, RETRANSFORMATION_ERROR, e);
+    private boolean hasEnabledInvokers() {
+        for (FlashlightProbe probe : probes.values()) {
+            if (probe.isEnabled() && probe.getInvokerCount() > 0) {
+                return true;
+            }
```
jbee: Instead of trying to maintain a "dirty" flag telling us if instrumentation needs to be updated the decision is now made based on the probe state. Anything else would cause us to remove instrumentation when there are active invokers waiting to be called.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -240,37 +190,35 @@ final synchronized void untransform() {
                 return null;
             }
 
-            if (enabled) {
+            if (hasEnabledInvokers()) {
                 // we still need to write out the class file in debug mode if it is
                 // disabled.
-                cw = new ClassWriter(ClassWriter.COMPUTE_FRAMES + ClassWriter.COMPUTE_MAXS);
+                ClassWriter cw = new ClassWriter(ClassWriter.COMPUTE_FRAMES + ClassWriter.COMPUTE_MAXS);
```
jbee: `cw` could be inlined.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -240,37 +190,35 @@ final synchronized void untransform() {
...
-                  count++;
-                }
-                ret = null;
-                Log.fine("untransformed", providerClass.getName());
+            if (_debug) {
+              ProbeProviderClassFileTransformer.writeFile(className.substring(className.lastIndexOf('/') + 1)+"supplied_"+count, classfileBuffer);
+              count++;
             }
+            Log.fine("untransformed", providerClass.getName());
+            return null;
```
jbee: FYI: `null` means transform back to the class bytes as given in the original file.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -240,37 +190,35 @@ final synchronized void untransform() {
...
         catch (Exception ex) {
             logger.log(Level.WARNING, REGISTRATION_ERROR, ex);
+            return null;
         }
-
-        return ret;
     }
 
-    private synchronized void addTransformer() {
+    private void addTransformer() {
```
jbee: Since all updates are run by the same thread no synchronisation is needed any more.

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -282,28 +230,18 @@ private static String makeKey(String name, String desc) {
...
-            fos = new FileOutputStream(new File(dir, name + ".class"));
-            fos.write(data);
+        File installRoot = new File(System.getProperty(SystemPropertyConstants.INSTALL_ROOT_PROPERTY));
+        File dir = new File(installRoot, "flashlight-generated");
+        if (!dir.isDirectory() && !dir.mkdirs()) {
+            logger.log(Level.WARNING, WRITE_ERROR, new RuntimeException("Can't create directory: " + dir));
+            return;
         }
-        catch (Throwable th) {
+        try (
```
jbee: Just using try-with-resources instead

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -323,7 +261,7 @@ public MethodVisitor visitMethod(int access, String name, String desc, String si
             MethodVisitor mv = super.visitMethod(access, name, desc, signature, exceptions);
 
             FlashlightProbe probe = probes.get(makeKey(name, desc));
-            if (probe != null) {
+            if (probe != null && probe.isEnabled() && probe.getInvokerCount() > 0) {
```
jbee: Important change to the way this class works: We decide if a method gets instrumented based on the state of the probe. As all state changes should trigger flag for update this is now safe to do.

jbee:
jenkins test please

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -79,141 +78,94 @@
...
     }
 
-    synchronized void addProbe(FlashlightProbe probe) throws NoSuchMethodException {
-        Method m = getMethod(probe);
-        probes.put(probe.getProviderJavaMethodName() + "::" + Type.getMethodDescriptor(m), probe);
-        
-        // probes can be added piecemeal after the initial transformation is done, flagging when probes are added to detect that
-        allProbesTransformed = false;
+    private void update() {
+        faggedForUpdate.set(true);
```
AlanRoth: Small typo

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -79,141 +78,94 @@
...
+        // make sure a updater is running
+        updater.updateAndGet(thread -> thread != null && thread.isAlive() ? thread : newUpdater());
     }
 
-    final synchronized void transform() {
-
-        Class providerClass = providerClassRef.get();
+    private final void runUpdate() {
+        faggedForUpdate.set(false);
```
AlanRoth: Small typo

jbee:
jenkins test please

nucleus/flashlight/framework/src/main/java/org/glassfish/flashlight/impl/client/ProbeProviderClassFileTransformer.java
```diff
@@ -79,141 +78,94 @@
...
+    private final Map<String, FlashlightProbe> probes = new ConcurrentHashMap<>();
+    private AtomicBoolean flaggedForUpdate = new AtomicBoolean(false);
     private boolean transformerAdded = false;
     private int count = 0;  // Only used for debug so we can look at the before/after class dumps for each iteration
     ///////////////  static variables  //////////////////
-    // weak map here so the classes don't prevent app classloaders from being reclaimed
-    private static Map<Class, ProbeProviderClassFileTransformer> instances =
-        new WeakHashMap<Class, ProbeProviderClassFileTransformer>();
+    // uses String as key so no reference to Class is held => allow being reclaimed
+    private static Map<String, ProbeProviderClassFileTransformer> instances = new ConcurrentHashMap<>();
```
jGauravGupta: Suspecting memory leak, @jbee may you share more detail.

jbee: I think you are technically right. The `ProbeProviderClassFileTransformer` instance and the `String` for the key are never cleaned even if the `Class` they are fore got reclaimed. I did not address it since there are so few classes actually used and the chance of reclaiming is quite superficial but then this as far as I know is a feature accessible to applications so theoretically there could be many such classes. I am going to add a cleanup to the worker thread loop.

jbee:
jenkins test please

jbee:
jenkins test please

nucleus/admin/monitor/src/main/java/org/glassfish/admin/monitor/StatsProviderRegistry.java
```diff
@@ -118,7 +118,8 @@ StatsProviderRegistryElement getStatsProviderRegistryElement(Object statsProvide
     }
 
     List<StatsProviderRegistryElement> getStatsProviderRegistryElement(String configElement) {
-        return (this.configToRegistryElementMap.get(configElement));
+        List<StatsProviderRegistryElement> plain = configToRegistryElementMap.get(configElement);
+        return plain == null ? Collections.emptyList() : new ArrayList<>(plain);
```
jbee: This is an unrelated fix for a problem I observed where the returned list is iterated and the iteration causes changes to the list causing a concurrent modification exception. Since the method is only used in this module by methods doing this kind of loop I concluded it is best to always return a copy that can be iterated without problem.

jbee:
@Pandrex247 might want to review this was well. Waiting for his feedback on it.

PR #4282, opened 2019-10-23
PAYARA-4130: Workaround hidden config service dependencies

pdudits:
# Description
This is a bug fix. 

`ConfigProviderResolverImpl` might provide instance to MP Config SPI at wrong moment, where its instance gets yet overwritten by instance of `ConfigProviderResolveSync` instance. This race condition manifests itself as `StackOverflowError` then, because Sync relied on Impl to set the instance.

This solution only masks a larger issue where various server internals cooperate with MP Config API to obtain configuration, however config impl is a service, that only works after proper initialization.

It would be best to use HK2 injection instead, but majority of the points of use are not HK2 services.

### Test suites executed

- Java EE7 Samples in `payara-micro-managed` profile, that manifested this behavior the most in nightly

### Testing Environment

Windows 10, Java 1.8.0_221, Maven 3.6.2


pdudits:
Jenkins test please

pdudits:
Jenkins test please

PR #4289, opened 2019-10-29
CUSTCOM-208 Suspended transaction should be resumed even when an exception is raised during commit()

albanf:
# Description
This fixes an issues when using @Transactional(Transactional.TxType.REQUIRES_NEW).

If an exception occurs during getTransaction().commit(), then the suspended transaction is not resumed.
The code flow should be similar to EJBContainerTransactionManager.postInvokeTx (resume always called, even when commit/rollback throws an exception).

# Testing

### Testing Performed
Testing was performed on an in-house application.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.cdi.transaction;
 
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
@@ -132,16 +132,16 @@ public class TransactionalInterceptorRequiresNew extends TransactionalIntercepto
                             "Managed bean with Transactional annotation and TxType of REQUIRES_NEW " +
                             "encountered exception during commit " + exception,
                             exception);
-                }
-
-                if (suspendedTransaction != null) {
-                    try {
-                        getTransactionManager().resume(suspendedTransaction);
-                    } catch (Exception exception) {
-                        throw new TransactionalException(
-                                "Managed bean with Transactional annotation and TxType of REQUIRED " +
-                                "encountered exception during resume " + exception,
-                                exception);
+                } finally {
+                    if (suspendedTransaction != null) {
+                        try {
+                            getTransactionManager().resume(suspendedTransaction);
+                        } catch (Exception exception) {
+                            throw new TransactionalException(
+                                    "Managed bean with Transactional annotation and TxType of REQUIRED " +
+                                    "encountered exception during resume " + exception,
+                                    exception);
+                        }
                     }
                 }
             }
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
@@ -109,43 +109,60 @@ public class TransactionalInterceptorRequiresNew extends TransactionalIntercepto
             }
 
             Object proceed = null;
-
+            Exception exception = null;
             try {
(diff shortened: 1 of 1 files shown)


lprimak:
Does this need to be done in EJB as well or does it work correctly there?

albanf:
It works correctly with EJB

appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
```diff
@@ -132,16 +132,16 @@ public Object transactional(InvocationContext ctx) throws Exception {
...
-
-                if (suspendedTransaction != null) {
-                    try {
-                        getTransactionManager().resume(suspendedTransaction);
-                    } catch (Exception exception) {
-                        throw new TransactionalException(
-                                "Managed bean with Transactional annotation and TxType of REQUIRED " +
-                                "encountered exception during resume " + exception,
-                                exception);
+                } finally {
```
dmatej: SonarQube has a rule for this - this would mask the possible exception from the try block, because finally is executed always. See https://www.oracle.com/technical-resources/articles/java/trywithresources.html

albanf: The exception from proceed would also be masked, so I changed the flow to keep the original exceptions as suppressed, throwing the last exception.

pdudits:
Jenkins test please

Cousjava:
Jenkins test please

albanf:
@dmatej: will you have time do review the latest commits soon?

albanf:
Thanks for approving it.
Just a question then: did it pass the JTA TCK before? If it did, should an issue be reported in the JTA TCK to improve the test suite?

pdudits:
Yes, it did pass the TCK before, and it's one of the smaller suites, so likely does not contain a testcase for this. I'm not sure about correct repository for filing such issue, probably https://github.com/eclipse-ee4j/jta-api for the time being.

I yet need to ask you to do some legal bits as listed in [Contributing](https://github.com/payara/Payara/blob/master/.github/CONTRIBUTING.md#legal-bits): updating copyright year, and signing the CLA, if you haven't yet done so.

albanf:
I've updated the copyright and will let you know when I have returned the CLA.

pdudits:
Jenkins test please

dmatej:
Hi, how is it with CLA now?

albanf:
I had totally forgotten about it... I just sent it now!

smillidge:
I can confirm we have received it

albanf:
After some more testing, I don't think it's safe to use transactions on CDI beans.
We ran into several issues, and I think the reason is that in many places (JavaEETransactionManagerSimplified for instance), transactional resources are enlisted/delisted against the current "component invocation" (Servlet, session bean, etc), which in the case of a transactional CDI bean annotated with REQUIRES_NEW is not the same.

Some resources are registered incorrectly, and some resources are cleaned up twice.
We've seen issues when trying to use ActiveMQ instead of OpenMQ,
And when calling a bean annotated ``@Transactional(Transactional.TxType.SUPPORTS)`` from a JAX-RS resource, we've seen random errors like:
```
java.lang.IllegalStateException: Attempting to execute an operation on a closed EntityManager.
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.verifyOpen(EntityManagerImpl.java:2062)
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.close(EntityManagerImpl.java:1903)
	at com.sun.enterprise.container.common.impl.EntityManagerWrapper$NonTxEMCleaner.beforePostInvoke(EntityManagerWrapper.java:1241)
	at org.glassfish.api.invocation.InvocationManagerImpl.postInvoke(InvocationManagerImpl.java:189)
	at com.sun.web.server.J2EEInstanceListener.handleAfterEvent(J2EEInstanceListener.java:317)
	... 66 common frames omitted
Wrapped by: java.lang.RuntimeException: Exception during processing of event of type AFTER_SERVICE_EVENT for web module StandardEngine[glassfish-web].StandardHost[server].StandardContext[]
```

Feel free to merge this PR, but we will revert to session beans for the time being for anything transactional.

smillidge:
OK thanks for the analysis. It looks like we need to look at this in more detail.

dmatej:
OK, so let's raise the bar higher - automatic test covering related scenarios required ... perhaps including your issues with CDI, so we would be sure we are going in right direction.

lprimak:
Does this issue happen in general or only when suspended transactions happen?
I wouldn't think any of the TCK tests would pass if there is a general issue without suspended transactions.

Suspended transactions are not used very much in real world (as far as I know) and it's very possible to have bugs like this.

Despite all of this, I thin it's prudent to merge this PR because it does fix one of the problems.

fturizo:
As it's been over a year since this request was implemented, we'll close the PR without proceeding to merge it as there are many questions surrounding the proposed changes and they lack the necessary tests as well.

PR #4292, opened 2019-10-30
PAYARA-4147 MP Metrics now use double for MBean metrics instead of long

### LINKED ISSUE #4230: MicroProfile Metrics API returns no data for vendor:system_cpu_load PAYARA-4147
# Description #
----------

Since Payara Micro 5.193 the /metrics endpoint does not return data for the field vendor:system_cpu_load anymore. The value is always zero.

## Expected Outcome

Calling GET {baseUrl}/metrics/vendor returns the field vendor:system_cpu_load with a value > 0.

## Current Outcome

Calling GET {baseUrl}/metrics/vendor the field vendor:system_cpu_load is _always_ zero.

Also the naming of the field changed.
Before 5.193 it is vendor:system_cpu_load
Since 5.193 it is vendor_system_cpu_load
(not that the : was replaced by _)

Execute java -jar payara-micro-5.193.jar --deploy anyApplication.war
Visit http://localhost:8080/metrics/vendor

### Samples

Payara Micro 5.191
  vendor:system_cpu_load 0.15454493453568963
  
Payara Micro 5.192
  vendor:system_cpu_load 0.5629938578656155
  
Payara Micro 5.193
  vendor_system_cpu_load 0

## Environment ##

- **Payara Version**: 5.193
- **Edition**: Micro
- **JDK Version**: Oracle Jdk, OpenJdk (Version 1.8)
- **Operating System**: Windows, Linux

Cousjava:
# Description
This is a bug fix 

Fixes #4230 
CPU Load was being displayed as 0 as gauge was used a long which would round off the actual cpu usage which occurs in the range [0,1]

# Testing

### Testing Performed
Started domain and admin console. Went to localhost:4848/metrics and checked system_cpu_load gauge, with this PR is is non-zero.

### Test suites executed
- Cargo Tracker

### Testing Environment
Zulu JDK 1.8_222 on Ubuntu 19.04 with Maven 3.6.0

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanGuageImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanGuageImpl.java
@@ -47,7 +47,7 @@ import org.eclipse.microprofile.metrics.Gauge;
  * @author jonathan coustick
  * @since 5.193
  */
-public class MBeanGuageImpl implements Gauge<Long> {
+public class MBeanGuageImpl implements Gauge<Double> {
 
     private final MBeanExpression mBean;
 
@@ -56,7 +56,7 @@ public class MBeanGuageImpl implements Gauge<Long> {
     }
 
     @Override
-    public Long getValue() {      
-        return mBean.getNumberValue().longValue();
+    public Double getValue() {      
+        return mBean.getNumberValue().doubleValue();
     }
 }


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanGuageImpl.java
```diff
@@ -47,7 +47,7 @@
  * @author jonathan coustick
  * @since 5.193
  */
-public class MBeanGuageImpl implements Gauge<Long> {
+public class MBeanGuageImpl implements Gauge<Double> {
```
jbee: Doesn't this mean you change the type of all MBean gauges? Or is this just used by the metric you wanted to change?

jbee:
I suggest we revert this change since now all internal MP metrics are doubles which is not very user friendly either.

PR #4294, opened 2019-10-31
PAYARA-4236 Various UX improvements and small Additions

jbee:
### Summary
Changes include:

* browser refresh no longer resets the currently shown page
* toolbar icons made into a toolbar menu with text and icon that also reflects state changes
* adding widgets no longer adds them to first column but distributes them to fill all columns
* adding widgets now uses a 2-step selection: namespace, metric in namespace
* experimental _instances_ filter setting was removed
* request tracing details page always shows sidebar when opened
* _Core_ page now shows a newly added metric for HTTP thread pool usage
* _HTTP_ page now shows busy HTTP threads instead of pool size
* data refresh interval is now a persistent global setting
* adding and removing widgets updates sidebar settings properly
* graphs within widgets are now scaled to available space in the widget
* widget sizing is now "smarter": As long as number of rows leaves a meaningful widget height for each widget all rows are made to fit the screen - if that is not possible a height is chosen so that first N rows fit the screen (scrolling is needed)
* added new feature: page rotation (each page is shown for some time, then page automatically switches to next one)
* adopted colours to Payara design system
* page delete only available for pages without a preset, reset only available for pages with a preset

### Note to Reviewers
Remember that `monitoring-console.js` is generated and does not need to be reviewed.

### Testing Performed
Manual testing of:

* Create new page
* rename the new page
* add widget(s) to page
* remove widget(s) from page
* move widgets on page
* delete page
* switch page layout (number of columns)
* reset page
* pause data refresh
* unpause data refresh
* slow/fast/normal data refresh interval
* enable page rotation
* disable page rotation
* setting a low _alarming_ threshold (e.g. Heap Usage 10%) and set a status message, check message is displayed above 10%
* setting a low _critical_ threshold (e.g. Heap Usage 20%) and set a status message, check message is displayed above 20%
* enable/disable Monitoring config for _Web Container_ and/or _HTTP Service_ and check that status message is displayed on _Core_ page


jbee:
jenkins test please

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-components.js
```diff
@@ -221,65 +229,106 @@ MonitoringConsole.View.Components = (function() {
...
-      function onCreation(model) {
-         if (!model.text) {
-            return $('<div/>', {'class': 'Indicator', style: 'display: none;'});
-         }
-         let html = model.text.replace(/\*([^*]+)\*/g, '<b>$1</b>').replace(/_([^_]+)_/g, '<i>$1</i>');
-         return $('<div/>', { 'class': 'Indicator status-' + model.status }).html(html);
+    function onCreation(model) {
+      let attrs = { 'class': 'Menu' };
+      if (model.id)
+        attrs.id = model.id;
```
AlanRoth: Some of the if statements look short enough to fit onto a single line, which I personally find more readable, just another nitpick

jbee: I opted to not do single line ifs since I also opted to not always do brackets so I made this compromise. Overall I still in the process of finding good format rules for the JS so it is not entirely consistent either but this is good feedback I definitely will consider in the future.

PR #4297, opened 2019-11-02
PAYARA-4248 Log Expired certificates from cacerts.jks with level warning

rdebusscher:
# Description
This is a feature request. 

Currently expired certificates are logged with SEVERE Level which confuses users. Server starts fine and users currently see this expired certificates logging as cause of some other problem .

# Important Info

### Testing Performed
Verified the changed log message in server.log


MeroRai:
Jenkins test please

PR #4298, opened 2019-11-05
PAYARA-2598 Hook up Soteria Identity Stores to Payara Realms

jGauravGupta:
# Description
This is a feature request which provides an implementation of Identity Store mapped to realm.

Following annotations are used to trigger realm identity store and authentication mechanism implementation:

### Existing Realm instance:
If the realm already exists (e.g default realm file, certificate, etc) in the server then following annotation can be used to create Identity Store mapping by proving the realm name:

-     @RealmIdentityStoreDefinition
-     @RealmIdentityStoreDefinitions

`@RealmIdentityStoreDefinition` annotation is repeatable hence authentication can be performed across multiple realms.

### Dynamic Realm instance:
The new instance of realm is created and registered using `create-auth-realm` asadmin command and Realm instance is fetched from `RealmManager` to perform authentication.

-     @FileIdentityStoreDefinition
Creates a new file realm instance and register it in DAS.
-     @CertificateAuthenticationMechanismDefinition
 To configure the Client Certificate Authentication just by using the annotation on a CDI bean.
-     @CertificateIdentityStoreDefinition
Creates a new certificate realm instance and register it in DAS.
-     @PamIdentityStoreDefinition
Creates a new pam realm instance and register it in DAS.
-     @SolarisIdentityStoreDefinition
Creates a new solaris realm instance and register it in DAS.

**Note:**  If `@CertificateIdentityStoreDefinition` annotation is defined then `@CertificateAuthenticationMechanismDefinition` annotation must also be defined on CDI bean.

**Limitation:**  `@CertificateIdentityStoreDefinition` annotation can only be used with other identity stores by providing the custom `HTTPAuthenticationMechanism` implementation which must be compatible across all used IdentityStores.

# Testing

### New tests
- Payara Samples testcase for existing and dynamic Realm instance.

### Testing Performed
- Manually Tested File Realm, Certificate Realm, JDBC Realm, and Pam Realm.

### Test suites executed
- Payara Samples

### Testing Environment
JDK 1.8.0_172, Windows 10, Ubuntu 18.04.3 LTS (Pam realm manual testing), Oracle Solaris 11.4 (Solaris realm mock manual test)


jGauravGupta:
jenkins test please

ggam:
Pretty cool. What happens when an application that created a realm is undeployed or redeployed? Is the realm removed/recreated?

jGauravGupta:
Hi @ggam,
On the undeployment of application, the newly created realm is not removed. And on the redeployment of application, Identity Store checks for the existing realm before creating the new realm instance.

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

ggam:
>  On the undeployment of application, the newly created realm is not removed. And on the redeployment of application, Identity Store checks for the existing realm before creating the new realm instance.

You probably have already discussed this stuff but what happens when redeploying an application with a modified realm? Would you need to remove the realm before deploying the application again? how would that work in case two applications use the same realm name?

Crazy idea: when using `@DatasourceDefinition`, `@MailSessionDefinition` and the likes, you define a JNDI namespace that can be application or server scoped. Redeploying always regenerates the resources. What about doing the same for realms? Behaviour would just be the same as for other resources, so it's predictable.

jGauravGupta:
> redeploying an application with a modified realm

It depends on which attribute/property modified:
- If name modified and no realm already exists with such name then-new realm created.
- If annotation changed (e.g FileIdentityStoreDefinition -> PamIdentityStoreDefinition) but the name remains the same and the realm already exists with the same name but different realm class then exception thrown on deployment.
- Other properties modification is ignored if the realm already exists with the correct realm class.


> Would you need to remove the realm before deploying the application again?

Not required, but If the realm already exist with the same name but different realm class then an exception is thrown and deployment fails.


> how would that work in case two applications use the same realm name?

The realm is managed by `RealmsManager` (server scoped), the same realm can be used across multiple applications.

> application or server scoped
Already did some POC on the application-scoped realms which are created on each deployment but the con is they cannot be managed from asadmin commands (or console) e.g `create-file-user` etc.

arjantijms:
Nice work @jGauravGupta :) Happy to see this issue, which internally took quite some preparation, has now finally been picked-up.

One remark, I originally thought of creating a mapping to just whatever realm has been configured as the default already. It was even planned for being put in the spec, but we ran out of time.

This is essentially what `javax.security.auth.message.callback.PasswordValidationCallback` does in JASPIC, though in JASPIC it has the limitation to being strongly restricted to username/password.

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/identitystores/PamRealmIdentityStore.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public static final Class<PamLoginModule> REALM_LOGIN_MODULE_CLASS = PamLoginModule.class;
+
+    public void init(PamIdentityStoreDefinition definition) {
+        configuration = PamRealmIdentityStoreConfiguration.from(definition);
+
+        try {
+            if (!Realm.isValidRealm(configuration.getName())) {
+                Properties props = new Properties();
+                props.put(JAAS_CONTEXT, configuration.getName());
```
pdudits: Same argument - default should be... oh there isn't predefined Login Module for pam, is it?

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/identitystores/SolarisRealmIdentityStore.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public static final Class<SolarisLoginModule> REALM_LOGIN_MODULE_CLASS = SolarisLoginModule.class;
+
+    public void init(SolarisIdentityStoreDefinition definition) {
+        configuration = SolarisRealmIdentityStoreConfiguration.from(definition);
+
+        try {
+            if (!Realm.isValidRealm(configuration.getName())) {
+                Properties props = new Properties();
+                props.put(JAAS_CONTEXT, configuration.getName());
```
pdudits: Default of `solarisRealm` more explicit name for jaas context in config.

appserver/tests/payara-samples/samples/realm-identity-stores/src/main/webapp/WEB-INF/beans.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pdudits: `beans.xml` should not be necessary (with scope annotations placed on relevant classes).

appserver/tests/payara-samples/samples/realm-identity-stores/src/main/webapp/WEB-INF/glassfish-web.xml
```diff
@@ -0,0 +1,17 @@
...
+
+    <security-role-mapping>
+        <role-name>ROLE_ADMIN</role-name>
+        <group-name>ROLE_ADMIN</group-name>
+    </security-role-mapping>
+    
+    <security-role-mapping>
+        <role-name>ROLE_USER</role-name>
+        <group-name>ROLE_USER</group-name>
+    </security-role-mapping>
```
pdudits: Also default mapping should not be necessary, as it is on by default.

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/identitystores/PamRealmIdentityStore.java
pdudits: Oh, there is - `pamRealm`

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/RealmUtil.java
```diff
@@ -0,0 +1,121 @@
...
+        return t.negate();
+    }
+
+    static void createAuthRealm(String name, String realmClass, String loginModule, Properties props) {
+        ServiceLocator serviceLocator = Globals.getDefaultHabitat();
+        InserverCommandRunnerHelper commandRunnerHelper = serviceLocator.getService(InserverCommandRunnerHelper.class);
+        EmbeddedSystemAdministrator administrator = serviceLocator.getService(EmbeddedSystemAdministrator.class);
+        ParameterMap parameters = new ParameterMap();
+        parameters.insert("authrealmname", name);
+        parameters.insert("property", props.entrySet().stream().map(Object::toString).collect(joining(":")));
```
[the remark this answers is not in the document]
jGauravGupta: Thanks, Checked asadmin command `create-auth-realm --property` attribute, escaping is needed for `:` & `=`.

jGauravGupta:
jenkins test please

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/RealmUtil.java
```diff
@@ -107,15 +106,25 @@ static void createAuthRealm(String name, String realmClass, String loginModule,
...
-            throw new IdentityStoreException("Error in creating Auth realm: " + name);
+        if (outreport.getActionExitCode() == ActionReport.ExitCode.FAILURE) {
+            throw new IllegalStateException("Error in creating Auth realm: " + name);
         }
     }
+    
+    static String escapeRealmProperty(String component) {
+        return component
+                .replaceAll("\\:", "\\\\:")
+                .replaceAll("\\=", "\\\\=");
```
pdudits: Are you sure no handling is needed for backslash? so when my password is `a\password`, won't it break?

appserver/tests/payara-samples/samples/realm-identity-stores/src/test/java/fish/payara/samples/realm/identity/store/ExistingRealmIdentityStoreDefinitionTest.java
```diff
@@ -89,7 +87,6 @@ public static WebArchive createDeployment() {
                         PersonController.class,
                         AuthoritiesConstants.class
                 )
-                .addAsWebInfResource(new File(WEBAPP_SOURCE, "WEB-INF/glassfish-web.xml"))
                 .addAsWebInfResource(EmptyAsset.INSTANCE, ArchivePaths.create("beans.xml"));
```
pdudits: Even empty beans.xml is superflous in my opinion

jGauravGupta: No classes added to test war with CDI scoped annotation.

appserver/security/realm-stores/src/main/java/fish/payara/security/realm/RealmUtil.java
jGauravGupta: Yes, I have checked, the property attribute only requires handling for colon and equal character.

PR #4300, opened 2019-11-06
PAYARA-4250 Prevent log spamming

jbee:
### Summary
The goal of this change is to prevent potential log spamming caused by the collection of metrics every second in case exception is thrown by a source. This is achieved in two ways:

* only log failing sources once (until they not fail any more)
* only collect metrics when monitoring service is enabled

### Solution
As discussed shortly during standup the solution to enable/disable via the existing setting was chosen to keep things simple both for the user and development.

The periodical collection task is submitted to the scheduler when monitoring is enabled initially or when a change event for the property changes is state to enabled. The task is cancelled should the state change to disabled.

### Testing Performed
With deployed MC the configuration _Monitoring_ => _Monitoring Service: Enabled_ is checked. The change can be observed on the consoles _Core_ page. Metrics start to not be for the recent 60s and eventually after 1 minute the "no data" state is reached. This was tested also for extra instances.
The change can also be observed by the info log messages when collection is started and stopped but this does not reflect if the job actually starts or stops. If the metrics disappear this means the job actually is no longer active.

To test if logging really logs failing sources only once a simple dummy class was added temporary that simulated a mostly failing source:
```java
@Service
class FlakyDataSource implements MonitoringDataSource {
  int x = 0;
  public void collect(MonitoringDataCollector collector) {
    if (++x % 10 != 0) {
      throw new RuntimeException();
    }
  }
}
```
(given in case reviewer wants to redo this; can be added to any module where HK2 `@Service` will be picked up). The failing source can be seen as metric: _Monitoring Console Internals_ => _Failed Collection Count_. To verify that the log is only written once after collection was successful enable `FINE` log level for the LOGGER `"monitoring-console-core"` or put a breakpoint at the log statement.

### Note for the Reviewer
Considering the changes done by #4303 it might be more logical to connect the data collection to a new separate configuration field that reflects the deployed state of the monitoring application so that collection is enabled/disabled in union with the application being deployed/undeployed. However, we might even want to split this into two (new) flags because it can make sense to collect data without having the application deployed to have a history available in case the application is turned on.
This can be discussed and potentially combined accordingly after this PR and #4303 are merged.

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
@@ -40,10 +40,13 @@
 
 package fish.payara.monitoring.store;
 
+import static java.lang.Boolean.parseBoolean;
 import static java.util.Arrays.asList;
 import static java.util.Arrays.copyOf;
 import static java.util.Collections.emptyList;
+import static org.jvnet.hk2.config.Dom.unwrap;
 
+import java.beans.PropertyChangeEvent;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -54,7 +57,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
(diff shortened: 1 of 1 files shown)


appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
```diff
@@ -101,6 +105,8 @@
      * The topic name used to share data of instances with the DAS.
      */
     private static final String MONITORING_DATA_TOPIC_NAME = "payara-monitoring-data";
+    private static final Logger LOGGER = Logger.getLogger("monitoring-console-core");
+    private final Set<String> sourcesFailingBefore = new HashSet<>();
```
jbee: The reasoning for using a `HashSet` and not a concurrent one was that the collection is run by a single worker thread. As we use a pool I later thought we have no guarantees that this is the same thread. Might be better to use concurrent nonetheless.

jbee:
jenkins test please

Cousjava:
How about not starting the worker thread unless monitoring is enabled?

jbee:
jenkins test please

jbee:
@Cousjava I now listen to config change events to start and stop the job that does the collection. (See also updated description)

MarkWareham:
Jenkins test please

PR #4301, opened 2019-11-06
PAYARA-4116: Properly shutdown server on startup failure

pdudits:
# Description
This is a bug fix 

Since cca summer 2017 server would hang when startup failure occurred, thread dump would show one or two threads hanging in 
```
  java.lang.Thread.State: TIMED_WAITING
	  at java.lang.Thread.sleep(Thread.java:-1)
	  at com.sun.enterprise.v3.admin.StopServer.doExecute(StopServer.java:74)
	  at com.sun.enterprise.v3.admin.StopDomainCommand.execute(StopDomainCommand.java:96)
	  at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:558)
	  at com.sun.enterprise.v3.admin.CommandRunnerImpl$2$1.run(CommandRunnerImpl.java:554)
	  at java.security.AccessController.doPrivileged(AccessController.java:-1)
	  at javax.security.auth.Subject.doAs(Subject.java:360)
	  at com.sun.enterprise.v3.admin.CommandRunnerImpl$2.execute(CommandRunnerImpl.java:553)
	  at org.glassfish.api.AsyncImpl$1$1.run(AsyncImpl.java:76)
```

This posed few interesting questions:
* Why there is no instance of `GlassFish` available and this command is stuck in endless loop?
* Why is there two of them?
* Why `AppServerStartup` delegates to command, which then delegates to `GlassFish` which then delegates to `AppServerStartup`?

The answers to those are:

`GlassFish` impplementations (that differ by runtime, e. g. server, embedded, micro) register themselves after kernel starts. It used to be an invariant, that kernel never throws an exception, so the registration always happened. But since 1d7999e kernel start may throw, and `StopServer` is out of luck.

There are multiple paths to `AppServerStartup#shutdown`. Failure during startup happens to hit more of them.

The mechanism of delegating shutdown to admin command, which happens to delegate back, comes from early design of Glassfish V3, where shutdown sequence was placed directly in the command. But this code very quickly moved next to startup code, and this delegation is superflous.

Therefore these changes:

* If kernel throws exception, immediately dispose the instance
* Stop kernel directly rather than delegating to stop command
* Guarantee thread safety of shutdown

Shutdown may be invoked synchronously as reaction to failed proceed, or asynchronously from RunLevelListener - it's hard to tell what the timing guarantees are, especially after startup completes.
Therefore a state machine in addition to synchronicity of `stop` now guards, that shutdown sequence in invoked only once.

PayaraExecutorService now may shutdown before it initialized.

# Testing

### Testing Performed
Injecting failures via ByteMan to Init and Startup Runlevel services, emulating startup errors common with 5.193; Injecting failures to grizzly server binds to emulate blocked ports. Server now consistenly shuts down.

### Test suites executed

- Payara samples in remote mode

### Testing Environment

Zulu JDK 1.8.0_192, Windows 10,  Maven 3.6.2

### DIFF
--- a/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
+++ b/nucleus/core/bootstrap/src/main/java/com/sun/enterprise/glassfish/bootstrap/GlassFishImpl.java
@@ -80,7 +80,16 @@ public class GlassFishImpl implements GlassFish {
             throw new IllegalStateException("Already in " + status + " state.");
         }
         status = Status.STARTING;
-        gfKernel.start();
+        try {
+            gfKernel.start();
+        } catch (RuntimeException | Error t) {
+            dispose();
+            if (t.getCause() instanceof GlassFishException) {
+                throw (GlassFishException)t.getCause();
+            } else {
+                throw t;
+            }
+        }
         status = Status.STARTED;
     }
 
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppServerStartup.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/AppServerStartup.java
@@ -60,6 +60,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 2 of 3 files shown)


pdudits:
jenkins test please

PR #4303, opened 2019-11-07
PAYARA-4245 Packaging MC with Payara (Server)

jbee:
### Problem
Package MC webapp in Payara distribution so it can be deployed through some sort of API.

### Solution
Using maven `unpack` unpacks the webapp war to `glassfish/lib/install/applications` (same folder as admingui). A new asadmin-command uses auto-deploy and undeploy to deploy the unpacked app from that folder when MC is enabled or disabled.

Note that this is a minimal solution that intends to be extended with a GUI in the admingui that uses the provided commands to enable/disable MC.

### Testing
Build and install the server. Enter asadmin command line:
```
asadmin> start-domain
asadmin> set-monitoring-console-configuration --enabled=true
```
MC should now run on `localhost:8080/monitoring-console/`
```
asadmin> set-monitoring-console-configuration --enabled=false
```
MC should no longer run on localhost.

### DIFF
--- a/appserver/distributions/payara/pom.xml
+++ b/appserver/distributions/payara/pom.xml
@@ -90,32 +90,6 @@
                     </execution>
                 </executions>
             </plugin>
-
-            <plugin>
-              <groupId>org.apache.maven.plugins</groupId>
-              <artifactId>maven-dependency-plugin</artifactId>
-              <version>2.8</version>
-              <executions>
-                <execution>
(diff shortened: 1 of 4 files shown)


jbee:
I am in the process of doing a packaging that can do auto-deploy via command so this should not be merged yet as we might replace it.

jbee:
jenkins test please

PR #4304, opened 2019-11-07
PAYARA-4256 MP Metrics are now always unregiestred after application undeployment

Cousjava:
# Description
This is a bug fix.

When undeploying an application that has capital letters in its name i.e. SNAPSHOT then it fails to be deregistered when it is removed.

# Important Info

# Testing

### New tests
New unit test added, see PR

### Testing Performed
Deployed the test application for PAYARA-4174, went to localhost:8080/metrics then undeployed the application and went to localhost:8080/metrics again. Before PR error thrown, afterwards no error.

### Test suites executed

### Testing Environment
"Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.0"


Cousjava:
Jenkins test please

Cousjava:
Jenkins test please

PR #4305, opened 2019-11-07
PAYARA-4171 Monitoring data of an Instance is not visible if it is in a Deployment Group

MeroRai:
# Description
This is a bug fix to an issue where monitoring data of an Instance was not visible if it was in a Deployment Group.

# Important Info
Although the issue was identified because of the monitoring issue, the underlying issue was caused by a command not being replicated to the instances in a deployment group. 

# Testing
Created a deployment group with an instance, enabled the monitoring service and checked if the monitoring data is available. 

### Testing Environment

# Documentation
No required


AlanRoth:
jenkins test please

PR #4308, opened 2019-11-08
PAYARA-4174 MP Metrics in web profile do not throw an exception on custom metrics

Cousjava:
# Description

This is a bug fix .

The /metrics endpoint is loaded as default-web-module which is a different context to the
user's application. In Web profile previously TCCLSingleton is used to hold the classloader,
which does not allow cross-application sharing. In Full profile, and now Web, ACLSingleton
is used which while is has separate classloader for each application does allow
cross-application communication and sharing of metrics.

# Important Info

# Testing

### Testing Performed
Used reproducer in Jira, throws an error in Payara Micro and Payara Web before this PR when going to /metrics and not afterwards.

### Test suites executed
- Payara Private Tests

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.0"-->

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/ACLSingletonProvider.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/ACLSingletonProvider.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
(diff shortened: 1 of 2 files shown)


Cousjava:
Jenkins test please

PR #4309, opened 2019-11-08
PAYARA-3940 Classloader whitelisting mechanism should isolate resources too

AlanRoth:
# Description
This is an improvement

# Important Info

The issue was caused by the service loader trying to load resources for already isolated classes when extreme classloader isolation is enabled, with this change, resources are only loaded if they are also whitelisted alongside the class.

# Testing

Build tests

### Testing Performed

Testing was performed by deploying the attached reproducer, with and without the workaround. 
Also deploying monitoring console to test the solution working with normal applications.

[jacksonpayaraproblem-whitelist.zip](https://github.com/payara/Payara/files/3824336/jacksonpayaraproblem-whitelist.zip)

### Testing Environment
Manjaro 18.0.4
Maven 3.6.1
Zulu JDK 1.8.0_222

# Notes for Reviewers
I did some formatting changes throughout the class just to keep consistency, the actual changes are in WebappClassloader `getResources(String name)` and DOLUtils.


AlanRoth:
Jenkisn test please

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -799,7 +794,7 @@ public synchronized void addRepository(String repository, File file) {
         // Note : There should be only one (of course), but I think we should
         // keep this a bit generic
 
-        if (repository == null)
+        if (repository == null) {
```
MeroRai: This will cause a compilation error. Curly braces will need to be closed.

AlanRoth:
Jenkins test please

AlanRoth:
jenkins test please

PR #4312, opened 2019-11-11
PAYARA-4238 Monitoring Console Hardening

jbee:
### Summary
Fixes a number of smaller issues found during testing:

* malformed series names (e.g. originating from an outdated config) do no longer cause server side exceptions when requesting data for them
* legend of for request tracing details view does not overlap with the graph any longer
* request tracing details view no longer shown "on top" of the grid but instead of it (avoid grid showing underneath) 
* names of pages with presets are now read-only
* low and high watermark of `RangeStatistic` are no longer tracked as a separate metric since each metric has a all time min/max already
* zero is always formatted as just `0` (has no specific unit as that sometimes is odd, e.g. saying 0gb for values that mostly are in kb range)
* decimal numbers (e.g. in legend) are always shown with unit
* formatting with units works correctly for negative numbers
* gaps above spanned column rows are not filled when possible
* widgets are now single select (select exclusively) since multi-select features have not been implemented yet
* added upscaling to allow corrects to value range as e.g. caused by #4292 

Improved settings inputs to prevent malformed user input and script errors as well as give user better feedback on what is valid or expected and what inputs do:

* added input `pattern` attribute where suitable
* added input title to describe the pattern where suitable
* where a `pattern` is used invalid input is highlighted with a red'ish background
* settings are updated when _Unit_ is changes since that has an effect on other setting fields
* range input can be cleared effectively (`NaN` vs `undefined` problem)
* settings inputs can have an id; if not provided an automatic unique id is assigned
* settings input and labels are linked via id so label can be clicked
* description can be provided for settings inputs that is shown as `title` attribute
* widgets can have a custom display name
* CSS: settings table does no longer expand should a value be wider than the table cell 

### Testing
Use the new command to deploy MC: 

    asadmin> set-monitoring-console-configuration enabled=true

Open MC at: http://localhost:8080/monitoring-console/

I tested all features manually. 
In particular this should now allow to add a series manually that has the name `ns: system. cpu.load`
without this causing the UI to show all widgets as "No Connection". Instead the added widget shows "No Data".


jbee:
jenkins test please

nucleus/common/internal-api/src/main/java/fish/payara/monitoring/collect/MonitoringDataCollection.java
```diff
@@ -108,10 +108,7 @@ private static boolean isGetter(Method method) {
 
     public static void collectRange(MonitoringDataCollector collector, RangeStatistic range) {
         collector
-            .collect(range.getName(), range.getCurrent())
-            .group(range.getName())
-                .collect("LowWaterMark", range.getLowWaterMark())
-                .collect("HighWaterMark", range.getHighWaterMark());
```
jbee: I was adding low and high watermark from the beginning just because they existed but soon realised that this is the same as all time min/max (of current) already provided for any metric so this just clutters the list of metrics for no good reason wherefore it now got removed again.

jbee:
jenkins test please

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-components.js
```diff
@@ -78,20 +81,25 @@ MonitoringConsole.View.Components = (function() {
...
         if (model.caption)
-          table.append(createHeaderRow({ label: model.caption }));
+          table.append(createHeaderRow({ label: model.caption, description: model.description }));
         return table;
       }
 
       function createRow(model, inputs) {
         let components = $.isFunction(inputs) ? inputs() : inputs;
-        if (typeof components === 'string') {
+        if (typeof components === 'string')
```
dmatej: I would vote for using curly braces - do we have some standard for javascript here?

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-components.js
```diff
@@ -61,7 +61,10 @@ MonitoringConsole.View.Components = (function() {
 
       function createHeaderRow(model) {
          let caption = model.label;
-         return $('<tr/>').append($('<th/>', {colspan: 2})
+         let config = {colspan: 2};
+         if (model.description)
```
dmatej: No curly braces and only single space indentation ...

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-units.js
```diff
@@ -192,6 +208,20 @@ MonitoringConsole.View.Units = (function() {
       return number.toString().padStart(2, '0');
    }
 
+   const DECIMAL_NUMBER_PATTERN = '([0-9]+\.)?[0-9]+';
```
dmatej: Wouldn't be better to declate constants on one place like we do in Java? It's only a question, not request.

appserver/monitoring-console/webapp/src/main/webapp/js/monitoring-console.js
```diff
@@ -866,7 +880,9 @@ MonitoringConsole.Model = (function() {
 	let Update = (function() {
 
 		function retainLastMinute(data) {
-			let startOfLastMinute = Date.now() - 65000;
+			if (!data)
```
dmatej: Curly braces or at least single line would be better for reading.

appserver/monitoring-console/webapp/src/main/webapp/js/monitoring-console.js
```diff
@@ -1322,6 +1349,17 @@ MonitoringConsole.View.Units = (function() {
       tb: 1024 * 1024 * 1024 * 1024,
    };
 
+   /**
+    * Factors by unit
+    */
+   const FACTORS = {
```
dmatej: Duplicit constant definition.  Same is in another file of the same diff (mc-view-units.js). Could they be shared?

appserver/monitoring-console/webapp/src/main/webapp/js/monitoring-console.js
```diff
@@ -1419,6 +1462,20 @@ MonitoringConsole.View.Units = (function() {
       return number.toString().padStart(2, '0');
    }
 
+   const DECIMAL_NUMBER_PATTERN = '([0-9]+\.)?[0-9]+';
```
dmatej: again, copy and paste. Functions below too.

appserver/monitoring-console/webapp/src/main/java/fish/payara/monitoring/web/MonitoringConsoleResource.java
```diff
@@ -77,10 +78,19 @@ private static RequestTraceStoreInterface getRequestTracingStore() {
         return Globals.getDefaultBaseServiceLocator().getService( RequestTracingService.class).getRequestTraceStore();
     }
 
+    private static Series seriesOrNull(String series) {
+        try {
+            return new Series(series);
+        } catch (RuntimeException e) {
+            return null;
```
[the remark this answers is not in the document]
dmatej: Better idea - I opened the Series constructor, you can remove the whole try/catch, because it is not required, The safe code can look like this:

1) Change the constructor of the Series class to provide better informations about possible error
```
        Objects.requireNonNull(key, "key");
        final String[] parts = key.split(SPLIT_PATTERN);
        // last part is the metric
        final int countOfTags = parts.length - 1;
        this.tags = new String[countOfTags];
        this.values = new String[countOfTags];
        this.metric = parts[parts.length - 1].intern();
        for (int i = 0; i < countOfTags; i++) {
            int eqIndex = parts[i].indexOf(TAG_ASSIGN);
            if (eqIndex < 0) {
                throw new IllegalArgumentException("The key does not have a proper syntax: " + key);
            }
            tags[i] = parts[i].substring(0, eqIndex).intern();
            values[i] = parts[i].substring(eqIndex + 1).intern();
        }

```

2) Log errors at least on finest level, but do not swallow stack traces:
```
    private static Series seriesOrNull(String series) {
        try {
            return new Series(series);
        } catch (RuntimeException e) {
            LOG.log(Level.WARNING, "Could not instantiate the Series class!", e);
            return null;
        }
    }
```
Side note - are those intern() calls required?

Javadoc with some example of the key variable syntax would be nice.

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-components.js
jbee: I am not aware of some so I used a style that I think fits the language. If I would force curlies everywhere code gets 2 times as long for no good reason. Its just a lot of ifs in JS because you have to check all the possible craziness that callers can task functions with and screen size is finite - I rather have functions on screen as a whole then curlies on every if and for statement.

appserver/monitoring-console/webapp/JS_DOCS.md
```diff
@@ -52,11 +52,13 @@ interval        = number
 ### Widget Model
 
 ```
-WIDGET     = { series, type, unit, target, grid, axis, options, decorations, status }
+WIDGET     = { series, type, unit, scaleFactor, target, grid, axis, options, decorations, status, displayName }
 series     = string
 target     = string
+displayName= string
```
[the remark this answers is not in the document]
jbee: It is "consistent" - just not by alphabet or feature context but by my mental model and where I see them fitting in best :D

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-units.js
jbee: The JS code uses the "directly called function module" pattern to do effective scoping. Constants are located in the "module" that uses them. Same as in java where we hide most constants in private fields of a class.

appserver/monitoring-console/webapp/src/main/java/fish/payara/monitoring/web/MonitoringConsoleResource.java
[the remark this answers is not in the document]
jbee: I made sure `Series` always throws a `IllegalArgumentException` with a text indicating what is wrong and documented the format and exception for the constructor. I added tests that check illegal inputs and added logging and further handling of a potential exception to all usages of `new Series` (apart from usages within tests).

jbee:
@dmatej addressed your comments on Java exception handling with added validation, logging and handling potential exceptions in different code paths. 

For JS formatting and code organisation I am not changing the style in the PR. I so far used what works good for me but that is also very much a _work in progress_. Surely we should consolidate this at some point. Maybe its good idea to have a look when more people get involved in MC. 

I am also aware that the merging of JS files isn't ideal but I wanted to keep it simple for us Java programmers not introducing JS stack just to do a simple file merge. We could look into having this treated as generated source file.

jbee:
jenkins test please

appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/model/SeriesTest.java
```diff
@@ -94,4 +95,23 @@ public void tagAssignmentIsSpecialCharacter() {
...
+        try {
+            assertNotNull(new Series("ns: system. cpu.load"));
+        } catch (IllegalArgumentException e) {
+            assertEquals("Malformed series key, `:` missing or misplaced in ns: system. cpu.load", e.getMessage());
+        }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void nullArgument() {
+        assertNotNull(new Series(null));
```
dmatej: here is correct to add fail(message] after the new Series and remove the assertNotNull which will never be called

appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/model/SeriesTest.java
```diff
@@ -94,4 +95,23 @@ public void tagAssignmentIsSpecialCharacter() {
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void emptyArgument() {
+        assertNotNull(new Series(""));
```
dmatej: Same as above

appserver/monitoring-console/core/src/test/java/fish/payara/monitoring/model/SeriesTest.java
jbee: I tend to use the `assertNotNull` as a way to do something with the outcome should it not throw an exception so static analysis does not throw out the statement and the exception never occurs for that reason. Maybe a message could make that clear.

PR #4315, opened 2019-11-12
CI-33 Add a JDK11 job

MarkWareham:
# Description
This is a CI addition. 

Adds a new Jenkins pipeline for testing PRs against JDK11

# Testing

### Testing Performed

It is not possible to test this without placing available in master for the Jenkins server to access.
Change is low risk. A copy of existing job with JDK versions changed.
Testing will be performed after this PR.

# Documentation
Internal wiki updated.

# Notes for Reviewers
Job is a copy of the existing job, except the tools block is moved to each stage as we want to build with JDK8 and test with JDK11

### DIFF
--- a/Jenkinsfile-JDK11
+++ b/Jenkinsfile-JDK11
@@ -48,7 +48,7 @@ pipeline {
                 jdk "zulu-11"
             }
             steps {
-                sh "rm *.zip"
+                sh "rm *.zip || true"
                 setupDomain()
             }
         }
--- a/Jenkinsfile-JDK11
+++ b/Jenkinsfile-JDK11
@@ -1,0 +1,210 @@
+#!groovy
+// Jenkinsfile for building a PR and running a subset of tests against it
+def pom
+def DOMAIN_NAME
+def payaraBuildNumber
+pipeline {
+    agent any
+    options {
+        disableConcurrentBuilds()
+    }
+    stages {
+        stage('Report') {
+            tools {
+                jdk "zulu-8"
+            }
+            steps {
+                script{
+                    pom = readMavenPom file: 'pom.xml'
+                    payaraBuildNumber = "PR${env.ghprbPullId}#${currentBuild.number}"
+                    DOMAIN_NAME = "test-domain"
+                    echo "Payara pom version is ${pom.version}"
+                    echo "Build number is ${payaraBuildNumber}"
+                    echo "Domain name is ${DOMAIN_NAME}"
+                }
+            }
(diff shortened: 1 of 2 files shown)


MarkWareham:
jenkins test jdk-11 please

MarkWareham:
Jenkins test JDK-11 please

MarkWareham:
test failed as the jenkinsfile not available. Must need testing after merge

Pandrex247:
@MarkWareham Are you sure?
`Jenkinsfile-jdk11` vs. `Jenkinsfile-JDK11`

Pandrex247:
Jenkins test JDK-11 please

Pandrex247:
Fixed it on Jenkins side, build is now progressing, but it seems to still be using JDK8...
`Java version: 1.8.0_212, vendor: Azul Systems, Inc., runtime: /usr/lib/jvm/zulu-8-amd64/jre`

MarkWareham:
@Pandrex247 I realised it was the caps. See my two sequential comments. It was a mistake in the regex for the trigger phrase

Pandrex247:
> 
> 
> @Pandrex247 I realised it was the caps. See my two sequential comments. It was a mistake in the regex for the trigger phrase

I meant caps in the Jenkins job itself, not the trigger phrase - Jenkins was looking for `Jenkinsfile-jdk11` but the file is `Jenkinsfile-JDK11`

Jenkinsfile-JDK11
```diff
@@ -0,0 +1,210 @@
...
+                    archiveArtifacts artifacts: 'appserver/extras/payara-micro/payara-micro-distribution/target/payara-micro.jar', fingerprint: true
+                }
+            }
+        }
+        stage('Setup for Quicklook Tests') {
+            tools {
+                jdk "zulu-11"
+            }
+            steps {
+                sh "rm *.zip"
```
Pandrex247: Build fails here

Pandrex247: This seems to be trying to delete the log files from previous runs: https://github.com/payara/Payara/pull/4315/files#diff-28788dd8800bef08d2db40a4380a2f1eR71

The following change might work, haven't tried it within a Jenkinsfile (and not sure on the placement of the quotes, pick which you think is right):
```suggestion
                sh "rm *.zip || true"
```
```suggestion
                sh "rm *.zip" || true
```

MarkWareham:
Jenkins test JDK-11 please

Issue #3042, opened 2018-08-14
OpenAPI JAX-RS Incomplete subresource support


vassterg:
# Description #
----------
Incomplete support of openapi when subresource is present. 
I have two classes Foo and Bar such that Foo is a jaxrs resource and Bar is a subresource.
When deployed, the openapi link is incomplete.

## Expected Outcome

I would expect in the openapi link, the Bar sub resource GET endpoint to show up in the openapi and the Bar sub resource to be included as a schema.

## Current Outcome
```
openapi: 3.0.0
info:
  title: Deployed Resources
  version: 1.0.0
servers:
- url: http://localhost:8080/openapi-test
  description: Default Server.
paths:
  /resources/foo:
    get:
      operationId: getMethod
      responses:
        default:
          description: Default Response.
          content:
            '*/*':
              schema:
                type: string
components: {}
```
No endpoint and schema for subresource bar.

## Steps to reproduce (Only for bug reports) 
Create a simple javaee 8 maven JAX-RS project. 
Example pom.xml
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.insurit</groupId>
    <artifactId>openapi-test</artifactId>
    <version>1.0-SNAPSHOT</version>
    <packaging>war</packaging>
    <dependencies>
        <dependency>
            <groupId>javax</groupId>
            <artifactId>javaee-api</artifactId>
            <version>8.0</version>
            <scope>provided</scope>
        </dependency>
    </dependencies>
    <build>
        <finalName>openapi-test</finalName>
    </build>
    <properties>
        <maven.compiler.source>1.8</maven.compiler.source>
        <maven.compiler.target>1.8</maven.compiler.target>
        <failOnMissingWebXml>false</failOnMissingWebXml>
    </properties>
    <name>openapi-test</name>
</project>
```
Add the following three classes:
```java
@ApplicationPath("resources")
public class JAXRSConfiguration extends Application {

}
```
```java
@Path("foo")
public class Foo {

    @Context
    private ResourceContext rc;

    @GET
    public String getMethod() {
        return "get foo";
    }
    
    @Path("bar")
    public Bar sub() {
        return rc.getResource(Bar.class);
[...]

PR #4317, opened 2019-11-12
PAYARA-4172 Rerun add openapi schemas to allow for sub-schemas to be …

Cousjava:
# Description
This is a bug fix.

If in an OpenAPI schema it references another schema then sometimes depending on loading the sub-schema is loaded yet and so it uses an empty schema definition. This adds another run of schema generation to processing so the sub-schemas (which will exist now) can be associated with where they are referenced.

# Testing

### Testing Performed
Ran provided test app (see Jira), redeploying it and going to the /openapi endpoint and checking it half a dozen times to ensure it works.

### Test suites executed

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.0


Cousjava:
Jenkins test please

jbee:
We should also cross-check if this solves #3042

Cousjava:
@jbee this does not solve that issue.

jbee:
And I thought we got lucky. Not this time. Thanks for checking anyway.

PR #4323, opened 2019-11-13
PAYARA-4197 Always print rollback+timeout warning

dmatej:
# Description

- print warning about timeout also in a case when there is more serious error if the tx timeouted too,
- when some bean tried to use timeouted transacion, a non-checked exception
  was thrown and it shadowed the original cause of rollback - the warning
  wasn't printed then. Now it is.

# Important Info

### Dependant PRs 

TBD (there will be done)

# Testing

### New tests

See future push in #4263  with integrated test case.

### Testing Performed

Manual using the JIRA example.

### Test suites executed

None of following yet:
- Quicklook
- Payara Samples
- Java EE7 Samples
- Java EE8 Samples
- Payara Private Tests
- Jakarta TCKs
- Mojarra
- Cargo Tracker

# Notes for Reviewers

The solution was designed after refactoring done in another PR, because present logging is not much useful and is rather confusing. But refactoring might introduce new errors and will need to pass at least TCK/EJB and TCK/JTA tests. So I created this minimal PR and the refactoring PR will follow.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
@@ -573,7 +573,7 @@ public class EJBContainerTransactionManager {
         try {
             container.forceDestroyBean(context);
         } finally {
-            transactionManager.rollback();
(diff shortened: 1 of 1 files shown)


dmatej:
Jenkins test please

PR #4324, opened 2019-11-14
PAYARA-3989 Unloaded Application will be now be unavailable as soon as they are disabled

MeroRai:
# Description
This is a bug fix to fix an issue where an application still response with an empty HTTP 200 for a few seconds when it is on the process of being disabled.

# Testing

### New tests
None

### Testing Environment

# Documentation
Not required

# Notes for Reviewers
This issue could not be reproduced by either the service team or I, hence I can't confirm my code changes will fix this issue. See [JIRA Issue](https://payara.atlassian.net/browse/PAYARA-3989) for more info.

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardWrapperValve.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardWrapperValve.java
@@ -56,6 +56,8 @@
  * limitations under the License.
  */
 
+// Portions Copyright [2019] [Payara Foundation and/or its affiliates]
+
 package org.apache.catalina.core;
 
 import java.io.IOException;
@@ -72,7 +74,6 @@ import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.UnavailableException;
 import javax.servlet.http.HttpServletResponse;
-
 import org.apache.catalina.Context;
 import org.apache.catalina.Globals;
 import org.apache.catalina.HttpRequest;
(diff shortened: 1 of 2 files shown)


MeroRai:
jenkins test please

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardWrapperValve.java
```diff
@@ -135,7 +140,23 @@ public int invoke(Request request, Response response)
         RequestFacade hreq = (RequestFacade) request.getRequest(true);
         HttpServletResponse hres =
             (HttpServletResponse) response.getResponse();
+        
+        String contextPath = context.getServletContext().getContextPath();
+        
+        if (!contextPath.isEmpty()) {
+            ServiceLocator services = org.glassfish.internal.api.Globals.getDefaultHabitat();
+            ApplicationRegistry appRegistry = services.<ApplicationRegistry>getService(ApplicationRegistry.class);
```
pdudits: This looks like quite expensive thing to do on each HTTP request

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardWrapperValve.java
```diff
@@ -135,7 +140,23 @@ public int invoke(Request request, Response response)
...
+            applicationInfo = appRegistry.get(context.getServletContext().getContextPath().replace("/", ""));
```
pdudits: This is only true, when context path matches application (i. e. not overriden in glassfish-web, not part of an EAR, ...)

PR #4325, opened 2019-11-15
PAYARA-4197 Minor but large cleanup in ejb-container and related things

dmatej:
- sorted imports, removed unused
- code formatting
  - preferred shorter block first
  - preferred positive if without negation
  - preferred fast return
  - collisions of these rules resolved appropriately
  - removed 10 years old TODOs
- using generics where it is clear
- trivial optimizations
  - declaring variables
  - preferred final instead of changing the default value
- removed redundant keywords (public method in interface, ie)
- removed redundant class typing
- using addSuppressed/cause rather than masking exceptions
- improved logging
  - log entry to important method at least as finest
  - removed some duplicit logs

# Description
This is an enhancement for better readibility and also to provide better informations in logs.
EJB timeout warnings are now printed immediately when the transaction is set to the rollback only state, and later are printed another informations about the rollback - why it is done. These messages are not duplicated, but together lead the developer to some idead what happened and why.

# Important Info

### Blockers 

This PR continues the work started in #4323 in much larger scale

# Testing

### New tests

Experimental test will be pushed in #4263 

### Test suites executed

- Quicklook - OK
- Payara Samples - not executed
- Java EE7 Samples - OK
- Java EE8 Samples - OK
- Payara Private Tests - not executed
- Payara Microprofile TCKs Runner - not executed
- Jakarta TCKs - ejb32 - fails one test at this moment, but it is not related.

### Testing Environment

Manual tests: Kubuntu 19.10, OpenJDK11
TestContainers: Debian 11, OpenJDK11
TCK: JDK8

# Notes for Reviewers

The code needs much more love :-)

### DIFF
--- a/appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
+++ b/appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
@@ -38,23 +38,31 @@
  * holder.
  */
 
+// Portions Copyright [2019] [Payara Foundation and/or its affiliates.]
+
 package com.sun.enterprise.connectors.work;
 
 
+import com.sun.appserv.connectors.internal.api.ConnectorRuntime;
 import com.sun.corba.ee.spi.threadpool.WorkQueue;
-import com.sun.enterprise.transaction.api.JavaEETransactionManager;
 import com.sun.enterprise.connectors.work.context.WorkContextHandlerImpl;
 import com.sun.enterprise.connectors.work.monitor.WorkManagementProbeProvider;
-import com.sun.appserv.connectors.internal.api.ConnectorRuntime;
-
-import javax.resource.ResourceException;
-import javax.resource.spi.ResourceAdapterAssociation;
-import javax.resource.spi.work.*;
-
-import org.glassfish.logging.annotation.LogMessageInfo;
+import com.sun.enterprise.transaction.api.JavaEETransactionManager;
 
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import javax.resource.ResourceException;
+import javax.resource.spi.ResourceAdapterAssociation;
+import javax.resource.spi.work.ExecutionContext;
+import javax.resource.spi.work.Work;
+import javax.resource.spi.work.WorkCompletedException;
+import javax.resource.spi.work.WorkEvent;
+import javax.resource.spi.work.WorkException;
+import javax.resource.spi.work.WorkListener;
+import javax.resource.spi.work.WorkRejectedException;
+
+import org.glassfish.logging.annotation.LogMessageInfo;
 
 /**
  * WorkCoordinator : Coordinates one work's execution. Handles all
@@ -91,10 +99,10 @@ public final class WorkCoordinator {
 
     private WorkManagementProbeProvider probeProvider = null;
 
-    private ConnectorRuntime runtime;
+    private final ConnectorRuntime runtime;
     private String raName = null;
 
-    private WorkContextHandlerImpl contextHandler;
+    private final WorkContextHandlerImpl contextHandler;
 
     /**
      * Constructs a coordinator
@@ -130,7 +138,7 @@ public final class WorkCoordinator {
     public String getRAName(){
         return raName;
     }
- 
+
     /**
      * Submits the work to the queue and generates a work accepted event.
      */
@@ -168,7 +176,7 @@ public final class WorkCoordinator {
      */
     public void preInvoke() {
 
-        // If the work is just scheduled, check whether it has timed out or not. 
+        // If the work is just scheduled, check whether it has timed out or not.
         if (waitMode == NO_WAIT && timeout > -1) {
             long elapsedTime = System.currentTimeMillis() - startTime;
 
@@ -217,7 +225,7 @@ public final class WorkCoordinator {
                     new WorkEvent(this, WorkEvent.WORK_STARTED, work, null));
         }
 
-        //set the unauthenticated securityContext before executing the work            
+        //set the unauthenticated securityContext before executing the work
         com.sun.enterprise.security.SecurityContext.setUnauthenticatedContext();
 
     }
@@ -253,9 +261,7 @@ public final class WorkCoordinator {
 
                     //If exception is not null, the work has already been rejected.
                     if (listener != null) {
-                        listener.workCompleted(
-                                new WorkEvent(this, WorkEvent.WORK_COMPLETED, work,
-                                        getException()));
+                        listener.workCompleted(new WorkEvent(this, WorkEvent.WORK_COMPLETED, work, getException()));
                     }
                 }
 
@@ -266,7 +272,7 @@ public final class WorkCoordinator {
             } catch(Exception e) {
 	            logger.log(Level.WARNING, e.getMessage());
             }finally{
-                //reset the securityContext once the work has completed            
+                //reset the securityContext once the work has completed
                 com.sun.enterprise.security.SecurityContext.setUnauthenticatedContext();
             }
         }
@@ -402,6 +408,7 @@ public final class WorkCoordinator {
      *
      * @return Unique identification concatenated by work object.
      */
+    @Override
     public String toString() {
         return id + ":" + work;
     }
@@ -412,6 +419,7 @@ public final class WorkCoordinator {
      * @param state CREATED or Either STARTED or COMPLETED or TIMEDOUT
      */
     public synchronized void setState(int state) {
+        logger.finest(() -> String.format("setState(state=%s)", state));
         this.state = state;
     }
 
@@ -465,7 +473,7 @@ public final class WorkCoordinator {
         }
         return ec;
     }
-    
+
     public static synchronized int increaseSeed() {
         return ++seed;
     }
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
@@ -92,7 +92,6 @@ public class EjbInvocation //
 
     private TransactionOperationsManager transactionOperationsManager;
 
-
     /**
      * The EJBObject/EJBLocalObject which created this EjbInvocation object.
      * This identifies the target bean.
@@ -268,10 +267,10 @@ public class EjbInvocation //
 
     public Method beanMethod;
 
-    // Only set for web service invocations.
+    /** Only set for web service invocations. */
     private WebServiceContext webServiceContext;
 
-    // Only set for EJB JAXWS
+    /** Only set for EJB JAXWS */
     // FIXME: private Message message = null;
     private Object message;
 
@@ -686,28 +685,30 @@ public class EjbInvocation //
             setWebServiceMethod(null);
             return true;
         }
-        Exception ie;
+        final Exception ie = authorizeWebServiceAndSetMethod(m);
+        if (ie == null) {
+            return true;
+        }
+        exception = ie;
+        throw ie;
+	}
+
+    private Exception authorizeWebServiceAndSetMethod(Method m) {
         try {
             this.method = m;
             if (((com.sun.ejb.Container) container).authorize(this)) {
                 // Record the method on which the successful
                 // authorization check was performed.
                 setWebServiceMethod(m);
-                ie = null;
-            } else {
-                ie = new Exception("Client not authorized for invocation of method {" + method + "}");
+                return null;
             }
+            return new Exception("Client not authorized for invocation of method {" + method + "}");
         } catch (Exception e) {
             String errorMsg = "Error unmarshalling method {" + method + "} for ejb ";
-            ie = new UnmarshalException(errorMsg, e);
+            // note: this exception is undeclared, but catched!
+            return new UnmarshalException(errorMsg, e);
         }
-        if (ie == null) {
-            return true;
-        }
-        // ie was only created or catched, it does not matter here
-        exception = ie;
-        throw ie;
-	}
+    }
 
     /**
      * @return true if the SecurityManager reports that the caller is in role
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
@@ -41,17 +41,27 @@
 
 package com.sun.ejb;
 
-//XXX: import javax.xml.rpc.handler.MessageContext;
-/* HARRY : JACC Changes */
-
-import com.sun.ejb.containers.*;
+import com.sun.ejb.containers.BaseContainer;
+import com.sun.ejb.containers.EJBContextImpl;
+import com.sun.ejb.containers.EJBLocalRemoteObject;
+import com.sun.ejb.containers.EjbContainerUtilImpl;
+import com.sun.ejb.containers.EjbFutureTask;
+import com.sun.ejb.containers.SimpleEjbResourceHandlerImpl;
 import com.sun.ejb.containers.interceptors.InterceptorManager;
+import com.sun.ejb.containers.interceptors.InterceptorManager.AroundInvokeContext;
 import com.sun.ejb.containers.interceptors.InterceptorUtil;
 import com.sun.enterprise.deployment.EjbBundleDescriptor;
 import com.sun.enterprise.deployment.MethodDescriptor;
 import com.sun.enterprise.transaction.spi.TransactionOperationsManager;
-import org.glassfish.api.invocation.ComponentInvocation;
-import org.glassfish.api.invocation.ResourceHandler;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.rmi.UnmarshalException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Predicate;
 
 import javax.ejb.EJBContext;
 import javax.ejb.Timer;
@@ -61,11 +71,10 @@ import javax.transaction.Transaction;
 import javax.xml.rpc.handler.MessageContext;
 import javax.xml.soap.SOAPMessage;
 import javax.xml.ws.WebServiceContext;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.Method;
-import java.rmi.UnmarshalException;
-import java.util.HashMap;
-import java.util.Map;
+
+import org.glassfish.api.invocation.ComponentInvocation;
+import org.glassfish.api.invocation.ResourceHandler;
+import org.glassfish.ejb.api.EJBInvocation;
 
 /**
  * The EjbInvocation object contains state associated with an invocation
@@ -75,32 +84,14 @@ import java.util.Map;
  * EJB(Local)Object/EJB(Local)Home before and after an invocation.
  */
 
-public class EjbInvocation
-    extends ComponentInvocation
-    implements InvocationContext, TransactionOperationsManager,
-         org.glassfish.ejb.api.EJBInvocation, InterceptorManager.AroundInvokeContext
-{
-
+public class EjbInvocation //
+    extends ComponentInvocation //
+    implements InvocationContext, TransactionOperationsManager, EJBInvocation, AroundInvokeContext {
 
     public ComponentContext context;
 
     private TransactionOperationsManager transactionOperationsManager;
 
-    EjbInvocation(String compEnvId, Container container) {
-        super.componentId = compEnvId;
-        super.container = container;
-        super.setComponentInvocationType(ComponentInvocation.ComponentInvocationType.EJB_INVOCATION);
-
-        EjbBundleDescriptor ejbBundleDesc = container.getEjbDescriptor().getEjbBundleDescriptor();
-        moduleName = ejbBundleDesc.getModuleName();
-        appName = ejbBundleDesc.getApplication().getAppName();
-        registrationName = ejbBundleDesc.getApplication().getRegistrationName();
-
-        //By default we enable TransactionOperationsManager checks. But EjbInvocation.clone()
-        //  clears transactionOperationsManager so that, be default, cloned invocations
-        //  doesn't enforce Transaction Operations checks.
-        transactionOperationsManager = this;
-    }
 
     /**
      * The EJBObject/EJBLocalObject which created this EjbInvocation object.
@@ -112,13 +103,13 @@ public class EjbInvocation
      * Local flag: true if this invocation was through the 2.x (or earlier)
      * Local client view, the 3.x local client view or a no-interface client view.
      */
-    public boolean isLocal=false;
+    public boolean isLocal = false;
 
     /**
      * True if this invocation was made through the 2.x (or earlier) Remote
      * client view or the 3.x remote client view.
      */
-    public boolean isRemote=false;
+    public boolean isRemote = false;
 
     /**
      * InvocationInfo object caches information about the current method
@@ -134,29 +125,29 @@ public class EjbInvocation
     /**
      * true if this is a web service invocation
      */
-    public boolean isWebService=false;
+    public boolean isWebService = false;
 
     /**
      * true if this is an ejb timeout method invocation
      */
-    public boolean isTimerCallback=false;
+    public boolean isTimerCallback = false;
 
     /**
      * true if this is a message-driven bean invocation
      */
-    public boolean isMessageDriven=false;
+    public boolean isMessageDriven = false;
 
     /**
      * true if this is an invocation on the home object
      * this is required for jacc.
      */
-    public boolean isHome=false;
+    public boolean isHome = false;
 
     /**
      * Home, Remote, LocalHome, Local, WebService, or business interface
      * through which a synchronous ejb invocation was made.
      */
-    public Class clientInterface;
+    public Class<?> clientInterface;
 
     /**
      * Method to be invoked. This is a method of the EJB's local/remote
@@ -189,21 +180,12 @@ public class EjbInvocation
      */
     public Throwable exceptionFromBeanMethod;
 
-
     /**
      * The client's transaction if any.
      * Set by the Container during preInvoke() and used by the Container
      * during postInvoke().
      */
     public Transaction clientTx;
-
-    /**
-     * The EJBContext object of the bean instance being invoked.
-     * Set by the Container during preInvoke() and used by the Container
-     * during postInvoke().
-     */
-    // Moved to com/sun/enterprise/ComponentInvocation
-    // public ComponentContext context;
 
     /**
      * The transaction attribute of the bean method. Set in generated
@@ -282,6 +264,38 @@ public class EjbInvocation
     // True if lock is currently held for this invocation
     private boolean holdingSFSBSerializedLock = false;
 
+    private int interceptorIndex;
+
+    public Method beanMethod;
+
+    // Only set for web service invocations.
+    private WebServiceContext webServiceContext;
+
+    // Only set for EJB JAXWS
+    // FIXME: private Message message = null;
+    private Object message;
+
+    private SOAPMessage soapMessage = null;
+
+    private Map<String, Object> contextData;
+
+
+    EjbInvocation(String compEnvId, Container container) {
+        super.componentId = compEnvId;
+        super.container = container;
+        super.setComponentInvocationType(ComponentInvocation.ComponentInvocationType.EJB_INVOCATION);
+
+        EjbBundleDescriptor ejbBundleDesc = container.getEjbDescriptor().getEjbBundleDescriptor();
+        moduleName = ejbBundleDesc.getModuleName();
+        appName = ejbBundleDesc.getApplication().getAppName();
+        registrationName = ejbBundleDesc.getApplication().getRegistrationName();
+
+        //By default we enable TransactionOperationsManager checks. But EjbInvocation.clone()
+        //  clears transactionOperationsManager so that, be default, cloned invocations
+        //  doesn't enforce Transaction Operations checks.
+        transactionOperationsManager = this;
+    }
+
     public ClassLoader getOriginalContextClassLoader() {
         return originalContextClassLoader;
     }
@@ -330,6 +344,7 @@ public class EjbInvocation
         this.doTxProcessingInPostInvoke = doTxProcessingInPostInvoke;
     }
 
+    @Override
     public EjbInvocation clone() {
         EjbInvocation newInv = (EjbInvocation) super.clone();
 
@@ -366,6 +381,7 @@ public class EjbInvocation
      * implementation should use this method rather than directly
      * accessing the ejb field.
      */
+    @Override
     public Object getJaccEjb() {
         Object bean = null;
         if( container != null ) {
@@ -392,17 +408,17 @@ public class EjbInvocation
     }
 
     /**
-     * Returns CachedPermission associated with this invocation, or
+     * @return CachedPermission associated with this invocation, or
      * null if not available.
      */
     public Object getCachedPermission() {
-        return (invocationInfo != null) ? invocationInfo.cachedPermission :
-            null;
+        return invocationInfo == null ? null : invocationInfo.cachedPermission;
     }
 
     /**
      * @return Returns the ejbCtx.
      */
+    @Override
     public EJBContext getEJBContext() {
         return (EJBContext) this.context;
     }
@@ -438,17 +454,16 @@ public class EjbInvocation
     }
 
     public void setTransactionOperationsManager(TransactionOperationsManager transactionOperationsManager) {
-        //Note: clone() clears transactionOperationsManager so that, be default, cloned invocations
-        //  doesn't enforce Transaction Operations checks.
+        // Note: clone() clears transactionOperationsManager so that, be default, cloned invocations
+        // doesn't enforce Transaction Operations checks.
         this.transactionOperationsManager = transactionOperationsManager;
     }
-
-    //Implementation of TransactionOperationsManager methods
 
     /**
      * Called by the UserTransaction implementation to verify
      * access to the UserTransaction methods.
      */
+    @Override
     public boolean userTransactionMethodsAllowed() {
         return ((Container) container).userTransactionMethodsAllowed(this);
     }
@@ -457,6 +472,7 @@ public class EjbInvocation
      * Called by the UserTransaction lookup to verify
      * access to the UserTransaction itself.
      */
+    @Override
     public void userTransactionLookupAllowed() throws NameNotFoundException {
         ((BaseContainer) container).checkUserTransactionLookup(this);
     }
@@ -464,35 +480,19 @@ public class EjbInvocation
     /**
      * Called by the UserTransaction when transaction is started.
      */
+    @Override
     public void doAfterUtxBegin() {
         ((Container) container).doAfterBegin(this);
     }
 
-    //Implementation of InvocationContext methods
-
-    private int interceptorIndex;
-
-    public Method   beanMethod;
-
-    // Only set for web service invocations.
-    private WebServiceContext webServiceContext;
-
-    // Only set for EJB JAXWS
-    //FIXME: private Message message = null;
-    private Object message;
-
-    private SOAPMessage soapMessage = null;
-
-    private Map      contextData;
-
     public InterceptorManager.InterceptorChain getInterceptorChain() {
-        return (invocationInfo == null)
-            ? null : invocationInfo.interceptorChain;
+        return (invocationInfo == null) ? null : invocationInfo.interceptorChain;
     }
 
     /**
      * @return Returns the bean instance.
      */
+    @Override
     public Object getTarget() {
         return this.ejb;
     }
@@ -500,6 +500,7 @@ public class EjbInvocation
     /**
      * @return Returns the timer instance.
      */
+    @Override
     public Object getTimer() {
         return timer;
     }
@@ -510,13 +511,16 @@ public class EjbInvocation
      *         method being invoked.  For lifecycle callback methods,
      *         returns null.
      */
+    @Override
     public Method getMethod() {
         return getBeanMethod();
     }
+
     public Method getBeanMethod() {
         return this.beanMethod;
     }
 
+    @Override
     public Constructor getConstructor() {
         return null;
     }
@@ -527,6 +531,7 @@ public class EjbInvocation
      * getParameters() returns the values to which the parameters
      * have been set.
      */
+    @Override
     public Object[] getParameters() {
         return this.methodParams;
     }
@@ -535,15 +540,17 @@ public class EjbInvocation
      * Set the parameters that will be used to invoke the business method.
      *
      */
+    @Override
     public void setParameters(Object[] params) {
         InterceptorUtil.checkSetParameters(params, getMethod());
         this.methodParams = params;
     }
 
-    /*
-     * Method takes Object to decouple EJBInvocation interface
+    /**
+     * Method takes Object to decouple {@link EJBInvocation} interface
      * from jaxws (which isn't available in all profiles).
      */
+    @Override
     public void setWebServiceContext(Object webServiceContext) {
         // shouldn't be necessary, but to be safe
         if (webServiceContext instanceof WebServiceContext) {
@@ -554,12 +561,14 @@ public class EjbInvocation
     /**
      * @return Returns the contextMetaData.
      */
+    @Override
     public Map<String, Object> getContextData() {
         if (this.contextData == null) {
-            if (webServiceContext != null)
+            if (webServiceContext != null) {
                 this.contextData = webServiceContext.getMessageContext();
-            else
-                this.contextData = new HashMap<String, Object>();
+            } else {
+                this.contextData = new HashMap<>();
+            }
         }
         return contextData;
     }
@@ -568,6 +577,7 @@ public class EjbInvocation
      * This is for EJB JAXWS only.
      * @param message  an unconsumed message
      */
+    @Override
     public <T> void setMessage(T message) {
         this.message = message;
     }
@@ -576,6 +586,7 @@ public class EjbInvocation
      * This is for EJB JAXWS only.
      * @return the JAXWS message
      */
+    @Override
     public Object getMessage() {
         return this.message;
     }
@@ -597,12 +608,8 @@ public class EjbInvocation
         return soapMessage;
     }
 
-    /* (non-Javadoc)
-     * @see javax.interceptor.InvocationContext#proceed()
-     */
-    public Object proceed()
-        throws Exception
-    {
+    @Override
+    public Object proceed() throws Exception {
         try {
             //TODO: Internal error if getInterceptorChain() is null
             interceptorIndex++;
@@ -623,8 +630,8 @@ public class EjbInvocation
     /**
      * Print most useful fields.  Don't do all of them (yet) since there
      * are a large number.
-     * @return
      */
+    @Override
     public String toString() {
 
         StringBuilder sbuf = new StringBuilder();
@@ -649,94 +656,100 @@ public class EjbInvocation
     }
 
     // Implementation of AroundInvokeContext
+    @Override
     public Object[] getInterceptorInstances() {
         return  ((EJBContextImpl)context).getInterceptorInstances();
     }
 
+    @Override
     public  Object invokeBeanMethod() throws Throwable {
         return ((BaseContainer) container).invokeBeanMethod(this);
     }
-
-    /*********************************************************/
-
-
 
     public com.sun.enterprise.security.SecurityManager getEjbSecurityManager() {
         return ((BaseContainer)container).getSecurityManager();
     }
 
+    @Override
     public boolean isAWebService() {
         return this.isWebService;
     }
 
+    @Override
     public Object[] getMethodParams() {
         return this.methodParams;
     }
 
+    @Override
     public boolean authorizeWebService(Method m) throws Exception {
-        Exception ie = null;
-        if (isAWebService()) {
-            try {
-                this.method = m;
-                if (!((com.sun.ejb.Container)container).authorize(this)) {
-                    ie = new Exception
-                    ("Client not authorized for invocation of method {" + method + "}");
-                } else {
-                    // Record the method on which the successful
-                    // authorization check was performed.
-                    setWebServiceMethod(m);
-                }
-            } catch(Exception e) {
-                String errorMsg = "Error unmarshalling method {" + method + "} for ejb ";
-                ie = new UnmarshalException(errorMsg);
-                ie.initCause(e);
+        if (!isAWebService()) {
+            setWebServiceMethod(null);
+            return true;
+        }
+        Exception ie;
+        try {
+            this.method = m;
+            if (((com.sun.ejb.Container) container).authorize(this)) {
+                // Record the method on which the successful
+                // authorization check was performed.
+                setWebServiceMethod(m);
+                ie = null;
+            } else {
+                ie = new Exception("Client not authorized for invocation of method {" + method + "}");
             }
-            if ( ie != null ) {
-                exception = ie;
-                throw ie;
-            }
-	    } else {
-		    setWebServiceMethod(null);
-	    }
-        return true;
+        } catch (Exception e) {
+            String errorMsg = "Error unmarshalling method {" + method + "} for ejb ";
+            ie = new UnmarshalException(errorMsg, e);
+        }
+        if (ie == null) {
+            return true;
+        }
+        // ie was only created or catched, it does not matter here
+        exception = ie;
+        throw ie;
 	}
 
-/**
-    * Implements the  method in org.glassfish.ejb.api.EJBInvocation
-    * @return true if the SecurityManager reports that the caller is in role
-    */
-   public boolean isCallerInRole(String role) {
-       return getEjbSecurityManager().isCallerInRole(role);
-   }
+    /**
+     * @return true if the SecurityManager reports that the caller is in role
+     */
+    @Override
+    public boolean isCallerInRole(String role) {
+        return getEjbSecurityManager().isCallerInRole(role);
+    }
 
+    @Override
     public void setWebServiceTie(Object tie) {
         webServiceTie = tie;
     }
 
+    @Override
     public Object getWebServiceTie() {
         return webServiceTie;
     }
 
+    @Override
     public void setWebServiceMethod(Method method) {
         webServiceMethod = method;
     }
 
+    @Override
     public Method getWebServiceMethod() {
         return webServiceMethod;
     }
 
+    @Override
     public void setMessageContext(MessageContext msgContext) {
        messageContext = msgContext;
     }
 
-   public ResourceHandler getResourceHandler() {
-       ResourceHandler rh = super.getResourceHandler();
-       if (rh == null) {
-           rh = context;
-       }
-
-       return rh;
-   }
+    @Override
+    public ResourceHandler getResourceHandler() {
+        ResourceHandler rh = super.getResourceHandler();
+        if (rh == null) {
+            rh = context;
+        }
+        return rh;
+    }
 
     public boolean isContainerStartsTx() {
         return containerStartsTx;
@@ -745,6 +758,14 @@ public class EjbInvocation
     public void setContainerStartsTx(boolean containerStartsTx) {
         this.containerStartsTx = containerStartsTx;
     }
+
+    /**
+     * @param classes must not be null
+     * @return true if the client interface is assignable to at least one of classes in parameter
+     */
+    public boolean isClientInterfaceAssignableToOneOf(final Class<?>... classes) {
+        Objects.requireNonNull(classes, "classes");
+        final Predicate<Class<?>> predicate = c -> c.isAssignableFrom(this.clientInterface);
+        return Arrays.stream(classes).anyMatch(predicate);
+    }
 }
-
-
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSessionContextImpl.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSessionContextImpl.java
@@ -37,41 +37,37 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2018-2019] [Payara Foundation and/or its affiliates]
 
 package com.sun.ejb.containers;
 
+import com.sun.ejb.EJBUtils;
 import com.sun.ejb.EjbInvocation;
 import com.sun.enterprise.deployment.EjbDescriptor;
 import com.sun.enterprise.deployment.EjbSessionDescriptor;
-import org.glassfish.api.invocation.ComponentInvocation;
-import org.glassfish.api.invocation.InvocationManager;
+
 import javax.ejb.SessionContext;
 import javax.ejb.TimerService;
 import javax.transaction.UserTransaction;
 import javax.xml.rpc.handler.MessageContext;
-import com.sun.ejb.EJBUtils;
+
+import org.glassfish.api.invocation.ComponentInvocation;
+import org.glassfish.api.invocation.InvocationManager;
 
 /**
  * Implementation of EJBContext for SessionBeans
  *
  * @author Mahesh Kannan
  */
+public abstract class AbstractSessionContextImpl extends EJBContextImpl implements SessionContext {
 
-public abstract class AbstractSessionContextImpl
-        extends EJBContextImpl
-        implements SessionContext {
-    
-    protected Object instanceKey;
-
-    protected String ejbName;
+    private final String ejbName;
+    private Object instanceKey;
 
 
     protected AbstractSessionContextImpl(Object ejb, BaseContainer container) {
         super(ejb, container);
-        EjbSessionDescriptor sessionDesc =
-                (EjbSessionDescriptor) getContainer().getEjbDescriptor();
-
+        EjbSessionDescriptor sessionDesc = (EjbSessionDescriptor) getContainer().getEjbDescriptor();
         this.ejbName = sessionDesc.getName();
     }
 
@@ -85,7 +81,7 @@ public abstract class AbstractSessionContextImpl
 
     @Override
     public String toString() {
-        return ejbName + "; id: " + instanceKey;
+        return ejbName + "; id: " + getInstanceKey();
     }
 
     @Override
@@ -93,7 +89,7 @@ public abstract class AbstractSessionContextImpl
 
         // Instance key is first set between after setSessionContext and
         // before ejbCreate
-        if (instanceKey == null) {
+        if (getInstanceKey() == null) {
             throw new IllegalStateException("Operation not allowed");
         }
 
@@ -107,8 +103,9 @@ public abstract class AbstractSessionContextImpl
         // was called from setSession/EntityContext. The instance key check
         // ensures that an exception is not thrown if this was called
         // from a stateless SessionBean's ejbCreate.
-        if ((state == BeanState.CREATED) && (instanceKey == null))
+        if (state == BeanState.CREATED && getInstanceKey() == null) {
             throw new IllegalStateException("Operation not allowed");
+        }
 
         return ((BaseContainer) getContainer()).getUserTransaction();
     }
@@ -118,18 +115,13 @@ public abstract class AbstractSessionContextImpl
         InvocationManager invManager = EjbContainerUtilImpl.getInstance().getInvocationManager();
         try {
             ComponentInvocation inv = invManager.getCurrentInvocation();
-
-            if ((inv != null) && isWebServiceInvocation(inv)) {
+            if (inv != null && isWebServiceInvocation(inv)) {
                 return ((EjbInvocation) inv).messageContext;
-            } else {
-                throw new IllegalStateException("Attempt to access " +
-                        "MessageContext outside of a web service invocation");
             }
         } catch (Exception e) {
-            IllegalStateException ise = new IllegalStateException();
-            ise.initCause(e);
-            throw ise;
+            throw new IllegalStateException("Could not get a message context.", e);
         }
+        throw new IllegalStateException("Attempt to access MessageContext outside of a web service invocation");
     }
 
     @Override
@@ -141,7 +133,7 @@ public abstract class AbstractSessionContextImpl
 
         // getBusinessObject not allowed for Stateless/Stateful beans
         // until after dependency injection
-        if (instanceKey == null) {
+        if (getInstanceKey() == null) {
             throw new IllegalStateException("Operation not allowed");
         }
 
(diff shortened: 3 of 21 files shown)


appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
```diff
@@ -91,10 +99,10 @@
 
     private WorkManagementProbeProvider probeProvider = null;
 
-    private ConnectorRuntime runtime;
+    private final ConnectorRuntime runtime;
```
dmatej: Eclipse automatically adds final to fields that are set only in constructor and never changed. (yes, you can still call their methods).

appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
```diff
@@ -217,7 +225,7 @@ public void preInvoke() {
                     new WorkEvent(this, WorkEvent.WORK_STARTED, work, null));
         }
 
-        //set the unauthenticated securityContext before executing the work
```
dmatej: Traling spaces are evil! Another automatic fix. And I have also script which could do it in all java sources, maybe I will execute it as a xmass present :-)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -75,32 +84,14 @@
...
-
+public class EjbInvocation //
+    extends ComponentInvocation //
+    implements InvocationContext, TransactionOperationsManager, EJBInvocation, AroundInvokeContext {
 
     public ComponentContext context;
 
     private TransactionOperationsManager transactionOperationsManager;
 
-    EjbInvocation(String compEnvId, Container container) {
```
dmatej: Constructor was moved behind fields. Too many fields. This class calls for refactoring and splitting.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -282,6 +264,38 @@
     // True if lock is currently held for this invocation
     private boolean holdingSFSBSerializedLock = false;
 
+    private int interceptorIndex;
+
+    public Method beanMethod;
+
+    // Only set for web service invocations.
+    private WebServiceContext webServiceContext;
```
dmatej: WARNING: this can be null. The EjbInvocation contains several overlapping implementations.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -282,6 +264,38 @@
...
+
+    // Only set for EJB JAXWS
+    // FIXME: private Message message = null;
+    private Object message;
```
dmatej: I said overlapping? Somebody tried to upgrade it, but ... ended in half of the path. Don't continue without tests, because it has processing stages.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -649,94 +656,100 @@ public String toString() {
...
-                }
-            } catch(Exception e) {
-                String errorMsg = "Error unmarshalling method {" + method + "} for ejb ";
-                ie = new UnmarshalException(errorMsg);
-                ie.initCause(e);
-            }
-            if ( ie != null ) {
-                exception = ie;
-                throw ie;
+        if (!isAWebService()) {
```
dmatej: End execution if you cannot continue - then you don't need `else`.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSessionContextImpl.java
```diff
@@ -85,15 +81,15 @@ public void setInstanceKey(Object instanceKey) {
 
     @Override
     public String toString() {
-        return ejbName + "; id: " + instanceKey;
+        return ejbName + "; id: " + getInstanceKey();
```
dmatej: It is weird, but instance key has overridable setter and getter. To be refactored later.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -94,8 +96,8 @@
     // Data members for Remote business view. Any objects representing the
     // Remote business interface are not subtypes of EJBObject.
     private EJBObjectImpl theRemoteBusinessObjectImpl = null;
-    private Map<String, java.rmi.Remote> theRemoteBusinessStubs =
-        new HashMap<String, java.rmi.Remote>();
+    private final Map<String, java.rmi.Remote> theRemoteBusinessStubs =
```
dmatej: Note full name of java.rmi.Remote - the ejb Remote annotation is not used here, but we are still in ejb module and if we would convert this to an import, it would confuse readers.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -490,7 +507,7 @@ private SingletonContextImpl createSingletonEJB()
                     }
                 }
                 else {
-                    context = (SingletonContextImpl)_constructEJBContextImpl(singletonMap.get(sessionKey));
+                    context = _constructEJBContextImpl(singletonMap.get(sessionKey));
```
dmatej: Typed map now.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -634,46 +653,48 @@ public boolean userTransactionMethodsAllowed(ComponentInvocation inv) {
...
     // default
-    public void activateEJB(Object ctx, Object instanceKey) {}
-
-/** TODO
-    public void appendStats(StringBuilder sbuf) {
-	sbuf.append("\nSingletonContainer: ")
-	    .append("CreateCount=").append(statCreateCount).append("; ")
-	    .append("RemoveCount=").append(statRemoveCount).append("; ")
-	    .append("]");
+    public void activateEJB(Object ctx, Object instanceKey) {
```
dmatej: Warning - there is just one method with same name but with three arguments (in other classes). But this one with two is not used.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -465,7 +457,7 @@
     // an EJB Remote interface to be a subtype of the Service Endpoint
     // Interface.  In that case, it's ambiguous to do a lookup based only
     // on a java.lang.reflect.Method
-    protected Map webServiceInvocationInfoMap = new HashMap();
+    protected Map<Method, InvocationInfo> webServiceInvocationInfoMap = new HashMap<>();
```
dmatej: Now I see, it should be final. But it is protected, so Eclipse was afraid that it could break something. It could, but not in Payara.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -966,20 +958,17 @@ public final void setStartedState() {
...
                 setInterceptorChain(next);
             }
-            for(Object o : this.webServiceInvocationInfoMap.values()) {
+            for (Object o : this.webServiceInvocationInfoMap.values()) {
                 InvocationInfo next = (InvocationInfo) o;
                 setInterceptorChain(next);
             }
-
-        } catch(Exception e) {
-            throw new RuntimeException(e);
```
dmatej: Don't use RuntimeException. What is it? Where we are and why? At least ISE is better - we are where we should not be. Why? I don't know. Wouldn't be better EjbException? But it has it's own processing it could change something. Tests required to do more.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -1816,49 +1765,46 @@ protected boolean isJCDIEnabled() {
         return ejbInfo;
     }
 
+    /**
+     * Must be overridden for any container that supports injection
+     *
+     * @param instance
+     * @return {@link EJBContextImpl}
+     */
     protected EJBContextImpl _constructEJBContextImpl(Object instance) {
```
dmatej: Classical example - comments in code often can be changed to javadoc.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -2186,27 +2132,28 @@ protected void postInvoke(EjbInvocation inv, boolean doTxProcessing) {
             releaseContext(inv);
         }
 
-        if ( inv.exception != null ) {
+        if (inv.exception != null) {
 
             // Unwrap the PreInvokeException if necessary
-            if ( inv.exception instanceof PreInvokeException ) {
-                inv.exception = ((PreInvokeException)inv.exception).exception;
```
dmatej: I changed the tiny PreInvokeException wrapper to use standard mechanism - it already is wrapper.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -2275,27 +2219,23 @@ protected void authorizeLocalMethod(int method) {
...
-    {
+    protected void authorizeRemoteMethod(int method) throws RemoteException {
         EjbInvocation inv = invFactory.create();
         inv.isLocal = false;
         inv.isHome = EJB_INTF_METHODS_INFO[method];
         inv.method = ejbIntfMethods[method];
         inv.invocationInfo = ejbIntfMethodInfo[method];
 
         if ( !authorize(inv) ) {
-            // TODO see note above about additional special exception handling needed
```
dmatej: Which note?

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -2362,103 +2302,81 @@ private void processEjbTimeoutMethod(Method method) throws Exception {
...
-
+    private Throwable mapRemoteException(final Throwable remoteException, final boolean async,
+        final boolean businessIface, boolean remoteClient) {
+        final Throwable originalException = remoteException;
         // If it's an asnyc invocation and we're mapping an exception it
         // means this is the thread of execution.  The exception won't directly
         // flow over the wire as a remote exception from the orb's perspective.
         // If it's asychronous we know it's a remote business interface, not the
         // 2.x client view.
-        if ( inv.invocationInfo.isAsynchronous() ) {
```
dmatej: Just another if/else excercise (Eclipse helps a lot!)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -2943,16 +2850,12 @@ public void assertValidRemoteObject(Object o) throws EJBException
...
     protected final int getTxAttr(Method method, String methodIntf) throws EJBException {
-
-        InvocationInfo invInfo = methodIntf.equals(EJB_WEB_SERVICE) ?
-                (InvocationInfo) webServiceInvocationInfoMap.get(method)
-                : (InvocationInfo) invocationInfoMap.get(method);
-
-        if (invInfo != null) {
-            return invInfo.txAttr;
-        } else {
+        InvocationInfo invInfo = methodIntf.equals(EJB_WEB_SERVICE) ? webServiceInvocationInfoMap.get(method)
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -4744,21 +4627,15 @@ public void onTermination() {}
...
-
+    protected final void preInvokeTx(EjbInvocation inv) throws Exception {
+        if (inv.invocationInfo == null) {
             inv.invocationInfo = getInvocationInfo(inv);
-
-            if ( inv.invocationInfo == null ) {
-                throw new EJBException("EjbInvocation Info lookup failed for " +
-                                       "method " + inv.method);
+            if (inv.invocationInfo == null) {
+                throw new EJBException("EjbInvocation Info lookup failed for " + "method " + inv.method);
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -4861,24 +4738,40 @@ protected void postInvokeNoTx(EjbInvocation inv) {
...
+    /**
+     * Checks if the exception is not null and if it is checked exception
+     * - uses ejb descriptor to resolve it.
+     *
+     * @param exception can be null
+     * @return true if the exception is not null and if it is not an unchecked exception
+     */
     protected boolean isApplicationException(Throwable exception) {
-        return !isSystemUncheckedException(exception);
+        return exception != null && !isSystemUncheckedException(exception);
```
dmatej: All calls of this method checked null first. The method should have the same "convention" as isSystemUncheckedException, so at least argument should be an exception at least. Null is not an exception and negation was not enough.
Bonus: I could remove all those null checks in calls.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -5244,13 +5137,15 @@ private static boolean isCosNamingObject(Object obj) {
      * PreInvokeException is used to wrap exceptions thrown
      * from BaseContainer.preInvoke, so it indicates that the bean's
      * method will not be called.
+     * <p>
+     * Use usual {@link PreInvokeException#getCause()} to get wrapped exception.
      */
     public final static class PreInvokeException extends EJBException {
 
-        Exception exception;
```
dmatej: No need for another field.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -96,12 +90,8 @@
         container = (BaseContainer)c;
         ejbDescriptor = ejbDesc;
         transactionManager = ejbContainerUtilImpl.getTransactionManager();
-
-        IASEjbExtraDescriptors iased = ejbDesc.getIASEjbExtraDescriptors();
-
-        if(iased.getCmtTimeoutInSeconds() != 0){
```
dmatej: Original code: set it to 0 and if something change the zero to another number.
Now: if something, set it to a value and to zero otherwise. And it is final.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -110,63 +100,63 @@
...
-            throw new EJBException("Transaction Attribute not found for method "
-                + md.prettyPrint());
-        }
-
         container.validateTxAttr(md, txAttr);
-
         return txAttr;
     }
 
+    private int getTransactionAttribute(final MethodDescriptor md) {
```
dmatej: This split allowed to use a single assignment to a final. It is also much more comprehensible - one method only looks for a TransactionAttribute value and the caller resolves how to use it.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -110,63 +100,63 @@
...
      */
     final void preInvokeTx(EjbInvocation inv) throws Exception {
+        LOG.finest(() -> String.format("preInvokeTx(inv=%s)", inv));
         // Get existing Tx status: this tells us if the client
         // started a transaction which was propagated on this invocation.
-        Integer preInvokeTxStatus = inv.getPreInvokeTxStatus();
-        int status = (preInvokeTxStatus != null) ?
-            preInvokeTxStatus.intValue() : transactionManager.getStatus();
+        final Integer preInvokeTxStatus = inv.getPreInvokeTxStatus();
+        final int status = (preInvokeTxStatus != null) ? preInvokeTxStatus.intValue() : transactionManager.getStatus();
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -331,42 +320,45 @@ private void startNewTx(Transaction prevTx, EjbInvocation inv) throws Exception
...
         } catch (Exception ex) {
+            final TransactionRolledbackLocalException toThrow = new TransactionRolledbackLocalException(ex.getMessage(), ex);
             try {
                 transactionManager.setRollbackOnly();
-            } catch ( Exception e ) {
-		//FIXME: Use LogStrings.properties
-		_logger.log(Level.FINEST, "", e);
-	    }
-            throw new TransactionRolledbackLocalException("", ex);
+            } catch (Exception e) {
```
dmatej: Logging of exception as finest (so practically ignoring it) replaced with the usage of suppressed. It will be visible in a stacktrace.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -428,46 +419,42 @@ protected void useClientTx(Transaction prevTx, EjbInvocation inv) {
...
-        Throwable newException = exception; // default
+        final int status = transactionManager.getStatus();
+        final int txAttr = inv.invocationInfo.txAttr;
+        final Throwable oriException = inv.exception;
 
         // Note: inv.exception may have been thrown by the container
         // during preInvoke (i.e. bean may never have been invoked).
 
         // Exception and Tx handling rules. See EJB2.0 Sections 17.6, 18.3.
+        final Throwable newException;
```
dmatej: Note: it is assigned only one. It is little bit more code, but it is still more comprehensible than it was before. Only one assignment.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBLocalHomeInvocationHandler.java
```diff
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
```
dmatej: Note: Unreadable diff.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbThreadPoolExecutor.java
```diff
@@ -86,35 +88,35 @@ public String toString() {
...
+                final int st = tm.getStatus();
+                logger.warning(() -> "NON-NULL TX IN AFTER_EXECUTE. TX STATUS: " + st);
+                if (st == Status.STATUS_ROLLEDBACK || st == Status.STATUS_COMMITTED || st == Status.STATUS_UNKNOWN) {
                     tm.clearThreadTx();
                 } else {
                     tm.rollback();
                 }
             }
         } catch (Exception e) {
-            e.printStackTrace();
```
dmatej: STDOUT -> server.log? Replaced with JUL.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/SessionContextImpl.java
```diff
@@ -324,40 +310,44 @@ public void setVersion(long newVersion) {
...
     public boolean existsInStore() {
         return existsInSessionStore ;
     }
 
+    @Override
     public void setExistsInStore(boolean val) {
         this.existsInSessionStore = val;
     }
 
-    public final void incrementRefCount() {
```
dmatej: The class is final, so this was redundant.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -375,70 +406,60 @@ public void startApplication(boolean deploy) {
...
-            new java.security.PrivilegedExceptionAction() {
-                public java.lang.Object run() throws Exception {
-                    if( !methodAccessible.isAccessible() ) {
-                        methodAccessible.setAccessible(true);
-                    }
-                    return null;
-                }
-         });
-
+        final PrivilegedExceptionAction<Void> action = () -> {
```
dmatej: Note: this is possible since Java8, no warnings and less code.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -454,31 +475,27 @@ protected void checkUnfinishedTx(Transaction prevTx, EjbInvocation inv) {
...
                         }
                     }
 
                     if (info.checkpointEnabled) {
                         if (_logger.isLoggable(Level.FINE)) {
-                            _logger.log(Level.FINE, "[SFSBContainer] "
-                                    + info.method + " MARKED for "
-                                    + "end-of-method-checkpoint");
+                            _logger.log(Level.FINE,
+                                "[SFSBContainer] " + info.method + " MARKED for " + "end-of-method-checkpoint");
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -544,22 +557,7 @@ protected EjbMonitoringStatsProvider getMonitoringStatsProvider(
         return statsProvider;
     }
 
-
```
dmatej: TODO died forgotten.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -786,11 +746,7 @@ private SessionContextImpl createBeanInstance()
             sessionBeanCache.put(sessionKey, context);
             context.setInstanceKey(sessionKey);
 
-
-            if (_logger.isLoggable(TRACE_LEVEL)) {
```
dmatej: TRACEL_LEVEL was masked FINE level.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1195,30 +1127,24 @@ protected void removeBean(EJBLocalRemoteObject ejbo, Method removeMethod,
...
                 // a transaction.
                 throw new RemoveException("Cannot remove EJB: transaction in progress");
             }
 
             // call ejbRemove on the EJB
-            if (_logger.isLoggable(TRACE_LEVEL)) {
-                _logger.log(TRACE_LEVEL, "[SFSBContainer] Removing "
-                        + "session: " + sc.getInstanceKey());
+            if (_logger.isLoggable(FINE)) {
+                _logger.log(FINE, "[SFSBContainer] Removing " + "session: " + sc.getInstanceKey());
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1253,28 +1176,27 @@ private void removeBean(EjbInvocation inv)
...
             // mark context as destroyed so no more invocations happen on it
             sc.setState(BeanState.DESTROYED);
-
             cleanupInstance(ctx);
 
-            if (_logger.isLoggable(TRACE_LEVEL)) {
-                _logger.log(TRACE_LEVEL, "[SFSBContainer] (Force)Destroying "
-                        + "session: " + sc.getInstanceKey());
+            if (_logger.isLoggable(FINE)) {
+                _logger.log(FINE, "[SFSBContainer] (Force)Destroying " + "session: " + sc.getInstanceKey());
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1424,10 +1326,9 @@ public void removeTimedoutBean(EJBContextImpl ctx) {
                     ctx.setInEjbRemove(false);
                 }
 
-                if (_logger.isLoggable(TRACE_LEVEL)) {
+                if (_logger.isLoggable(FINE)) {
                     SessionContextImpl sc = (SessionContextImpl) ctx;
-                    _logger.log(TRACE_LEVEL, "[SFSBContainer] Removing TIMEDOUT "
-                            + "session: " + sc.getInstanceKey());
+                    _logger.log(FINE, "[SFSBContainer] Removing TIMEDOUT " + "session: " + sc.getInstanceKey());
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1550,58 +1444,42 @@ EJBLocalObjectImpl getOptionalEJBLocalBusinessObjectImpl(Object sessionKey) {
...
      * Check if the given EJBObject/LocalObject has been removed.
      *
      * @throws NoSuchObjectLocalException if the object has been removed.
      */
+    @Override
     protected void checkExists(EJBLocalRemoteObject ejbObj) {
-        if (ejbObj.isRemoved())
+        if (ejbObj.isRemoved()) {
             throw new NoSuchObjectLocalException("Bean has been removed");
-    }
```
dmatej: Pushed down in the same class

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -2750,71 +2497,69 @@ private void repopulateEEMMapsInContext(Object sessionKey,
...
 
             // Now see if there's already a *different* extended
             // persistence context within this transaction for the
             // same EntityManagerFactory.
-            PhysicalEntityManagerWrapper physicalEM = (PhysicalEntityManagerWrapper) clientJ2EETx.getExtendedEntityManagerResource(emf);
-            if( (physicalEM != null) && entry.getValue().getEM() != physicalEM.getEM() ) {
-                throw new EJBException("Detected two different extended persistence contexts for the same EntityManagerFactory within a transaction");
+            PhysicalEntityManagerWrapper physicalEM = (PhysicalEntityManagerWrapper) clientJ2EETx
+                .getExtendedEntityManagerResource(emf);
+            if ((physicalEM != null) && entry.getValue().getEM() != physicalEM.getEM()) {
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/WebServiceInvocationHandler.java
```diff
@@ -99,27 +89,44 @@ public WebServiceInvocationHandler(Class ejbClass,
...
-                invokeJavaObjectMethod(this, method, args);    
+            container.onEnteringContainer();
+            // NOTE : be careful with "args" parameter. It is null
+            // if method signature has 0 arguments.
+            final Class<?> methodClass = method.getDeclaringClass();
+            if (methodClass == java.lang.Object.class) {
+                return InvocationHandlerUtil.invokeJavaObjectMethod(this, method, args);
+            }
+
+            // Invocation was created earlier in the web service dispatching
```
dmatej: Question to github: where is the - section?

dmatej: Aha, response: I extracted part of the method to a standalone private method. That confused even author as a reviewer :dancing_men:

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/WebServiceInvocationHandler.java
```diff
@@ -129,96 +136,76 @@ public Object invoke(Object proxy, Method method, Object[] args)
...
-            inv.invocationInfo = (InvocationInfo)
-                invocationInfoMap_.get(inv.method);
+            inv.invocationInfo = (InvocationInfo) invocationInfoMap_.get(inv.method);
 
-            if( inv.invocationInfo == null ) {
-                throw new EJBException
-                    ("Web service Invocation Info lookup failed for " +
-                     "method " + inv.method);
+            if (inv.invocationInfo == null) {
+                throw new EJBException("Web service Invocation Info lookup failed for " + "method " + inv.method);
```
dmatej: FIXME

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/WebServiceInvocationHandler.java
```diff
@@ -129,96 +136,76 @@ public Object invoke(Object proxy, Method method, Object[] args)
...
+            return container.intercept(inv);
+        } catch (NoSuchMethodException nsme) {
             inv.exception = nsme;
-        } catch(InvocationTargetException ite) {
+            return null;
+        } catch (InvocationTargetException ite) {
             inv.exception = ite.getCause();
-        } catch(Throwable c) {
+            return null;
+        } catch (Throwable c) {
```
dmatej: Catching Throwable is a bad idea. Always. Even in application servers. Maybe we should give it some attention in Payara 6.

appserver/ejb/ejb-full-container/src/main/java/org/glassfish/ejb/mdb/MessageBeanContainer.java
```diff
@@ -748,7 +754,7 @@ private static boolean isNoMethodsListenerInterface(Class<?> messageListenerType
     }
 
     @Override
-    protected EJBContextImpl _constructEJBContextImpl(Object instance) {
+    protected MessageBeanContextImpl _constructEJBContextImpl(Object instance) {
```
dmatej: API change. But this is protected scope and classes are final. Maybe I could do even more changes.

appserver/ejb/ejb-internal-api/src/main/java/org/glassfish/ejb/api/EJBInvocation.java
```diff
@@ -47,83 +47,99 @@
...
-
-
 public interface EJBInvocation {
 
-    public EJBContext getEJBContext();
-    
+    /**
+     * @return runtime {@link EJBContext} of this invocation
+     */
+    EJBContext getEJBContext();
```
dmatej: In Java all methods in interfaces are public by default and nobody can change it.
I added also some javadocs.

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionImpl.java
```diff
@@ -198,26 +213,29 @@ private void cancel() {
         }
     }
 
+    @Override
     public boolean isTimedOut() {
         return timedOut;
     }
     // END: local transaction timeout
 
-    private static synchronized long getNewTxId() {
```
dmatej: No synchronization needed with AtomicLong

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionImpl.java
```diff
@@ -679,29 +721,32 @@ public void registerSynchronization(Synchronization sync)
...
             syncs.add(sync);
+        } else {
+            jtsTx.registerSynchronization(sync);
+        }
     }
 
+    @Override
     public void setRollbackOnly() throws IllegalStateException, SystemException {
         checkTransationActive();
-        if ( jtsTx != null )
```
dmatej: Say no to if/else in negation!

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1893,31 +1745,25 @@ protected void postInvokeTx(EjbInvocation inv) throws Exception {
...
+    @Override
     public void releaseContext(EjbInvocation inv) {
         SessionContextImpl sc = (SessionContextImpl) inv.context;
 
         // Make sure everything is within try block so we can be assured that
         // any instance lock is released in the finally block.
         try {
             // check if the bean was destroyed
-            if (sc.getState() == BeanState.DESTROYED)
+            if (sc.getState() == BeanState.DESTROYED) {
```
jbee: Maybe `sc.getState()` should be extracted to a local var to be deterministic throughout the method?

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -3191,26 +2907,23 @@ public void trimEvent(Object primaryKey, Object context) {
...
-                            (new java.security.PrivilegedAction() {
-                                public java.lang.Object run() {
-                                    currentThread.setContextClassLoader(myClassLoader);
-                                    return null;
-                                }
-                            });
+                    PrivilegedAction<Void> action = () -> {
+                        currentThread.setContextClassLoader(myClassLoader);
+                        return null;
+                    };
```
jbee: Shouldn't that also be possible as 

```java
PrivilegedAction<Void> action = () -> currentThread.setContextClassLoader(myClassLoader);
```
?

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatelessSessionContainer.java
```diff
@@ -111,43 +118,37 @@
     // Data members for Remote business view. Any objects representing the
     // Remote business interface are not subtypes of EJBObject.
     private EJBObjectImpl theRemoteBusinessObjectImpl = null;
-    private Map<String, java.rmi.Remote> theRemoteBusinessStubs =
-        new HashMap<String, java.rmi.Remote>();
+    private final Map<String, Remote> theRemoteBusinessStubs = new HashMap<>();
```
jbee: another non concurrent map where I'd expect something else. hm.

appserver/persistence/entitybean-container/src/main/java/org/glassfish/persistence/ejb/entitybean/container/EntityContainer.java
```diff
@@ -504,55 +522,46 @@ public void appendStats(StringBuilder sbuf) {
...
-	if (readyStore != null) {
-	    maxSize = (cacheProp.maxCacheSize <= 0)
-		? Integer.MAX_VALUE
-		: cacheProp.maxCacheSize;
-	}
-
-	return maxSize;
+        int maxSize = 0;
+        if (readyStore != null) {
+            maxSize = (cacheProp.maxCacheSize <= 0) ? Integer.MAX_VALUE : cacheProp.maxCacheSize;
```
jbee: why not return that directly?

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionImpl.java
jbee: It all depends. Readability is not about following a set of rules. Its a property of the whole. There might be cases where a negation reads better and nobody should start asking to apply some rule. Here on the other hand flipping makes it more readable :)

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
```diff
@@ -714,32 +742,38 @@ public void componentDestroyed(Object instance, ComponentInvocation inv) {
...
+    @Override
     public boolean isTimedOut() {
         JavaEETransaction tx = transactions.get();
-        if ( tx != null)
+        if ( tx != null) {
             return tx.isTimedOut();
-        else
+        } else {
             return false;
+        }
```
jbee: Please...

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
```diff
@@ -773,16 +808,19 @@ public void checkTransactionExport(boolean isLocal) {
         }
 
         JavaEETransaction tx = transactions.get();
-        if ( tx == null )
+        if ( tx == null ) {
             return;
+        }
 
-        if ( !tx.isLocalTx() ) // a JTS tx, can be exported
+        if ( !tx.isLocalTx() ) {
```
jbee: Merge with block above?

appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
```diff
@@ -412,6 +419,7 @@ public String toString() {
      * @param state CREATED or Either STARTED or COMPLETED or TIMEDOUT
      */
     public synchronized void setState(int state) {
+        logger.finest(() -> String.format("setState(state=%s)", state));
```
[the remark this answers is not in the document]
dmatej: I would defend suppliers, because the code is optimized by JVM and I believe it is still faster than if block with JUL formatter; especially when you enable finest.
But we can argue about nanoseconds, the JUL logging is extremely slow by default because of poor implementation of handlers and formatters, concatenating strings, etc.
If I remember well, the lambda here is created pretty fast, because it is not generated in runtime b some other method, it uses some template and even JIT can inline it.

But for sure, I will create some test for that, because I don't believe too much what I wrote :)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -75,32 +84,14 @@
...
  */
 
-public class EjbInvocation
-    extends ComponentInvocation
-    implements InvocationContext, TransactionOperationsManager,
-         org.glassfish.ejb.api.EJBInvocation, InterceptorManager.AroundInvokeContext
-{
-
+public class EjbInvocation //
+    extends ComponentInvocation //
```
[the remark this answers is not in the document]
dmatej: In fact I don't know any other option to have stable formatting. This only prevents manually formatted code before automatic reformatting which will someone do and would not notice that something breaks when he tries to fix something else ...

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -554,12 +561,14 @@ public void setWebServiceContext(Object webServiceContext) {
...
+    @Override
     public Map<String, Object> getContextData() {
         if (this.contextData == null) {
-            if (webServiceContext != null)
+            if (webServiceContext != null) {
                 this.contextData = webServiceContext.getMessageContext();
-            else
-                this.contextData = new HashMap<String, Object>();
+            } else {
+                this.contextData = new HashMap<>();
```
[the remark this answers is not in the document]
dmatej: Note that I had to be very fast - I could not spend too much time with every line. I'm sure we will need to come back to nearly every module, but all this I made in several hours before moving to something requested by management.
Braces were automatic, I only trid to read those thousands lines after.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
```diff
@@ -649,94 +656,100 @@ public String toString() {
...
+            return true;
+        }
+        Exception ie;
+        try {
+            this.method = m;
+            if (((com.sun.ejb.Container) container).authorize(this)) {
+                // Record the method on which the successful
+                // authorization check was performed.
+                setWebServiceMethod(m);
+                ie = null;
```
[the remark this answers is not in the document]
dmatej: Now you forced me to switch branches again :D
(I already wanted to do that ;) )

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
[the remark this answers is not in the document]
dmatej: Well, this is surprising even for me ...
Junit test:

* beforeclass three logs to wakeup JUL manager, 
* always logging a message with one parameter which is always changed in each iteration (required to be assigned to a final variable for supplier)
* 1000 iterations warmup, 
* 100000000 iterations test
* logger level set to INFO

Results:

* finest (-> skipped), 
```
concat: 
  warmup: 1 ms
   test: 3150 ms
julformatter: .... this is really bad
  warmup: 0 ms
   test: 321 ms
julformatterBehindIf: 
  warmup: 0 ms
   test: 82 ms
supplierAndConcat: 
  warmup: 1 ms
   test: 65 ms
supplierAndStringFormat: 
  warmup: 0 ms
   test: 61 ms
```

* info (-> output to a console, only 10000 iterations):

concat: 
  warmup: 100-155 ms
   test: 387-458 ms
julformatter: 
  warmup: 58-63 ms
   test: 463-480 ms
julformatterBehindIf: 
  warmup: 21-33 ms
   test: 416-497 ms
supplierAndConcat: 
  warmup: 24-29 ms
   test: 380-421 ms
supplierAndStringFormat: 
  warmup: 34-67 ms
   test: 379-412 ms

So I think the code in the patch is clear winner if it is not enabled, if it is, it nearly does not matter (I will close both eyes for conditioned jul formatter).

And more, it seems JUL formats the message before sending it to a log formatter, so parameters slow down the processing even when the log level is too low. That is strictly opposite as in SLF4J and LOG4J2, even SLF4J and ald LOG4J1 use "logging events" and message formatting is deferred as is is possible ...
Blah! We should use anything else than JUL :-(
it seems JUL formats the message before sending it to a log forma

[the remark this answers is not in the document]
dmatej: OpenJDK8 via Maven

* level finest (skipped output)
```
concat: 
  warmup: 1/1 ms
   test: 4198/3460 ms
julformatter: 
  warmup: 1/1 ms
   test: 480/339 ms
julformatterBehindIf: 
  warmup: 0 ms
   test: 85/81 ms
supplierAndConcat: 
  warmup: 2/3 ms
   test: 88/85 ms
supplierAndStringFormat: 
  warmup: 1/1 ms
   test: 96/89 ms
```

* level info (with log output):
```
concat: 
  warmup: 334/188 ms
   test: 756/509 ms
julformatter: 
  warmup: 165/70 ms
   test: 1940/1284 ms
julformatterBehindIf: 
  warmup: 37/27 ms
   test: 2421/1909 ms
supplierAndConcat: 
  warmup: 43/26 ms
   test: 2282/1824 ms
supplierAndStringFormat: 
  warmup: 59/57 ms
   test: 2598/1892 ms
```

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSessionContextImpl.java
dmatej: Actually it does not matter in real time, but it is an abstract class and it should expect that this method would be overriden.
This is very aften case when I always have to override also toString method whenever I override any getter. Generally it depends on the whole concept, ie. equals can use getters; if it doesn't toString should not do it too to avoid confusion. But even this is not a law, sometimes toString prints only an id, sometimes classname and hashcode ...
Also - the old version - if anyone would override the getter, parent would still use the field and child would use different value.

We could also make getInstanceKey final, but ... it is an Object, it has both setter and getter ... I think this is better.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
dmatej: True, there is too much synchronization (or none here). But I'm afraid of that, TCK is sequential and slow, but I'm not sure with applications with thousands of threads. The code is not much well prepared for concurrency and we don't have much tests for that.
I will try to create some later ...

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
dmatej: ISE is suitable exactly for situations where we reach some illegal state, something is broken and there is no prepared reaction.
I don't change anything without searching - in this case, there are only Exception and Throwable catches. Using instanceof for ISE would be a crime (but yes, catching Throwable is a crime too).

All exceptions from here will end up in this ugly catch block:
```
            } catch (Throwable loadException) {
                logger.log(Level.SEVERE, KernelLoggerInfo.lifecycleException, loadException);
                report.failure(logger, "Exception while loading the app", null);
                report.setFailureCause(loadException);
                tracker.actOn(logger);
            } finally {
```

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
dmatej: I made code review also to myself :D

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
dmatej: concat

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
dmatej: inverted

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
dmatej: I finally moved these checks into the methods in else branches

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionImpl.java
dmatej: This is not the case. Maybe I should use smileys :)
Usually negation is maybe most usual cause of misunderstandings. Especially when I am tired and it is used in a sequence of ternary operators ...

appserver/connectors/work-management/src/main/java/com/sun/enterprise/connectors/work/WorkCoordinator.java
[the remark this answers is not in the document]
dmatej: See test results I posted somewhere in the begining. It surprised even me - you can safely use lambdas where posiible, I'm pretty sure that if you don't use a method to generate new lambda, it is still the same.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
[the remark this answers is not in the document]
dmatej: Not true, a tthat point it is unset, not null. Both old and new code uses same trick - using exception object, but throwing it only in the end. I agree it is still ugly.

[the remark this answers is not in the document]
dmatej: I splitted it into another method, still not perfect but we can use `final` at least.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -94,8 +96,8 @@
     // Data members for Remote business view. Any objects representing the
     // Remote business interface are not subtypes of EJBObject.
     private EJBObjectImpl theRemoteBusinessObjectImpl = null;
-    private Map<String, java.rmi.Remote> theRemoteBusinessStubs =
-        new HashMap<String, java.rmi.Remote>();
+    private final Map<String, java.rmi.Remote> theRemoteBusinessStubs =
+        new HashMap<>();
```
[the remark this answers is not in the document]
dmatej: It seems these containers are initialized by a single thread but then used by multiple threads only for reading (at least some parts). So it is from some point of view correct (we would see bugs here for a long time if not).

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -451,8 +468,8 @@ public ComponentContext instantiateSingletonInstance() {
     }
 
     @Override
-    protected EJBContextImpl _constructEJBContextImpl(Object instance) {
-	return new SingletonContextImpl(instance, this);
+    protected SingletonContextImpl _constructEJBContextImpl(Object instance) {
```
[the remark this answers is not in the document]
dmatej: Benefit is removing class unchecked class casting in some places.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/AbstractSingletonContainer.java
```diff
@@ -683,72 +704,46 @@ protected void doConcreteContainerShutdown(boolean appBeingUndeployed) {
...
-                        remoteBusinessIntfInfo.values()) {
-                    next.referenceFactory.destroyReference
-                        (theRemoteBusinessObjectImpl.getStub
-                            (next.generatedRemoteIntf.getName()),
-                         theRemoteBusinessObjectImpl.getEJBObject
-                            (next.generatedRemoteIntf.getName()));
+            if (hasRemoteBusinessView) {
+                for (RemoteBusinessIntfInfo next : remoteBusinessIntfInfo.values()) {
+                    next.referenceFactory.destroyReference(
+                        theRemoteBusinessObjectImpl.getStub(next.generatedRemoteIntf.getName()),
```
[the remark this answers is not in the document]
dmatej: It is a field. But the class is not serializable, so yeah, I think we can do that!

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -966,20 +958,17 @@ public final void setStartedState() {
...
-
             initializeInterceptorManager();
 
-            for(Object o : invocationInfoMap.values()) {
-                InvocationInfo next = (InvocationInfo) o;
+            for (InvocationInfo next : invocationInfoMap.values()) {
                 setInterceptorChain(next);
             }
-            for(Object o : this.webServiceInvocationInfoMap.values()) {
+            for (Object o : this.webServiceInvocationInfoMap.values()) {
```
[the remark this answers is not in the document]
dmatej: Done

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -1256,37 +1245,31 @@ protected void initializeHome()
...
 
             invocationHandler.setContainer(this);
-            Object servant = Proxy.newProxyInstance
-                    (loader, new Class[] { serviceEndpointIntfClass },
-                    invocationHandler);
+            Object servant = Proxy.newProxyInstance(loader, new Class[] {serviceEndpointIntfClass}, invocationHandler);
 
             // starting in 2.0, there is no more generated Ties
-            if (webServiceEndpoint.getTieClassName()!=null) {
+            final Class<?> tieClass;
```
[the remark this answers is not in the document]
dmatej: I don't like multiline ternary operators. So I extracted it.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -2362,103 +2302,81 @@ private void processEjbTimeoutMethod(Method method) throws Exception {
...
         }
+        return protocolException;
+    }
 
-        return (mappedException != null) ? mappedException : t;
-
+    private Throwable mapLocal3xException(final Throwable t) {
+        if (t instanceof TransactionRolledbackLocalException) {
+            return new EJBTransactionRolledbackException(t.getMessage(), (Exception) t);
+        } else if (t instanceof TransactionRequiredLocalException) {
```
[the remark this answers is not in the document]
dmatej: I think it is same clear, but I agree that optically are standalone ifs nicer :)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -3178,18 +3076,12 @@ private InvocationInfo addInvocationInfo(Method method, String methodIntf,
...
-                        EJBLocalObject.class.isAssignableFrom(originalIntf));
-            eligibleForAsync = !is2xClientView;
+        if (methodIntf.equals(MethodDescriptor.EJB_LOCAL) || methodIntf.equals(MethodDescriptor.EJB_REMOTE)) {
+            // ejb1/2
+            return !EJBObject.class.isAssignableFrom(originalIntf) //
+                && !EJBLocalObject.class.isAssignableFrom(originalIntf);
         }
-
-        return eligibleForAsync;
+        return false;
```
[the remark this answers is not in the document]
dmatej: I dislike like both versions same and the third too ...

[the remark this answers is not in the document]
dmatej: Two pairs of booleans mixed in AND + OR.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -110,63 +100,63 @@
...
+            return Container.TX_SUPPORTS;
+        } else if (attr.equals(ContainerTransaction.REQUIRED)) {
+            return Container.TX_REQUIRED;
+        } else if (attr.equals(ContainerTransaction.REQUIRES_NEW)) {
+            return Container.TX_REQUIRES_NEW;
+        } else if (attr.equals(ContainerTransaction.MANDATORY)) {
+            return Container.TX_MANDATORY;
+        } else if (attr.equals(ContainerTransaction.NEVER)) {
+            return Container.TX_NEVER;
+        } else {
```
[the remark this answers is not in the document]
dmatej: In this case I don't believe it would be better. I would like to use enums instead of strings and ints. Too far to go.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -110,63 +100,63 @@
...
+            return -2;
```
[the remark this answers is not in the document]
dmatej: It is only a dummy result, anything under a zero is an error and would cause an EJBException without this value. So it does not matter, maybe with debugger.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -181,32 +171,33 @@ final void preInvokeTx(EjbInvocation inv) throws Exception {
         // (i.e. a tx context with a null Coordinator objref)
         // or if this server's tx interop mode flag is false.
         // Follow the tables in EJB2.0 sections 19.6.2.2.1 and 19.6.2.2.2.
-        boolean isNullTx = false;
+        final boolean isNullTx;
```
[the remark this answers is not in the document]
dmatej: Ok, ternary instead of if/else.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -294,7 +283,7 @@ final void preInvokeTx(EjbInvocation inv) throws Exception {
      * no-op in those containers that do not need this callback
      */
     private void startNewTx(Transaction prevTx, EjbInvocation inv) throws Exception {
-
+        LOG.finest(() -> String.format("startNewTx(prevTx=%s, inv=%s)", prevTx, inv));
```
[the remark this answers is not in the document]
dmatej: Nope. It would only slow down. See test result.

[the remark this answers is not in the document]
dmatej: (another surprise was that String.format is not so fast as I thought, but Suppliers are)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
```diff
@@ -388,37 +380,36 @@ protected void useClientTx(Transaction prevTx, EjbInvocation inv) {
...
                 try {
                     transactionManager.setRollbackOnly();
-                } catch ( Exception e ) {
-					//FIXME: Use LogStrings.properties
-					_logger.log(Level.FINEST, "", e);
+                } catch (Exception e) {
+                    toThrow.addSuppressed(e);
 				}
-                throw new TransactionRolledbackLocalException("", ex);
+                throw toThrow;
```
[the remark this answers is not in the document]
dmatej: It is tiny change - and it passed TCK at night. Except one test broken also in master. That's my next issue.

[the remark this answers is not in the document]
dmatej: These changes are because EJB container sometimes drops exceptions. As you have seen, it will need much more work, but I'm pretty sure this will not break anything.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -302,33 +336,28 @@
...
 
     /*
      * Cache for keeping ref count for shared extended entity manager.
      * The key in this map is the physical entity manager
      */
 
-    private static final Map<EntityManager, EEMRefInfo> extendedEMReferenceCountMap
-            = new HashMap<EntityManager, EEMRefInfo>();
+    private static final Map<EntityManager, EEMRefInfo> extendedEMReferenceCountMap = new HashMap<>();
+    private static final Map<EEMRefInfoKey, EntityManager> eemKey2EEMMap = new HashMap<>();
```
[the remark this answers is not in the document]
dmatej: Concurrent maps are not silver bullet. But I too wonder how this module can work safely for so many years. Except it swallows exceptions when there is already some processed. From Java 7 it can be added as suppressed so we don't lose information.
I will not change it until I would have proof it is broken. JMH + TestContainers will do the job. Or maybe JMH with JUnit and some unit test only. Maybe.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -454,31 +475,27 @@ protected void checkUnfinishedTx(Transaction prevTx, EjbInvocation inv) {
...
 
+
     protected void loadCheckpointInfo() {
         try {
             if (isHAEnabled) {
-                Iterator iter = invocationInfoMap.values().iterator();
+                Iterator<InvocationInfo> iter = invocationInfoMap.values().iterator();
                 while (iter.hasNext()) {
-                    InvocationInfo info = (InvocationInfo) iter.next();
+                    InvocationInfo info = iter.next();
```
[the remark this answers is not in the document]
dmatej: done

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
```diff
@@ -1372,40 +1280,34 @@ private void destroyExtendedEMsForContext(SessionContextImpl sc) {
...
-                utMethodsAllowed = (sc.getInstanceKey() != null);
-            } else {
-                utMethodsAllowed = true;
+                return sc.getInstanceKey() != null;
             }
+            return true;
         }
-
-        return utMethodsAllowed;
+        return false;
```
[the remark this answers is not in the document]
dmatej: Yes, but pretty ugly. Generics would help, instead of class typing ... or at least getter.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
dmatej: What if author made it intentionaly? The status changes in each step?

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
dmatej: Nope, because the action must have a return type. I tried that again now ...
But we could make a special method in some utility module.

appserver/persistence/entitybean-container/src/main/java/org/glassfish/persistence/ejb/entitybean/container/EntityContainer.java
dmatej: I avoided to make more changes in this class, it would take me whole day and yet another after review ...

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
dmatej: The second class I wanted to avoid ...

dmatej:
Jenkins test please

dmatej:
Good result:

image

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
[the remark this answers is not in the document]
jbee: I guess this response belongs to the logging comment.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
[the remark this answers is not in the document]
jbee: Was just a suggestion. Your change is a good improvement that is all that counts :)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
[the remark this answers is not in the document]
jbee: In that case either extract a constant or put a comment explaining it - otherwise its just you knowing this.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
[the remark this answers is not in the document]
jbee: I don't say don't fix what is broken, just split a systematic cleanup from a goal oriented bugfix. I know that cleaning when looking for a fix is also needed and a good way to find and fix problems but usually such changes are somewhat limited in scope and do not change 20 classes each having hundreds of changing lines.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
jbee: You are right, we should keep such a change out of this PR. In this old code one never knows.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/StatefulSessionContainer.java
jbee: ah yes, now I remember. the void ugliness :D

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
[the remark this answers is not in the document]
jbee: "Unset" is `null` for references. Its just some IDEs making a difference speculating on programmers intent asking for this or that change to make it explicit.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBContainerTransactionManager.java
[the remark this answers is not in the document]
dmatej: I will leave it as it is this time to spare some time of possible next iteration, because it is simple to find the only place where it is called and it is a private method not providing subzero results to other than nearest caller (which will throw an exception then).
But I agree with you :-)

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/EjbInvocation.java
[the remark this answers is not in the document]
dmatej: That applies only for fields, not for local variables ;)
Compiler does not allow it without explicit null.

PR #4327, opened 2019-11-18
PAYARA-4048: Inject MDBs as non-contextual

dmatej:
Cherrypick of the commit from 5.193.1

### Test suites executed
- Jakarta TCKs
  - `./tck.sh ejb32/mdb/modernconnector` - failed before this commit, passed after it

### Notes

- original commit was only in BitBucket, that is why I could not find it here :-)


dmatej:
Jenkins test please

PR #4328, opened 2019-11-18
PAYARA-4166 MicroProfile Metrics v2.2 implementation

jGauravGupta:
# Description
This is a feature, MicroProfile Metrics v2.2 implementation.

# Important Info

### Dependant PRs 
- https://github.com/payara/Payara_PatchedProjects/pull/284
- https://github.com/payara/patched-src-microprofile-metrics/pull/2

### Related PRs
- https://github.com/payara/MicroProfile-TCK-Runners/pull/67
- https://github.com/payara/patched-src-microprofile-metrics/pull/1

### Patched API & TCK: `2.2.payara-p1`
- CDI version changed to `2.0`
- Few fix in TCK: https://github.com/eclipse/microprofile-metrics/pulls

# Testing

### Testing Performed
- MicroProfile Metrics API TCK 2.2
- MicroProfile Metrics REST TCK 2.2

### Test suites executed

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
@@ -38,7 +38,6 @@
  *     holder.
  *
  */
-
 package fish.payara.microprofile.metrics.impl;
 
 import java.time.Instant;
@@ -46,12 +45,14 @@ import java.time.temporal.ChronoUnit;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.LongAdder;
 import javax.enterprise.inject.Vetoed;
 import org.eclipse.microprofile.metrics.ConcurrentGauge;
 
 /**
  * Implementation of ConcurrentGauge from Microprofile Metrics
+ *
  * @see ConcurrentGauge
  * @since 5.193
  */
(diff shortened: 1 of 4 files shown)


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
```diff
@@ -62,25 +63,53 @@
...
+     * A map containing all the min count values that the count of the gauge has
+     * been at in the last and current minutes. The key is the time in minute
+     * that the value changed, the value is the min value of the count at the
+     * moment before it changed
      */
-    private Map<Instant, Long> lastCounts = new ConcurrentHashMap<>();
-    
+    private final ConcurrentMap<Instant, Long> minCounts
+            = new ConcurrentHashMap<>();
+
```
pdudits: This is nice simplification, but when I understand correctly these maps should always have just single entry, since it's regularly cleared. So we might be ok just with simple variables.

Or due to concurrency, two `MinMax` instances - one for previously completed minute, and one for updating.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
```diff
@@ -62,25 +63,53 @@
...
-        }
-        count.increment();
+        Instant currentMin = getCurrentMinute();
+
+            count.increment();
+            long latestCount = count.longValue();
+            Long value = maxCounts.getOrDefault(currentMin, 0L);
+            if (latestCount > value) {
+                maxCounts.put(currentMin, latestCount);
+            }
```
pdudits: These needs to utilize update functions. Otherwise the moment you arrive at `maxCounts.put`, any of the following might have happened:
* latestCount is less than actual value of counter
* value no longer represents captured maximum
* other thread already put updated maximum, greater than latestCount

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
jbee: I agree that the `Map` should be replaced with simple fields. It makes this appear more complicated than it is and also has overheads and higher risk of concurrency issues we can avoid. If I understand the API correctly 2 atomic longs and 2 volatile longs are a good way.
The atomics to count/set min/max of the not yet completed minute whenever inc/dec are called. If we see that the method is called when a new minute has completed just copy over the current value of the atomics to the volatile min/max fields for "last completed minute" before doing the atomic update. the getters can simply return the volatile min/max fields. No map or cleanup should be needed.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/impl/ConcurrentGaugeImpl.java
```diff
@@ -95,59 +98,79 @@ public long getCount() {
...
-        Instant previousMinute = Instant.now().truncatedTo(ChronoUnit.MINUTES).minus(1, ChronoUnit.MINUTES);
-        if (previousMinute.equals(lastInstant)) {
-            //already called in this minute
-            return;
+    private class MinMax {
+
+        final AtomicLong min;
+        final AtomicLong max;
+        final Instant minute;
+        boolean finished;
```
jbee: Since this is only relevant to the same thread setting it I am pretty sure this does not need any form of synchronisation. But its always tricky. A way to avoid the question altogether would be in `currentStats` first do `openStats.get()` and keep that in a local var, then do the `getAndUpdate`. If the update returns another `MinMax` instance it is outdated. The `finished` is no longer needed.

pdudits: `finished` is more optimization for the common case - usually the minute doesn't roll over, and then you only need to interact with atomic reference once.

appserver/pom.xml
```diff
@@ -206,7 +206,7 @@
         <microprofile-fault-tolerance.version>2.0.1</microprofile-fault-tolerance.version>
         <microprofile-jwt-auth.version>1.1.payara-p1</microprofile-jwt-auth.version>
         <microprofile-healthcheck.version>2.1</microprofile-healthcheck.version>
-        <microprofile-metrics.version>2.0.1.payara-p1</microprofile-metrics.version>
+        <microprofile-metrics.version>2.2.payara-p1</microprofile-metrics.version>
```
pdudits: I merged in BOM now, version properties moved to top-level bom

Cousjava:
Jenkins test please

PR #4332, opened 2019-11-19
PAYARA-4261 Show a Warning in Request Tracing Store Configuration

jbee:
### Summary
Adds a simple warning text to the request tracing store size and historical store size configuration as shown below.

PAYARA-4261

### Note to the Reviewer
I had to use `style` attribute to do a bit of padding. I tried existing CSS classes that seemed fitting but they would also override the colour gained from the `ConWrnLblTxt_sun4` class. Since this should be a minimal and temporary fix I did not want to extend the change and change general CSS so I added the padding manually using `style` attribute. I felt the text was bit to blocky in its normal size so I reduced that a bit as well.

Note that the message cannot be put below the usual help text as the parent component always puts the text after the inner content.

### Testing
Build and install the server, open admin GUI and browse to the request tracing configuration.

### DIFF
--- a/appserver/admingui/payara-console-extras/src/main/resources/requestTracing/requestTracing.jsf
+++ b/appserver/admingui/payara-console-extras/src/main/resources/requestTracing/requestTracing.jsf
@@ -168,6 +168,9 @@
                       helpText="$resource{i18nrt.requestTracing.configuration.traceStoreSizeLabelHelpText}">
             <sun:textField id="traceStoreSizeField" columns="$int{40}"  styleClass="integer" maxLength="30" 
                            text="#{pageSession.valueMap['traceStoreSize']}" />
+            <f:verbatim>
+            <div class="ConWrnLblTxt_sun4" style="padding-top: 10px; font-size: 90%;">Please note that from 5.194 onwards the store size refers to the size of a single store shared by all cluster members. Any configuration change regarding the store should be applied to all configurations equally to prevent unbalanced sharing. All affected configuration are extracted and moved to a central configuration in a future release.</div>
+            </f:verbatim>
         </sun:property>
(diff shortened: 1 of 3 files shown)


jbee:
jenkins test please

appserver/admingui/payara-console-extras/src/main/resources/requestTracing/requestTracing.jsf
```diff
@@ -189,6 +192,9 @@
                       helpText="$resource{i18nrt.requestTracing.configuration.historicTraceStoreSizeLabelHelpText}">
             <sun:textField id="historicTraceStoreSizeField" columns="$int{40}"  styleClass="integer" maxLength="30" 
                            text="#{pageSession.valueMap['historicTraceStoreSize']}" />
+            <f:verbatim>
+            <div class="ConWrnLblTxt_sun4" style="padding-top: 10px; font-size: 90%;">Please note that from 5.194 onwards the store size refers to the size of a single store shared by all cluster members. Any configuration change regarding the store should be applied to all configurations equally to prevent unbalanced sharing. All affected configuration are extracted and moved to a central configuration in a future release.</div>
```
cubastanley: Could I suggest changing the final part of the sentence to "All _affected configurations will be_ extracted and moved to a central configuration in a future release." It's picky but it would just make the tense consistent

jbee: Changed.

jbee:
jenkins test please

Pandrex247:
Is there any warning printed when running the asadmin command? Would be nice if not... :)

jbee:
@Pandrex247 No, should I look into adding such a message?

jbee:
@Pandrex247 merged this - if we want warning for the command we can do this in a new PR.

PR #4333, opened 2019-11-19
PAYARA-4265: Fix cleanup issues in DirWatcher

pdudits:
# Description
This is a bug fix. 

Log messages `Watch service attempted to get checker for untracked item...` that appeared during deployment in recent build hinted, that the lifecycle around directory watchers (deployment optimizations from #4279) was not correct:

Same path can be registered multiple times. Therefore we cannot remove a watcher while there is still a classloader interested in the directory -- this is what caused the message to be logged.
We now keep track of subscription count and remove only if there are no subscribers left.

On some occasions, multiple levels of parent directories might be non-existing, therefore we need to register parent watchers until we find an existing directory.

# Important Info

### Dependant PRs 
#4279 - Original implementation

# Testing

### Testing Performed

Executing the test suite and watching for references to `watch`.

### Test suites executed

- Java EE7 Samples

### Testing Environment

Zulu JDK 1.8_222 on Windows 10 with Maven 3.6.2

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/DirWatcher.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/DirWatcher.java
@@ -116,10 +116,14 @@ class DirWatcher {
             // Quite often the root path doesn't even exist yet (and registering for watch would fail)
             // In such case we register watcher for parent directory so it would register the main watcher
             // when directory is created
-            processors.computeIfAbsent(path, p -> ItemChecker.registerNonExisting(p, watchService));
+            processors.compute(path, (p, watcher) -> watcher == null ? ItemChecker.registerNonExisting(p, watchService) : watcher.subscribe());
             processors.computeIfAbsent(path.getParent(), p -> new ParentWatcher(p, watchService));
+            while (Files.notExists(path.getParent())) {
+                path = path.getParent();
+                processors.computeIfAbsent(path.getParent(), p -> new ParentWatcher(p, watchService));
+            }
         } else {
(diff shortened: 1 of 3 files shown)


pdudits:
Jenkins test please

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/DirWatcher.java
```diff
@@ -171,15 +175,17 @@ private synchronized void syncWatches() {
...
         }
     }
 
     private void unregisterFromFilesystem(Path path) {
-        WatchProcessor processor = processors.remove(path);
+        WatchProcessor processor = processors.get(path);
         if (processor != null) {
-            processor.cancelled();
+            if (processor.unsubscribe()) {
+                processor.cancelled();
```
jbee: Maybe the cancel should always occur as part of `unsubscribe` when the count gets to zero?

nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/WatchProcessor.java
```diff
@@ -60,6 +60,7 @@
     protected boolean registered;
     protected boolean overflowed;
     protected boolean cancelled;
+    private int subscribers = 1;
```
jbee: I take it that creating an instance is also considered subscribing to it. As this cannot get clear from the API alone I think a comment on the `subscribe` method is the best to point that out. WDYT?

pdudits:
Jenkins test please

pdudits:
Jenkins test please

pdudits:
Jenkins test please

PR #4338, opened 2019-11-21
PAYARA-4215: Enable persistent timers in Deployment Group targets

pdudits:
# Description
This is a bugfix / feature (bit of both). 

Method annotated with `@Schedule` weren't executed when application was deployed to a deployment group. It was quite correct thing to do, because Deployment Groups are loose groupings of instances that do not necessarily share configuration, inclusive Timer Service configuration.

It is task of DAS to register timers defined by annotations and xml descriptor, and in general it couldn't know which timer service defined in domain to use.

However, from now on we will assume, that user is doing the right thing and his instances' timer service configurations are compatible. E. g. point to same external database, or use DataGrid timer service. At this point, `server-config` needs to have same timer service configuration as well (separate issue).
DAS will use one of instance's configuration to create the timers.

When such application is deployed a warning is printed on DAS:

```
[2019-11-20T20:54:53.787+0100] [Payara 5.194] [WARNING] [] [javax.enterprise.system.container.ejb.org.glassfish.ejb.startup] [tid: _ThreadID=57 _ThreadName=admin-thread-pool::admin-listener(1)] [timeMillis: 1574279693787] [levelValue: 900] [[
  Deployment targets deployment group localDG, it is assumed that timer service configuration is consistent accross all members of the group]]

[2019-11-20T20:54:53.836+0100] [Payara 5.194] [INFO] [] [javax.enterprise.system.container.ejb.com.sun.ejb.containers] [tid: _ThreadID=57 _ThreadName=admin-thread-pool::admin-listener(1)] [timeMillis: 1574279693836] [levelValue: 800] [[
  Timer Service configuration local-instance-config will be used to prepare timers]]
```

Persistent timer will now migrate whenever deployment group member disappears (instance shuts down). Therefore, following are limitations to use persistent timers properly and have them correctly migrate:

* Timer service config is the same for all instances of deployment group and DAS (same database or DataGrid)
* DataGrid is enabled and instances see each other
* Deployment Groups do not form complex topologies. Timers migrate to any instance that is part of any deployment group the parting instance was member of. Any target instance should have same set of (persistence-timer enabled) applications.
* No application management is done on instance level - disabling application, or undeploying it from instance may break timer migration.

# Testing

### Testing Performed
Used reproducer from [payara-4125.zip](https://github.com/payara/Payara/files/3871154/payara-4125.zip).
Create deployment group with multiple instances and deploy the application.
After deployment, all instancess will print 3 messages per second:

```
  Non-Persistent annotated timer runs
  Non-Persistent Programmatic timer inst1-177f171c-137e-45a4-b3ac-e82efc5f67db runs
  Persistent Programmatic timer inst1 runs
```
where `inst1` is name of instance. One of the instances will also print
```
  Persistent annotated timer runs
```

Stopping or killing instances cause persistent timers to migrate to different instance.

### Testing Environment

Zulu JDK 1.8_222 on Windows 10 with Maven 3.6.2

# Documentation

T.B.D.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbContainerUtilImpl.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbContainerUtilImpl.java
@@ -43,6 +43,7 @@ package com.sun.ejb.containers;
 
 import com.sun.ejb.base.io.EJBObjectInputStreamHandler;
 import com.sun.ejb.base.io.EJBObjectOutputStreamHandler;
+import com.sun.enterprise.config.serverbeans.Server;
 import com.sun.enterprise.container.common.spi.util.JavaEEIOUtils;
 import com.sun.enterprise.transaction.api.JavaEETransaction;
 import com.sun.enterprise.transaction.api.JavaEETransactionManager;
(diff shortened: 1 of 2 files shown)


pdudits:
Jenkins test please

PR #4351, opened 2019-11-26
PAYARA-4277 : Uniform jvm-option in server-config and default-config

rdebusscher:
#4277 Description
This is a bug fix.

Both -client and -server are defined in default-config of domain.1 domain.xml.  This PR removes the -server option line so that options are the same in server-config and default-config.


chrjohn:
Does it make sense to use the `-client` option on a server?
On 64bit systems this is ignored anyway and `-server` is used, but just asking...

smillidge:
Yeah maybe better to delete client rather than server if you are seeking consistency

rdebusscher:
I updated to use `-server` as option in both config section of domain.xml.

AlanRoth:
jenkisn test please

AlanRoth:
jenkins test please

Issue #4197, opened 2019-09-04
Allow steady pool size with 0 again


deratzmann:
# Description #
----------
For our test environments we create connection-pools for a postgres DB which might not be available. Therefore to avoid errors while startup we created the pools with a steady size of 0.
here is the cli file:
`create-jdbc-connection-pool --validateatmostonceperiod=5 --creationretryattempts=3 --creationretryinterval=20 --maxwait=2000 --datasourceclassname org.postgresql.ds.PGConnectionPoolDataSource --restype javax.sql.ConnectionPoolDataSource --isconnectvalidatereq true --validationmethod auto-commit --steadypoolsize 0 --maxpoolsize 30 --poolresizequantity 15 --idletimeout 3600 --leakreclaim false --statementtimeout 60 --leaktimeout 120 --statementleakreclaim=false --statementleaktimeout 30 --failconnection false --property user="${DAYSNAP_username}":password="${ALIAS=daysnap_password}":url="${DAYSNAP_url}":characterEncoding="UTF-8":ssl="true":sslmode="verify-ca" DAYSNAP2`

Expected: successfully created connectionpool.

Current:
`$ ./asadmin -p 1048 multimode --file t.txt
remote failure: JDBC connection pool DAYSNAP2 creation failed. Constraints for this JdbcConnectionPool configuration have been violated: on property [  ] violation reason [ Max-pool-size has to be greater than or equal to steady-pool-size ]
Command create-jdbc-connection-pool failed.
Command multimode failed.
`
If I change steadypoolsize to 1, it works.
So is this a bug or a new constraint?

## Environment ##

- **Payara Version**: 5.193
- **Edition**: Full
- **JDK Version**: Openjdk 8
- **Operating System**: windows
- **Database**: Postgres

rdebusscher:
Hi @deratzmann 

There is indeed now a check on the value of Max Pool Size and Steady Pool Size that it can't be 0 or smaller (but without the correct error message)

This is the code added in 5.193
https://github.com/payara/Payara/commit/eec452eed1d3da01ffc9a428dad38b1d0c33138d#diff-95dc2efc07f48f0d9760406d695c3e06R140

I have created the internal issue PAYARA-4205 to correct the error message. But I agree with the change because a pool size of 0 doesn't make much sense.

The create command will succeed even when the database is not up, only deployment of applications will fail if they access the JDBC pool and database cannot be contacted.

sergitrujillo:
Hi everyone,

[...]

PR #4357, opened 2019-11-28
CUSTCOM-27: Fix that steady-pool-size can be 0

sergitrujillo:
# Description
This is a bug fix
https://github.com/payara/Payara/issues/4197

# Notes for Reviewers
It's my first pull request to payara. Sorry if if did some mistake in some convenion.

### DIFF
--- a/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
+++ b/appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
@@ -138,7 +138,7 @@ public class JdbcConnectionPoolValidator
                 }
                 
                 if (maxPoolSizeValue < steadyPoolSizeValue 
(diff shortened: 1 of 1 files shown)


jbee:
jenkins test please

jbee:
@smillidge Do we need a CLA for this small of a change?

smillidge:
No that's too small and obvious to be covered by copyright.

jbee:
@rdebusscher you seemed to know the context of PAYARA-4205 - maybe you can have a look and see if the change is the solution you were looking for in your comments.

dmatej:
It is rather PAYARA-4207, 4205 was about the error message. Both issues should join in correct solution = correct check allowing zero steady pool size and correct error message.
I will finish the test case for that tomorrow.

rdebusscher:
@jbee The PR is good for the fix of the steady pool size. The 'issue' that we perform 2 checks (maxPoolSizeValue should be larger then steadyPoolSizeValue and values larger then 0) and show only 1 error message still needs to be addressed.

dmatej:
Tested both with asadmin and DAS UI.

PR #4358, opened 2019-11-28
Addresses Various Sonar Issues

jbee:
### Summary
I tried to filter out more severe issues and then cleaned up the file. So there isn't a common theme other than each file had one or more sonar warnings I tried to get rid of. Then it become lots of adding generics as there weren't any in lots of the files. And with some added other needed adding so it cascaded on me. At that point I limited the cleanup to generics and pure style changes like cleaning condition expressions, removing unnecessary else blocks or use lambdas instead of anonymous classes. 

### Testing
Non special. Just the build and PR tests.


appserver/deployment/client/src/main/java/org/glassfish/deployapi/SunDeploymentManager.java
```diff
@@ -363,15 +362,9 @@ private void setWebApplicationTargetModuleIDInfo(TargetModuleID tmid, HostAndPor
     */
     @Override
     public DeploymentConfiguration createConfiguration(DeployableObject dObj) throws InvalidModuleException {
-        try {
-            SunDeploymentConfiguration deploymentConfiguration = new SunDeploymentConfiguration(dObj);
-            deploymentConfiguration.setDeploymentManager(this);
-            return deploymentConfiguration;
-        } catch(ConfigurationException e) {
```
jbee: The constructor would no longer throw that exception (as it didn't) so it doesn't need handling.

jbee:
jenkins test please

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -401,21 +391,11 @@ public void unloadManagedBeans(Application app) {
...
-
-            eligible = (bundle instanceof WebBundleDescriptor) ||
-                (bundle instanceof EjbBundleDescriptor);
-
-        } else if( processType == ProcessType.ACC ) {
-            eligible = (bundle instanceof ApplicationClientDescriptor);
-        }
-
-        return eligible;
+        return processType.isServer() && (bundle instanceof WebBundleDescriptor || bundle instanceof EjbBundleDescriptor)
```
dmatej: I would rather use something in between ... the expression is too complicated to read, so I would split it to more blocks with returns:
```
        if (processType.isServer() ) {
            return (bundle instanceof WebBundleDescriptor) || (bundle instanceof EjbBundleDescriptor);
        } 
        if (processType == ProcessType.ACC) {
            return bundle instanceof ApplicationClientDescriptor;
        }
        return false;
```
Also it looks weird that one branch calls isServer method and second compares type.
Another problem is using && and || on the same level without braces `A && B || C` - for newbies and sleepy developers it is hard to read, they make mistakes.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -472,18 +454,16 @@ private boolean bundleEligible(BundleDescriptor bundle) {
...
             // Need to keep track of context in order to destroy properly
-            Map<Object, JCDIService.JCDIInjectionContext> bundleNonManagedObjs =
+            Map<Object, JCDIService.JCDIInjectionContext<?>> bundleNonManagedObjs =
                 jcdiManagedBeanInstanceMap.get(bundleDescriptor);
             bundleNonManagedObjs.put(callerObject, jcdiContext);
 
-        } else {
-
-
+        } else if (desc != null) {
```
dmatej: Hmm, 458/440 check desc == null, so it can be null. Original code would throw NPE in this branch, now it would return null callerObject. So maybe NPE would be thrown from calling method, but not from here, so yes, Sonar cannot see it now and this method is really better than before.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -659,15 +631,12 @@ public void destroyManagedBean(Object managedBean, boolean validate)  {
...
-                    java.security.AccessController.doPrivileged(
-                    new java.security.PrivilegedExceptionAction() {
-                        public java.lang.Object run() throws Exception {
-                            if( !finalF.isAccessible() ) {
-                                finalF.setAccessible(true);
-                            }
-                            return null;
-                        }
-                    });
+                java.security.AccessController.doPrivileged((PrivilegedExceptionAction<Object>) () -> {
```
dmatej: If you would also extract the action to the variable, you don't need casting it.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/connectors/util/ConnectionPoolObjectsUtils.java
```diff
@@ -330,12 +332,10 @@ public static Subject createSubject(ManagedConnectionFactory mcf,
                         new PasswordCredential(prin.getName(),
                                 password.toCharArray());
                 pc.setManagedConnectionFactory(mcf);
-                AccessController.doPrivileged(new PrivilegedAction() {
-                    public Object run() {
-                        tempSubject.getPrincipals().add(prin);
-                        tempSubject.getPrivateCredentials().add(pc);
-                        return null;
-                    }
+                AccessController.doPrivileged((PrivilegedAction<Object>) () -> {
```
dmatej: dtto - action to the variable. Also better use type Void instead of Object:
```
PrivilegedAction<Void> action = () -> { xxx ... return null;};
AccessController.doPrivileged(action);
```

appserver/deployment/client/src/main/java/org/glassfish/deployapi/config/SunDeploymentConfiguration.java
```diff
@@ -73,7 +70,7 @@
 	  new LocalStringManagerImpl(SunDeploymentConfiguration.class);
     
     /** Creates a new instance of SunDeploymentConfiguration */
-    public SunDeploymentConfiguration(DeployableObject deployObject) throws ConfigurationException {
```
dmatej: It has no children, maybe it was some original idea, you are right.

appserver/deployment/client/src/main/java/org/glassfish/deployapi/config/DConfigBeanRootFactoryImpl.java
```diff
@@ -1,71 +0,0 @@
...
-    }
-  
-    /**
-     * @return the right DConfigBeanRoot implementation depending on the 
-     * passed module type
-     * 
-     * @param the module type 
-     * @return the correspondind DConfigBeanRoot instance
-     */
-    public static DConfigBeanRoot getDConfigBeanRootForType(ModuleType moduleType) {
```
dmatej: Well, this was really teethless zombie class :-)

appserver/deployment/client/src/main/java/org/glassfish/deployapi/SunDeploymentManager.java
dmatej: I am confirming that.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -119,36 +121,37 @@
...
         processType = processEnv.getProcessType();
     }
 
-    public void event(Event event) {
+    @Override
+    public void event(Event<?> event) {
 
          if (event.is(Deployment.APPLICATION_LOADED) ) {
-             ApplicationInfo info =  Deployment.APPLICATION_LOADED.getHook(event);
+             ApplicationInfo info =  Deployment.APPLICATION_LOADED.getHook((Event<ApplicationInfo>) event);
```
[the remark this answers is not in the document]
jbee: There is no else since this is a general event listener mechanism so you will get all kinds of events even if you are only interested in some. Implementations then just pick out those they originally wanted to be connected to. This is a shortcoming of the current `EventListener` mechanism.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
jbee: Yeah, I am unsure if this is the right fix but it was the best I could see. Maybe @pdudits has an opinion on this?

jbee:
jenkins test please

appserver/appclient/client/acc/src/main/java/org/glassfish/appclient/client/acc/ACCClassLoader.java
```diff
@@ -262,49 +234,40 @@ public ACCClassLoader run() {
...
-            final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try (InputStream is = getResourceAsStream(resourceName);
+                ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+            if (is == null) {
+                throw new ClassNotFoundException(className);
+            }
             final byte[] buffer = new byte[8196];
             int bytesRead;
             while ( (bytesRead = is.read(buffer)) != -1) {
                 baos.write(buffer, 0, bytesRead);
```
jbee: If we move beyond Java9 the whole block could be replace with a transfer found in JDK.

jbee:
@dmatej and @pdudits this is ready for a re-review.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
pdudits: I'd rather not have one :) But reading from larger height I see this:

1. if CDI is enabled, delegate managed bean creation to CDI
2. If CDI is not enabled, we follow old Managed Bean spec. That mandates that a bean has `@ManagedBean` annotation, and therefore a desc was likely created for it

Also, interestingly the javadoc of the method says: "desc can be null if JCDI enabled bundle.".

If you want to be completely sure, add an else branch throwing `IllegalArgumentException("CDI not enabled and no managed bean of class $managedBeanClass found")`.

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -549,17 +541,15 @@ private boolean bundleEligible(BundleDescriptor bundle) {
...
             // Need to keep track of context in order to destroy properly
-            Map<Object, JCDIService.JCDIInjectionContext> bundleNonManagedObjs =
+            Map<Object, JCDIService.JCDIInjectionContext<?>> bundleNonManagedObjs =
                 jcdiManagedBeanInstanceMap.get(bundleDescriptor);
             bundleNonManagedObjs.put(callerObject, jcdiContext);
 
-        } else {
-
-
+        } else if (desc != null) {
```
pdudits: In such case, and same exception thrown here if `desc` is null

Isn't there an opportunity to extract common parts of these methods?

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
```diff
@@ -119,42 +122,48 @@
...
         processType = processEnv.getProcessType();
     }
 
-    public void event(Event event) {
+    @Override
+    public void event(Event<?> event) {
 
          if (event.is(Deployment.APPLICATION_LOADED) ) {
-             ApplicationInfo info =  Deployment.APPLICATION_LOADED.getHook(event);
+             ApplicationInfo info =  Deployment.APPLICATION_LOADED.getHook(castEvent(event));
```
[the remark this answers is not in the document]
jbee: You mean the `getHook` method should accept `Event<?>`?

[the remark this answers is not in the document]
pdudits: Yes. Idiom is to do `is` test before, and `getHook` returns `null` when `is` check fails. Therefore enforcing type of event hook, especially when these events are broadcasted untyped is unhelpful.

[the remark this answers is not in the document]
jbee: I  created a new method that takes a consumer which is run in case the type matches. WDYT?

appserver/common/container-common/src/main/java/com/sun/enterprise/container/common/impl/managedbean/ManagedBeanManagerImpl.java
jbee: I changed to throwing the exception you suggested.
I also looked into the duplication and from the method signatures I'd assume those should be almost the same but there are strange differences in the details. It might just be that code paths were not updated properly but I felt this is too big of a change to fit into the scope of this PR so I left it as is.

jbee:
jenkins test please

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorManager.java
```diff
@@ -270,11 +271,11 @@ public void registerRuntimeInterceptor(Object o) {
     }
 
     public InterceptorManager.InterceptorChain getAroundTimeoutChain(
-            MethodDescriptor mDesc, Method beanMethod) {
+            MethodDescriptor mDesc, @SuppressWarnings("unused") Method beanMethod) {
```
dmatej: Generally I learned to avoid using this annotation with unused, because - the warning is true and valid, so it should be visible even if we don't plan to change it.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorManager.java
```diff
@@ -594,23 +596,22 @@ private void initEjbCallbackIndices()
         callbackChain = new CallbackChainImpl[size];
         for (CallbackType eventType : CallbackType.values()) {
             int index = eventType.ordinal();
-            CallbackInterceptor[] interceptors = (CallbackInterceptor[])
-                    callbacks[index].toArray(new CallbackInterceptor[callbacks[index].size()]);
+            CallbackInterceptor[] interceptors = callbacks[index].toArray(new CallbackInterceptor[callbacks[index].size()]);
```
dmatej: The value on the index position is resolved twice, JIT will probably simplify it, but it could be simplified directly ... also then you would not need the index variable. 
Blah ... array of arraylist, who invented this? :D ... nevermind ...

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/deployment/BeanMethodCalculatorImpl.java
```diff
@@ -92,22 +94,22 @@ public Vector getPossibleCmpCmrFields(ClassLoader cl,
             int nextModifiers = next.getModifiers();
             if( Modifier.isAbstract(nextModifiers) ) {
                 if( nextName.startsWith("get") &&
-                    nextName.length() > 3 ) {
+                        nextName.length() > 3 ) {
```
dmatej: I expect you replaced tabs with spaces, but the formatting is still weird.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorManager.java
jbee: I get that. I tend to suppress  it in files that are otherwise free of warnings to better see when a warning comes up but for a team with mixed IDEs this might not make as much sense as just leaving the warning in.

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/interceptors/InterceptorManager.java
jbee: There is a lot of more cleaning possible. At some point I tried to keep this focused on generics and warnings I came across while doing that.

appserver/ejb/ejb-container/src/main/java/org/glassfish/ejb/deployment/BeanMethodCalculatorImpl.java
jbee: Yeah, I didn't follow up on all bad formatting and weird indenting. Maybe its better to actually run a format all on full module at a time to get those changes out of the way.

PR #4360, opened 2019-12-01
PAYARA-4161 Fixed full logging - increased buffer capacity

dmatej:
Note: To be refactored to be much more effective under PAYARA-4276, this is a minimal fix

The buffer accepts all messages even before the logging handlers are configured, but - if the capacity is reached, the threads are blocked until some thread would start process records from buffer - this will never be done, so this is a bit risky - this issue does not fix it, only increases the capacity.

Under PAYARA-4276 I will refactor and optimize the logging, I already gained 10-50% throughput in yet uncommited changes (depends on system configuration and state).

# Testing

### New tests

See VerboseLoggingITest in https://github.com/payara/Payara/pull/4263

### Testing Environment
Kubuntu 19.10, OpenJDK11


dmatej:
Jenkins test please

PR #4361, opened 2019-12-02
Monitoring Console Colour Configuration and Colour Schemes

jbee:
### Summary
Makes graph colours configurable by

* adding color input type to `Settings` component that can be used for single or multiple values
* adds configuration for waterline, alerting and critical colours per widget
* adds new settings group _Colors_ which allows to modify the list of colours used for data lines in graphs as well as the defaults for waterline, alerting and critical colours. 

PR also makes settings more flexible by allowing to mark a group as `collapsed` which will not show the settings of the section initially. This was added to hide the colour settings so they don't move the widget settings out of normal screen.

Besides making colours configurable for the user this PR also makes sure that all graphs and pages are consistent in their use of colour for a particular source. With the PR colour used is not simply picked by index of the data series but based on the key property of the series. Mostly this is the instance name. This means graphs will show data belonging to the same instance in the same colour. There are exceptions like request tracing where instance name is not the distinguishing property. This is configured now using the new `Coloring` settings of widgets. 

In addition the changes make sure that duplicate colours for lines in a chart are highly unlikely.
Should there be less colours in the colour palette defined by the user new colours are derived based on the colours in the palette. The algorithm has an intentional bias towards bright colours - dark colours and greyish ones are avoided as they are hard to see or distinguish in a graph.

### Note for Reviewers
As "usual" let me point out that the `monitoring-console.js` file is a merge of the other JS files that does not need a review.

### Testing Performed
Manual testing.

Checking compatibility: clear local browser storage for MC, reload page
Checking colour scheme feature: open settings, chose a scheme and check effect of colour change on _Core_ and _Request Tracing_ page including _Request Tracing_ details page.

Checking defaults:
1. change a default colour, e.g. waterline
2. select widget _Heap Usage_ on _Core_ page and check it has the changed default for waterline
3. change the widgets waterline colour and check that changing the default does not change the widgets custom colour
4. reset the widgets waterline colour to default by clicking on the box right to the colour picker.
5. change the default for waterline again and check that the widget also changes its waterline colour

Testing generated colours:
1. open _Color_ settings
2. remove all colours in the palette but one. 
3. look at a chart with multiple lines (e.g. start additional instances)
4. also add colours to the palette will give the sequence of generated colours derived **only** from the first colour

Testing opacity:
1. open _Color_ settings
2. change _Opacity_ to 80
3. check the background colour in graphs now is almost non-transparent

### DIFF
--- a/appserver/monitoring-console/webapp/JS_DOCS.md
+++ b/appserver/monitoring-console/webapp/JS_DOCS.md
@@ -214,11 +214,11 @@ Describes the model expected by the `Settings` component.
 
 ```
 SETTINGS    = [GROUP]
-GROUP       = { id, caption, entries }
+GROUP       = { id, caption, entries, collapsed }
 id 		    = string
 caption     = string
 entries     = [ENTRY]
-ENTRY       = { label, type, input, value, unit, min, max, options, onChange, description, defaultValue } 
+ENTRY       = { label, type, input, value, unit, min, max, options, onChange, description, defaultValue, collapsed } 
 label       = string
 type        = undefined | 'header' | 'checkbox' | 'range' | 'dropdown' | 'value' | 'text' | 'color'
 unit        = string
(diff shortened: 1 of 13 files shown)


jbee:
jenkins test please

PR #4365, opened 2019-12-04
CUSTOM-54: NullPointerException when JMX attribute is null

rdebusscher:
CUSTOM-54: NullPointerException when JMX attribute is null within the rest-monitoring endpoint

# Description
This is a bug fix

When JMX attributes are null, it results in NPE when accessed from the rest-monitoring endpoint.

The fix  checks if the attribute value is null and return JsonValue.NULL in that case instead of the 'generic' method who tries to convert the instance to a JsonValue.

# Testing

### Testing Performed

Manual testing

- /asadmin start-domain 
- deploy simple jms example (https://github.com/payara/Payara-Examples/tree/master/javaee/simple-jms-example)
- ./asadmin set-monitoring-level --level=HIGH --module=jmsService --target=server-config
- ./asadmin set-amx-enabled --dynamic=true --enabled=true --target=server-config 
- ./asadmin set-monitoring-service-configuration --mbeanEnabled=true --monitoringEnabled=true --dtraceEnabled=false --target=server-config
- ./asadmin set-rest-monitoring-configuration --securityenabled=false --contextroot=/rest-monitoring --enabled=true --applicationname=__restmonitoring --target=server-config 
- http://localhost:4848/rest-monitoring/rest/read/com.sun.messaging.jms.server:type=Destination,subtype=Monitor,desttype=q,name="simpleQ"

### Testing Environment
Oracle 1.8.0_181 on Mac 10.14.6 with Maven 3.5.4

# Documentation
N/A

### DIFF
--- a/appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
+++ b/appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
@@ -89,8 +89,7 @@
         <!-- JSONP Provider -->
         <dependency>
             <groupId>org.glassfish</groupId>
-            <artifactId>javax.json</artifactId>
-            <version>1.1</version>
+            <artifactId>jakarta.json</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
+++ b/appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
@@ -71,6 +71,28 @@
             <version>7.0</version>
             <scope>provided</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>${mockito.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- JSONP Provider -->
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.1</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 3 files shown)


rdebusscher:
@MarkWareham Added Unit tests

appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
```diff
@@ -71,6 +71,28 @@
...
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- JSONP Provider -->
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+            <version>1.1</version>
```
pzygielo: Why not use the same version and from jakarta as https://github.com/payara/Payara/blob/2e6edd27d6f1ac9837bf072eda4d794eae0bb04a/pom.xml#L186
seems to be used for example in https://github.com/payara/Payara/blob/2e6edd27d6f1ac9837bf072eda4d794eae0bb04a/api/payara-bom/pom.xml#L798-L802

appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
```diff
@@ -71,6 +71,28 @@
...
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.9.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- JSONP Provider -->
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
```
MarkWareham: I believe this has now changed to Jakarta. It may be in the Payara POM

appserver/payara-appserver-modules/rest-monitoring/rest-monitoring-war/pom.xml
rdebusscher: I though the `org.glassfish:jakarta.json` pointed to the API, and not an actual implementation., as only the API artifacts should be 'jakartified'.
Fixed in new commit.

rdebusscher:
jenkins test please

PR #4373, opened 2019-12-10
CUSTCOM-53 Fixed jdbc pool validation

dmatej:
# Description

Fixes CUSTCOM-53. Splits validation of steady pool size and max pool size, and adds unit test. Also fixes the javascript error message.

**Note**: I did not manage to translate the connpool.poolsize.max error message to japanese, chinese and korean language, remaining languages should be reviewed, but I believe they are alright.

# Important Info

### Dependant PRs 

Extends https://github.com/payara/Payara/pull/4357 and fixes also issue created by a customer.

# Testing

### New tests

TDD used in practice ;)
JdbcConnectionPoolValidatorTest (94% coverage of the module)

### Testing Performed

Manual testing - create and update the JDBC pool, try to set negative or zero errors, read messages.

### Test suites executed
- Quicklook
- Payara Samples

### Testing Environment

Firefox, Linux


dmatej:
Jenkins test please

appserver/connectors/connectors-internal-api/src/main/java/org/glassfish/connectors/config/ConnectorConnectionPool.java
```diff
@@ -85,31 +85,24 @@
...
-/* @XmlType(name = "", propOrder = {
-    "description",
-    "securityMap",
-    "property"
-}) */
-
 @Configured
 @ConnectionPoolConstraints({
-    @ConnectionPoolConstraint(value = ConnectionPoolErrorMessages.MAX_STEADY_INVALID)
+    @ConnectionPoolConstraint(value = ConnectionPoolErrorMessages.POOL_SIZE_STEADY),
```
dmatej: I split the common validation into two standalone. The relation between those two values is validated in max pool size validation.

appserver/connectors/connectors-internal-api/src/main/java/org/glassfish/connectors/config/validators/ConnectionPoolValidator.java
```diff
@@ -49,49 +50,58 @@
...
+ * <li>Max pool size to be always higher than steady pool size
+ * <li>Check if statement wrapping is on when certain features are enabled.
+ * </ul>
+ *
  * @author Shalini M
  */
-public class ConnectionPoolValidator
-    implements ConstraintValidator<ConnectionPoolConstraint, ResourcePool> {
-    
+public class ConnectionPoolValidator implements ConstraintValidator<ConnectionPoolConstraint, ResourcePool> {
```
dmatej: This class was updated too, because it uses same constants and same mechanism, despite it uses it's own implementation. I didn't create unit test for this.

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -41,204 +41,234 @@
...
-    public String getParsedVariable(String variableToRetrieve) {
 
-        String[] variableReference = variableToRetrieve.split("=");
+    @Override
+    public boolean isValid(final ResourcePool pool, final ConstraintValidatorContext context) {
+        if (!JdbcConnectionPool.class.isInstance(pool)) {
+            return true;
+        }
+        // original code allowed this. Why?
+        if (this.poolFaults == null) {
```
dmatej: That's the question. I didn't change this behavior.

appserver/connectors/connectors-internal-api/src/main/java/org/glassfish/connectors/config/validators/ConnectionPoolValidator.java
```diff
@@ -49,49 +50,58 @@
...
+            return true;
+        }
+        final ConnectorConnectionPool connPool = (ConnectorConnectionPool) pool;
+        if (poolFaults == ConnectionPoolErrorMessages.POOL_SIZE_STEADY) {
+            final int steadyPoolSize = getSteadyPoolSize(connPool);
+            return steadyPoolSize >= 0;
+        } else if (poolFaults == ConnectionPoolErrorMessages.POOL_SIZE_MAX) {
+            final int steadyPoolSize = getSteadyPoolSize(connPool);
+            final int maxPoolSize = getMaxPoolSize(connPool);
+            return maxPoolSize > 0 && maxPoolSize >= steadyPoolSize;
```
Pandrex247: Is `final` really necessary here?
It's essentially just noise since you're immediately returning.

You could even make it a one liner with `return getSteadyPoolSize(connPool) >= 0;`

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -41,204 +41,234 @@
...
+
+    private boolean validateMaxPoolSize(final JdbcConnectionPool jdbcPool) {
+        final Integer steadyPoolSize = getSteadyPoolSize(jdbcPool);
+        final String propertyValue = jdbcPool.getMaxPoolSize();
+        final String value = propertyValue == null ? Constants.DEFAULT_MAX_POOL_SIZE : propertyValue;
+        final Integer maxPoolSize = toInteger(resolve(value));
+        if (maxPoolSize == null || maxPoolSize < 1) {
+            return false;
+        }
+        if (steadyPoolSize != null && steadyPoolSize > maxPoolSize) {
```
rdebusscher: if can be replaced by single return

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -41,204 +41,234 @@
...
-            return System.getProperty(variableReference[0]);
+
+    private boolean validateSteadyPoolSize(final JdbcConnectionPool jdbcPool) {
+        final Integer steadyPoolSize = getSteadyPoolSize(jdbcPool);
+        return steadyPoolSize != null && steadyPoolSize >= 0;
+    }
+
+
+    private boolean validateMaxPoolSize(final JdbcConnectionPool jdbcPool) {
+        final Integer steadyPoolSize = getSteadyPoolSize(jdbcPool);
```
rdebusscher: final for local variables are overhead

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
```diff
@@ -41,204 +41,234 @@
...
+        if (!wrappingEnabled && isPositiveInt(stmtCacheSize)) {
+            return false;
         }
-        
-        //If this point is reached, the variable value could not be found
-        return null; 
+        if (!wrappingEnabled && isPositiveInt(stmtLeakTimeout)) {
+            return false;
+        }
+        if (!wrappingEnabled && isTrue(jdbcPool.getStatementLeakReclaim())) {
```
rdebusscher: If can be replaced by single return

appserver/connectors/connectors-internal-api/src/main/java/org/glassfish/connectors/config/validators/ConnectionPoolValidator.java
dmatej: They are not necessary, but in usual editors (Netbeans/Idea/Eclipse) it is more natural to have all variables visible. It also forces good architecture - you will rather find a good name for private method than recycle variables, which is much harder to understand.
Code with final variables can be very simply optimized by a compiler (not only JIT), it depends on CPU/memory/OS architecture how deep it is effective to go. But today compilers can optimize also non-final variables - and also nanoseconds are not so strong argument, when nearly every code loses hundreds of miliseconds on GC vs. too big variable lifespans, bad synchronizations, resource management ...

dmatej: Fixed.

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
dmatej: Yes, I could merge both ifs to one, but it would be less understandible. Now you see on the first sight what it does.

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
dmatej: Additional info to the response I wrote to @Pandrex247, surprising even myself: https://stackoverflow.com/a/21485260/375449
It seems finals have much bigger positive effect then I thought ... (I thought on JVMs after JDK5 or 6 it is nearly same, but it is not. We should use final variables much more than we do ...)

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
dmatej: It can, but it would be harder to read. Reading conditions on more than single line is harder than two if conditions with implicit else.
If the first IF is resolved to false, then it does not return anything and evaluates the second IF. 

Compiler optimization or at least JIT optimizer can do that automatically and even if it would not do that, this code still may be faster than joined evaluation (but one nanosecond is not worth to analyze that, we have hundred of miliseconds in other places)..

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
dmatej: (this is rather argumentation foul, but it is still interesting result)
https://dzone.com/articles/all-about-finals

dmatej:
Jenkins test please

appserver/jdbc/jdbc-config/src/main/java/org/glassfish/jdbc/config/validators/JdbcConnectionPoolValidator.java
rdebusscher: >We should use final variables much more than we do ...)

Well, then we should have some code rules and update all code in the repository to be uniform. Now we start creating a mixture which is 'ugly'.

And it is misleading, final doesn't mean the variable can't be changed, just the reference (so in most cases the 'content' of the object can be updated - unless it is created as immutable).

dmatej: There are no standards in Payara/Glassfish code, but I agree we can standardize it (then we must use tools to enforce that). We can discuss it privately on monday, this is not right place for that.
The `final` keyword was in Java always about references, not internal states of objects.

PR #4378, opened 2019-12-12
CUSTCOM-76: Fix Thread synchronization in WebappClassloader

rdebusscher:
# Description
This is a bug fix 

Customer reports occasionally a NullPointerException in `org.glassfish.web.loader.WebappClassLoader` when starting up his Payara Micro (5.193.1) Application.

> Caused by: java.lang.NullPointerException
> at org.glassfish.web.loader.WebappClassLoader.findResourceInternalFromJars(WebappClassLoader.java:2908)
> at org.glassfish.web.loader.WebappClassLoader.findResourceInternal(WebappClassLoader.java:2761)
> at org.glassfish.web.loader.WebappClassLoader.findClassInternal(WebappClassLoader.java:2651)

There is a synchronized access on a non final property causing multithreading issues in the addJar() and findResourceInternalFromJars()

# Testing

### Testing Performed
- Manual smoke testing with a few customer reperoducers to verify normal classloader behaviour

### Test suites executed
- Payara Samples

### Testing Environment
Oracle 8u181, mvn 3.5.4

# Documentation
N/A

# Notes for Reviewers

Also added {} to if statements if missing.

# Update 16/12
I used to the (already present)  jarFilesLock object within the `findResourceInternalFromJars` method since that one was already protecting the access to property `jarFiles`.  
This means that accessing the jar files is now synchronized with start up but also with shutdown.

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -55,7 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.web.loader;
 
@@ -64,20 +64,15 @@ import com.sun.appserv.ClassLoaderUtil;
 import com.sun.appserv.server.util.PreprocessorUtil;
 import com.sun.enterprise.deployment.Application;
 import com.sun.enterprise.deployment.util.DOLUtils;
-import com.sun.enterprise.util.io.FileUtils;
-import org.apache.naming.JndiPermission;
-import org.apache.naming.resources.DirContextURLStreamHandler;
-import org.apache.naming.resources.JarFileResourcesProvider;
-import org.apache.naming.resources.ProxyDirContext;
-import org.apache.naming.resources.WebDirContext;
-import org.apache.naming.resources.Resource;
-import org.apache.naming.resources.ResourceAttributes;
-import org.glassfish.api.deployment.InstrumentableClassLoader;
-import org.glassfish.web.util.ExceptionUtils;
-import org.glassfish.web.util.IntrospectionUtils;
-import org.glassfish.hk2.api.PreDestroy;
 import com.sun.enterprise.security.integration.DDPermissionsLoader;
 import com.sun.enterprise.security.integration.PermsHolder;
+import com.sun.enterprise.util.io.FileUtils;
+import org.apache.naming.JndiPermission;
+import org.apache.naming.resources.*;
+import org.glassfish.api.deployment.InstrumentableClassLoader;
+import org.glassfish.hk2.api.PreDestroy;
+import org.glassfish.web.util.ExceptionUtils;
+import org.glassfish.web.util.IntrospectionUtils;
 
 import javax.naming.Binding;
 import javax.naming.NameClassPair;
@@ -179,7 +174,7 @@ public class WebappClassLoader
      * All permission.
      */
     private static final Permission ALL_PERMISSION = new AllPermission();
-    
+
     private static final String META_INF_SERVICES = "META-INF/services/";
 
 
@@ -257,6 +252,11 @@ public class WebappClassLoader
     protected JarFile[] jarFiles = new JarFile[0];
 
     /**
+     * Used in synchronize blocks since we can't use jarFiles directly (mutable!)
+     */
+    private final Object jarFilesPropertyLock = new Object();
+
+    /**
      * Lock to synchronize closing and opening of jar
      */
     protected final Object jarFilesLock = new Object();
@@ -295,7 +295,7 @@ public class WebappClassLoader
      * is for a web application context.
      */
     private final ConcurrentLinkedQueue<Permission> permissionList = new ConcurrentLinkedQueue<Permission>();
-    
+
     //holder for declared and ee permissions
     private PermsHolder permissionsHolder;
 
@@ -381,7 +381,7 @@ public class WebappClassLoader
      * resources.
      */
     boolean antiJARLocking = false;
-    
+
     /**
      * Reference to the JDBC Leak Prevention class.
      * Held uniquely due to the way it is accessed outside the normal
@@ -478,9 +478,9 @@ public class WebappClassLoader
         }
 
         @Override
-        public ClassLoader run() {       
+        public ClassLoader run() {
             return clazz.getClassLoader();
-        }           
+        }
     }
 
     // START PE 4985680
@@ -610,7 +610,7 @@ public class WebappClassLoader
         }
 
         if (securityManager != null) {
-            
+
             securityManager.checkSecurityAccess(DDPermissionsLoader.SET_EE_POLICY);
 
             Permission permission = null;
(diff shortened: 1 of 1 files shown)


appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -64,20 +64,15 @@
...
+import com.sun.enterprise.security.integration.PermsHolder;
 import com.sun.enterprise.util.io.FileUtils;
 import org.apache.naming.JndiPermission;
-import org.apache.naming.resources.DirContextURLStreamHandler;
-import org.apache.naming.resources.JarFileResourcesProvider;
-import org.apache.naming.resources.ProxyDirContext;
-import org.apache.naming.resources.WebDirContext;
-import org.apache.naming.resources.Resource;
-import org.apache.naming.resources.ResourceAttributes;
+import org.apache.naming.resources.*;
```
dmatej: I would stay with a list of classes if it is less than 11.

Pandrex247: That's up for debate - personally I hate using `*` unless I'm literally using every single class :P

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -837,56 +838,58 @@ public synchronized void addJar(String jar, JarFile jarFile, File file)
             logger.log(Level.FINER, "addJar({0})", jar);
         }
 
-        // See IT 11417
-        super.addURL(getURL(file));
+        synchronized (jarFilesPropertyLock) {
```
dmatej: The method is already synchronized, how could this help? Two locks can cause deadlock ... but ... in previous version jarFiles are used also as a lock, so that's the same ...

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -2711,7 +2729,7 @@ protected ResourceEntry findResourceInternal(String name, String path) {
         entry = findResourceInternalFromRepositories(name, path);
 
         if (entry == null) {
-            synchronized (jarFiles) {
```
dmatej: Wouldn't be better to extract the synchronized block to a synchronized method, so it would use the same "default" lock?
Or all accesses to the final field should use own but same lock ... or use Concurrent implementation if it would make sense (I don't think it does).

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
dmatej: Yes, my experience is that if you use more than 10 classes from the same package in imports, usually the new class asks to be refactored, so finally we end up in the same nice state :)

pdudits:
Were you able to confirm that the race condition is between `findResourceInternalFromJars` and `addJar`?

My hypothesis would be that it is race condition between `findResourceInternalFromJars` and `closeJars`.

 `findResourceInternalFromJars` confirms that jars are still open by invoking `openJars`:
https://github.com/payara/Payara/blob/c2905122f21f49ed1455612bdc3aded5b1d1f2ce/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L2838

`openJars` synchronizes on `jarFilesLock`:

https://github.com/payara/Payara/blob/c2905122f21f49ed1455612bdc3aded5b1d1f2ce/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L2561

But before `findResourceInternalFromJars` proceeds to accessing the jars (or while it is doing it), `closeJars` get called, and clears the array:

https://github.com/payara/Payara/blob/c2905122f21f49ed1455612bdc3aded5b1d1f2ce/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java#L1929

It synchronizes on same lock liken `openJars`, but that method exited a while ago. So if something needs synchronization, it might be `findResourceInternalFromJars` itself.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
rdebusscher: the method addJar() is synchronized, but other parts of the class requires synchronized access to the property jarFiles.  But the synchronization on the method does not synchronize on the property.

In this class, there are synchronization block on 6 different items.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
rdebusscher: I don't want to change the synchronization fundamentally since there seems a lot of synchronization logic in the code. And changing those things means that we probably break the server functionality.

rdebusscher:
@pdudits `closeJars` is also a possibility, but unlikely since issue occurs during startup of application within a @Singleton @Startup @PostConstruct method. So runtime is not stopping.

But I'll add it since it can lead to some issues in the future.

pdudits:
> 
> 
> @pdudits `closeJars` is also a possibility, but unlikely since issue occurs during startup of application within a @singleton @startup @PostConstruct method. So runtime is not stopping.

But `closeJars` is not only called on stopping. Quite the opposite, it's called at the end of startup to fight windows' habit of sticking to file references and preventing their deletion:

https://github.com/rdebusscher/Payara/blob/b1e5172/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java#L5788

rdebusscher:
I used to the (already present) jarFilesLock object within the findResourceInternalFromJars method since that one was already protecting the access to property jarFiles.
This means that accessing the jar files is now synchronized with start up but also with shutdown.

appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
```diff
@@ -837,56 +838,58 @@ public synchronized void addJar(String jar, JarFile jarFile, File file)
...
-                result[i] = paths[i];
-            }
-            result[paths.length] = jar;
-            paths = result;
+                String[] result = new String[paths.length + 1];
+                for (int i = 0; i < paths.length; i++) {
+                    result[i] = paths[i];
+                }
+                result[paths.length] = jar;
+                paths = result;
```
pdudits: These 4 idioms for appending arrays are quite disturbing here. 

The design might stem from pre-1.4 era, where there was nothing similar to `CopyOnWriteArrayList<PathEntry>`, and is instead simulated with multiple arrays. But these should at least be refactored into form `paths = appendArray(paths, entry)`. And internally use `Arrays.copyOf(arr, arr.length+1)`.

rdebusscher: Correct, but the goal of this PR is not to update some already existing code constructs.

rdebusscher:
>But closeJars is not only called on stopping. Quite the opposite, it's called at the end of startup to fight windows' habit of sticking to file references

But jars are reopened immediately again to have the classloader working. So not an efficient part of this old code.

rdebusscher:
Customer verified custom build with this changes and reports issue is fixed.  Good to be merged in master

Pandrex247:
Jenkins test please

PR #4380, opened 2019-12-12
CUSTCOM-77 Payara-Samples integrated to the Payara build

dmatej:
# Description

Examples how to run:
```
 # run only samples, payara instance managed by Maven
mvn clean install -Ppayara-samples -pl :payara-samples -amd -fae

# run payara build and samples directly at once
mvn clean install -PQuickBuild,payara-samples -fae 

# run samples and all it's modules
mvn clean install -PQuickBuild,payara-samples -pl :payara-samples -amd -fae 

# with different implementation; some modules will fail because groupIds changed.
mvn clean install -PQuickBuild,payara-samples -pl :payara-samples -amd -Dpayara.version=5.193.1 -fae 

# in payara-samples directory
cd appserver/tests/payara-samples;
mvn clean verify -fae 

# test embedded payara and jersey-client running in the same JVM (currently fails on 5.193.1 and newer)
mvn clean verify -Ppayara-samples -fae -rf :payara-samples -Dpayara.embedded.version=5.193.1

# remote tests, payara managed manually - 6 modules should fail, due to some undeployment bug will fail one more if tried more times on the same domain
mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:payara-samples -Ppayara-server-remote -amd -Dpayara.home=/app/payara5

# micro
mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:payara-samples -amd -Ppayara-micro-managed 

# 
```
# Results

remote: 5/6 failures on master
managed: 2 failures on master, 1 with -Dpayara.version=5.193
micro: 5 failures on master

# Warning

The build uses target in the current directory of the reactor, so changing current directory still affects results, because clean does not delete unpacked payara if used with -rf or -pl skipping project in current directory.
This will be resolved after finishing PAYARA-4176 after which we will probably use different mechanism.

# Remaining failures

mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:embedded-vs-jersey
- fails always, will be fixed under CUSTCOM-70

mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:openid 
- fails on remote due to hostname verification
- passes on managed
- on micro ignored

mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:jaxws-tracing
- fails on all profiles
- on managed+remote - tracing is not enabled
- on micro - wsdl is not on the expected context

mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:formauth
- fails only on micro, HTTP 200 instead of 403

mvn clean verify -Ppayara-samples -fae -pl fish.payara.samples:microprofile-secure-endpoints 
- fails on remote (hostname verification)
- fails on micro (No https URL could be created from http://localhost:8910/dcf53e26-e8a6-47a9-b67f-317ad93139ad/)

### DIFF
--- a/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
+++ b/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
@@ -55,41 +55,9 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>jakarta.platform</groupId>
+            <artifactId>jakarta.jakartaee-api</artifactId>
+        </dependency>
     </dependencies>
-
-    <profiles>
-        <profile>
-            <id>payara4</id>
-            <activation>
-                <property>
-                    <name>payara.version</name>
-                    <value>/4.+/</value>
-                </property>
-            </activation>
-            <dependencies>
-                <dependency>
-                    <groupId>fish.payara.arquillian</groupId>
-                    <artifactId>payara-client-ee7</artifactId>
-                    <scope>test</scope>
-                </dependency>
-            </dependencies>
-        </profile>
-    
-        <profile>
-            <id>payara5</id>
-            <activation>
-                <property>
-                    <name>payara.version</name>
-                    <value>/5.+/</value>
-                </property>
-            </activation>
-            <dependencies>
-                <dependency>
-                    <groupId>fish.payara.arquillian</groupId>
-                    <artifactId>payara-client-ee8</artifactId>
-                    <scope>test</scope>
-                </dependency>
-            </dependencies>
-        </profile>
-    </profiles>
 </project>
--- a/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
+++ b/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
@@ -55,9 +55,5 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
-        <dependency>
-            <groupId>jakarta.platform</groupId>
-            <artifactId>jakarta.jakartaee-api</artifactId>
-        </dependency>
     </dependencies>
 </project>
--- a/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
+++ b/appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
@@ -44,13 +44,30 @@
 
     <parent>
         <groupId>fish.payara.samples</groupId>
-        <artifactId>classpath</artifactId>
+        <artifactId>samples-classpath-checked</artifactId>
         <version>5.201-SNAPSHOT</version>
     </parent>
 
     <artifactId>embedded-vs-jersey</artifactId>
+    <name>Payara Samples - Running Jersey Client with Payara Embedded</name>
+
+    <properties>
+        <payara.embedded.version>${payara.version}</payara.embedded.version>
+    </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-client</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>fish.payara.extras</groupId>
+            <artifactId>payara-embedded-all</artifactId>
+            <!-- forced override required to test different versions -->
+            <version>[${payara.embedded.version}]</version>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -13,7 +13,7 @@
     </parent>
 
     <groupId>fish.payara.samples</groupId>
-    <artifactId>payara-samples-aggregator</artifactId>
+    <artifactId>payara-samples</artifactId>
     <packaging>pom</packaging>
 
     <name>Payara Samples - root</name>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -67,10 +67,6 @@
         </repository>
     </repositories>
 
-    <dependencies>
-        <!-- DO NOT ADD DEPENDENCIES -->
-    </dependencies>
-
     <dependencyManagement>
         <dependencies>
             <dependency>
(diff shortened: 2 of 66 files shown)


dmatej:
Jenkins test please

dmatej:
(Jenkins build should pass, because payara-samples profile is not enabled -> payara-samples module will not be executed).

appserver/tests/payara-samples/pom.xml
```diff
@@ -103,29 +118,11 @@
             <dependency>
                 <groupId>io.opentracing</groupId>
                 <artifactId>opentracing-api</artifactId>
-                <version>0.30.0</version>
+                <version>0.32.0</version>
```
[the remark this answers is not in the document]
dmatej: Not sure. What do you think, @pdudits ?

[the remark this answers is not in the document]
dmatej: Btw - BOM maybe should use payara.version property instead of project.version ... but then samples cannot be executed with older version. There will be always issues with compatibility if we would try to run samples with another version than the current.
I would prefer using jakartaee-bom here and not payara.bom, because we use directly the zip file, payara-bom should not be needed here.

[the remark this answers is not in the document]
pdudits: The answer is "yes" to both points ;)

Opentracing API is included in the BOM

https://github.com/payara/Payara/blob/master/api/payara-bom/pom.xml#L586

Samples should use payara.version to import the BOM. Because it is not just about Jakarta APIs, it is also about microprofile APIs, opentracing apis, and jersey implementations for client-side tests (which break in embedded mode if they differ).

appserver/tests/payara-samples/pom.xml
```diff
@@ -53,24 +52,35 @@
         <module>classpath</module>
     </modules>
 
-    <dependencies>
-        <!-- DO NOT ADD DEPENDENCIES -->
-    </dependencies>
```
pdudits: This served as warning, that adding dependency here will break classpath module, and I think it should remain here.

appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
```diff
@@ -56,40 +56,4 @@
...
-                    <value>/5.+/</value>
-                </property>
-            </activation>
-            <dependencies>
-                <dependency>
-                    <groupId>fish.payara.arquillian</groupId>
-                    <artifactId>payara-client-ee8</artifactId>
-                    <scope>test</scope>
-                </dependency>
-            </dependencies>
```
pdudits: the dependency was part of test, should remain here.

appserver/tests/payara-samples/pom.xml
```diff
@@ -53,24 +52,35 @@
...
+    </repositories>
 
     <dependencyManagement>
         <dependencies>
             <dependency>
                 <groupId>fish.payara.api</groupId>
                 <artifactId>payara-bom</artifactId>
-                <!-- A workaround for case when module redefines its version (rest-monitoring-war) -->
-                <version>${project.parent.version}</version>
+                <version>${project.version}</version>
```
pdudits: Yes, using `${payara.version}` is the best approach here.

Then almost all of the dependency management section is then unnecessary, with exception to internal module references.

appserver/tests/payara-samples/samples-programmatic/pom.xml
```diff
@@ -19,6 +19,10 @@
             <groupId>fish.payara.extras</groupId>
             <artifactId>payara-micro</artifactId>
         </dependency>
+        <dependency>
+            <groupId>fish.payara.server.internal.extras</groupId>
+            <artifactId>payara-micro-boot</artifactId>
+        </dependency>
```
pdudits: contents of micro boot is within `payara-micro`. Why is explicit dependency to internal module necessary?

appserver/tests/payara-samples/pom.xml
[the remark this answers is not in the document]
pdudits: On the other hand, the reason for moving samples into the main tree was, that you should be running samples against same matching payara version. But adding indirection with `payara.version` allows for some wiggle room between snapshot and release versions.

appserver/tests/payara-samples/pom.xml
dmatej: Warning not needed, there is maven enforcer plugin to detect it.
Generally there should not be any implicit dependencies, now remained junit, hamcrest and com.sun:tools-jar.

appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
dmatej: It was not used. Currently we have reported issue starting from 5.193.1, test passes on 5.193 and 5.192, so we broke it. Now is not possible to use embedded payara (web/all) with jersey-client and even it's own jersey-client. Fails on some runlevel issue in request tracing filter.
I will try to fix it next week under CUSTCOM-70

appserver/tests/payara-samples/samples-programmatic/pom.xml
dmatej: Not true. Micro's dependencies are all optional, so they have to be explicitly declared.

appserver/tests/payara-samples/pom.xml
dmatej: Partially true, except the fact that bom brings compile scope to anything. Real time scope then depends on the "root" artifact. But even with these rules it helped much :)

dmatej:
Jenkins test please

appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
MattGill98: I'm fairly sure that's a legit issue that this test tests for 😉

MarkWareham:
why the do not merge label @dmatej ?

appserver/tests/payara-samples/classpath/embedded-vs-jersey/pom.xml
dmatej: It is and it is broken, as I wrote, there already exists issue for that, CUSTCOM-70. The dependency was not used, tests work same with or without it (passes for 5.192, 5.193, fails for 5.193.1, 5.194 and master).

dmatej:
Rebased, fixed conflicts. Blocked by reviewers.

MarkWareham:
jenkins test please

rdebusscher:
In addition, I have also the following error which needs to be fixed later on.

> (configure-domain-ejb-invoker) on project client: Execution configure-domain-ejb-invoker of goal org.glassfish.build:glassfishbuild-maven-plugin:3.2.20.payara-p2:exec failed: Execute failed: java.io.IOException: Cannot run program "/glassfish/bin/asadmin" (in directory "/Users/rubus/payara/payara-server/appserver/tests/payara-samples/samples/ejb-http-remoting/client/target"): error=2, No such file or directory

PR #4379, opened 2019-12-12
CUSTCOM-56 Fix production domain.xml

pzygielo:
# Description
This is a bug fix.

#4366

# Testing

### New tests
Validation of specified xmls done in the build process.

### Testing Performed
Full build (`mvn clean install`) on travis.

### Testing Environment
Ubuntu 16.04.6 LTS
OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1ubuntu1~16.04.1-b10)
Maven 3.6.1

PR #4383, opened 2019-12-13
CUSTCOM-56 Verify Domain.xml on Domain Creation

MattGill98:
# Description
This is a bug fix.

The Web Profile Production domain.xml wasn't well-formed, so has been fixed. 

Domain xml verification has been added to domain creation. This should prevent invalid or badly formed XML documents being added again.

 

 

# Testing

### New tests
Sort of? The domain creation in the build will fail, which is sort of a test in and of itself.

### Testing Performed

Running domain creation manually with badly formed xml and invalid xml.

### Test suites executed

N/A

# Documentation

N/A

# Notes for Reviewers

### DIFF
--- a/appserver/admin/production_domain_template_web/src/main/resources/config/domain.xml
+++ b/appserver/admin/production_domain_template_web/src/main/resources/config/domain.xml
@@ -405,11 +405,11 @@
         <jvm-options>[9|]--add-opens=java.base/java.lang=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/java.net=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/java.nio=ALL-UNNAMED</jvm-options>
-        <jvm-options>[9|]--add-opens=java.base/java.util=ALL-UNNAMED</jvm-options
+        <jvm-options>[9|]--add-opens=java.base/java.util=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/sun.nio.ch=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.management/sun.management=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.naming/javax.naming.spi=ALL-UNNAMED</jvm-options>
-        <jvm-options>[9|]--add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED</jvm-options
+        <jvm-options>[9|]--add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/sun.net.www.protocol.jrt=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/sun.net.www.protocol.jar=ALL-UNNAMED</jvm-options>
         <jvm-options>-XX:+UseG1GC</jvm-options>
--- a/nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
+++ b/nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
@@ -37,21 +37,30 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
- * Portions Copyright [2017] Payara Foundation and/or affiliates
+ * Portions Copyright [2017-2019] Payara Foundation and/or affiliates
  */
 
 package org.glassfish.internal.api;
 
-import com.sun.enterprise.module.ModulesRegistry;
-import com.sun.enterprise.module.single.StaticModulesRegistry;
-import org.jvnet.hk2.annotations.Service;
-import org.glassfish.common.util.Constants;
-import org.glassfish.hk2.api.Rank;
-import org.glassfish.hk2.api.ServiceLocator;
-import org.glassfish.hk2.runlevel.RunLevel;
-import com.sun.enterprise.config.serverbeans.ConfigBeansUtilities;
+import java.io.File;
+import java.net.URLClassLoader;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.ArrayList;
+
 import javax.inject.Inject;
 import javax.inject.Singleton;
+
+import com.sun.enterprise.config.serverbeans.ConfigBeansUtilities;
+import com.sun.enterprise.module.ModulesRegistry;
+import com.sun.enterprise.module.single.StaticModulesRegistry;
+import com.sun.enterprise.util.SystemPropertyConstants;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+import org.glassfish.hk2.api.ServiceLocator;
+import org.glassfish.hk2.runlevel.RunLevel;
+import org.jvnet.hk2.annotations.Service;
 
 /**
  * Global class for storing the service locator for all hk2 services
@@ -65,8 +74,9 @@ import javax.inject.Singleton;
 public class Globals {
 
     private static volatile ServiceLocator defaultHabitat;
+    private static volatile ServiceLocator defaultHabitatWithModules;
 
-    private static Object staticLock = new Object();
+    private static final Object STATIC_LOCK = new Object();
     
     // dochez : remove this once we can get rid of ConfigBeanUtilities class
     @SuppressWarnings("unused")
@@ -134,7 +144,7 @@ public class Globals {
      */
     public static ServiceLocator getStaticHabitat() {
         if (defaultHabitat == null) {
-            synchronized (staticLock) {
+            synchronized (STATIC_LOCK) {
                 if (defaultHabitat == null) {
                     ModulesRegistry modulesRegistry = new StaticModulesRegistry(Globals.class.getClassLoader());
                     defaultHabitat = modulesRegistry.createServiceLocator("default");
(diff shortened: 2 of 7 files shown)


MattGill98:
jenkins test please

appserver/admin/production_domain_template_web/src/main/resources/config/domain.xml
```diff
@@ -405,11 +405,11 @@
         <jvm-options>[9|]--add-opens=java.base/java.lang=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/java.net=ALL-UNNAMED</jvm-options>
         <jvm-options>[9|]--add-opens=java.base/java.nio=ALL-UNNAMED</jvm-options>
-        <jvm-options>[9|]--add-opens=java.base/java.util=ALL-UNNAMED</jvm-options
+	      <jvm-options>[9|]--add-opens=java.base/java.util=ALL-UNNAMED</jvm-options>
```
dmatej: Use spaces, not tabs!

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigParser.java
```diff
@@ -161,36 +176,44 @@ public DomDocument parse(URL source, DomDocument document, Dom parent) {
...
-                    if (Logger.getAnonymousLogger().isLoggable(Level.FINE)) {
-                        Logger.getAnonymousLogger().fine("closing element type " + localName);
-                    }
+                    handleError("Closing element type " + localName, Level.FINE);
                     depth--;
                 }
-                if (Logger.getAnonymousLogger().isLoggable(Level.FINE) && tag==START_ELEMENT) {
-                    Logger.getAnonymousLogger().fine("Jumping over " + in.getLocalName());
+                if (tag == START_ELEMENT) {
+                    handleError("Jumping over " + in.getLocalName(), Level.FINE);
```
dmatej: This would have poor performance, string will be created always, but if the level would be configured to INFO, the expensive message would be dropped. Use suppliers, they have much better performance.
Also this message is not an error ... name of the method is confusing.
```
private void log(Level level, Supplier<String> supplier) {...}
log(Level.FINE, () -> "Jumping over " + in.getLocalName());

```

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
```diff
@@ -144,7 +154,50 @@ public static ServiceLocator getStaticHabitat() {
...
+        File mdir = new File(idir, "modules");
+        for (File f : mdir.listFiles()) {
+            if (f.toString().endsWith(".jar")) {
+                urls.add(f.toURI().toURL());
+            }
+        }
+
+        final URL[] urlsA = urls.toArray(new URL[urls.size()]);
+
+        ClassLoader cl = (ClassLoader) AccessController.doPrivileged(new PrivilegedAction<Object>() {
```
dmatej: Set the action to the variable -> you don't need any casting then, and also you don't need an anonymous class, you can use lambda.
```
PrivilegedAction<URLClassLoader> action = () -> {...};
ClassLoader cl = AccessController.doPrivileged(action);
```

dmatej: Also ... I will have conflicts in this class, but that's my problem ;-)

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
```diff
@@ -144,7 +154,50 @@ public static ServiceLocator getStaticHabitat() {
...
+            @Override
+            public Object run() {
+                return new URLClassLoader(urlsA, Globals.class.getClassLoader());
+            }
+        });
+
+        ModulesRegistry registry = new StaticModulesRegistry(cl);
+        return registry.createServiceLocator("default");
```
dmatej: I doubt that this is correct - the StaticModulesRegistry will exist only until it would be destroyed by the GC? How long after exiting this method will locator work correctly?

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigParser.java
```diff
@@ -74,16 +74,27 @@
  * @author Kohsuke Kawaguchi
  */
 public class ConfigParser {
+
+    private static final Logger LOGGER = Logger.getAnonymousLogger();
```
dmatej: I don't like using anonymous logger - it cannot be configured well in logging.properties. Better would be `Logger.getLogger(ConfigParser.class.getName());`

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigParser.java
```diff
@@ -74,16 +74,27 @@
...
 
     public ConfigParser(ServiceLocator habitat) {
-        this.habitat = habitat;
+        this(habitat, true);
     }
 
+    public ConfigParser(ServiceLocator habitat, boolean logErrors) {
+        this.habitat = habitat;
+        this.logErrors = logErrors;
+        this.errors = new ArrayList<>();
```
dmatej: Somewhere in javadoc should be said that the parser is not reusable (errors have state before and after first use)

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
MattGill98: You're right sorry, I'd been working on Payara 4 so I'd been in the mindset of not using Java 8 features 😆

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
MattGill98: What do you mean by this?

MarkWareham:
jenkins test please

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
dmatej: StaticModuleRegistry serves as a register of modules, it has also shutdown method ... it does not implement Closeable, because the implementation is very old, but I tried to investigate a bit ... it has shutdown() method and contains a Register which can use timers which are interrupted on shutdown ... so at least it should call shutdown in finally. But then I'm not sure if the provided service object is still valid or not (it might be also in inconsistent state if it would have a reference to the registry and trying to use it).
The registry does not have finalize method implementation, so when it is garbage collected, nothing stops the timer ... so this can lead to resource leak ... (how do we work with *ModuleRegistry in other cases?)

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigParser.java
```diff
@@ -161,36 +182,55 @@ public DomDocument parse(URL source, DomDocument document, Dom parent) {
...
+     *                       Object...)}
+     * @param parameters     the parameters to be filled into the base log message
+     */
+    private void log(Level level, String baseLogMessage, Object... parameters) {
+        if (level != null && LOGGER.isLoggable(level)) {
+            String logMessage = String.format(baseLogMessage, parameters);
+            if (logErrors) {
+                LOGGER.log(level, logMessage);
+            }
+            this.errors.add(logMessage);
```
dmatej: GOTO LABEL1: so if I would set the logging level to ALL, this would add even FINE messages to errors.

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/ConfigParser.java
```diff
@@ -161,36 +182,55 @@ public DomDocument parse(URL source, DomDocument document, Dom parent) {
...
-                    if (Logger.getAnonymousLogger().isLoggable(Level.FINE)) {
-                        Logger.getAnonymousLogger().fine("closing element type " + localName);
-                    }
+                    log(FINE, "Closing element type %s", localName);
                     depth--;
                 }
-                if (Logger.getAnonymousLogger().isLoggable(Level.FINE) && tag==START_ELEMENT) {
-                    Logger.getAnonymousLogger().fine("Jumping over " + in.getLocalName());
+                if (tag == START_ELEMENT) {
+                    log(FINE, "Jumping over %s", in.getLocalName());
```
dmatej: LABEL 1: logging fine message as an error, because log method will add this message to errors if it is loggable.

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
MattGill98: Hmm I hadn't spotted the shutdown method. Do you think a static reference should not be kept then?

dmatej: Why it cannot just use Globals.getStaticBaseServiceLocator()?

MattGill98: Because then the xml tags aren't recognised. The method is javadocced, I just put it somewhere common so that both sources could use the same function. If it's a big issue, I can just duplicate the code

MattGill98:
Closing as fixed by https://github.com/payara/Payara/pull/4379

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/Globals.java
dmatej: Hmm, tags are not recognized with getStaticBaseServiceLocator? That's weird ... because the method should do the same ... or I am missing something. I will try to checkout this PR after I will get rid of P4 build ...

PR #4389, opened 2019-12-16
CUSTCOM-83 Update Grizzly Dependencies to 2.4.3.payara-p10

MattGill98:
# Description
This is a bug fix. In some scenarios, an SSL connection is terminated before data is fully sent. In this case, the SSLBaseFilter will currently keep passing the incomplete data down the chain. An infinite loop was caused by the data handling filters deciding there's not enough data, and restarting the chain. This change makes sure that the SSLBaseFilter doesn't try to repeatedly process incomplete data on a terminated SSL connection.

# Important Info

### Dependant PRs 
https://github.com/payara/Payara_PatchedProjects/pull/285 for the artifact
https://github.com/payara/patched-src-grizzly/pull/20 for the source

# Testing

### Testing Performed

- Grizzly unit tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,7 +121,7 @@
         <!-- BOM-referenced versions -->
         <jakartaee.api.version>8.0.0</jakartaee.api.version>
         <servlet-api.version>4.0.2</servlet-api.version>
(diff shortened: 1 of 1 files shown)


MattGill98:
jenkins test please

PR #4394, opened 2019-12-18
CLOUD-13 Prevent Hazelcast from being started twice on Payara Micro

Cousjava:
# Description
This is a bug fix

As PayaraMicroImpl#L1060 will start Hazelcast whatever, having it
enabled in the domain.xml means that Hazelcast will be started earlier
then restarted when the bootstrap says to enable Hazelcast.

# Testing
### Testing Performed
Start Payara Micro, check the Hazelcast has only been started once.

### Test Suites Run
 - JavaEE7-Samples
 - JavaEE8-Samples

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.2

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/resources/MICRO-INF/domain/domain.xml
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/resources/MICRO-INF/domain/domain.xml
@@ -37,7 +37,7 @@
     and therefore, elected the GPL Version 2 license, then the option applies
     only if the new code is made subject to such option by the copyright
     holder.
-Portions Copyright [2016-2018] [Payara Foundation and/or its affiliates]
+Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
 -->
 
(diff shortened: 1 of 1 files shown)


Cousjava:
Jenkins test please

pdudits:
I think the reasons for the current behavior needs to be inspected more.

We did have straight `enabled=false` in bootup sequence before e11ebf3, there must have been a reason that led to this side-line signaling to HazelcastCore. Probably the startup was single-threaded back then, so this thread-local flag was enough.

My gut feeling is, that this negatively affects persistent EJB timers, as access to timer service (and to datagrid) is necessary during PREPARE phase. On the other hand, it did negatively affect it before that commit as well, but there was no DataGrid timer service back then. This bug counterbalanced that.

If we start disabled, it would likely be best to enable DataGrid before deployment.

Cousjava:
I have tested this, and it all works fine using the reproducer for PAYARA-2150.  Initialisation of ejbtimers occurs after datagrid is started, so all works fine.

Cousjava:
Testing this off master today and I can't find any problems with this. Tested with the reproducer for CUSTCOM-24 and I didn't have any issues.

PR #4396, opened 2019-12-18
CUSTCOM-56 Extend domain.xml validation to every domain.xml file

MattGill98:
The validation added in https://github.com/payara/Payara/pull/4379 was only performed for the production domain of payara-web. This commit extends the validation to every packaged domain.xml.

## Tests performed

`mvn validate` on each of the changed modules

`mvn validate` fails when the domain.xml isn't well formed

### DIFF
--- a/appserver/admin/gf_template/pom.xml
+++ b/appserver/admin/gf_template/pom.xml
@@ -130,6 +130,17 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <!-- Validate XML resources -->
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>xml-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>validate-xml-files</id>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 </project>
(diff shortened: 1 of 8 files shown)


MattGill98:
jenkins test please

pzygielo:
With multiple modules using the same plugin I propose to manage execution in top pom as well.

MattGill98:
I would tend to agree, but I often try to exclude execution information from top-level poms as it makes it less clear what the plugins are doing when you read the child poms. @dmatej what do you think?

pzygielo:
> I would tend to agree, but I often try to exclude execution information from top-level poms as it makes it less clear what the plugins are doing when you read the child poms.

Perhaps name/id execution in parent and explicitly use it in children:
https://github.com/MattGill98/Payara/compare/CUSTCOM-56-maven-plugin-completion...pzygielo:custcom56

pzygielo:
>  The right phase for this should be verify

My mistake of changing default `test` phase in first place.

dmatej:
Yes, test may be suitable phase too (perhaps even better)

MattGill98:
I've made a change which I think addresses all of the suggestions quite well, let me know what you both think

MattGill98:
jenkins test please

PR #4398, opened 2019-12-19
CUSTCOM-109 Injection of parameters into array will default to 0-length array

Cousjava:
# Description
This is a bug fix.

When an asadmin command has a Param that is an array and is optional then it it defaulted to null, which in some cases caused an NPE when Payara tried to iterate over it.

# Testing

### New tests
New `InjectionTest` unit test added.

### Testing Performed
Manual test - `send-asadmin-command` fails  before this PR, and succeeds afterwards.

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 18.04 with Maven 3.6.2


Cousjava:
Jenkins test please

PR #4399, opened 2019-12-19
CUSTCOM-153 maven plugin completion xml validation

dmatej:
# Description

In #4379 @pzygielo came with the idea to validate broken domain.xml in module we broke in one release.
@MattGill98 extended it in #4396  to validate all source domain.xml files.
I am expanding it here to validate all output xml files in modules where it was already enabled by @MattGill98 . I could be even enabled to all modules, unfortunately xml plugin does not have option to accept non-existing directories as valid set of files.

# Tests performed

Build everything and check if the plugin was called
mvn clean install -PBuildExtras,BuildDockerImages

Break some domain.xml and run 
mvn clean verify -pl :xxx
- were xxx is artifactId where is the domain.xml file used

# Note

There is also a set of templates which are invalid, because they contains a placeholder with `<` characters. They are excluded from tests.
If the maven plugin would be extended, we can move the configuration even higher.

### DIFF
--- a/appserver/admin/gf_template_web/pom.xml
+++ b/appserver/admin/gf_template_web/pom.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
-  
+
    Copyright (c) 2017-2019 Payara Foundation and/or its affiliates. All rights reserved.
-  
+
    The contents of this file are subject to the terms of either the GNU
    General Public License Version 2 only ("GPL") or the Common Development
    and Distribution License("CDDL") (collectively, the "License").  You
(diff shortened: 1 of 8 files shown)


appserver/admin/gf_template_web/pom.xml
```diff
@@ -124,12 +124,11 @@
                             <attach>false</attach>
                             <appendAssemblyId>false</appendAssemblyId>
                             <useProjectArtifact>false</useProjectArtifact>
-                        </configuration>                        
+                        </configuration>
                     </execution>
                 </executions>
             </plugin>
 
-            <!-- Validate XML resources -->
```
MattGill98: Why are comments being removed?

dmatej:
@MattGill98 We should validate what we distribute, not sources. Sources are usually already validated by editors (Eclipse, Idea), but files may be filtered - then your validation wouldn't detect problems.

But this my validation has another problem - it validates all xml files (except those excluded in one case), so it validates also junit xml outputs, which is not desired.

appserver/admin/gf_template_web/pom.xml
dmatej: This comment only duplicated information already visible in the id: <id>validate-xml-files</id>

MattGill98:
Okay I'll wait for an extra commit

dmatej:
Closed, because we created extra issue for this under CUSTCOM-153.

PR #4400, opened 2019-12-20
CUSTCOM-109 SendAsadminCommand now does a check for if supplied instances is null

Cousjava:
# Description
This is a bug fix.

# Important Info
This is an alternative to #4398

As the other commands which accept multiple operands have it as non-optional then I think a better approach is to only check in the command that does it this way, rather than doing a check every single time (which is less efficient).


Cousjava:
Jenkins test please

PR #4402, opened 2019-12-20
CUSTCOM-130 Docker images integrated and tested 

dmatej:
# Description

I needed this to test another issue for which I had no docker image.
See JIRA issue for more info.

Now it is possible to change the basic image, payara version used inside, create JDK8/11 image from the same code.

# Examples

## Build docker images from current snapshot (must be in maven cache) using defaults (will run also TestContainers tests)
`mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd;`

### Test it manually
`docker run -it -p 14848:4848/tcp -p 18080:8080/tcp payara/server-full:5.202-SNAPSHOT`

## Build docker images from current snapshot (must be in maven cache) based on openjdk11 on debian

After last commits must be executed individually, because Micro is Alpine based, but Server and Node are Debian based.

```
mvn clean install -PBuildExtras,BuildDockerImages -pl :server-full-docker-image, -amd 
-Ddockerfile.noCache=true 
-Dpayara.version=5.202-SNAPSHOT 
-Ddockerfile.tag=5.202-tmp-openjdk11 
-Ddocker.java.repository=adoptopenjdk/openjdk11
-Ddocker.java.tag=x86_64-debian-jdk-11.0.5_10-slim
```

## ... on ubuntu
```
mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd 
-Ddockerfile.build.noCache=true 
-Dpayara.version=5.202-SNAPSHOT 
-Ddockerfile.tag=5.202-tmp-openjdk11 
-Ddocker.java.repository=adoptopenjdk/openjdk11
-Ddocker.java.tag=x86_64-ubuntu-jdk-11.0.5_10-slim

```

### Test it
```
mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd 
-rf :docker-images-tests -Ddockerfile.tag=5.202-tmp-openjdk11
```

### Test it manually
`docker run -it -p 14848:4848/tcp -p 18080:8080/tcp payara/server-full:5.202-tmp-openjdk11 `

## Build docker image from maven installed payara 5.191.11-SNAPSHOT:

1. Remove IgnoreUnrecognizedVMOptions from Dockerfile
2. Run
```
mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd 
-Dpayara.version=5.191.11-SNAPSHOT
````

# Notes

I changed Dockerfiles a bit
- to have more maintable syntax without warnings I use default true value available in each ux shell - it prevents long diffs when adding or removing lines in commands
- now they are filtered by maven, so we can simply change some values without changing Dockerfile source
- I added JVM option to tolerate unsupported JVM options between JDK8 and 10 - without that is harder to experiment with new possibilities
- I added some packages available by default in ubuntu/zulu, but missing in debian, etc.
- payara is not downloaded, but used from Maven

### CODE REVIEW

**Comment by rdebusscher** [comment]
How is the created docker image pushed to [DockerHub](https://hub.docker.com/r/payara/server-full/) ?

File: appserver/extras/docker-images/server-full/src/main/docker/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,93 @@
...
+    && ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} create-jvm-options '-XX\:+UnlockExperimentalVMOptions:-XX\:+IgnoreUnrecognizedVMOptions:-XX\:+UseCGroupMemoryLimitForHeap:-XX\:MaxRAMFraction=1' \
+    && ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} set-log-attributes com.sun.enterprise.server.logging.GFFileHandler.logtoFile=false \
+    && ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} stop-domain ${DOMAIN_NAME} \
+    # Cleanup unused files
+    && rm -rf \
+        /tmp/tmpFile \
+        payara.zip \
+        ${PAYARA_DIR}/glassfish/domains/${DOMAIN_NAME}/osgi-cache \
+        ${PAYARA_DIR}/glassfish/domains/${DOMAIN_NAME}/logs \
+        ${PAYARA_DIR}/glassfish/domains/domain1 \
```
rdebusscher: domain name hardcoded

rdebusscher:
Docker image is 709 MB, almost 200 Mb more than 5.194 (which was 513 MB)

This is due to the COPY and in next RUN command the mv which duplicates the Payara content in the Docker Image

dmatej:
> How is the created docker image pushed to [DockerHub](https://hub.docker.com/r/payara/server-full/) ?

That must be resolved by someone else, purpose of this was to use it even with snapshots in tests, but I think it should be principially same as with the "brother" module, payaraserver-node.

Generally I would expect that it could be done by some script in maven release phase, or run additionaly.

dmatej:
@rdebusscher Aha, I see it! Thank you! Yep, I replaced the downloding with copying ...

appserver/extras/docker-images/pom.xml
```diff
@@ -10,64 +10,150 @@
...
+                                        </includes>
+                                    </resource>
+                                </resources>
+                            </configuration>
+                        </execution>
+                    </executions>
+                </plugin>
+                <plugin>
+                    <groupId>com.spotify</groupId>
+                    <artifactId>dockerfile-maven-plugin</artifactId>
```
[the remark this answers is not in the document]
dmatej: There is no plugin with such id. This is the only one:
https://mvnrepository.com/artifact/com.spotify/dockerfile-maven-plugin

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: It is correct - this is a list of files and directories which should be removed.

dmatej:
The sizes should be alright now.
The plugin is correct and it comes from the repository - we use latest version of it.
I fixed several other issues.

The server-full depends on debian/ubuntu based JDK images, because it uses apt-get, which is not supported on alpine or "rpm" distributions. But I have found that tiny can be installed on ubuntu from apt too, but changing this is not a target of this issue/pr.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
rdebusscher: But when a Docker image is created with domain1, this will remove the domain we actually need.

rdebusscher: By either copying the domain1 or production domain into the image, another 3MB can be gained. (by filtering it out in the maven-resources-plugin execution)
Probably can do this later.

Pandrex247:
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 6.454 s <<< FAILURE! - in fish.payara.distributions.docker.PayaraServerFullTest
[ERROR] fish.payara.distributions.docker.PayaraServerFullTest  Time elapsed: 6.454 s  <<< ERROR!
java.lang.ExceptionInInitializerError
        at fish.payara.distributions.docker.PayaraServerFullTest.<clinit>(PayaraServerFullTest.java:77)
Caused by: java.util.NoSuchElementException: No value present
        at fish.payara.distributions.docker.PayaraServerFullTest.<clinit>(PayaraServerFullTest.java:77)
```
#doesntworkonmymachine

dmatej:
> ```
> [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 6.454 s <<< FAILURE! - in > #doesntworkonmymachine

What exact command you executed?

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: Good idea!

dmatej: It does not create domains, it only uses domain production which is already in zip file dependency together with domain1 which is not needed. 
But to filter out the domain1 even before running docker plugin is good idea.
Second thing is that tiny installation - it could be somehow merged from it's own image OR installed by system command instead dlownloading it by wget which must be too installed first ... I don't know how much we can spare with that (and if) ...
But filter out the domain should be simple :)

appserver/extras/docker-images/pom.xml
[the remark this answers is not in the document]
dmatej: Aha, now I understand ... I did not know it had yet another repo. We use latest version, at least I checked it yesterday :-)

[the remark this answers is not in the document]
dmatej: The risky thing is that thay say even here that they are fixing only critical things and they don't expect further evolution ... so in some future we should be aware of alternatives ...

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
rdebusscher: >Second thing is that tiny installation - it could be somehow merged from it's own image OR installed by system command instead dlownloading it by wget which must be too installed first ..

It depends on what the tiny Image is based. Bit since wget and install (and gpg verification) is in one RUN command, and thus in one layer, you can't optimize it more than it already is today.

Pandrex247:
> What exact command you executed?

Just `mvn clean install -PQuickBuild,BuildDockerImages`

It got to the point of testing the Docker stuff and then barfed - it downloaded some layers and then just failed on that with no other logs immediately visible.

dmatej:
> > What exact command you executed?
> 
> Just `mvn clean install -PQuickBuild,BuildDockerImages`
> 
> It got to the point of testing the Docker stuff and then barfed - it downloaded some layers and then just failed on that with no other logs immediately visible.

It seems you have wring operating system! :dagger: https://github.com/testcontainers/testcontainers-java/issues/666#issuecomment-385344506

What version of docker do you have? Mine is 
```
docker -v
Docker version 19.03.2, build 6a30dfca03

```

Pandrex247:
```
andrew@OpenSUSE:~/Git/Payara> docker -v  
Docker version 19.03.1, build 74b1e89e8ac6
```

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: I moved filtering to maven phase - server-full is 3 MBs lighter, server-node is 6 MB lighter.

dmatej:
Jenkins test please

dmatej:
I'm leaving DO NOT MERGE label still set, until we will not find out why @Pandrex247 s build fails or if we can ignore it.

MarkWareham:
> I'm leaving DO NOT MERGE label still set, until we will not find out why @Pandrex247 s build fails or if we can ignore it.

Have you made any fixes, or does the problem require investigation?

pdudits:
> Built a new VM since I'm now working from home and it works :)

The usual cause of failures was (and will be) import of GPG keys to verify `tini`, which tends to time out. The current version uses larger pool of servers so it is a bit more reliable, but `tini` should be probably built as a supplementary image outside of the build.

dmatej:
Another cause are those "tiny" downloads - usually timeout.
`https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini`

dmatej:
Jenkins test please

Pandrex247:
Disclaimer: didn't realised you'd force-pushed a load of other stuff since yesterday.
I was reviewing yesterday's commit 1249a54ff8957f619d8be7dec406984ee179800e (Fixed executable file permissions in builds on windows machines)

PR #4405, opened 2020-01-01
Thread Naming for easier Thread Dump analysis - CUSTCOM-145

rdebusscher:
# CUSTCOM-145: Assign meaningful thread names

During Thread dump analysis, it was not always easy to identify what threads are due to their name (Thread-[n] or pool-[x]-thread-[y])

After this update, threads name are mostly set. Some threads are still un named and could be done in

osgi-javaee-base org.glassfish.osgijavaeebase.ExtenderManager.GlassFishServerTracker#addingService
org.glassfish.osgijavaeebase.JavaEEExtender.HybridBundleTrackerCustomizer#addingBundle

and Apache Felix

# Description
This is a improvement

Should be easier to identify threads in Thread Dump.

# Testing

### Testing Performed

Manual testing to identify the list of threads through JConsole.

### Test suites executed
- Quicklook

### Testing Environment

Mac OS

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingOutputStream.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingOutputStream.java
@@ -134,7 +134,7 @@ public class LoggingOutputStream extends ByteArrayOutputStream {
                 }
             }
         };
-        pump.setName("logging output pump");
+        pump.setName("Logging output pump");
         pump.setDaemon(true);
         pump.start();
     }
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/JobCleanUpService.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/JobCleanUpService.java
@@ -100,21 +100,20 @@ public class JobCleanUpService implements PostConstruct,ConfigListener {
     public void postConstruct() {
         logger.log(Level.FINE,KernelLoggerInfo.initializingJobCleanup);
 
-
+        if (Version.getFullVersion().contains("Micro")) {
+            //if Micro we don't have any jobs to cleanup
+            return;
+        }
 
         scheduler = Executors.newScheduledThreadPool(10, new ThreadFactory() {
             @Override
             public Thread newThread(Runnable r) {
                 Thread result = new Thread(r);
                 result.setDaemon(true);
+                result.setName("Job Cleanup Service");
                 return result;
             }
         });
-        
-        if (Version.getFullVersion().contains("Micro")) {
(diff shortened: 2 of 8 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/JobCleanUpService.java
```diff
@@ -100,21 +100,20 @@
     public void postConstruct() {
         logger.log(Level.FINE,KernelLoggerInfo.initializingJobCleanup);
 
-
+        if (Version.getFullVersion().contains("Micro")) {
```
rdebusscher: The scheduler is also *not* used in Payara Micro so no need to create it.

jbee: Shouldn't the log output also be affected by the guard?

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/LoggingOutputStream.java
```diff
@@ -134,7 +134,7 @@ public void run() {
                 }
             }
         };
-        pump.setName("logging output pump");
+        pump.setName("Logging output pump");
```
dmatej: I will have conflict here, but no problem...

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/JobCleanUpService.java
dmatej: I agree with Jan here, the logger does not log "postConstruct" entry here, but on Micro lies that it is initializing the cleanup service, but it does not do that.

dmatej: Btw on line 151 cleanupFuture variable resulting from scheduling can be removed.

rdebusscher: @dmatej Correct regarding the cleanupFuture variable

But there are a lot more code 'issues'. (12 in total) Unused injection point, unused class variables, unused method parameter, unnecessary return types, unnecessary switch  etc ...

But cleanup of the class should not be the goal of this PR. (as it will hide later on what actually was changed)

dmatej: I agree, that was only first thing I noticed. Other things may have side effects, so should be postponed.

dmatej:
Jenkins test please

PR #4409, opened 2020-01-06
CUSTOM-142 Delete JVM options with min max versions.

rdebusscher:
This is a bug fix

When you try to delete a JVM option which has defined a min and/or max version, the option cannot be found (and thus not deleted).

This PR fixes the delete logic of the JVM Options.

# Important Info

No Unit Test created due to proxy requirements of JavaConfig within `org.jvnet.hk2.config.ConfigSupport#apply(org.jvnet.hk2.config.SingleConfigCode<T>, T)` 

# Testing

### Testing Performed

./asadmin start-domain
./asadmin create-jvm-options '[1.8.123|1.8.231]-DtestOption'
./asadmin delete-jvm-options '[1.8.123|1.8.231]-DtestOption'

- Quicklook

### Testing Environment

Maven 3.5.4
JDK 1.8u181

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/commands/DeleteJvmOptions.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/commands/DeleteJvmOptions.java
@@ -162,7 +162,8 @@ public final class DeleteJvmOptions implements AdminCommand, AdminCommandSecurit
         SingleConfigCode<JvmOptionBag> scc = (JvmOptionBag bag1) -> {
             List<String> jvmopts = new ArrayList<>(bag1.getJvmRawOptions());
             int orig = jvmopts.size();
-            boolean removed = jvmopts.removeIf(option -> toRemove.contains(new JvmOption(option).option));
+            // using new JvmOption(option).toString() (instead op option directly) to make sure the correct formatting is applied.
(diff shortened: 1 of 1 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/commands/DeleteJvmOptions.java
```diff
@@ -162,7 +162,8 @@ private void deleteX(final JvmOptionBag bag, final List<String> toRemove, final
         SingleConfigCode<JvmOptionBag> scc = (JvmOptionBag bag1) -> {
             List<String> jvmopts = new ArrayList<>(bag1.getJvmRawOptions());
             int orig = jvmopts.size();
-            boolean removed = jvmopts.removeIf(option -> toRemove.contains(new JvmOption(option).option));
+            // using new JvmOption(option).toString() (instead op option directly) to make sure the correct formatting is applied.
+            boolean removed = jvmopts.removeIf(option -> toRemove.contains(new JvmOption(option).toString()));
```
dmatej: This method has expectations ... it expects that formatting is done in option.toString() and it's result will be same as strings in bag1.getJvmRawOptions(). 
Seems fragile to me, 

getJvmRawOptions does not have javadoc, but it looks it is the content of the XML element.
JvmOption.toString prints also minVersion and maxVersion, seems it is really symetric at this moment ... ok.

dmatej:
Jenkins test please

PR #4419, opened 2020-01-09
CUSTCOM-168 OpenID Connect: Fixed simultaneous redirects and invalidation of session

parysto:
# Description
This is a bug fix / feature. 

1. If a page contains multiple protected resources and authentication (via OpenID Connect) is activated, multiple redirects to the OpenID Connect provider are performed (one for each protected resource). This is especially the case when auto refresh of the token is activated, the token expires and a new access token must be requested. To prevent multiple redirects, I have synchronized the authentication/refresh logic on the user's session.
2. If auto refresh is enabled for the access token, the principal was registered in the session again after  refreshing the token. This caused the currently active session to be invalidated (destroys all session scoped beans and session attributes).

# Testing

### Testing Performed 
Deployed developed app on instance with fix installed and compared/validated behavior while using the app (especially when refresh of token happened).

### Testing Environment
OpenJDK 11 on Ubuntu 19.10

### DIFF
--- a/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
+++ b/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
@@ -147,7 +147,7 @@ public class OpenIdAuthenticationMechanism implements HttpAuthenticationMechanis
     private StateController stateController;
 
     private static final Logger LOGGER = Logger.getLogger(OpenIdAuthenticationMechanism.class.getName());
-    
+
     private static final String SESSION_LOCK_NAME = OpenIdAuthenticationMechanism.class.getName();
 
     /**
@@ -183,41 +183,39 @@ public class OpenIdAuthenticationMechanism implements HttpAuthenticationMechanis
         this.configuration = configurationController.buildConfig(definition);
         return this;
     }
-    
+
     @Override
     public AuthenticationStatus validateRequest(
             HttpServletRequest request,
             HttpServletResponse response,
             HttpMessageContext httpContext) throws AuthenticationException {
 
-        synchronized (this.getSessionLock(request)) {
-            if (isNull(request.getUserPrincipal())) {
-                LOGGER.fine("UserPrincipal is not set, authenticate user using OpenId Connect protocol.");
-                // User is not authenticated
-                // Perform steps (1) to (6)
-                return this.authenticate(request, response, httpContext);
+        if (isNull(request.getUserPrincipal())) {
+            LOGGER.fine("UserPrincipal is not set, authenticate user using OpenId Connect protocol.");
+            // User is not authenticated
+            // Perform steps (1) to (6)
+            return this.authenticate(request, response, httpContext);
+        } else {
+            // User has been authenticated in request before
+
+            // Try-catch-block taken from AutoApplySessionInterceptor
+            // We cannot use @AutoApplySession, because validateRequest(...) must be called on every request
+            // to handle re-authentication (refreshing tokens)
+            // https://stackoverflow.com/questions/51678821/soteria-httpmessagecontext-setregistersession-not-working-as-expected/51819055
+            // https://github.com/javaee/security-soteria/blob/master/impl/src/main/java/org/glassfish/soteria/cdi/AutoApplySessionInterceptor.java
+            try {
+                httpContext.getHandler().handle(new Callback[]{
+                    new CallerPrincipalCallback(httpContext.getClientSubject(), request.getUserPrincipal())}
+                );
+            } catch (IOException | UnsupportedCallbackException ex) {
+                throw new AuthenticationException("Failed to register CallerPrincipalCallback.", ex);
+            }
+
+            if (configuration.isTokenAutoRefresh()) {
+                LOGGER.log(Level.FINE, "UserPrincipal is set, check if Access Token is valid.");
+                return this.reAuthenticate(request, response, httpContext);
             } else {
-                // User has been authenticated in request before
-
-                // Try-catch-block taken from AutoApplySessionInterceptor
-                // We cannot use @AutoApplySession, because validateRequest(...) must be called on every request
-                // to handle re-authentication (refreshing tokens)
-                // https://stackoverflow.com/questions/51678821/soteria-httpmessagecontext-setregistersession-not-working-as-expected/51819055
-                // https://github.com/javaee/security-soteria/blob/master/impl/src/main/java/org/glassfish/soteria/cdi/AutoApplySessionInterceptor.java
-                try {
-                    httpContext.getHandler().handle(new Callback[]{
-                        new CallerPrincipalCallback(httpContext.getClientSubject(), request.getUserPrincipal())}
-                    );
-                } catch (IOException | UnsupportedCallbackException ex) {
-                    throw new AuthenticationException("Failed to register CallerPrincipalCallback.", ex);
-                }
-
-                if (configuration.isTokenAutoRefresh()) {
-                    LOGGER.log(Level.FINE, "UserPrincipal is set, check if Access Token is valid.");
-                    return this.reAuthenticate(request, response, httpContext);
-                } else {
-                    return SUCCESS;
-                }
+                return SUCCESS;
             }
         }
     }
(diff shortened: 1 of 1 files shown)


appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
```diff
@@ -332,7 +333,10 @@ private AuthenticationStatus refreshTokens(HttpMessageContext httpContext, Refre
             updateContext(tokensObject);
             OpenIdCredential credential = new OpenIdCredential(tokensObject, httpContext, configuration);
             CredentialValidationResult validationResult = identityStoreHandler.validate(credential);
-            // Register session manually (if @AutoApplySession used, this would be done by its interceptor)
+            
+            // Dot not register session, as this will invalidate the currently active sessions (with all of its attributes)!
+            // httpContext.setRegisterSession(validationResult.getCallerPrincipal().getName(), validationResult.getCallerGroups());
+
             httpContext.setRegisterSession(validationResult.getCallerPrincipal().getName(), validationResult.getCallerGroups());
```
dmatej: Same line is duplicated and commented out ... was it tested? :-)

parysto: Ah, sorry. Copied the changes from my test-env without removing the line. Will fix it.

jGauravGupta:
jenkins test please

jGauravGupta:
Hi @parysto ,

Thanks for the PR . A import is missing in `OpenIdAuthenticationMechanism` class, please fix the compilation error:
````
project openid-client-integration: Compilation failure
/var/lib/jenkins/workspace/PayaraQuickBuildAndTest/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java:[371,9] cannot find symbol
  symbol:   class HttpSession
  location: class fish.payara.security.openid.OpenIdAuthenticationMechanism
````

And can you also share the sample application with multiple protected resources?

**Comment by parysto** [comment]
Hi @jGauravGupta ,

have pushed the sample application to: https://github.com/parysto/OpenIDJsfDemo
It's a JSF application that includes some JS (JQuery to increase size/load time) files. All resources, except callback.xhtml, (index.xhtml and *.js) are protected.
Please set providerURI, clientID and clientSecret in Configuration.java to your OpenID connect provider.

To trigger the race condition, set the validity time of the access token to a small value (like 1 min). Open the sample app, log in and wait until the token expired. Press F5 repeatedly, so that multiple requests are sent to Payara at the same time. Due to the race condition, you will be logged out and redirected to callback.xhtml. If you enabled logging for OpenIdAuthenticationMechanism, you will see log messages like:

> FEIN:   UserPrincipal is set, check if Access Token is valid.
> FEIN:   UserPrincipal is set, check if Access Token is valid.
> FEIN:   UserPrincipal is set, check if Access Token is valid.
> FEIN:   Access Token is expired. Request new Access Token with Refresh Token.
> AM FEINSTEN:   Sending the request to the userinfo endpoint
> FEIN:   UserPrincipal is not set, authenticate user using OpenId Connect protocol.
> AM FEINSTEN:   Redirecting for authentication to https://(...)

Will fix the missing imports ;-)

**Comment by parysto** [comment]
jenkins test please

**Comment by jGauravGupta** [comment]
jenkins test please

**Comment by parysto** [comment]
Hi @jGauravGupta ,

I will check if synchronizing `OpenIdAuthenticationMechanism#reAuthenticate` is sufficient.

**Comment by parysto** [comment]
Hi @jGauravGupta,

hadn't any issues synchronizing on `OpenIdAuthenticationMechanism#reAuthenticate` only :-)
Have updated the PR

**Comment by jGauravGupta** [comment]
Hi @parysto,

Thanks for updating the PR, Can you perform the locking operation only in case of token expired as mentioned above.

**Comment by parysto** [comment]
That's a good point (to improve performance). Will update the PR.

**Comment by jGauravGupta** [comment]
jenkins test please

**Comment by jGauravGupta** [comment]
Hi @parysto,

Please fix the folowing syntax error:
````
/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java:[327,5] missing return statement
````

**Comment by jGauravGupta** [comment]
jenkins test please

**Comment by parysto** [comment]
Hi @jGauravGupta,

is there something more I can do to complete this merge request?

File: appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java [hunk 2]
```diff
@@ -364,4 +368,20 @@ private void updateContext(JsonObject tokensObject) {
         }
     }
 
+    private Object getSessionLock(HttpServletRequest request) {
```
dmatej: Is really the plan to put the lock to the session object and leave it there forever? What if the token expires for second time? The lock will be still in session attributes from previous iteration.

Also replication in cluster will have some latency, is it or isn't it a problem? (probably it isn't)

I don't say it is incorrect, I am only asking ;-)

pzygielo: Interesting...
I know raw Object is not Serializable. I do not remember perfectly, but I do wonder what happens at HttpSession passivation...

parysto: Hi @dmatej, hi @pzygielo 
> Is really the plan to put the lock to the session object and leave it there forever? What if the token expires for second time? The lock will be still in session attributes from previous iteration.

You mean the lock-object? Yes, when the token expires again, the same instance of the lock will be reused. So, we do not need to set the "lock-attribute" again as long as the user's session is not invalidated. When the user's session is invalidated, the lock will be removed. Or am I wrong?

> Also replication in cluster will have some latency, is it or isn't it a problem? (probably it isn't)

Object itself does not have any fields that need to be transferred. So there shouldn't be a huge impact on latency. Although I haven't measured that ;-)

> I know raw Object is not Serializable. I do not remember perfectly, but I do wonder what happens at HttpSession passivation... Or if session migrates to other VM...

Good point, haven't thought about that. Yes, Object is not Serializable, so passivation or replication will fail. Will change the lock to an "empty" implementation of Serializable.

> I don't say it is incorrect, I am only asking ;-)

I appreciate your feedback :-)

dmatej:
Jenkins test please

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
dmatej: @pzygielo I think I investigated in years ago when I was fixing SSO session replication and I found that nonserializable objects are simply ignored if there is no special implementation for them.

pzygielo: @dmatej thanks for checking
>  I think I investigated in years ago when I was fixing SSO session replication and I found that nonserializable objects are simply ignored if there is no special implementation for them.

However it's not so clear for me, given section 7.7.2 of Servlet 4.0 spec, and:
> The distributed servlet container must throw an IllegalArgumentException for
objects where the container cannot support the mechanism necessary for migration
of the session storing them.

thus I'm not so sure about simply ignoring such session attributes. So it seems I have to investigate myself :wink:

pzygielo: Here is one implementation piece I found, that confirms your statement, about "how it is": https://github.com/payara/Payara/blob/bdb468a6bbde2f4f17b3a5b5de7e9a50309fc420/appserver/web/web-core/src/main/java/org/apache/catalina/session/StandardSession.java#L2172-L2178

jGauravGupta:
Jenkins test please

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
dmatej: "the container cannot support the mechanism necessary" ... such mechanism might be even ignoring objects which are managed by some mechanism on the other side.
But ... yes, I think there are some "blank spaces" in impl ... the problem is that implementing it would crash nearly all older+larger applications I have seen :D

dmatej:
Jenkins test please

PR #4421, opened 2020-01-10
CUSTCOM-150 Leaking Managed Connector Dependency

MattGill98:
See commit message for details.

I've added the optional flag to make none of the connectors include themselves transiently.


MattGill98:
jenkins test please

dmatej:
Optional -> force module to declare this dependency if it is required. So ... why not removing it if it is not required? Or is it some transitive dependency? Then it is on classpath even when using other profiles!!!

PR #4422, opened 2020-01-10
APPSERV-14 Slow SQL tracing views

jbee:
### Summary
Adds on server side:

* a general data annotation mechanism to data collection (see details below)
* SQL max execution time metric based on SQL tracing
* a watch for SQL max execution time with threshold from pool configuration
* extends watch conditions
* adds series tag name wildcards, ala `?:foo` (any tag name with value `foo` including no tag) to allow explicit selection of family of series where some have further tags and others have not like used for lists of health check alerts or a general alert list
* REST API for data requests supports better selection of the needed type of data (points, alerts, annotations, watches) to avoid sending data between client and server that isn't used by the client
* extends request tracing metric so it does allow adding useful watch
* adds a watch to cause alerts in case request tracing is above threshold
* adds "compaction" for alert frames to limit memory consumption and client-server data exchange

Adds on client side:

* shows thresholds of multiple watches in charts as stacked threshold indicator lines/areas
* fixes threshold indicator gradient and line alignment
* fixes alert condition formatting
* adds annotations to alerts with setting to hide them
* adds new widget type _Annotations_ that lists just the matching annotations in a list or table
* adds new widget setting _Mode_ to switch between _Table_ or _List_ layout for _Annotations_ widgets
* widgets now have a page wide unique id so that same series (pattern) can be used multiple times on the same page as done for SQL, request tracing or alerts.
* adds a new _SQL_ preset page that show slow SQL tracing information
* fixes alert table vertical overflow by using a scrollbar when needed
* adds list of tracing data and tracing metric with alerts to request tracing page
* adds a general _Alerts_ page preset that shows all alerts organised in groups

### Data Annotations
Annotations are a new concept in collection of metrics to attach key-value data to a particular metric at a particular point in time. Such annotation should only be done in situations that are in some way noteworthy such as a metric's value exceeding a particular threshold. In general annotations are event like information that are stored per series in a queue of fixed size. Newer annotations eventually replace older annotations to avoid number of annotations growing out of control.
Current limit is set at 20 per metric series.

The annotation data is used by the client to enhance views such as lists of alerts with more data for the period of the alert or to allow widgets of type `annotation` that for example can list the slow SQL meta data in table.

Annotations provide a general mechanism that makes such values available for any metric and from all instances. To verify the general nature of the concept it has also been applied to request tracing which allows to also list the trace span attributes in a table.

### Testing
New unit tests were added for 
* extended variants of `Condition` logic 
* extended `Series` patterns using the added `?` wild-card
* `Alert.Frame` compaction 

#### Testing Performed
Manual testing of various features. This PR does change general parts and extends others. It makes sense to test usual workflows even though they were not an explicit target of the changes.

#### Testing Instructions
General Setup:
5. check that following pages do exist: _SQL_ and _Alerts_ (if not most likely a browser cache issue - or get in touch)

Testing slow SQL tracing:
1. Open admin console at http://localhost:4848/
2. navigate to _Resources_ => _JDBC_ => _JDBC Connection Pools_
3. select _H2Pool_ tab _Advanced_
4. change _Slow Query Log Threshold_ from `-1` (disabled) to `0.001` (1ms)
5. open page _SQL_ in MC
6. deploy e.g. https://github.com/javaee/tutorial-examples/tree/master/persistence/order (some app using the default pool)
7. deployment should already cause alert(s) and annotations being shown on the _SQL_ page.
8. use the order app some and see how "slow" queries cause alerts and SQL is available in the table on _SQL_ page

NOTE: the `order` app causes errors during deploy on server restart. Undeploy the app and redeploy with server running. AFAIK this is a shortcoming of the app's setup/scripts.

Testing Request Tracing (again):
1. Open admin console at http://localhost:4848/
2. navigate to _Configurations_ => _server-config_ => _Request Tracing_
3. check _Enabled_
4. set _Target Count_ to `2`, _Time Value_ to `20`, _Time Unit_ to `SECONDS`
5. set _Threshold Value_ to somewhere between `5` to `20`, _Threshold Unit_ to `MILLISECONDS`
6. save changes
7. open MC _Request Tracing_ page. The polling of MC itself should be enough to cause some data to appear in the widgets

Testing Health Checks (again):
1. Open admin console at http://localhost:4848/
2. navigate to _Configurations_ => _server-config_ => _HealthCheck_
3. open tab _CPU Usage_ , check _Enabled_ and save changes
4. eventually open other tabs to enable more checks, the _Enabled_ checkbox on _General_ tab does **not** have to be checked.
5. open _Health Checks_ page in MC and check that the graphs for the enabled checks do show data and thresholds
6. go back to admin console configuration and change a threshold, e.g. heap usage to a level that is in between the low and high point currently present for your GC cycle (see graph).
7. check that changed threshold is updated in graph and that alerts are caused when heap usage is exceeded (note that the alert condition requires the threshold to be exceeded for average of last 15 points) 

Further things in MC to test (again)
* create a new page, name it, add some widgets, remove some widgets
* reset preset pages after changes
* change settings of widgets
* change colour scheme 
* clear local storage for the page and reload the page (full reset or client side settings)


jbee:
jenkins test please

jbee:
jenkins test please

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/ManagedConnectionFactoryImpl.java
```diff
@@ -1199,22 +1197,21 @@ public void setSqlTraceListeners(String sqlTraceListeners) {
...
-    
+
     public void setSlowQueryThresholdInSeconds(String seconds) {
         spec.setDetail(DataSourceSpec.SLOWSQLLOGTHRESHOLD, seconds);
         double threshold = Double.parseDouble(seconds);
         if (threshold > 0) {
             if (sqlTraceDelegator == null) {
                 sqlTraceDelegator = new SQLTraceDelegator(getPoolName(), getApplicationName(), getModuleName());
             }
-            sqlTraceDelegator.registerSQLTraceListener(new SlowSQLLogger((long)(threshold * 1000), TimeUnit.MILLISECONDS));
```
jbee: NB. This was not needed and even could cause outdated threshold to be used as listeners installed will not be replaced when a listener of same class is attempted to install again (which would happen later when a connection is created).

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/com/sun/gjc/spi/ManagedConnectionImpl.java
```diff
@@ -506,7 +511,7 @@ public Object getConnection(Subject sub, javax.resource.spi.ConnectionRequestInf
         
         if (sqlTraceDelegator == null) {
             if ((requestTracing != null && requestTracing.isRequestTracingEnabled())
-                    || (connectionPool != null && isSlowQueryLoggingEnabled())) {
+                    || (isSlowQueryLoggingEnabled())) {
```
jbee: NB. `isSlowQueryLoggingEnabled` includes the `null` check.

appserver/jdbc/jdbc-ra/jdbc-core/src/main/java/fish/payara/jdbc/SQLTraceStoreAdapter.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * An adapter between the {@link SQLTraceListener} abstraction that is registered with implementation class as key and a
+ * managed instance of the {@link SQLTraceStore}.
+ * 
+ * @author Jan Bernitt
+ */
+public class SQLTraceStoreAdapter implements SQLTraceListener {
+
+    private static ThreadLocal<SQLQuery> currentQuery = new ThreadLocal<>();
```
jbee: NB. Sadly this cannot share code with `SQLTraceLogger` which does almost the same as this has to be its own thread local otherwise the manipulation of the query happens multiple times if both listeners are installed which messes up the query.

appserver/jdbc/jdbc-runtime/src/main/java/fish/payara/jdbc/SQLTraceStoreImpl.java
```diff
@@ -0,0 +1,163 @@
...
+
+import fish.payara.monitoring.collect.MonitoringData;
+import fish.payara.monitoring.collect.MonitoringDataCollector;
+import fish.payara.monitoring.collect.MonitoringDataSource;
+import fish.payara.monitoring.collect.MonitoringWatchCollector;
+import fish.payara.monitoring.collect.MonitoringWatchSource;
+
+@Service
+@Singleton
+public class SQLTraceStoreImpl implements SQLTraceStore, MonitoringDataSource, MonitoringWatchSource {
```
jbee: NB. Most interesting part is the class location. It had to be in a module that supports HK2 stuff and which ideally already depends on modules dealing with SQL tracing.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryAlarmService.java
```diff
@@ -108,11 +108,14 @@
     public void init() {
         isDas = serverEnv.isDas();
         changedConfig(parseBoolean(serverConfig.getMonitoringService().getMonitoringEnabled()));
-        Watch watch = new Watch("Metric Collection Duration", new Metric(new Series("ns:monitoring CollectionDuration")))
-                .red(800L, 3, true, 800L, 5, false)
-                .amber(600L, 3, true, 600L, 5, false)
-                .green(-400L, 1, false, null, null, false);
-        addWatch(watch);
+        addWatch(new Watch("Metric Collection Duration", new Metric(new Series("ns:monitoring CollectionDuration")))
```
jbee: NB. The watch for a metric provided by `InMemoryMonitoringDataRepository` had to be added here as this would otherwise cause a cyclic dependency between the two.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryAlarmService.java
```diff
@@ -123,12 +126,12 @@ void changedConfig(boolean enabled) {
         if (!enabled) {
             checker.stop();
         } else {
-            checker.start(executor, 2, SECONDS, this::checkWatches);
+            checker.start(executor, 1, SECONDS, this::checkWatches);
```
jbee: NB. Initially I thought evaluating alerts every two seconds is good enough. While this is true in general this does allow to miss value spikes that should cause an alert in case of most basic condition that only looks at the latest value. So it was changed to 1 second to make sure each value is considered.

jbee:
jenkins test please

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/jdbc/SQLTraceStore.java
```diff
@@ -0,0 +1,48 @@
...
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+package org.glassfish.api.jdbc;
+
+import org.jvnet.hk2.annotations.Contract;
+
+@Contract
+public interface SQLTraceStore {
```
jbee: NB. Main reason this interface is needed is because of module dependencies.

nucleus/common/internal-api/src/main/java/fish/payara/monitoring/collect/MonitoringDataCollector.java
```diff
@@ -88,14 +109,26 @@ default MonitoringDataCollector prefix(CharSequence prefix) {
...
+                return this;
             }
 
             @Override
-            public MonitoringDataCollector collect(CharSequence key, long value) {
-                return self.collect(new StringBuilder(prefix).append(key), value);
+            public MonitoringDataCollector annotate(CharSequence metric, long value, String... attrs) {
+                prefixed.setLength(prefix.length());
+                self.annotate(prefixed.append(metric), value, attrs);
+                return this;
```
jbee: NB. It was a bug to not return `this` for any of the methods of `MonitoringDataCollector` because when the instance is used with chaining it should be the wrapper that is called (so it does its prefix thing), not the instance returned by `self` (that would not do the prefix thing). In practice this bug never had an effect as the only usage would not use chaining but for future this now works properly.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HeapMemoryUsageHealthCheck.java
```diff
@@ -102,7 +102,7 @@ private static MemoryUsage getMemoryUsage() {
 
     @Override
     public void collect(MonitoringWatchCollector collector) {
-        collectUsage(collector, "ns:health HeapUsage", "Heap Usage", 10, true);
```
jbee: NB. 10 turned out to be too flaky so I increased it.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/SeriesAnnotation.java
```diff
@@ -0,0 +1,99 @@
...
+ * {@link #instance} and pointing to a {@link #time} that is in the range of {@link SeriesDataset}.
+ * 
+ * The meta data attached is a list of key-value pairs.
+ * 
+ * @author Jan Bernitt
+ * @since 5.201
+ */
+public final class SeriesAnnotation implements Serializable {
+
+    public final long time;
```
Pandrex247: Personally I prefer Bean style: private fields with getters & setters.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/RequestTracingService.java
```diff
@@ -563,26 +574,70 @@ public RequestTraceStoreInterface getRequestTraceStore() {
     }
 
     @Override
-    public void collect(MonitoringDataCollector rootCollector) {
-        MonitoringDataCollector tracingCollector = rootCollector.in("trace");
+    public void collect(MonitoringWatchCollector collector) {
+        if ("true".equals(configuration.getEnabled())) {
```
Pandrex247: Reason there is for Yoda-style? <(°.°)>

appserver/monitoring-console/webapp/JS_DOCS.md
```diff
@@ -267,7 +316,7 @@ Mandatory members of `ENTRY` depend on `type` member. Variants are:
 'text'     : { label, value, onChange }
 'color'    : { label, value, defaultValue, onChange }
 ```
-* `onChange` may be ommitted for _text_ inputs which makes the field _readonly_.
+* `onChange` may be omitted for _text_ inputs which makes the field _readonly_.
```
Pandrex247: 👌

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/RequestTracingService.java
dmatej: It is nullsafe and despite thera are two groups of developers, I like this more :D

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/SeriesAnnotation.java
dmatej: So only with getter here, because it is final.

nucleus/payara-modules/requesttracing-core/src/main/java/fish/payara/nucleus/requesttracing/RequestTracingService.java
jbee: yes, its `null` safe, in contrast to `someVar.equals("stringConstant")`

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/model/SeriesAnnotation.java
jbee: I prefer less code and a getter for an immutable `final` field seems to me has no benefit and only costs but I can bring myself to add one anyway and call it a sacrifice for "consistency" and (other readers) "readability"  :D

jbee:
jenkins test please

jbee:
@Pandrex247 addressed your comments.

I also fixed a potential memory leak I noticed. When SQL traces would hit the new store but monitoring would be disabled the queues could grow without limit. I applied same technique I used for request tracing where the queue is limited to 50 entries. Should it be that size when new entry is added the oldest entry is thrown away.

PR #4423, opened 2020-01-10
CUSTCOM-13 A REST management DELETE command returns 415 code instead of 404

MattGill98:
This is a bug fix. Starting a fresh domain in 5.194 and running the following curl command:

~~~
curl -X DELETE -v -H "Content-Type: application/x-www-form-urlencoded" -H "X-Requested-By:Payara" http://localhost:4848/management/domain/servers/server/server/application-ref/non-existing-app-ref
~~~

will cause a HTTP 415 to return, where it should be a 404. Note that in 5.193 this worked.

# Testing

### New tests

Second commit. I've made the testing module fairly generic, as I see it as something that could be useful in the future.

### Testing Performed

The new test against 5.193 and the SNAPSHOT on micro, managed and remote profiles.

### Test suites executed
- Payara Samples (only single test)
- Java EE7 Samples (only JAX-RS tests against remote)
- Java EE8 Samples (only JAX-RS tests against remote)

# Notes for Reviewers
Read the commit messages ;)

I've deliberately ignored the previous fix that regressed. I hope this does the same job but with lower technical debt.

### DIFF
--- a/appserver/tests/payara-samples/samples/rest-management/src/test/java/fish/payara/samples/rest/management/HttpResponsesTest.java
+++ b/appserver/tests/payara-samples/samples/rest-management/src/test/java/fish/payara/samples/rest/management/HttpResponsesTest.java
@@ -1,0 +1,66 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) [2020] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://github.com/payara/Payara/blob/master/LICENSE.txt
+ * See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ * GPL Classpath Exception:
+ * The Payara Foundation designates this particular file as subject to the "Classpath"
+ * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ * file that accompanied this code.
+ *
+ * Modifications:
+ * If applicable, add the following below the License Header, with the fields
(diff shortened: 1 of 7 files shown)


MattGill98:
jenkins test please

Cousjava:
Copyright notices need updating.

MattGill98:
Good spot, I'm in 2019 mode still :wink:

MattGill98:
jenkins test please

nucleus/admin/rest/rest-service/src/main/java/org/glassfish/admin/rest/readers/FormReader.java
```diff
@@ -64,22 +65,22 @@
...
 @Provider
-public class FormReader implements MessageBodyReader<HashMap<String, String>> {
+public class FormReader implements MessageBodyReader<Map<String, String>> {
     
     private static final String DEFAULT_CHARSET = StandardCharsets.UTF_8.toString();
 
     @Override
     public boolean isReadable(Class<?> type, Type genericType, Annotation[] annotations, MediaType mediaType) {
-        return type.equals(HashMap.class);
+        return type.isAssignableFrom(HashMap.class);
```
jbee: Shouldn't it be `Map` now?

dmatej:
Tested also:
`mvn clean verify -pl :payara-samples -amd -Ppayara-samples -fae`
- two tests are failing, but we know about them, they will be fixed in another issue.

dmatej:
Jenkins test please

appserver/tests/payara-samples/samples/rest-management/src/test/java/fish/payara/samples/rest/management/HttpResponsesTest.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+@NotMicroCompatible
+public class HttpResponsesTest extends RestManagementTest {
+
+    /**
+     * Tests that when deleting a non-existent resource a 404 is returned instead of
+     * a 415. See CUSTCOM-13 for details.
+     */
+    @Test
+    public void when_DELETE_non_existent_resource_expect_404() {
```
dmatej: camel case would be better, but ok.

MattGill98: Noooo slaps on the wrist are owed for camel case in unit tests

PR #4426, opened 2020-01-11
CUSTCOM-200 Get PKs for every timer found in timers

### LINKED ISSUE #4356: CUSTCOM-200 Unused timer var in HazelcastTimerStore.cancelTimersByKey
# Description #
----------
I have an issue with understanding 
https://github.com/payara/Payara/blob/c4f9338fed0a627999e3285760e12420605de73d/appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java#L267-L288
especially usage of `timers` and `timer` loop variable in https://github.com/payara/Payara/blob/c4f9338fed0a627999e3285760e12420605de73d/appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java#L277-L282

## Expected Outcome
`timer` is used in loop body

## Current Outcome
`timers` is used to get timer by the same `primaryKey` for `timers.size()` times from `pkCache` (`Map`). So I see it as it's getting the same `hzTimer` over and over.
As it is then placed (possibly, if `hzTimer.getTimedObjectPk().equals(primaryKey)` is satisfied which I do not know yet) in `timersToCancel` (`Set`) - then I think 
https://github.com/payara/Payara/blob/c4f9338fed0a627999e3285760e12420605de73d/appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java#L285 removes one timer only.

If it's all as it should be I will be fine with no detailed explanation.

pzygielo:
# Description
This is a bug fix.
I took the liberty to interpret _Easy to Contribute_ this way.

Fixes #4356.

# Testing

### Test suites executed
Included in `mvn clean build` only.

### Testing Environment
GNU/Linux, OpenJDK Runtime Environment (build 1.8.0_232-b09), Maven 3.6.3

### DIFF
--- a/appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
+++ b/appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2016-2019 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016-2020 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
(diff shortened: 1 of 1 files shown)


smillidge:
jenkins test please

PR #4429, opened 2020-01-13
CUSTCOM-153 Validate XML resources

dmatej:
- validate resources from own modules to fail fast
- validate distributed resources to fail before install

# Description
This is a build enhancement

# Important Info

# Testing

### Testing Performed

Tested manually on each module by braking some XML syntax, running the build of a module and watching when the check is done and if it fails the build.

# Notes for Reviewers

See comments in diff

### DIFF
--- a/appserver/admin/gf_template/pom.xml
+++ b/appserver/admin/gf_template/pom.xml
@@ -40,10 +40,10 @@
     holder.
 
 -->
-<!-- Portions Copyright [2019] [Payara Foundation and/or its affiliates] -->
+<!-- Portions Copyright [2020] [Payara Foundation and/or its affiliates] -->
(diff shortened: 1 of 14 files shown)


appserver/admin/gf_template/pom.xml
```diff
@@ -65,6 +65,15 @@
 
     <build>
         <plugins>
+            <plugin>
```
dmatej: The plugin execution is configured in plugin management, but each execution is configured explicitly in a module where it is required.

appserver/admin/production_domain_template_web/pom.xml
```diff
@@ -135,26 +135,9 @@
...
                 <executions>
                     <execution>
-                        <goals>
-                            <goal>validate</goal>
-                        </goals>
-                        <phase>validate</phase>
+                        <id>validate-xml-files</id>
                     </execution>
                 </executions>
-                <configuration>
```
dmatej: Replaced by default execution configuration - which validates all xml files in given validation root directory, with several exceptions (templates which are not valid xml files, because contain placeholders not passing the validation or are empty).

dmatej:
Jenkins test please

dmatej:
Jenkins test please

PR #4430, opened 2020-01-13
CUSTCOM-76 WebappClassLoader Synchronisation Test

MattGill98:
# Description
Synchronisation issues in the WebappClassLoader have already been fixed,
but not tested properly. This commit adds a test for the synchronisation
issues that caused CUSTCOM-76.

# Testing

### Testing Performed
Ran the unit test with `-Dsurefire.rerunFailingTestsCount=500` before and after the changes in https://github.com/payara/Payara/pull/4378.

### DIFF
--- a/appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
+++ b/appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
@@ -1,0 +1,178 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) [2020] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://github.com/payara/Payara/blob/master/LICENSE.txt
+ * See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ * GPL Classpath Exception:
+ * The Payara Foundation designates this particular file as subject to the "Classpath"
+ * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ * file that accompanied this code.
+ *
+ * Modifications:
+ * If applicable, add the following below the License Header, with the fields
+ * enclosed by brackets [] replaced by your own identifying information:
+ * "Portions Copyright [year] [name of copyright owner]"
+ *
+ * Contributor(s):
+ * If you wish your version of this file to be governed by only the CDDL or
+ * only the GPL Version 2, indicate your decision by adding "[Contributor]
+ * elects to include this software in this distribution under the [CDDL or GPL
+ * Version 2] license."  If you don't indicate a single choice of license, a
+ * recipient has the option to distribute your version of this file under
+ * either the CDDL, the GPL Version 2 or to extend the choice of license to
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+package org.glassfish.web.loader;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+import org.apache.naming.resources.FileDirContext;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class WebappClassLoaderTest {
+
+    private static CyclicBarrier lock;
+    private static ExecutorService executor;
+    private static File junitJarFile;
+
+    @BeforeClass
+    public static void setup() throws URISyntaxException {
+        // Run 3 methods at the same time, and make the pool large enough to increase
+        // the chance of a race condition
+        lock = new CyclicBarrier(3);
+        executor = Executors.newFixedThreadPool(60);
+
+        // Fetch any JAR to use for classloading
+        junitJarFile = new File(Test.class.getProtectionDomain().getCodeSource().getLocation().toURI());
+    }
+
+    @AfterClass
+    public static void shutdown() {
+        if (executor != null) {
+            executor.shutdownNow();
+        }
+    }
+
+    @Test
+    public void check_findResourceInternalFromJars_thread_safety() throws Exception {
+        final ClassLoader classLoader = this.getClass().getClassLoader();
+        final WebappClassLoader webappClassLoader = new WebappClassLoader(classLoader, null);
+        webappClassLoader.start();
+        webappClassLoader.setResources(new FileDirContext());
+
+        CompletableFuture<Exception> result = new CompletableFuture<>();
+
+        // Create the tasks, and have them each run at the same time
+        // using the cyclic barrier
+        Runnable lookupTask = waitAndDo(lock, result, () -> lookup(classLoader, webappClassLoader));
+        Runnable addTask = waitAndDo(lock, result, () -> add(classLoader, webappClassLoader));
+        Runnable closeTask = waitAndDo(lock, result, () -> webappClassLoader.closeJARs(true));
+
+        try {
+            // Run the methods at the same time
+            for (int i = 0; i < 100; i++) {
+                executor.execute(addTask);
+                executor.execute(lookupTask);
+                executor.execute(closeTask);
+            }
+            // Check to see if any completed exceptionally
+            Exception ex = result.get(200, TimeUnit.MILLISECONDS);
+            if (ex != null) {
+                throw ex;
+            }
+        } catch (TimeoutException ex) {
+            // Success!
+        } finally {
+            webappClassLoader.close();
+        }
+    }
+
+    private void add(ClassLoader realClassLoader, WebappClassLoader webappClassLoader) throws IOException {
+        List<JarFile> jarFiles = findJarFiles(realClassLoader);
+
+        for (JarFile j : jarFiles) {
+            try {
+                webappClassLoader.addJar(junitJarFile.getName(), j, junitJarFile);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private void lookup(ClassLoader realClassLoader, WebappClassLoader webappClassLoader) throws Exception {
+        for (JarFile jarFile : findJarFiles(realClassLoader)) {
+            for (JarEntry entry : Collections.list(jarFile.entries())) {
+                webappClassLoader.findResource(entry.getName());
+                // System.out.println("Looked up " + resourceEntry);
+                Thread.sleep(0, 100);
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


MattGill98:
jenkins test please

dmatej:
Test will **pass** if all tasks complete without exception (-> future not completed -> TimeoutException -> swallowed -> pass)
Test will **pass** if all tasks end up in deadlock (-> future not completed -> TimeoutException -> swallowed ->pass)

- is this what you wanted?

Test will **fail** if any task complete with an exception (any) (-> future completed -> rethrowing the cause)

Nevermind, the **test passed** even with 600 threads in the pool, 100000 tasks (*3) and 30 second timeout and Thread.yield instead of sleep.

Then I played bit more with that ... see comments in code. I'm still not sure with the target of the test (maybe add javadoc with some human readable description of the purpose).

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        // Fetch any JAR to use for classloading
+        junitJarFile = new File(Test.class.getProtectionDomain().getCodeSource().getLocation().toURI());
+    }
+
+    @AfterClass
+    public static void shutdown() {
+        if (executor != null) {
+            executor.shutdownNow();
+        }
```
dmatej: As javadoc suggests, you should add following line to avoid interaction with other tests:
executor.awaitTermination(30, TimeUnit.SECONDS);

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+            Exception ex = result.get(200, TimeUnit.MILLISECONDS);
+            if (ex != null) {
+                throw ex;
+            }
+        } catch (TimeoutException ex) {
+            // Success!
+        } finally {
+            webappClassLoader.close();
+        }
+    }
```
dmatej: I would add this, but it depends on the purpose of the test - if deadlock was acceptable result, then this would only break the test:
```
        List<Runnable> notExecutedYet = executor.shutdownNow();
        assertTrue("some jobs never executed: " + notExecutedYet.size(), notExecutedYet.isEmpty());
```
It will fail if the executor pool is so small that it can execute some of tasks and some must wait in a queue. If tasks are not ending, the pool will not be capable to run more ...

So now we have 60 threads in pool for 300 tasks (3*100). Easy.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+            }
+        }
+    }
+
+    private void lookup(ClassLoader realClassLoader, WebappClassLoader webappClassLoader) throws Exception {
+        for (JarFile jarFile : findJarFiles(realClassLoader)) {
+            for (JarEntry entry : Collections.list(jarFile.entries())) {
+                webappClassLoader.findResource(entry.getName());
+                // System.out.println("Looked up " + resourceEntry);
+                Thread.sleep(0, 100);
```
dmatej: This causes slowing down of all lookup tasks. When I tried to run much more threads to cause deadlock, I replaced this with Thread.yield(). But it is not generally better, it depends on what we hunt.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        try {
+            // Run the methods at the same time
+            for (int i = 0; i < 100; i++) {
+                executor.execute(addTask);
+                executor.execute(lookupTask);
+                executor.execute(closeTask);
+            }
+            // Check to see if any completed exceptionally
+            Exception ex = result.get(200, TimeUnit.MILLISECONDS);
```
rdebusscher: This also mean that all tasks must be finished 200 ms after the lasts task is scheduled.

I tested this on my machine and the time between the end of the last task finished and the timeout was only about 35 ms.

So this should help 

            boolean stopped = executor.awaitTermination(500, TimeUnit.MILLISECONDS);
            if (!stopped) {
                fail("Tasks are not finished in a timely manner");
            }

But it revealed another issue, all the tasks are not finished when retrieving the Future.get() :  the lookup task never ends and seems blocked in a deadlock.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param lock   the lock to wait on before execution
+     * @param result where to store any encountered exceptions
+     * @param task   the task to run
+     * @return a new task
+     */
+    private static Runnable waitAndDo(final CyclicBarrier lock, final CompletableFuture<Exception> result,
+            final ExceptionalRunnable task) {
+        return () -> {
+            try {
+                lock.await();
```
dmatej: GOTO for cycle with executor.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static CyclicBarrier lock;
+    private static ExecutorService executor;
+    private static File junitJarFile;
+
+    @BeforeClass
+    public static void setup() throws URISyntaxException {
+        // Run 3 methods at the same time, and make the pool large enough to increase
+        // the chance of a race condition
+        lock = new CyclicBarrier(3);
```
rdebusscher: This will only synchronize the start of the first 3 tasks. The 4th tasks will start immediately. Comment is a bit misleading in that regard.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+        // Create the tasks, and have them each run at the same time
+        // using the cyclic barrier
+        Runnable lookupTask = waitAndDo(lock, result, () -> lookup(classLoader, webappClassLoader));
+        Runnable addTask = waitAndDo(lock, result, () -> add(classLoader, webappClassLoader));
+        Runnable closeTask = waitAndDo(lock, result, () -> webappClassLoader.closeJARs(true));
+
+        try {
+            // Run the methods at the same time
+            for (int i = 0; i < 100; i++) {
+                executor.execute(addTask);
```
dmatej: LABEL for cycle with executor: note that the pool uses a queue, so first three tasks will get their time and then all other (do I remember it well?), until the pool will be empty. Following requests will wait in the queue for free (finished) threads.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
dmatej: > Running the test on my system suggests that there is a deadlock.

I had to change values, but I can cause the deadlock too under some circumstances ...

dmatej: > This also mean that all tasks must be finished 200 ms after the lasts task is scheduled.

It does not, it waits only to first `result.complete(ex)` call.

rdebusscher: It seems not to be a deadlock, but the lookup tasks just takes to long by the use of

    Thread.sleep(0, 100);

So that lookup tasks are not finished.

By replacing it with

    Thread.yield();

Tests runs normal and all tasks are finished 500 ms after the last one is launched (with .awaitTermination() )

dmatej: With more threads it still does not finish ... I will try yet to use jcmd if it would see the deadlock or not, but even my cpu load dropped after a while far before timeout.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,178 @@
...
+                executor.execute(lookupTask);
+                executor.execute(closeTask);
+            }
+            // Check to see if any completed exceptionally
+            Exception ex = result.get(200, TimeUnit.MILLISECONDS);
+            if (ex != null) {
```
dmatej: If you would use completeExceptionally, you don't need this check.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,183 @@
...
+    private File junitJarFile;
+
+    @Before
+    public void setup() throws URISyntaxException {
+        // Run 3 methods at the same time, and make the pool large enough to increase
+        // the chance of a race condition
+        executor = Executors.newFixedThreadPool(60);
+
+        // Require a minimum number of executions before completing
+        latch = new CountDownLatch(EXECUTION_COUNT);
```
rdebusscher: This means we just wait fore the first 100 tasks of the 300 tasks to be finished.  With the current code (`Thread.sleep(0, 100);` in the lookup() method) this means only the 'add' and 'closeJar' tasks are finished. No 'lookup' task is finished by the time the test checks for an error.

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,183 @@
...
+
+        // Fetch any JAR to use for classloading
+        junitJarFile = new File(Test.class.getProtectionDomain().getCodeSource().getLocation().toURI());
+    }
+
+    @After
+    public void shutdown() throws InterruptedException {
+        if (executor != null) {
+            executor.shutdownNow();
+            assertTrue("Executor could not shutdown. This could mean there is a deadlock.",
```
rdebusscher: executor.awaitTermination termination always return true, also in the case the tasks isn't finished. In that case, the executor send an interrupted signal and InterruptedException occurs (verified with try catch in lookup method).

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,183 @@
...
+        try {
+            // Run the methods at the same time
+            for (int i = 0; i < EXECUTION_COUNT; i++) {
+                executor.execute(addTask);
+                executor.execute(lookupTask);
+                executor.execute(closeTask);
+            }
+
+            // Wait for tasks to execute
+            assertTrue("The tasks didn't finish in the allowed time.",
```
rdebusscher: It waits only for 100 tasks of the 300 submitted. Since 'add' and 'closeJar' are small/fast tasks, this will never be an issue (except in case of deadlock)

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+    }
+
+    private void lookup(WebappClassLoader webappClassLoader) throws Exception {
+        for (JarFile jarFile : findJarFiles()) {
+            for (JarEntry entry : Collections.list(jarFile.entries())) {
+                webappClassLoader.findResource(entry.getName());
+                // System.out.println("Looked up " + resourceEntry);
+                Thread.sleep(0, 100);
```
rdebusscher: This sleep means that only a 'few' lookups are performed (in my test, around 3000 in the 130 ms the test is running).

By replacing it with `Thread.yield()`, the number of lookup actions is increased to around 21000 which mean the chance of having a concurrency issue (which we like to detect with this test) is increased significantly.

MattGill98: `Thread.yield()` causes a huge amount of false positives on the test. I imagine this is because it reduces the total time spent on this task

appserver/web/war-util/src/test/java/org/glassfish/web/loader/WebappClassLoaderTest.java
MattGill98: Yep, this is there exclusively for deadlocks

MattGill98:
jenkins test please

PR #4384, opened 2019-12-13
APPSERV-18 Use application name provided by a user when a deployment descriptor does not define the name

MeroRai:
# Description
This is an improvement to use application name provided by a user if a deployment descriptor does not define the name. 

# Testing

### Testing Performed
Tested both the `deploy` and `redeploy` command to make sure the improvement works as intended. 

### Testing Environment
Zulu JDK 1.8_222 on Elementary OS 0.4.1 Loki with Maven 3.5.4>

# Documentation
TODO

# Notes for Reviewers

PR #4432, opened 2020-01-14
APPSERV-18 Test for testing get application name

MeroRai:
# Description
This PR adds unit tests for this PR: https://github.com/payara/Payara/pull/4384

### Testing Environment
Zulu JDK 1.8_222 on Elementary OS 0.4.1 Loki with Maven 3.5.4>

### DIFF
--- a/nucleus/common/internal-api/src/test/java/org/glassfish/internal/deployment/GetApplicationNameTest.java
+++ b/nucleus/common/internal-api/src/test/java/org/glassfish/internal/deployment/GetApplicationNameTest.java
@@ -1,0 +1,124 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) [2020] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


MarkWareham:
Without deeply diving into the code, I would suggest some tests when USER_PROVIDED_APP_NAME has not been provided (if that's possible)

MeroRai:
@jbee I am testing the behaviour of `genericHandler.getDefaultApplicationName` method. I have mocked the behaviour of its dependencies so it doesn't rely on them.

jbee:
@MeroRai  Missed that the stub extends the implementation class.

nucleus/common/internal-api/src/test/java/org/glassfish/internal/deployment/GetApplicationNameTest.java
```diff
@@ -0,0 +1,124 @@
...
+    @Test
+    public void testWhenDDDoesNotHaveAppName() {
+        when(applicationInfoProvider.getNameFor(readableArchive, null)).thenReturn(null);
+        assertEquals(USER_PROVIDED_APP_NAME, genericHandler.getDefaultApplicationName(readableArchive, null, USER_PROVIDED_APP_NAME));
+    }
+
+    @Test
+    public void testWhenDDAppNameIsSetToDefaultArchiveName() {
+        when(applicationInfoProvider.getNameFor(readableArchive, null)).thenReturn(DEFAULT_ARCHIVE_NAME);
+        when(readableArchive.getName()).thenReturn(DEFAULT_ARCHIVE_NAME);
```
jbee: I feel like a `assertSomething` helper method should be extended and used by each of the 3 test methods which should make them more readable in the the way: this input has that expected output. It should also become clear why only this text needs mockup of the `readableArchive.getName()`.

MeroRai:
jenkins test please

PR #4437, opened 2020-01-16
CUSTCOM-78 Fix Redeployment For Deployment Groups

MattGill98:
# Description
This is a bug fix.

When redeploying applications for deployment groups, a HK2 bean locking exception would be thrown as the server attempts to deploy the application to the deployment group as well as each of the instances within that deployment group. This change means that only the deployment group is targeted so the application can deploy normally.

The broken functionality was caused by 81f87239ac175e26affc4fc2fbe0c5338a618696, but the included fix does seem appropriate for the original issue. This change aims to complement
that commit.

# Testing

### New tests
N/A

### Testing Performed

Redeploying to deployment groups spread across multiple notes, as well as redeploying to individual instances within that deployment group.

### Test suites executed

N/A

### DIFF
--- a/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Domain.java
+++ b/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Domain.java
@@ -38,7 +38,7 @@
  * holder.
  */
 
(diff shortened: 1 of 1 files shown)


MattGill98:
jenkins test please

PR #4439, opened 2020-01-17
CUSTCOM-11 Update Jersey Version to 2.29.1.payara-p2

MattGill98:
# Description
Based on https://github.com/payara/Payara_PatchedProjects/pull/289.

When an endpoint with method POST/PUT is defined, and it is called with no payload, it results in an Exception.

~~~
javax.ws.rs.ProcessingException: Error deserializing object from entity stream.
        at org.glassfish.jersey.jsonb.internal.JsonBindingProvider.readFrom(JsonBindingProvider.java:77)
~~~~
However, according to the JAX-RS specification, an 'empty' object should be created:

----
4.2.4 Standard Entity providers

        When reading zero-length message entities all pre-packaged MessageBodyReader implementations, except the JAXB one and those for the (boxed) primitive types above, MUST create a corresponding Java object that represents zero-length data.
----

# Testing

### New tests

https://github.com/payara/patched-src-javaee7-samples/pull/50

### Test suites executed

- Payara Samples
- Java EE7 Samples
- Java EE8 Samples

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
         <servlet-api.version>4.0.2</servlet-api.version>
         <grizzly.version>2.4.3.payara-p10</grizzly.version>
         <jax-rs-api.impl.version>2.1.5</jax-rs-api.impl.version>
-        <jersey.version>2.29.1.payara-p1</jersey.version>
+        <jersey.version>2.29.1.payara-p2</jersey.version>
         <jakarta.validation.version>2.0.2</jakarta.validation.version>
         <hibernate.validator.version>6.0.16.Final</hibernate.validator.version>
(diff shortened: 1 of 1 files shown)


smillidge:
Has this been raised upstream?

dmatej:
This is already fixed in upstream Jersey 2.30, but I found that in all versions are still bugs related to primitive JSON requests - they are not reported, probably because nobody uses JSON for single Integer values. Source of these bugs is Yasson and are reproducible even on latest 1.0.6.

I found them when I tried to explicitly use JSON specification rules against several versions of Payara + Yasson + Jersey. Only good to know.

PR #4441, opened 2020-01-17
[FISH-1017] Payara can close JarFile instances used by current URLClassLoaders

### LINKED ISSUE #4414: Payara can close JarFile instances used by current URLClassLoaders
# Description #
----------

Several places in the Payara code call JarFileUtils.closeCachedJarFiles (Weld deployment, undeployment, etc) in order to clean up no longer needed open files.  The problem is that the code cannot determine why the JarFile is open.  Unfortunately in certain circumstances URLClassLoader can open JarFile objects that are stored in that cache and forcing the closure breaks the class loader for all further class loading attempts.

## Expected Outcome

Classloaders should continue working until the classloader itself is closed.

## Current Outcome

Under certain circumstances (URLClassLoaders with a URL NOT in the "file:" schema), the JarFile cleanup causes the class loader to break.

The circumstances we saw this was in using JRuby based servlet applications (Ruby on Rails) where it was using internal JRuby SSL code (in the form of the jopenssl.jar file embedded in the jruby-stdlibs jar).

## Environment ##

- **Payara Version**: 5.194
- **Edition**: Full
- **JDK Version**: Open JDK 8
- **Operating System**: RedHat Linux

bjetal:
# Description
Changes the closeCachedJarFiles behavior to only close JarFile instances associated with a "file:" URL.  All others are removed from the cache so they can be closed when no longer in use.  The assumption is that others might be owned by a URLClassLoader, but "file:" is most likely owned by the process requesting the close.  This should allow redeployment of the vast majority of applications on Windows without having to worry about file locks, while letting applications which need unusual Jar file URLs to work.

fixes: https://github.com/payara/Payara/issues/4414

# Testing

### New tests
None provided

### Testing Performed
Tested that it fixes the issue using the application which showed it.

### Test suites executed
None

### Testing Environment
OS: RedHat Linux 7
JDK: OpenJDK 8.0.201
Maven: 3.6.2

# Documentation
N/A

# Notes for Reviewers
N/A

### DIFF
--- a/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/JarFileUtils.java
+++ b/nucleus/admin/util/src/main/java/com/sun/enterprise/admin/util/JarFileUtils.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2015-2018 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015-2021 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
@@ -41,8 +41,11 @@ package com.sun.enterprise.admin.util;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.URL;
 import java.util.ConcurrentModificationException;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


rdebusscher:
@bjetal We are awaiting the signed CLA (https://github.com/payara/Payara/blob/master/PayaraCLA.pdf) otherwise we need to close this PR.

bjetal:
I apologize, I thought I had emailed it, but in looking at it I realized I had sent it to the wrong email address.  I have emailed it again, this time to the correct address.

AlanRoth:
jenkins test please

lprimak (changes requested):
please update copyright year, otherwise looks good to me

lprimak:
jenkins test

lprimak:
@bjetal please merge payara master into your PR and re-push it, your master is out-of-date and doesn't pass security checks anymore,
thanks1

lprimak:
Marking this as abandoned, but think it needs to be revived as internal, as it is useful

bjetal:
Sorry to be so slow getting back to this.  It has been updated to the latest master and should be ready to merge.

lprimak (approved):

lprimak:
jenkins test

bjetal:
@lprimak What do you mean by "jenkins test"?  If all you are looking for a unit test, I can probably do that, although it may be a few days until I can get to it.  If you are looking for a test that this "fixes" the overall bug, then I doubt if I will be able to find time to do that.

lprimak:
That just kicks off Payara internal test suite

PR #4442, opened 2020-01-19
fixes issue #4169 - Display node uptime in a human readable format

### LINKED ISSUE #4169: Development request (node uptime) /CUSTCOM-187
# Description #
----------

In node uptime the time is displayed in milliseconds, which is meaning nothing to nobody. It would be great to format node uptime to some human readably time.

obraz

bbulk:
# Description

Converted the node uptime display format from milliseconds to a human readable format.
There was already a handler for this purpose that could be used. Since a node can also be stopped and then has no uptime, the handler needed a null check.

fixes #4169 

# Testing

### Testing Performed
Created a new instance and opened its standaloneInstanceGeneral.jsf page.
Checked the result when the instance is running & stopped.

### Testing Environment
Zulu JDK 1.8_232
Windows 10
Maven 3.6.1


MeroRai:
Jenkins test please

PR #4452, opened 2020-01-23
APPSERV-19 Adds monitoring of stuck and hogging threads to monitoring console

jbee:
### Summary
Changes on server side:
* extends the annotation mechanism added in APPSERV-14 with the `keyed` flag (details below)
* fixes calculation of duration a thread is _working_ in stuck thread calculations
* changes stuck thread calculations to be based on millisecond timestamps and duration only (nanosecond precision makes no sense as this would also require running the check below 1ms intervals)
* adds collection of stuck threads metrics and annotations
* adds watch for stuck threads (based on threshold duration)
* cleanup and extension of hogging thread calculation
* lowers hogging thread retry minimum to `0` (no retry)
* adds collection of hogging threads metrics and annotations
* adds watch for hogging threads (based on count of hogging threads)

Changes on client side:
* adds a _Threads_ page preset that shows stuck and hogging thread incidents

### Keyed Annotations
Usually annotations are stored in a fixed size queue where newest annotation eventually overrides the oldest. When annotations are _keyed_ this behaviour is altered slightly. For a keyed annotation the first annotation attribute value is considered as a key. All annotations of the same key are automatically removed from the queue when an annotation is added. Only if the size limit is still exceeded the newest still replaces the oldest. This addition allows to effectively update annotation on the same _thing_ identified by the key without necessarily removing annotations on other _things_.
This allows annotations on many _things_ for the same metric series which avoids unnecessarily detailed metrics for each _thing_. 

In the context of this PR this means annotations on one thread (usually) do not replace annotations on another thread for the same metric, here stuck thread duration or hogging thread duration.

### Testing
New unit tests have been added for the general mechanics of keyed annotations.

Reading the documentation changes https://github.com/payara/Payara-Server-Documentation/pull/699 might help to understand the context of the feature.

The feature was tested manual according to below test instructions:

#### General Setup:

#### Testing Stuck Threads Health Checks
By using debug mode:
0. Start MC in **debug mode**
1. Open admin console at http://localhost:4848/
2. navigate to _Configurations_ => _server-config_ => _HealthCheck_
3. open tab _Stuck Threads_, check _Enabled_, set a Threshold of 1 second and save
4. put a breakpoint in a method you know is called (e.g. `fish.payara.monitoring.web.MonitoringConsoleResource.getSeriesData(SeriesRequest)`)
5. after some seconds at the break-point continue execution. 
6. check in MC _Threads_ page that annotation(s) occurs in widget _Stuck Thread Incidents_
7. check in MC _Alerts_ and _Health Checks_ page that an related alert exists

Alternatively one could deploy an app with a known REST API method that does take several seconds to complete. 

#### Testing Hogging Threads Health Checks
With a testing app:
1. Open admin console at http://localhost:4848/
2. navigate to _Configurations_ => _server-config_ => _HealthCheck_
3. open tab _Hogging Threads_, check _Enabled_, set a Threshold of 50% and a _Retry Count_ of 0-1 and save
4. deploy a test application (see JIRA) with a method that does a busy loop for a given length of time
5. invoke the method for some seconds
6. check in MC _Threads_ page that widget _Hogging Thread Incidents_ contains entries, check that _Method_ refers to your method with the loop
7. check in MC _Alerts_ and _Heath Checks_ page that an alert exists

By restarting (there is a chance):
1. Open admin console at http://localhost:4848/
2. navigate to _Configurations_ => _server-config_ => _HealthCheck_
3. open tab _Hogging Threads_, check _Enabled_, set a Threshold of **1%** and a _Retry Count_ of 0-1 and save
4. restart the server, and wait until server is up again
5. check in MC _Threads_ page that widget _Hogging Thread Incidents_ contains entries
6. check in MC _Alerts_ and _Heath Checks_ page that an alert exists


jbee:
jenkins test please

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HoggingThreadsHealthCheck.java
```diff
@@ -91,65 +142,107 @@ public String getDescription() {
...
+            if (record == null) {
+                record = new ThreadCpuTimeRecord();
+                recordsById.put(threadId, record);
             } else {
-                times.setStartCpuTime(times.getEndCpuTime());
-                times.setStartUserTime(times.getEndUserTime());
-                times.setEndCpuTime(c);
-                times.setEndUserTime(u);
-
-                long checkTime = getOptions().getUnit().toMillis(getOptions().getTime());
```
jbee: NB. using the interval from the options is only an approximation of the actual time passed in the measurement interval. As this now can be different intervals for the check run by the health check service and the monitoring data collection I changed this to use the actual time passed. This is also more accurate.

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/StuckThreadsHealthCheck.java
```diff
@@ -59,58 +69,118 @@
...
+        return action + info.getLockInfo().toString();
+    }
+
+    private void acceptStuckThreads(StuckThreadConsumer consumer) {
+        ThreadMXBean bean = ManagementFactory.getThreadMXBean();
+        long thresholdInMillis = getThresholdInMillis();
+        long now = System.currentTimeMillis();
         ConcurrentHashMap<Long, Long> threads = stuckThreadsStore.getThreads();
-        for (Long thread : threads.keySet()){
-            Long timeHeld = threads.get(thread);
```
jbee: NB. As discussed in chat the `timeHeld` here was a semantic confusion. The map contained the timestamp when the thread started the work. I changed the algorithm accordingly and also changed all the computation to be based on milliseconds as nanosecond level only would make sense if we would run the check often and fast enough (every t with a t < 1ms)

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HoggingThreadsHealthCheck.java
```diff
@@ -91,65 +142,107 @@ public String getDescription() {
...
+                            "Thread", thread, //
+                            "Usage%", String.valueOf(percentage), //
+                            "Threshold%", String.valueOf(threshold), //
+                            "Method", initialMethod, //
+                            "Exited", String.valueOf(!initialMethod.equals(getMethod(info))));
+                    hoggingThreadCount.incrementAndGet();
+                    hoggingThreadMaxDuration.updateAndGet(value -> Math.max(value, totalTimeHogging));
+                });
+        collector
+        .collect("HoggingThreadCount", hoggingThreadCount)
```
Pandrex247:
```suggestion
                .collect("HoggingThreadCount", hoggingThreadCount)
```
Indenting

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/HoggingThreadsHealthCheck.java
```diff
@@ -91,65 +142,107 @@ public String getDescription() {
...
+        .collect("HoggingThreadDuration", hoggingThreadMaxDuration);
```
Pandrex247:
```suggestion
                .collect("HoggingThreadDuration", hoggingThreadMaxDuration);
```
Indenting 2: The tab vs. space wars

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/StuckThreadsHealthCheck.java
Pandrex247: I think this possibly needs a bit more cleanup, as you can still configure it with a threshold of 5 nanoseconds.

Particularly in the monitoring console you get a funny situation where you get a Threshold listed as `0` in comparison to something like `5ms`. In this case, you could possibly change it to have it say the threshold is `<1ms`

appserver/monitoring-console/webapp/src/main/webapp/js/mc-model.js
```diff
@@ -262,6 +262,24 @@ MonitoringConsole.Model = (function() {
...
+					numberOfColumns: 4,
+					widgets: [
+						{ series: 'ns:health StuckThreadDuration', type: 'annotation', mode: 'table', unit: 'ms',
+							displayName: 'Stuck Thread Incidents',
+							grid: {column: 0, item: 1, colspan: 3, rowspan: 1},
+							fields: ["Thread", "Started", "Value", "Threshold", "Suspended", "Locked", "State"]},
+						{ series: 'ns:health HoggingThreadDuration', type: 'annotation', mode: 'table', unit: 'ms',
+							displayName: 'Hogging Thread Incidents',
+							grid: {column: 0, item: 2, colspan: 3, rowspan: 1},
+							fields: ["Thread", "When", "Value", "Usage%", "Threshold%", "Method", "Exited"]},
```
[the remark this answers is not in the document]
jbee: yes, that was intentional.

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/StuckThreadsHealthCheck.java
jbee: Agreed, I think this should possibly be handled by a separate PR.

jbee:
jenkins test please

nucleus/payara-modules/healthcheck-stuck/src/main/java/fish/payara/nucleus/healthcheck/stuck/StuckThreadsHealthCheck.java
Pandrex247: 👍 I'll create a separate Jira

Issue #4103, opened 2019-07-24
Persistent EJB timers not restored after instance restart


g-manfredi:
# Description #
----------

When an application with persistent EJB timers is deployed on a Payara Micro instance inside a cluster, restarting the instance will cause the timers to be lost. The timers are found on startup (as the logs state), but they are not restored.

This affects  Payara Micro 5.x at least since 5.181, but not e.g. 4.1.2.181.

I digged a bit into the source code of Payara (payara-server-5.183). I tried avoiding the return statement in fish.payara.ejb.timer.hazelcast.HazelcastTimerStore.restoreTimers() so that the _restoreTimers() method is called. However the restoration is skipped because the EJB container IDs of the bean don't match.

## Expected Outcome

When restarting a clustered Payara Micro instance, persistent EJB timers are restored.

https://docs.payara.fish/documentation/payara-micro/persistent-ejb-timers.html

## Current Outcome

The timers are found on startup (as the logs state), but they are not restored.

## Steps to reproduce (Only for bug reports) 

To reproduce, you can use this sample project: https://github.com/g-manfredi/persistent-timers

## Context (Optional)

Originally thought to be related to running Payara in Docker an raised here: https://github.com/payara/docker-payaramicro/issues/48

## Environment ##

- **Payara Version**: 5.x at least since 5.181
- **Edition**: Micro
- **JDK Version**: 8.0_202 - Oracle
- **Operating System**: Windows 10

smillidge:
Are you specifying an instance name with the ``--name`` command line option?

g-manfredi:
Yes, although through Docker: https://github.com/g-manfredi/persistent-timers/blob/master/services/sample-service/docker/Dockerfile

krystianity:
Any updates on this issue?

BirolYildiz:
I'm able to reproduce this issues as well on Kubernetes and the latest Micro Release 5.193.

smillidge:
Is this only when run in a Docker container or also when run normally via the command line?

krystianity:
We do not really know if this also occurs without docker, however I dont see why this should only be a docker issue.

g-manfredi:
It is not specific to Docker, please see @OndroMih 's comment on the original post: https://github.com/payara/docker-payaramicro/issues/48#issuecomment-514562347

smillidge:
As @OndroMih says this is reproducible I have moved this to Engineering under PAYARA-4258

krystianity:
[...]

PR #4457, opened 2020-01-27
CUSTCOM-24 Restore Persistent EJB Timers on Payara Micro Restart

MattGill98:
# Description
This is a bug fix. 

When stopping and starting a Payara Micro instance with a static name (`--name`) and an EJB timer running, the timer will be stored to be resumed, but won't actually run. This is caused by an inconsistent EJB container ID, which in turn is caused by an inconsistent application unique ID. This re-implements old functionality to remove the CDI event that fires to reconfigure the application unique ID from the clustered store.

Fixes https://github.com/payara/Payara/issues/4103

# Testing

### Testing Performed

Reproducer in linked issue (just the application outside of docker).

To repeat: run two Payara Micro instances with:

~~~
java -jar payara-micro-5.201-SNAPSHOT.jar --name micro1 --autobindhttp --deploy sample-service.war
~~~

~~~
java -jar payara-micro-5.201-SNAPSHOT.jar --name micro2 --autobindhttp --deploy sample-service.war
~~~

Trigger the timer on one of the instances:

~~~
curl -X GET localhost:8080/timers/start-timer
~~~

Restart the instance with the timer running, and the timer should start again.

### Test suites executed

- Java EE7 Samples

# Notes for Reviewers

The commits reference several commits that introduced the bug, but they should be taken into account in the review.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -1049,12 +1049,14 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
             postBootCommands.executeCommands(gf.getCommandRunner());
             this.runtime = new PayaraMicroRuntimeImpl(gf, gfruntime);
 
-            // load all applications, but do not start them until Hazelcast gets a chance to initialize
-            deployAll();
+            // Enable hazelcast
             if (!noCluster) {
                 gf.getCommandRunner().run("set-hazelcast-configuration", "--enabled", "true", "--dynamic", "true", "--target", "server-config", "--hostawarepartitioning", Boolean.toString(hostAware), "--lite", Boolean.toString(liteMember));
             }
 
+            // deploy all applications and then initialize them
+            deployAll();
+            // These steps are separated in case any steps need to be done in between
             gf.getCommandRunner().run("initialize-all-applications");
 
             postDeployCommands.executeCommands(gf.getCommandRunner());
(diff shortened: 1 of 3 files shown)


appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1043,7 +1043,6 @@ public PayaraMicroRuntime bootStrap() throws BootstrapException {
 
             // boot the server
             preBootCommands.executeCommands(gf.getCommandRunner());
-            HazelcastCore.setThreadLocalDisabled(true);
             try {
```
pdudits: If this solves the issue and we agree, that this is how hazelcast bootstrap should work (pre-boot, rather than post boot), then entire code for handling "threadlocaldisabled" should be removed as well, as it has no other usage.

MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1043,19 +1043,14 @@ public PayaraMicroRuntime bootStrap() throws BootstrapException {
...
-                // load all applications, but do not start them until Hazelcast gets a chance to initialize
-                deployAll();
-            } finally {
-                HazelcastCore.setThreadLocalDisabled(false);
-            }
+            // load all applications, but do not start them until Hazelcast gets a chance to initialize
+            deployAll();
             if (!noCluster) {
                 gf.getCommandRunner().run("set-hazelcast-configuration", "--enabled", "true", "--dynamic", "true", "--target", "server-config", "--hostawarepartitioning", Boolean.toString(hostAware), "--lite", Boolean.toString(liteMember));
             }
```
pdudits: If all the reproducers we mentioned work with this configuration, would you let try swapping these two statements, finalizing config before deploy?

MattGill98: I've swapped them and my reproducer seems to work fine, so I agree this is a good change. I'm going to commit it, I just need to test the other reproducers before I merge this

MattGill98:
jenkins test please

PR #4461, opened 2020-01-28
APPSERV-42 Add Option to Enable Data Grid Encryption from Admin Console

Pandrex247:
# Description
Adds a button specifically to the domain data grid configuration page
that enables or disables encryption, plus associated help text.

# Testing
Manually tested by enabling encryption, restarting, and then disabling again.


Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Pandrex247:
You do yes, I mentioned as such in the help text.

> Whether or not to encrypt the data grid. Requires an encryption key generated by the generate-encryption-key command. Requires restart of all instances.

I didn't do it from here because a) generation requires the domain to not be running and b) requires the master password, not the admin password.

Pandrex247:
What _should_ have happened in that case Jan is that Hazelcast should have failed to start. The instance still starts and you can navigate the DAS etc. but Hazelcast services won't be available.

Pandrex247:
I could change it to be a warning, but that is arguably also misleading because the command _did_ work.

Pandrex247:
APPSERV-54 raised to track printing a warning if data grid encryption is enabled without a key already being generated.

PR #4463, opened 2020-01-30
APPSERV-47 Adds Custom Watches to Monitoring Console

jbee:
### Summary
Adds custom watches to client. In contrast to watches collected from `MonitoringWatchSource`s a custom watch is a user defined `Watch` instance. The monitoring client is used to compose such a watch. This can be done in the new page _Watches_ that also lists all watches and allows to edit, delete or disable and enable them.

A disabled watch is simply not actively evaluating data, it is ignored but it will show in the graphs as a outline box instead of a filled one. All watches can be disabled and enabled. This is stored in the added configuration so that the disabled state can be recreated on server startup.

Only custom watches can be delete. Collected (or programmatic) watches cannot be deleted as they originate from the server's code.

Custom watches can be edited. Programmatic watches can be duplicated, which changes the name of the duplicate to _Copy of {Name}_. So this becomes another watch but allows to use the programmatic watch as a template.

To recreate the custom watches on server-startup the watches need to be stored in some form.
I decided to write custom to and from JSON methods for the affected classes. This will allow to put in the proper code that can handle changes to the data structure. The strategy here is to be able to handle missing fields (that would be added) and ignore fields (that would be removed).
The conversion is covered with unit tests.

On server side:
* adds monitoring console configuration to store disabled watches as well as custom watch data
* REST and service API to enable/disable watches
* distinction between custom watches and programmatic watches
* watch states are shared when deriving a watch
* REST and service API to create custom watches
* REST and service API to delete custom watches
* `WatchData` REST API class is not writable to receive custom watch data send by client 

On client side:
* adds a _Watches_ configuration page that lists existing watches, allows to edit, delete or disable/enable them.
* extracts model constants to a separate file for better readability and clarity
* extracts a controller file that encapsulation the client-server communication on client side for better readability and clarity
* fixes/updates formatting of watch conditions (e.g. in alert tables) and watch lists
* uses payara design system colors with better contrast
* fixes alignment of indicator (text message above graph)
* uses configured colours for status messages
* now also triggers status messages from alerts (red=critical, amber=alerting)
* adds new colour defaults for _error_ and _missing_ status
* use of `const` instead of `let` (TLDR; `const` should be default, `let` only when variable is actually changed, `var` is basically never needed or better suited in our "non-legacy" codebase)
* watch indicator at right side of the graph now shows the indicator as outline if the watch is disabled
* `Setting` input with `unit` now can use a function to provide a changing unit.

### `WriteableView` and `ProtectedList`
Tried to improve it to make it more usable. Code also could do with some cleanup.
* added generics
* added missing `@Override` annoations
* removed unnecessary `else` nesting
* made static helpers `static`
* added some `List` methods that would otherwise throw a `UnsupportedOperationException` on the basis of existing method (so no actual transaction management was needed)

As mentioned some of the existing methods of the `ProtectedList` still show some odd behaviour.
When using `set(int, E)` the value at the index would be null next time I loaded the list.

### Tests
Unit tests were added to verify the correctness of the to/from JSON conversion of the `Watch` class and all classes involved.

General information can be taken from the documentation https://github.com/payara/Payara-Server-Documentation/pull/705.

The custom watches feature was tested manually following the steps below:

#### General Setup:
5. check that following pages do exist: _Watches_ (if not most likely a browser cache issue - or get in touch)

#### Testing Disabling Existing Watches
1. Open page _Watches_ in MC
2. click checkbox of one of the listed watches to disable it, e.g. _Metric Collection Duration_
3. for _Metric Collection Duration_ watch open page _Monitoring_ and check the watch indicator on right side of the graph is now drawn as outline instead of filled
4. restart the server, e.g. using asadmin commands
5. refresh page _Watches_ in MC and check disabled watch is still disabled
6. enable the watch again
7. for _Metric Collection Duration_ watch open page _Monitoring_ again and check the indicator is now drawn filled

#### Testing Disabling Custom Watches
1. Open page _Watches_ in MC
2. create a new watch named _Heap Check_ by checking _Unhealthy_, filling in: _If_ `ns:jvm HeapUsage` _in_ `Percent` _is_ `>` _(some value clearly below your current heap usage, see _Core_ page)_, and press _Save or Update_
3. goto _Core_ page and check that the graph shows an alert because heap usage is above the watch level
4. go back to _Watches_ page and disable the created watch by clicking the checkbox on the left or using the cogs menu
5. go back to _Core_ and check that alert is gone (and that indicator is now as outline)
6. go to _Alerts_ page and check the alert is shown in _Past Unacknowledged Alerts_
7. go to _Watches_ page again an enable the watch once more, check on _Core_ page that an alert is triggered again

#### Testing Custom Watches
1. create a custom watch as given in steps for _Testing Disabling Custom Watches_
2. restart the server
3. open _Watches_ page and check the custom watch created is still there

#### Testing Listing of collected Watches
1. open _Watches_ page of MC, with default settings two watches should be listed
2. open admin console and browse to _Configurations_ => _server-config_ => HealthCheck_ tab _CPU Usage_ and check _Enabled_
3. go back to _Watches_ page and check a new watch _CPU Usage_ is listed now

#### Testing new admin command parameters are hidden
run:
```
asadmin> set-monitoring-console-configuration --help
NAME
     set-monitoring-console-configuration

SYNOPSIS
     Usage: set-monitoring-console-configuration [--enabled=true|
     false] 

OPTIONS
     --enabled

Command set-monitoring-console-configuration executed successfully.
```

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2019 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019-2020 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
@@ -40,10 +40,14 @@
 
 package fish.payara.monitoring.admin;
 
+import static java.util.Collections.singletonList;
 import static org.glassfish.config.support.CommandTarget.DAS;
 
+import java.beans.PropertyVetoException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
+import java.util.logging.Logger;
 
 import javax.inject.Inject;
 
@@ -53,6 +57,8 @@ import org.glassfish.api.admin.AdminCommand;
 import org.glassfish.api.admin.AdminCommandContext;
 import org.glassfish.api.admin.CommandRunner;
 import org.glassfish.api.admin.ExecuteOn;
+import org.glassfish.api.admin.RestEndpoint;
+import org.glassfish.api.admin.RestEndpoints;
 import org.glassfish.api.admin.RuntimeType;
 import org.glassfish.config.support.TargetType;
 import org.glassfish.deployment.autodeploy.AutoDeployer;
@@ -61,21 +67,40 @@ import org.glassfish.deployment.autodeploy.AutoUndeploymentOperation;
 import org.glassfish.hk2.api.PerLookup;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.jvnet.hk2.annotations.Service;
+import org.jvnet.hk2.config.ConfigSupport;
+import org.jvnet.hk2.config.SingleConfigCode;
+import org.jvnet.hk2.config.TransactionFailure;
 
 import com.sun.enterprise.config.serverbeans.Domain;
 import com.sun.enterprise.util.SystemPropertyConstants;
+
+import fish.payara.monitoring.configuration.MonitoringConsoleConfiguration;
 
 @Service(name = "set-monitoring-console-configuration")
 @PerLookup
 @ExecuteOn({RuntimeType.DAS})
 @TargetType({DAS})
+@RestEndpoints({
+    @RestEndpoint(configBean = Domain.class,
+            opType = RestEndpoint.OpType.POST,
+            path = "set-monitoring-console-configuration",
+            description = "Set Monitoring Console Configuration")
+})
 public class SetMonitoringConsoleConfigurationCommand implements AdminCommand {
+
+    private static final Logger LOGGER = Logger.getLogger(SetMonitoringConsoleConfigurationCommand.class.getName());
 
     private static final String MONITORING_CONSOLE_APP_NAME = "__monitoringconsole";
     private final static String GLASSFISH_LIB_INSTALL_APPLICATIONS = "glassfish/lib/install/applications";
 
     @Param(optional = true)
     private Boolean enabled;
+
+    @Param(optional = true, alias = "disable-watch")
+    private String disableWatch;
+
+    @Param(optional = true, alias = "enable-watch")
+    private String enableWatch;
 
     @Inject
     protected CommandRunner commandRunner;
@@ -88,12 +113,44 @@ public class SetMonitoringConsoleConfigurationCommand implements AdminCommand {
 
     @Override
     public void execute(AdminCommandContext context) {
-        if (enabled != null) {
+        MonitoringConsoleConfiguration config = domain.getExtensionByType(MonitoringConsoleConfiguration.class);
+        if (config == null) {
+            context.getActionReport().failure(LOGGER, "Monitoring Console configuration does not exist.");
+            return;
+        }
+        if (enabled != null && Boolean.parseBoolean(config.getEnabled()) != enabled.booleanValue()) {
             if (enabled.booleanValue()) {
                 deployMonitoringConsole(context.getActionReport());
             } else {
                 undeployMonitoringConsole(context.getActionReport());
             }
+        }
+        try {
+            ConfigSupport.apply(new SingleConfigCode<MonitoringConsoleConfiguration>(){
+                @Override
+                public Object run(MonitoringConsoleConfiguration configProxy) throws PropertyVetoException {
+                    if (enabled != null) {
+                        configProxy.setEnabled(enabled.toString());
+                    }
+                    if (disableWatch != null && !disableWatch.isEmpty()) {
+                        List<String> disabledWatchNames = configProxy.getDisabledWatchNames();
+                        if (disabledWatchNames == null) {
+                            configProxy.setDisabledWatchNames(singletonList(disableWatch));
+                        } else if (!disabledWatchNames.contains(disableWatch)) {
+                           disabledWatchNames.add(disableWatch);
+                        }
+                    }
+                    if (enableWatch != null && !enableWatch.isEmpty()) {
+                        List<String> disabledWatchNames = configProxy.getDisabledWatchNames();
+                        if (disabledWatchNames != null) {
+                            disabledWatchNames.remove(enableWatch);
+                        }
+                    }
+                    return null;
+                }
+            }, config);
+        } catch (TransactionFailure ex) {
+            context.getActionReport().failure(LOGGER, "Failed to update Monitoring Console configuration", ex);
         }
     }
 
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
@@ -40,7 +40,6 @@
 
 package fish.payara.monitoring.admin;
 
-import static java.util.Collections.singletonList;
 import static org.glassfish.config.support.CommandTarget.DAS;
 
 import java.beans.PropertyVetoException;
@@ -102,6 +101,15 @@ public class SetMonitoringConsoleConfigurationCommand implements AdminCommand {
     @Param(optional = true, alias = "enable-watch")
     private String enableWatch;
 
+    @Param(optional = true, alias = "add-watch-name")
+    private String addWatchName;
+
+    @Param(optional = true, alias = "add-watch-json")
+    private String addWatchJson;
+
+    @Param(optional = true, alias = "remove-watch")
+    private String removeWatch;
+
     @Inject
     protected CommandRunner commandRunner;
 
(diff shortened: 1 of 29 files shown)


appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
```diff
@@ -61,22 +66,50 @@
...
+
+    @Param(optional = true, alias = "add-watch-name")
+    private String addWatchName;
+
+    @Param(optional = true, alias = "add-watch-json")
+    private String addWatchJson;
+
+    @Param(optional = true, alias = "remove-watch")
+    private String removeWatch;
+
```
jbee: NB. I opted for a parameter for each action. This could also be done by adding a general parameter `watch` for the name and `data` for the JSON and another parameter `action` or so. But I found this easier to use. WDYT?

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/WriteableView.java
```diff
@@ -317,16 +318,16 @@ public synchronized boolean canCommit(Transaction t) throws TransactionFailure {
         return currentTx==t;
     }
 
-    private void handleValidationException(Set constraintViolations) throws ConstraintViolationException {
+    private void handleValidationException(Set<? extends ConstraintViolation<?>> constraintViolations) throws ConstraintViolationException {
 
         if (constraintViolations != null && !constraintViolations.isEmpty()) {
-            Iterator<ConstraintViolation<ConfigBeanProxy>> it = constraintViolations.iterator();
```
jbee: NB. Filling in generics the `ConfigBeanProxy` has shown to be not correct. While this might be the actual type the generics along the path to this point do not actually provide this guarantee.

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/WriteableView.java
```diff
@@ -394,43 +396,8 @@ else if ( s.startsWith("<") ) {
...
-                            }
-                        }
-                        else {
-                              throw new IllegalArgumentException();
-                        }
-                    }
-                    appliedChanges.add(event);
-                }
+            for (ProtectedList<?> entry :  changedCollections.values())  {
+                commitListChanges(entry, appliedChanges);
```
jbee: NB. The loop body got extracted to the `commitListChanges` method because I needed a bindable type variable to use generics.

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/WriteableView.java
```diff
@@ -472,32 +479,37 @@ public synchronized void abort(Transaction t) {
             throw new TransactionFailure("Not part of a transaction", null);
         }
         ConfigBean newBean = bean.allocate(type);
-        WriteableView writeableView = bean.getHabitat().<ConfigSupport>getService(ConfigSupport.class).getWriteableView(newBean.getProxy(type), newBean);
+        bean.getHabitat().<ConfigSupport>getService(ConfigSupport.class);
+        WriteableView writeableView = ConfigSupport.getWriteableView(newBean.getProxy(type), newBean);
```
jbee: NB. Changed to static access or the static method `getWriteableView`.

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/WriteableView.java
```diff
@@ -597,12 +610,28 @@ public int size() {
...
+
+    @Override
+    public int lastIndexOf(Object object) {
+        return proxied.lastIndexOf(object);
+    }
+
+    @Override
+    public String toString() {
+        return proxied.toString();
+    }
```
jbee: NB. These actually work now :man_dancing:

nucleus/hk2/hk2-config/src/main/java/org/jvnet/hk2/config/WriteableView.java
```diff
@@ -736,11 +762,31 @@ public synchronized boolean remove(Object object) {
...
+        for (E e : proxied) {
+            if (filter.test(e) && remove(e)) {
+                removedAny = true;
+            }
+        }
+        return removedAny;
+    }
+
+    @Override
+    public E set(int index, E object) {
```
jbee: NB. While the above two now work I think `set` is not behaving properly. I looked into that but decided to not mess with it and avoid using it.

jbee:
jenkins test please

jbee:
jenkins test please

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-units.js
```diff
@@ -115,6 +115,21 @@ MonitoringConsole.View.Units = (function() {
       percent: PERCENT_FACTORS,
    };
 
+   const UNIT_NAMES = {
+      count: 'Count', 
+      ms: 'Milliseconds', 
+      ns: 'Nanoseconds', 
+      bytes: 'Bytes',
```
Pandrex247: Is bytes necessarily the best unit vs. kilo or mega?
An concrete example in our monitoring doesn't jump immediately to mind but just inciting the thought.

appserver/monitoring-console/webapp/src/main/webapp/css/monitoring-console.css
```diff
@@ -37,16 +37,31 @@
    only if the new code is made subject to such option by the copyright
    holder.
 */
+:root {
+  /* Color RGB values to be used in other variable definitions */
```
Pandrex247: `Color`
(╯°□°）╯︵ ┻━┻

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/alert/Circumstance.java
```diff
@@ -137,4 +141,30 @@ public String toString() {
         }
         return str.toString();
     }
+
+    public JsonValue toJSON() {
```
Pandrex247: I know this class already exists, but a description of what it's used for wouldn't go amiss - circumstance isn't an intuitive name to me in the context of alerts (in comparison to something like _event_).

appserver/monitoring-console/webapp/src/main/webapp/js/mc-view-units.js
jbee: Its more a dimension or base unit. KB or MB are multiples of this base unit. Which is expressed further up in the table of conversions. Also metrics themselves are actually given in bytes not kilo or mega. For the user this has no significance as values are automatically shown in the appropriate unit(s). For example 5KB or 10MB300Kb.  Actual metrics such as used memory are in bytes and are only converted to something more fitting in the UI.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/configuration/MonitoringConsoleConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+import org.jvnet.hk2.config.Element;
+
+import com.sun.enterprise.config.serverbeans.DomainExtension;
+
+/**
+ * Configuration for the monitoring console core.
+ * This is first of all the data and watch collection and evaluation.
+ * 
+ * @author Jan Bernitt
+ * @since 5.204
```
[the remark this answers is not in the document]
jbee: Right, was again confused.

jbee:
> You have it mentioned in your testing guidelines but I don't think a server restart is required for custom watches to take affect from my own testing - don't know if there's some documentation you need to edit around that.

That was referring to a change made via the asadmin command. If you use the monitoring console webapp the change takes effect immediately.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/alert/Alert.java
```diff
@@ -84,6 +84,10 @@
         public boolean isLessSevereThan(Level other) {
             return ordinal() > other.ordinal();
         }
+
+        public static fish.payara.monitoring.alert.Alert.Level parse(String level) {
```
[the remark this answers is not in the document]
jbee: nope. good catch.

jbee:
jenkins test please

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/alert/Circumstance.java
jbee: Added javadoc.

appserver/monitoring-console/webapp/src/main/webapp/css/monitoring-console.css
jbee: Changed to `Colour` ;)

jbee:
@Pandrex247 Addressed your comments. 

Changed the input+dropdown to a single input field where any number without unit is understood as count while any number with unit is understood as time.

Removed the icon in the list. Also never liked it but wanted to have some more clear indicator. Now the checkbox indicates the _disabled_ state. Should a watch be _stopped_ (which usually not occurs) it is also greyed out and the title will say `(stopped)` in addition to the name to make that clear as it could be checked but not effective when stopped.

Parameter for command are now hidden ones (one underscore).

Other small comments were addressed as well.

Pandrex247:
> Changed the input+dropdown to a single input field where any number without unit is understood as count while any number with unit is understood as time.

Just checking that I understand this correctly, but what exactly is meant by `Count`? Is it "checks" / "updates"? So dependent on how frequently you have the data refresh polling?

jbee:
It is the number of points considered. Those are always the most recent n points available. So it has some connection to the frequency of data collection (not the client polling) on the server which is once per second. For example a count of 5 means take the last 5 points and check if the condition holds for them whereas 5sec would mean take any number of points in the last 5 seconds (usually that are 6) compare those. This can make a bigger difference since not all metrics do collect once per second as the `@MonitoringData` annotation can be used to increase this to any number of seconds.

PR #4464, opened 2020-01-30
APPSERV-30 Exception when Adding an Instance to a Deployment Group with an Application Deployed

MeroRai:
# Description
When you create and add a new instance to a deployment group with an application deployed to it, an error is thrown and the application is not deployed

# Important Info
**Steps to reproduce:**
- Start a clean domain
- Create a deployment group with no instances
- Deploy an application to the deployment group
- Create a new instance and add it to the deployment group
- Start the instance and look at the logs

### Testing Performed
Followed the above reproducer to make sure the fix worked as intended.

### Test suites executed
Build Test

### Testing Environment

### DIFF
--- a/nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/AddInstanceToDeploymentGroupCommand.java
+++ b/nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/AddInstanceToDeploymentGroupCommand.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2017 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) [2017-2020] Payara Foundation and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/AddInstanceToDeploymentGroupCommand.java
```diff
@@ -164,4 +172,25 @@ public void execute(AdminCommandContext context) {
...
+
+        StringJoiner virtualServers = new StringJoiner(",");
+        if (config != null) {
+            HttpService httpService = config.getHttpService();
+            if (httpService != null) {
+                List<VirtualServer> hosts = httpService.getVirtualServer();
+                if (hosts != null) {
+                    for (VirtualServer host : hosts) {
+                        if (("__asadmin").equals(host.getId())) {
+                            continue;
```
jbee: It might be personal preference but I think use of `continue` should be avoided if possible. Here it is easy to avoid by negating the condition and moving the add into the block.

MeroRai:
jenkins test please

PR #4475, opened 2020-02-07
Addresses sonar warnings of recent code changes

jbee:
### Summary
Had a look at new sonar warning I recently caused by code changes and addressed them (hopefully).

Mostly this is 

* using constants, 
* extracting methods to reduce complexit
* use a more appropriate method for a test (`indexOf` vs. `includes`, etcetera)
* use lambdas with _expression body_ syntax where possible

### Testing 
Manually tested that (since there where non-automatic changes applied)

* monitoring data collection from non-DAS instances still works.
* SQL tracing in MC still works
* the MC JS generally does not complain about some error

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
@@ -95,18 +95,23 @@ public class SetMonitoringConsoleConfigurationCommand implements AdminCommand {
     @Param(optional = true)
     private Boolean enabled;
 
+    @SuppressWarnings("squid:S116")
     @Param(optional = true, alias = "disable-watch")
     private String _disableWatch;
 
+    @SuppressWarnings("squid:S116")
     @Param(optional = true, alias = "enable-watch")
     private String _enableWatch;
 
+    @SuppressWarnings("squid:S116")
     @Param(optional = true, alias = "add-watch-name")
     private String _addWatchName;
 
+    @SuppressWarnings("squid:S116")
     @Param(optional = true, alias = "add-watch-json")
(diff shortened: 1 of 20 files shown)


appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/store/InMemoryMonitoringDataRepository.java
```diff
@@ -426,7 +433,7 @@ private static boolean isRelevantSet(SeriesDataset set, Set<String> instanceFilt
         String[] series;
         long[] values;
         // annotations
-        List<SeriesAnnotation> annotations;
+        ArrayList<SeriesAnnotation> annotations;
```
jbee: NB. because of serialisation the type here must be a type known to be `Serializable`

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
```diff
@@ -95,18 +95,23 @@
     @Param(optional = true)
     private Boolean enabled;
 
+    @SuppressWarnings("squid:S116")
```
jbee: NB. This is so sonar will no more complain about the `_` in the name which makes those hidden parameters.

appserver/monitoring-console/webapp/src/main/java/fish/payara/monitoring/web/ApiRequests.java
```diff
@@ -50,6 +50,7 @@
  * 
  * @author Jan Bernitt
  */
+@SuppressWarnings("squid:S1104")
```
jbee: NB. This stops sonar from complaining about non private fields that are not final which is needed for JSON IO.

jbee:
jenkins test please

Issue #4281, opened 2019-10-22
Cyclic dependency between SSO classes causes deadlock


sgflt:
# Description #
----------
Issue may relate to https://github.com/payara/Payara/issues/2881 and https://github.com/payara/Payara/issues/4280

During debugging why SSO never expires I have found a path, that can cause denial of service.

## Expected Outcome
No DoS

## Current Outcome
```
"ContainerBackgroundProcessor[StandardEngine[glassfish-web].StandardHost[server].StandardContext[]]" #248 daemon prio=5 os_prio=0 tid=0x00007f4945cd6000 nid=0x18dd waiting for monitor entry [0x00007f497cf16000]
   java.lang.Thread.State: BLOCKED (on object monitor)
at org.apache.catalina.authenticator.SingleSignOn.lookup(SingleSignOn.java:616)
- waiting to lock <0x00000000f4d78408> (a java.util.HashMap)
at org.glassfish.web.ha.authenticator.HASingleSignOn.removeSession(HASingleSignOn.java:186)
at com.sun.enterprise.security.web.GlassFishSingleSignOn.sessionEvent(GlassFishSingleSignOn.java:333)
at org.apache.catalina.session.StandardSession.fireSessionEvent(StandardSession.java:2367)
at org.apache.catalina.session.StandardSession.expire(StandardSession.java:992)
- locked <0x00000000f7611558> (a org.glassfish.web.ha.session.management.FullHASession)
at org.apache.catalina.session.StandardSession.expire(StandardSession.java:862)
at org.apache.catalina.session.StandardSession.isValid(StandardSession.java:779)
at org.apache.catalina.session.StandardSession.getAttributeNames(StandardSession.java:1392)
at org.glassfish.web.ha.session.management.BaseHASession.superToString(BaseHASession.java:239)
at org.glassfish.web.ha.session.management.BaseHASession.toString(BaseHASession.java:256)
at org.apache.catalina.authenticator.SingleSignOnEntry.removeSession(SingleSignOnEntry.java:124)
- locked <0x00000000f76113e0> (a org.glassfish.web.ha.authenticator.HASingleSignOnEntry)
at org.glassfish.web.ha.authenticator.HASingleSignOnEntry.removeSession(HASingleSignOnEntry.java:140)
- locked <0x00000000f76113e0> (a org.glassfish.web.ha.authenticator.HASingleSignOnEntry)
[...]

PR #4478, opened 2020-02-09
CUSTCOM-260 Fixed possible deadlock in StandardSession

sgflt:
# Description
This is a bug fix of deadlock described in #4281 

I could not create simple reproducer because race conditions are hard to reproduce. 
EDIT dmatej: the possible deadlock is well visible from sources.

# Notes for Reviewers
Reviewer should proove that the cycle between classes existed in previous version and now is fixed correctly.

@dmatej


rdebusscher:
@sgflt  We are awaiting the signed CLA (https://github.com/payara/Payara/blob/master/PayaraCLA.pdf) otherwise we need to close this PR.

sgflt:
CLA on the way :)

smillidge:
We've got the CLA

dmatej:
Jenkins test please

PR #4480, opened 2020-02-10
CUSTCOM-135 Fixed race condition on ConfigProviderResolver initialization

dmatej:
- instane() is synchronized
- setInstance is NOT synchronized, so it ignores any locks
- preferring @Inject over ConfigProvider in HealthCheckService
- using @ContractsProvided to allow both @Inject target types
- more logging, but only when configured
- tested in extremely slowed TestContainers docker container

# Important Info

Removed the timeout and countdouwn latch, because 

- it is more simple to diagnose deadlock without timeout
- deadlock should never happen - if it happened, something is broken again => timeout is not a way to get out of it
- this PR fixed the cause of timeout
- this PR still allows inverted (incorrect) order of initialization.

# Testing

### New tests
Not pushed reproducer in PAYARA-4176 branch (not production quality, but stacktraces in logs visible)

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
@@ -57,6 +57,7 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 import javax.inject.Named;
 import fish.payara.nucleus.microprofile.config.converters.CharacterConverter;
@@ -139,10 +140,17 @@ public class ConfigProviderResolverImpl extends ConfigProviderResolver {
     private Config serverLevelConfig;
 
     /**
-     * Sets the global {@link ConfigProviderResolver#instance()} to this instance.
+     * Logs constructor as finest - may be useful to watch sequence of operations.
      */
     public ConfigProviderResolverImpl() {
         LOG.finest("ConfigProviderResolverImpl()");
+    }
+
+    /**
+     * Sets the global {@link ConfigProviderResolver#instance()} to this instance.
+     */
+    @PostConstruct
+    public void postConstruct() {
         // the setInstance is not synchronized, but instance() method body is.
         // this will block possible concurrent access.
         synchronized (ConfigProviderResolver.class) {
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) [2017-2019] Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) [2017-2020] Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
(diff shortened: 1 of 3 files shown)


nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverSync.java
```diff
@@ -96,4 +95,17 @@ public void registerConfig(Config config, ClassLoader classLoader) {
...
+
+    private ConfigProviderResolver await() {
+        LOG.log(Level.WARNING, "Payara Microprofile Config requested too early, the HK2 service is not initialized yet."
+            + " Waiting until it will be active.");
+        while (true) {
+            final ConfigProviderResolver resolver = instance();
+            if (resolver != null && resolver != this) {
+                return resolver;
+            }
+            Thread.yield();
```
rdebusscher: .yield() is a hint, which means that control isn't necessarily  passed on to another Thread.  A sleep with very small amount should be used.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
```diff
@@ -129,31 +126,34 @@
...
+    /**
+     * Sets the global {@link ConfigProviderResolver#instance()} to this instance.
+     */
     public ConfigProviderResolverImpl() {
+        LOG.finest("ConfigProviderResolverImpl()");
+        // the setInstance is not synchronized, but instance() method body is.
+        // this will block possible concurrent access.
+        synchronized (ConfigProviderResolver.class) {
+            LOG.log(Level.CONFIG, "Setting global ConfigProviderResolver instance to {0}", this);
+            ConfigProviderResolver.setInstance(this);
```
rdebusscher: This is executed too early if you ask me. The current instance is set as value BEFORE the instance is completely setup as HK2 service. So MP config calls can get routed to this class before HK2 is finished processing, resulting in errors (NPE since the @Inject isn't processed?)

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverSync.java
dmatej: I used it on Linux+Intel/AMD and on Solaris/SPARC, even old versions, and it always worked. Also it is only few blocks after startup, so it does not slow down the Payara Startup. Sleep is unnecessary overkill, because I don't need to sleep running thread, I only tell the scheduler "at this moment I have nothing to do, but if other threads have some work, they can".
https://www.geeksforgeeks.org/java-concurrency-yield-sleep-and-join-methods/

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
dmatej: I agree, I was a bit too agressive and this is an antipattern (providing this before finishing constructor). But ... if we don't have other possibility, isn't the @PostConstruct method same situation?
From what I tested, @Inject will not cause NPE here, because HK2 will block any injection until it finishes PostConstruct method (constructor already finished). Bigger problem is programmatic usage of Globals class.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverSync.java
pdudits: `instance()` also has synchronized block, so there's another yield point there, so I believe this might be correct.

But still, I trust `CountDownLatch` more, but might be overkill for 1 producer to 1 consumer scenario.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
pdudits: Semantics of postconstruct guarantees that fields are set. The situation we're dealing with here is that someone triggered mp config API before this service was initialised. Postconstruct is therefore almost proper time as it is considered initialized after the method exits

Injections of this class via globals is indeed wrong

dmatej:
Jenkins test please

PR #4486, opened 2020-02-12
CUSTCOM-135: Move config initialization before Startup run level

pdudits:
Many services have implicit dependency on Microprofile config. We therefore need
to assure that it is already available as services enter startup run level.

Expansion of domain.xml attributes should also refer to our implementation rather
than to MP Config API, as that might lock up in certain scenarios, like single-threaded
startup.

# Description
This is a bug fix. 

During development of other feature, it became apparent that during load of applications on Payara Micro the startup is single threaded. Therefore no amount of wait would resolve implicit dependency to Microprofile config provider, it simply needs to be started earlier.

Another potential danger point was evalutation of `${MPCONFIG=key}`  while our provider didn't yet start.

# Important Info

# Testing

### Testing Performed

Prepared Payara Micro with an deployed application, then restarted it. Before, in 100% of attempts it resulted in 

```
[WARNING] [] [fish.payara.nucleus.microprofile.config.spi.ConfigProviderResolverImpl] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1581503313800] [levelValue: 900] [[  
  Timeout out waiting for Microprofile Config startup
java.lang.IllegalStateException: Payara Microprofile Config needs running server environment to work. Either it's not running, or you're experiencing a race condition
```

The condition is not happening anymore.

### Test suites executed
- MP Config TCK payara-micro-managed

### Testing Environment

OpenJ9 1.8.0_232 on Windows 10, Maven 3.6.3

### DIFF
--- a/nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java
+++ b/nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java
@@ -37,11 +37,12 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2020] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.config.support;
 
 import com.sun.enterprise.security.store.DomainScopedPasswordAliasStore;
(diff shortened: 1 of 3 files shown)


pdudits:
Jenkins test please

pdudits:
Jenkins test please

PR #4443, opened 2020-01-20
CUSTCOM-70 Fixed jersey-client issues with embedded distros of Payara

dmatej:
# Description

# Testing

```
mvn clean install -TC4 -PQuickBuild,fastest
mvn clean install -PBuildExtras,QuickBuild,fastest -pl :payara-embedded-all
mvn clean install -Ppayara-samples -pl :embedded-vs-jersey
```

### Test suites executed

- Quicklook Y
- Payara Samples Y
- Java EE7 Samples N
- Java EE8 Samples N
- Payara Private Tests N
- Payara Microprofile TCKs Runner N
- Jakarta TCKs - jaxrs
- Mojarra N
- Cargo Tracker N

PR #4491, opened 2020-02-13
CUSTCOM-70 Additional Fix to JAXRS Client With Embedded Dependency

MattGill98:
# Description
This is a bug fix.

When Payara Embedded is used as a library rather than starting the embedded server, an exception will be thrown when using the JAX-RS client caused by the request tracing service not having initialised.

The previous CUSTCOM-70 fix allowed a JAX-RS client request to be built, but executing the request would still throw the same exception. This change allows the request to be executed as well.

# Important Info

This PR is an extension to https://github.com/payara/Payara/pull/4443, which didn't address the initial problem. The test has been amended to match.

# Testing

### New tests

Payara Sample test amendment.

### DIFF
--- a/appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/PayaraTracingServices.java
+++ b/appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/PayaraTracingServices.java
@@ -40,111 +40,125 @@
 
 package fish.payara.requesttracing.jaxrs.client;
 
-import fish.payara.nucleus.requesttracing.RequestTracingService;
-import fish.payara.opentracing.OpenTracingService;
-
-import io.opentracing.Tracer;
-
 import org.glassfish.api.invocation.InvocationManager;
+import org.glassfish.hk2.api.ServiceHandle;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.glassfish.internal.api.Globals;
+import org.glassfish.internal.deployment.Deployment;
+
+import fish.payara.nucleus.requesttracing.RequestTracingService;
+import fish.payara.opentracing.OpenTracingService;
+import io.opentracing.Tracer;
 
 /**
- * This is a class hiding internal mechanism of lookup of HK2 services.
- * The lookup is lazy, done with first request, but this may be simply changed later.
- * <p>
- * The lazy lookup prevents problems with embedded distributions, when jersey lookups
- * may detect filters in this package and try to use them before Payara started, which
- * is not supported use case.
+ * This is a class hiding internal mechanism of lookup of HK2 services. The
+ * required services will be eagerly initialised using their service handles,
+ * and throwing an exception if the handle is available but not the service
+ * itself.
  *
  * @author David Matejcek
  */
 public final class PayaraTracingServices {
 
-    private static volatile boolean initialized;
+    private final RequestTracingService requestTracingService;
+    private final OpenTracingService openTracingService;
 
-    private static ServiceLocator basicServiceLocator;
-    private static RequestTracingService requestTracingService;
-    private static OpenTracingService openTracingService;
-
-
-    private static void checkInitialized() {
-        if (initialized) {
-            return;
-        }
-        synchronized (PayaraTracingServices.class) {
-            if (initialized) {
-                return;
-            }
-            basicServiceLocator = Globals.getStaticBaseServiceLocator();
-            requestTracingService = basicServiceLocator.getService(RequestTracingService.class);
-            openTracingService = basicServiceLocator.getService(OpenTracingService.class);
-            initialized = true;
-        }
-    }
-
+    private final InvocationManager invocationManager;
+    private final Deployment deployment;
 
     /**
-     * @return default service locator, same as {@link Globals#getStaticBaseServiceLocator()}.
+     * Initialise the tracing services if they are available.
+     * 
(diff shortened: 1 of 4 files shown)


MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/tests/payara-samples/classpath/embedded-vs-jersey/src/test/java/fish/payara/samples/classpath/embeddedvsjersey/JerseyClasspathTest.java
```diff
@@ -54,6 +54,6 @@
     @Test
     public void jerseyShouldInitialize() {
         Client client = ClientBuilder.newClient();
-        client.target("http://localhost:8080").request().buildGet();
+        client.target("http://localhost:8080").request().get();
```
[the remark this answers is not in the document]
MattGill98: I think the response is irrelevant. The test still seems to work if you just fill the URL with a random string

appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/PayaraTracingServices.java
```diff
@@ -40,111 +40,117 @@
...
     }
 
-
     /**
      * @return {@link InvocationManager}
      */
     public InvocationManager getInvocationManager() {
-        checkInitialized();
-        return basicServiceLocator.getService(InvocationManager.class);
+        if (checkInitialized()) {
```
dmatej: Simply return the field, you don't need the check here

appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/PayaraTracingServices.java
```diff
@@ -40,111 +40,117 @@
...
-
     /**
-     * @return actually active {@link Tracer} for the current application.
+     * @return actually active {@link Tracer} for the current application, or null
+     *         if the tracing service is not available.
      */
     public Tracer getActiveTracer() {
         final String applicationName = getApplicationName();
-        if (applicationName == null) {
+        if (applicationName == null || !checkInitialized()) {
```
dmatej: No need to call the check - it will be null if applicationName is null and applicationName is null if check ...

appserver/payara-appserver-modules/jaxrs-client-tracing/src/main/java/fish/payara/requesttracing/jaxrs/client/PayaraTracingServices.java
```diff
@@ -40,111 +40,117 @@
...
-        if (otService == null) {
-            return null;
+        return openTracingService.getTracer(applicationName);
+    }
+
+    /**
+     * @return true if the Request Tracing services are available and have been
+     *         initialised, or false if the services are not available.
+     */
+    private boolean checkInitialized() {
```
dmatej: rename to isTracingAvailable and make it public.

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/JerseyOpenTracingAutoDiscoverable.java
```diff
@@ -58,7 +57,7 @@
     @Override
     public void configure(FeatureContext context) {
         // Only register for application deployments (not the admin console)
-        final ServiceLocator serviceLocator = new PayaraTracingServices().getBasicServiceLocator();
+        final ServiceLocator serviceLocator = Globals.getStaticBaseServiceLocator();
```
dmatej: This is wrong - autodiscoverable is the trap causing problems, Jersey will find it even without Payara runing and try to use it. So safe should be:
final PayaraTracingServices services = new PayaraTracingServices();
```
Deployment deployment = services.getDeployment(); // as other services
if (deployment == null || deployment.getCurrentDeploymentContext() == null) {
  return;
}
...
```
note: in fact the deployment will be a singleton instance of ApplicationLifecycle, says my Eclipse ...

This is about architecture and responsibilities - PayaraTracingServices is a border class, where on one side is Jersey and on the other is HK2. The border covers if it will provide those services or not, and Jersey then has no other dependency on HK2 services. This class has control and responsibility over it.

appserver/payara-appserver-modules/microprofile/opentracing/src/main/java/fish/payara/microprofile/opentracing/jaxrs/client/JaxrsClientBuilderDecorator.java
```diff
@@ -165,7 +164,7 @@ public Client build() {
 
     private boolean requestTracingPresent() {
         try {
-            ServiceLocator locator = PAYARA_SERVICES.getBasicServiceLocator();
+            ServiceLocator locator = Globals.getStaticBaseServiceLocator();
```
dmatej: This will start HK2 base service locator initialization. Don't use it. I hate Globals API and sooner or later I will delete that class, because I must always open it and look on it's content :-D
Better use new PayaraTracingServices().isTracingAvailable(). That should be safe (if Globals.getDefaultBaseServiceLocator() OR same getDefaultHabitat() returns null -> all services are null -> false)

dmatej:
Jenkins test please

PR #4493, opened 2020-02-14
APPSERV-36 Extracts Monitoring Console to its own Repository

jbee:
### Summary
Three modules got extracted:

* **MC API**: Contains interfaces that make data and watch collection possible. It is implemented by a couple of classes (services) in the server; the modules of those are directly dependent on the API module
* **MC Implementation**: Contains the processing of data and watch collection and alert evaluation. A "adapter" module in the server has this as runtime dependency where the implementation is connected via API using Java service loader.
* **MC Webapp**: Contains the web application. As this is still deployed from commands the server packaging includes this as a build dependency.

There is one module left in the server that does the bootstrapping of the monitoring console by using its API. This module is not included in e.g. micro when monitoring should not be included.

So in summary a server that does not include monitoring console does only have to include the MC API module. A server that does include it includes all 3 MC modules and the bootstrapping module in the server. From the point of the server those 3 MC modules are external dependencies like any other library is. As such they are referenced by non snapshot versions which are configured via 3 new maven properties in the main `pom.xml`.

Corresponding PR in new repository is https://github.com/payara/monitoring-console/pull/1

### Notes for the Reviewer
* The API dependency is added to all module that share metrics. While this make for more work this is cleaner in terms of the OSGi dependency graph.
```xml
<dependency>
  <groupId>fish.payara.monitoring-console</groupId>
  <artifactId>monitoring-console-api</artifactId>
</dependency>
``` 
* (almost) all removed files are moved to monitoring-console repository https://github.com/payara/monitoring-console/pull/1
* `MonitoringConsoleRuntimeImpl` is the service bootstrapped on startup. It implements the low level grunt work for as requested by the API. In its post construction initialisation it uses the `MonitoringConsoleFactory` of the API to also bootstrap the `MonitoringConsole` instance that is the control abstraction for the "engine" of monitoring console.
* the _process_ and _webapp_ modules are runtime dependencies of the server that includes the monitoring console wherefore those are included by the packager
* as there was no central module that also depended on API for `CountStatistic` or `RangeStatistic` those now are registered by the modules that rely on them being registered. 

### Special Note on Bootstrapping
Due to changes in how modules are composed as server the old bootstrapping did no longer trigger creation of the `MonitoringConsoleRuntimeImpl` which is a `@Service`. This is most likely to be expected behaviour as no other part of the server needs this service so it is never created by HK2.
To work around this the class now implements the `ApplicationLifecycleInterceptor` `@Contract` which is resolved via `getAllServices(ApplicationLifecycleInterceptor.class)` at some point during server bootstrapping causing the creation of the `MonitoringConsoleRuntimeImpl` service.
As this is probably _too early_ it uses the `Events` and registers itself as a `EventListener` to wait with the actual bootstrapping of the console until the `SERVER_READY` event is received.

### Testing
Build and start the server. 

Check:
* MC can be started via command
* MC shows data
* MC can be disabled by disabling the server's monitoring service
* MC request tracing detail data is available when enabling RT
* MC custom watches can be created and remain after restart of the server
* MC shows data from other instances

Build and start Payara Micro. Check no problems due to missing dependencies occur during startup.


Pandrex247:
Couple of conflicts 😥

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/runtime/MonitoringConsoleRuntimeImpl.java
```diff
@@ -0,0 +1,336 @@
...
+
+    public void init() {
+        if (!initialised.compareAndSet(false, true) ) {
+            return;
+        }
+        try {
+            LOGGER.info("Bootstrapping Monitoring Console Runtime");
+            boolean isDas = serverEnv.isDas();
+            config = domain.getExtensionByType(MonitoringConsoleConfiguration.class);
+            String instanceName = "server"; // default
```
Pandrex247: Am I understanding this correctly? It seems off to me.

So if Hazelcast isn't enabled, despite an instance for example being called "Insty1", it will go to the `MonitoringConsoleFactory` and eventually get given an `InMemorySeriesRepository` which registers the instance name as _server_ and adds that name to its set of instances. 

Is that correct? I haven't dug deep down into the guts but an instance registering itself with an incorrect name perked an eyebrow 🤔

appserver/payara-appserver-modules/microprofile/metrics/pom.xml
```diff
@@ -108,5 +108,10 @@
             <groupId>jakarta.platform</groupId>
             <artifactId>jakarta.jakartaee-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>fish.payara.monitoring-console</groupId>
+            <artifactId>api</artifactId>
+            <version>1.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
jbee: Yes, missed to clean this after I added the property. :roll_eyes:

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/runtime/MonitoringConsoleRuntimeImpl.java
jbee: So far I extracted the name from hazelcast from the `HazelcastCore.INSTANCE_ATTRIBUTE` as I found that is done at other places. Happy to use some other reliable source if you can point me at one.

pom.xml
```diff
@@ -205,6 +205,9 @@
         <concurrent-api.version>1.1.2</concurrent-api.version>
         <concurrent.version>1.0.payara-p2</concurrent.version>
         <asm.version>7.2</asm.version>
+        <monitoring-console-api.version>1.0</monitoring-console-api.version>
+        <monitoring-console-process.version>1.0-SNAPSHOT</monitoring-console-process.version>
```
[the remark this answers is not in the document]
jbee: Yes, during development I used SNAPSHOT but now it should be 1.0. I also changes the webapp version to 1.0.

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/runtime/MonitoringConsoleRuntimeImpl.java
Pandrex247: I believe `ServerEnvironment.getInstanceName()` is fairly reliable.
Ottomh I'm not sure that the name returned by that is 100% always the name used in Hazelcast, but I'm nitpicking what happens if Hazelcast isn't enabled here so should be fine (assuming it works).

Although having said all that - does the monitoring even work without Hazelcast?

jbee:
jenkins test please

appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/runtime/MonitoringConsoleRuntimeImpl.java
jbee: I changed it to `ServerEnvironment.getInstanceName()`. It does work without Hazelcast in a single instance (DAS) configuration.

jbee:
As expected CI build fails because of missing dependencies.

jbee:
jenkins test please

jbee:
jenkins test please

MarkWareham:
jenkins test please

MarkWareham:
jenkins test please

MarkWareham:
jenkins test please

Pandrex247:
Jenkins test please

jbee:
Yeha. The honour is all yours @Pandrex247 .

pzygielo:
When one can expect this to be in central/other repository?
```
[ERROR] Failed to execute goal on project hazelcast-bootstrap: Could not resolve dependencies for project fish.payara.server.internal.payara-modules:hazelcast-bootstrap:glassfish-jar:5.201-SNAPSHOT: Could not find artifact fish.payara.monitoring-console:monitoring-console-api:jar:1.0 in central (https://repo.maven.apache.org/maven2) 
```

dmatej:
@pzygielo I hope it will be resolved today, emergency solution is to build this repo:
https://github.com/payara/monitoring-console , this commit: https://github.com/payara/monitoring-console/commit/1a5be6406f53aa82a28f6002cc7ff0ac889f16db

pzygielo:
> I hope it will be resolved today,

Thanks for checking.

> emergency solution is to build this repo:
> https://github.com/payara/monitoring-console , this commit: [payara/monitoring-console@1a5be64](https://github.com/payara/monitoring-console/commit/1a5be6406f53aa82a28f6002cc7ff0ac889f16db)

Yeah, it works indeed, that's what I did on my local machine on Friday :smile: 
(But I have to remember to delete MC from local repo as it's not snapshot, and will not be re-fetched from central...)

jbee:
@pzygielo https://search.maven.org/search?q=g:fish.payara.monitoring-console

Issue #4489, opened 2020-02-13
[microprofile] Openapi - Implementation of TreeSet comparator is wrong. /CUSTCOM-197


Berger1994:
The compartor implementation for the used TreeSet in the class `fish.payara.microprofile.openapi.impl.visitor.OpenApiWalker` ([here](https://github.com/payara/Payara/blob/a1dda93443b92f2c3ca54b8dbb1786a775727d5d/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java#L120)) is wrong. It depends on how many classes are filled in the list. (the order has a problem too)

I created a simple test class with the same compartor and the result is false althoug the class is inside the tree.

```
public static void main(String... args) {
		Set<Class<?>> clazzes = new TreeSet<>((class1, class2) -> {
			System.out.print("Compare " + class1.getSimpleName() + " with " + class2.getSimpleName() + " | ");
			if (class1.equals(class2)) {
				System.out.println("0");
				return 0;
			}
			// Subclasses first
			if (class1.isAssignableFrom(class2)) {
				System.out.println("-1");
				return -1;
			}
			// Non contextual objects at the start
			if (!class1.isAnnotationPresent(ApplicationPath.class) && !class1.isAnnotationPresent(Path.class)) {
				System.out.println("-1");
				return -1;
			}
			// Followed by applications
			if (class1.isAnnotationPresent(ApplicationPath.class)) {
				System.out.println("-1");
				return -1;
			}
			// Followed by everything else
			System.out.println("1");
			return 1;
		});

		clazzes.add(String.class);
		clazzes.add(Long.class);
		clazzes.add(Number.class);

		System.out.println("------");
		System.out.println(clazzes.contains(String.class));
	}
```

Output:
```
Compare String with String | 0
Compare Long with String | -1
Compare Number with String | -1
Compare Number with Long | -1
------
Compare String with Long | -1
Compare String with Number | -1
false
```

The contains method have to work. Otherwise there are troubles with the openapi file generation ([example](https://github.com/payara/Payara/blob/37695197d289b1cc2dcf5750aa47682e38305e41/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java#L1030))

rdebusscher:
Hi @Berger1994,

[...]

PR #4496, opened 2020-02-15
CUSTCOM-204 - Fixed comparator of OpenApiWalker's internal Tre…

Pscheidl:
…eSet of allowed classes

Attempts to fix https://github.com/payara/Payara/issues/4489

# Description
This is a bug fix for  https://github.com/payara/Payara/issues/4489.

As described in JavaDoc above the test for this fix, a TreeSet has been used inside `OpenApiWalker` class to ensure low lookup times during the OpenApi schema creation/generation process. Which is a clever choice, as a `TreeSet` ensures `O(log(n))` asymptotic complexity [source](https://www.baeldung.com/java-collections-complexity).

The previous comparator, with it's limited number of outputs for the whole set of allowed classes with typically high cardinality, was effectively making some classes not foundable on lookup, even though those classes were present. On lookup, the comparator sent the pointer to a completely different direction/tree node.

I believe a cheap and optimal solution would be to compare distance among classes using Strings. To quote the JavaDoc above the test created in this PR:

```
The compared objects inside the internal comparator supplied to the Set implementation are of type "Class".
And classes have no natural distinct ordinal numbers in Java. One way to get an ordinal number for a Class is
to use it's fully qualified name and compare the distance to the other compared class in the space consisting
of other class names. This way, "more similar" strings have closer cartesian distance. And as String 
representation of a Classs is used, fast lookup is ensured.
```
The implementation of `String.compareTo` method is [fast & efficient](http://jniosocket.sourceforge.net/result.html) and is able to quickly send the lookup pointer the right way in the tree by using the whole space of all possible class names. The opposite would be comparing by +1/-1 steps could also mean linear search in worst case.

# Important Info

# Testing

### New tests
`OpenApiWalkerTest` class with one new test named `testClassOrderingAndSearch`.

This class tests:
1. All inserted classes are found afterwards in the internal `TreeSet`
1. Internal contract is defined as `Set`
1. The implementation used is a `TreeSet` - debatable. If anyone changes the implementation, this test will fail and the person changing the code will be forced to change the test and thus cinciously consider the lookup complexity.

### Testing Performed
Created a JUnit test.

### Testing Environment
Oracle JDK 1.8.0_241
Ubuntu 19.10

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
@@ -55,6 +55,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.Comparator;
 import java.util.logging.Logger;
 
 import javax.ws.rs.ApplicationPath;
@@ -69,7 +70,6 @@ import javax.ws.rs.OPTIONS;
 import javax.ws.rs.PATCH;
 import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
-import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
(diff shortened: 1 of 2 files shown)


appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -117,25 +117,7 @@
...
-            if (!class1.isAnnotationPresent(ApplicationPath.class) && !class1.isAnnotationPresent(Path.class)) {
-                return -1;
-            }
-            // Followed by applications
-            if (class1.isAnnotationPresent(ApplicationPath.class)) {
-                return -1;
-            }
-            // Followed by everything else
-            return 1;
-        });
```
jbee: I think @MattGill98 knows about the logic here and if the order had any real significance other than performance.

MattGill98: Looks like a legitimate hangover from when the schema searching was less sophisticated. I don't think this serves much of a purpose any longer. As the TCK passes, our tests pass and my manual testing seems all okay, I'm happy with this change

jbee:
jenkins test please

Pscheidl:
Thank you guys.

PR #4497, opened 2020-02-17
CUSTCOM-75 Fix Payara System Property REST API

MattGill98:
# Description
This is a bug fix. 

This change allows instance property overrides to be created without deleting all the other instance properties by using the REST endpoint correctly. It also allows empty property fields to be submitted to remove the property.

Firstly the PUT and POST methods were the wrong way around, so they've been swapped. Secondly, the POST method didn't allow empty fields, which meant that instance property overrides couldn't be removed. To stop this, empty properties are now treated as properties to be removed.

# Testing

### Testing Performed
Setting up domain with:

~~~
asadmin copy-config default-config test-config
asadmin create-instance --node localhost-domain1 --config test-config instance1
asadmin create-system-properties --target instance1 ASADMIN_LISTENER_PORT=12345
asadmin create-system-properties --target instance1 HTTP_LISTENER_PORT=23456
~~~

From the admin console tried:

- removing an instance property from test-config -> system-properties -> instance values
- creating an instance property from test-config -> system-properties -> instance values
- removing an instance property from instances -> instance1 -> properties
- creating an instance property from instances -> instance1 -> properties

### Test suites executed

N/A

### DIFF
--- a/appserver/admingui/common/src/main/resources/configuration/sysInstanceValues.jsf
+++ b/appserver/admingui/common/src/main/resources/configuration/sysInstanceValues.jsf
@@ -38,6 +38,7 @@
     only if the new code is made subject to such option by the copyright
     holder.
 
+    Portions Copyright [2020] [Payara Foundation and/or its affiliates]
 -->
 
 <!initPage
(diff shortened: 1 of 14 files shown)


MattGill98:
jenkins test please

MattGill98:
jenkins test please

PR #4504, opened 2020-02-18
CUSTCOM-199 Fix REST management JSON requests

MattGill98:
# Description
This is a bug fix. 

The REST management interface couldn't be invoked via a JSON request, since the parameters would be read into the map as a JSONStringImpl, which still contains the wrapping quotation marks. This means that all of the parameters would be wrong. This commit uses the toString method
of the values read rather than hard casting to make sure the correct values are parsed.


MattGill98:
jenkins test please

MattGill98:
jenkins test please

dmatej:
Jenkins test please

PR #4505, opened 2020-02-18
QACI-95 Samples fixes

MarkWareham:
Removed problematic modules, with the intention of adding them back in later once problems have been addressed. Tracked internally with QACI-95


MarkWareham:
check of testing not applicable.

PR #4510, opened 2020-02-18
APPSERV-73 Upgrade Bouncy Castle to 1.64

MeroRai:
# Description
This is a component upgrade of the  Bouncy Castle to 1.64


MeroRai:
Jenkins test please

PR #4526, opened 2020-02-20
APPSERV-87 Remove trailing space

Pandrex247:
This is a bug fix.

The "Wide-eyed" adjective for the auto name generation has a trailing space, leading it to create names like "Wide-eyed -Ladyfish", which breaks stuff.

# Testing
None


jbee:
jenkins test please

PR #4534, opened 2020-02-25
APPSERV-55 Addresses causes of FT & Metrics TCK failures

jbee:
### Background
Both FT and Metrics TCK contain tests that are are not stably successful running against Payara.
In particular the test experience unexpected timeouts and a breech of the bulkhead limit.

The timeouts seem related to overall _too slow_ processing of the tests. Investigations suggest that this is due to lack of a worker actually starting the asynchronous task. The obvious reason would be that the executor service is busy running other tasks and/or that it does not offer the number of threads (and therefore concurrency) expected by the tests.

### Summary
Goal of this PR is to improve the MicroProfile FT and Metrics implementation so that they pass the TCK stably. 

### Causes
Findings on some potential causes:

* executor service used to run asynchronous execution did not offer sufficient parallelism and throughput. This is especially important since the TCK tests use `@Asynchronous` as part of the test _program_ to effectively call the method under test from multiple caller threads and because the executor used so far was also doing other tasks within the server.
* performance of processing FT semantics was degraded by substantial amount of individual map lookups some of which would create temporary intermediate data structures. Also this unnecessarily increased the surface area for race conditions as there were many paths accessing different aspects of the FT state for the same annotated method which were made thread-safe at the finest possible level of granularity.
* processing of bulkhead semantics was hard to reason about because of the way it nested and had to use try-finally blocks.
* `MetricRegistry` was not thread-safe. The _get or create_ semantics would not be _atomic_, that is a single point in execution would determine thread-safe and correctly for all concurrent threads if the lookup was a get or a create and which thread would end up creating the instance that should be received by others. Also there were some minor issues with `Metadata` validation edge cases and the thread-safe consistency between `Metric`s and `Metadata`. 
* **THE** race condition causing breaching of the bulkhead was introduced by use of the `InvocationManager` to get current application name which become part of the key to lookup the state for the FT annotated method. It is **not** thread-safe (even though it really should be). An internal issue is opened to address the race condition itself.

### Improvements

* the `ExecutorService` that runs asynchronous execution is exclusive and tuned towards a high throughput for many short living tasks. This is given by using a `ThreadPoolExecutor` (same as returned by `Executors.newCachedThreadPool()`) which dynamically scales the number of threads in the pool between 0 and some maximum depending on demand for running tasks. The maximum is now configured via asadmin commands. Changes are applied directly.
* the `ScheduledExecutorService` to delay execution and detect timeouts is now equally exclusive making sure that this executor is not busy with long running tasks scheduled by other sources.
The default pool size is 20. Changes to the size require a restart as the scheduler does not support changes after creation.
* **THE** race condition caused by use of `InvocationManager` is circumvented by not using it any longer. Instead the state for each method is solely identified by an identifier derived from its signature and target object. No longer needed method state is disposed using a cleanup thread. The cleanup is scheduled using the delay scheduler. This is fine as the cleanup is equally a very fast task with little actual work as identifying a no longer needed state object can be done by reading a counter and comparing a timestamp. Also this will always only be a single task. Cleanup runs once every minute. This has the benefit of getting rid of outdated state more aggressively then before where all state would always exist until the application would be undeployed. 
* The organisation of FT state was simplified a lot from multiple maps with several lookups (with lazy creation) per method during processing to a single state object for each method that is created at the beginning of an invocation and is effectively immutable during the execution. This both reduced surface area for race conditions and contributes positive to less thread contention and faster execution. It also is significantly easier to understand and follow when reading the code. 
* The FT bulkhead stage was re-implemented using a completely different approach. Originally done to exclude the somewhat hard to reason about semaphore based solution as the origin of the race condition causing TCK failures. While we now know that the semaphore based implementation was most likely not source of errors I kept the new solution because it is more elegant and easy to reason about regarding its correctness - of limiting threads to a certain number. It does so by treating the bulkhead as if it was a fairground ride with a certain maximum capacity which include both the people on the ride and those queuing. One counter counts people as they enter and exit the ride "area". Then a FIFO queue tracks people on the ride and the queuing people are waiting to get into this queue representing the ride. The big advantage of this way of counting and waiting is that it is the very same logic for both styles of bulkheads just that the _semaphore style_ (spec name) has a waiting queue length of zero. So anyone not fitting on the ride would immediately give up.
* The `MetricRegistryImpl` internal organisation of data was revised so that lookups and creation is _atomic_ and thread-safe while also keeping `Metadata` and `Metric`s consistent with each other. To ensure correctness of behaviour new tests were added. As those changes meant that some of the read-only collection accessor methods would need to compute the returned collections the usage of such methods was avoided as far as possible and replaced with direct lookup methods. As the registry lacks methods for key based lookup in its API this would done via service which internally has access to additional methods of the implementation class. A request  https://github.com/eclipse/microprofile-metrics/issues/543 has been made (and received positive) to FT metrics to improve the `MetricRegistry` API so it offers key based access.

### Tests
During the investigation a whole range of unit tests testing the correctness of the bulkhead in concurrent scenarios were added. These were not able to find problems as the race condition was outside the FT logic. When `InvocationManager` was returning wrong app name it would lead to creation of more then one state object for the same method allow twice as much threads in a bulkhead.

The revised `MetricRegistryImpl` got covered 78%. This should cover every possible illegal case and most of the possible legal cases. These do not check thread-safety but correctness of behaviour. 

#### Tests Performed
Running the FT TCK 2.0. `tck-suite2.0-stable.xml` is successful. `tck-suite2.0.xml` can fail on `BulkheadMetricTest` methods with a `java.util.NoSuchElementException: No value present`. This is from `getBulkheadConcurrentExecutions().get()` on an empty `Optional` which should have hold the number of concurrent executions from the gauge provided by the bulkhead. As far as I can tell this is a race condition in the tests and I suggest to ignore those tests for now and hope for TCK improvements. When run individually they often succeed so the general behaviour is fine.
Also there is still a change to get the odd _Work is not being done simultaneously enough_ error. Nothing we can do about that except giving it more threads and a bigger machine to run on. Again we have to hope on TCK improvements which are discussed in FT group in regards to this.

Side note: TCK execution time went down from 20isch minutes to about 7 on my laptop. :tada: 

The changed asadmin commands where tested manually. 
```
asadmin get-fault-tolerance-configuration
asadmin set-fault-tolerance-configuration --async-max-pool-size=2000
``` 

The cleanup of FT state can be verified from logs. After running FT TCK there should appear messages like 

> Cleaned 55 expired FT method contexts

after a while (runs once per minute)

### Logging
The server will now print a log statement on `INFO` level each time a new FT method state is created, an example:

> Creating FT method context for 1b75bfbc@org.eclipse.microprofile.fault.tolerance.tck.interceptor.xmlInterceptorEnabling.InterceptorComponent$Proxy$_$$_WeldSubclass.asyncGetString

and a summary entry, each time such objects are cleaned, for example:

>  Cleaned 10 expired FT method contexts

I decided that these would be good on `INFO` level as they would not occur too frequent and help as a basis of and FT related problem analysis done "post-mortem".

Also adds a warning in case configuration had a non-default value for the outdated configuration of managed executor names.


appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
```diff
@@ -200,9 +205,17 @@ public FaultToleranceConfig getConfig(InvocationContext context, Stereotypes ste
...
     @Override
     public FaultToleranceMetrics getMetrics(InvocationContext context) {
+
         FaultToleranceApplicationState appState = getApplicationState(getApplicationContext(context));
-        return appState.getMetrics().updateAndGet(
-                metrics -> metrics != null ? metrics : new BindableFaultToleranceMetrics()).bindTo(context);
+        return appState.getMetrics(this::getApplicationMetricRegistry, FaultToleranceUtils.getCanonicalMethodName(context));
+    }
+
+    private MetricRegistry getApplicationMetricRegistry() {
```
svendiedrichsen: How about using an optional here?

jbee: PR is still in development. Let me get back when I feel ready.

svendiedrichsen: sure

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadata.java
```diff
@@ -259,16 +233,43 @@ public MetricType getTypeRaw() {
...
-    
+
     public void addTags(List<XmlTag> tags) {
         if (this.tags == null) {
             this.tags = new ArrayList<>();
         }
         this.tags.addAll(tags);
     }
 
+    @Override
```
jbee: NB. Both methods are copied straight from `DefaultMetadata` so they follow the same logic. This is needed because the registry checks for illegal reuse comparing _new_ and _existing_ `Metadata` with `equals`. Not having it implemented fails the test and considers them _not equal_ and therefore illegal for reuse. It is beyond me how this did not pop up earlier.

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadataHelper.java
```diff
@@ -88,7 +88,7 @@
     public List<MBeanMetadata> registerMetadata(MetricRegistry metricRegistry,
             List<MBeanMetadata> metadataList, boolean isRetry) {
 
-        if (!metricRegistry.getMetadata().isEmpty() && !isRetry) {
+        if (!metricRegistry.getNames().isEmpty() && !isRetry) {
```
jbee: NB. `getNames()` is a view, whereas `getMetadata()` is computed collection, therefore the change.

jbee:
jenkins test please

jbee:
jenkins test please

jbee:
@Pandrex247 there was one more small change to make ;) Just one to be on the safe side but important nonetheless.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
jbee: @svendiedrichsen I had a look now. The return value is just used internally. I didn't feel that `Optional` would have improved the code. This might be because I am not a big fan personally or because `Optional` isn't use in our codebase generally. I don't feel starting here is the way to go. Not even sure I ever want to start using it but if this should be something broadly applied to the codebase. To go from having to expect `null` basically everywhere to _there should not be null references_ and if that means you need an `Optional` then use one.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/cdi/FaultToleranceInterceptor.java
```diff
@@ -74,9 +77,12 @@ public Object intercept(InvocationContext context) throws Exception {
         try {
             FaultToleranceService env =
                     Globals.getDefaultBaseServiceLocator().getService(FaultToleranceService.class);
-            FaultTolerancePolicy policy = FaultTolerancePolicy.get(context, () -> env.getConfig(context, this));
+            AtomicReference<FaultToleranceConfig> lazyConfig = new AtomicReference<>();
```
jbee: NB. Makes sure multiple calls to the supplier function always result in the same config instance.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
```diff
@@ -148,112 +171,63 @@ public void event(Event<?> event) {
...
     }
 
     private static void collectCircuitBreakerState(MonitoringDataCollector collector, CircuitBreakerState state) {
+        if (state == null) {
+            return;
+        }
         collector
-            .collect("halfOpenSuccessFul", state.getHalfOpenSuccessFulResultCounter())
-            .collect("state", state.getCircuitState().name().charAt(0));
+            .collect("circuitBreakerHalfOpenSuccessFul", state.getHalfOpenSuccessFulResultCounter())
```
Pandrex247: Only just noticed this since apparently it's already there in the CircuitBreakerState for _some_ method names: _SuccessFul_ shouldn't have a capital 'F'

o((>ω< ))o

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
```diff
@@ -122,20 +113,52 @@
...
-        serviceConfig = serviceLocator.getService(FaultToleranceServiceConfiguration.class);
         invocationManager = serviceLocator.getService(InvocationManager.class);
         requestTracingService = serviceLocator.getService(RequestTracingService.class);
-        InitialContext context = new InitialContext();
-        defaultExecutorService = (ManagedExecutorService) context.lookup("java:comp/DefaultManagedExecutorService");
-        defaultScheduledExecutorService = (ManagedScheduledExecutorService) context
-                .lookup("java:comp/DefaultManagedScheduledExecutorService");
+        config = serviceLocator.getService(FaultToleranceServiceConfiguration.class);
+        delayExecutorService = Executors.newScheduledThreadPool(getMaxDelayPoolSize());
+        asyncExecutorService = new ThreadPoolExecutor(0, getMaxAsyncPoolSize(), 60L, TimeUnit.SECONDS,
```
Pandrex247: Just some food for thought: do we want the timeout and core pool size to be configurable? Even as hidden parameters of the set command so the uninitiated don't use them.

Probably fine as is but just planting the seed of thought (and/or doubt) 🙂

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
```diff
@@ -122,20 +113,52 @@
...
+                new SynchronousQueue<Runnable>(true)); // a fair queue => FIFO
+        delayExecutorService.scheduleAtFixedRate(this::cleanMethodContexts, 1L, 1L, TimeUnit.MINUTES);
```
Pandrex247: See note above - do we want to have some hidden configuration knobs we can turn if this rate isn't ideal?

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceServiceConfiguration.java
```diff
@@ -39,22 +39,38 @@
...
  * Configuration for the Fault Tolerance Service.
- * @author Andrew Pielage
+ *
+ * @author Andrew Pielage (initial)
+ * @author Jan Bernitt (change to pool size)
  */
 @Configured(name = "microprofile-fault-tolerance-configuration")
 public interface FaultToleranceServiceConfiguration extends ConfigExtension {
 
-    @Attribute(defaultValue = "concurrent/__defaultManagedExecutorService", dataType = String.class)
```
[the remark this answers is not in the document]
Pandrex247: OK just did a quick check and no it doesn't throw an error if the domain.xml has unrecognised entries on startup.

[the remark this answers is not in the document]
jbee: I this good news, meaning I can just erase them?

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/test/java/fish/payara/microprofile/faulttolerance/policy/BulkheadTckAsynchRetryTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Runs test similar to {@code org.eclipse.microprofile.fault.tolerance.tck.bulkhead.BulkheadAsynchRetryTest} as JUnit
+ * tests.
+ * 
+ * Uses the same test method names and tries the duplicate the test as good as possible on the
+ * {@link FaultTolerancePolicy} level. For many identifiers the terminology of the TCK test is used to make them easier
+ * to compare.
+ */
+public class BulkheadTckAsynchRetryTest extends AbstractBulkheadTest {
```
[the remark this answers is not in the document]
jbee: Same here. I can rename it for our sake :)

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
jbee: Going to fix that.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
jbee: I'll add one for the keep alive time (time before an idle worker is removed).

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
jbee: Good idea, I add one.

jbee:
jenkins test please

jbee:
@Pandrex247 added hidden configurations for async pool keep alive and the cleanup interval. Renamed the _asnch_ test to _async_ and corrected the _successFul_ to _successful_.

I also fixed a bug in the cleanup of expired state. It was really just half way implemented. Now it is reasonable and also considers that a circuit breaker can only be cleaned if it is closed without failures in the queue that might cause a state change. I also use a `WeakReference` on the target object to see if that got collected. In that case the FT state is outdated and can be cleaned. This e.g. should detect the case where an application is undeployed.

All in all this means you should not just check the changes for as-admin command but also rerun the TCK and check the cleanup is running.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceServiceConfiguration.java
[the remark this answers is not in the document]
Pandrex247: The only thing that's stopping me from saying _yes_ is "what happens when someone updates from 201 and they were using this to use a non-default executor".
It might be nice to check if it's been configured to use a non-default executor on startup and print a warning saying it's not used any more - cleanest way I can think of to do this would be to retain the attribute and read it using the get method rather than doing xml parsing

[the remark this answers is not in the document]
jbee: I kept the fields and added the `obsolete = true`.

jbee:
jenkins test please

jbee:
jenkins test please

jbee:
@Pandrex247 Overall I am confident that the improvements in this PR move the stability issue from an overall issue to 1-2 smaller "local" issues.

I was mostly running TCK with max of 2000 (as I forgot I had changed it to that) and that mostly got me through. As you say and as I also pointed out in description the bulkhead metrics have a tendency to fail with `NoSuchElement` exception. I cannot say 100% if this is race condition on tests or on our side but its certainly a minor metrics related issue. My suggestion was to exclude those tests for now from TCK runner.

I'd say we accept this PR as a good step forward and see if we want and can spend some more time for metrics issue and overall stability with regards of executor capacity. WDYT?

Pandrex247:
Sorry missed that bit in your description.
OK I'm fairly happy to take this as a step towards getting it perfect as it has numerous improvements.

I've tested that your log messages appear correctly and the set command seems to work as expected - the only thing that's missing that I've just noticed is an update to the admin console: it still gives drop-down boxes for selecting executor services.

configs > MicroProfile > Fault Tolerance

jbee:
jenkins test please.

jbee:
jenkins test please

fschuerer:
Hi,

is it possible to use at FaultToleranceServiceImpl::getTargetMethodId(InvocationContext context) 
"target.hashCode()"
instead of 
"System.identityHashCode(target)"

jbee:
@fschuerer for what reason?

fschuerer:
The reason is, I can use my own hashCode() method. 
For example in the method  which is annotated with CircuitBreaker, I' am calling different servers per REST. If one service is down, the circuitbreaker opend only for the service which is down and all other service are reachable. This works perfect before this PR.  Now, if one service is down, the circuitbreaker is opend for all services.

jbee:
I feel I miss something. What the method tries to do is give you a unique ID for a particular method of a particular class and a particular instance. If any of the 3 is different this should be another FT context and as such must compute to another ID.

Now, say we would use `target.hashCode()` I assume the returned value would be different for actual different instances which should be the same behaviour as `System.identityHashCode`. If on the other hand different instances would return the same `hashCode()` value all would map to the same FT context. I am fairly certain is that this isn't what the specification intends to be valid. It might very well be a grey-area that isn't clearly defined. But before we dig into this, is that what you used to do? Returning same `hashCode()` value and thereby mapping different actual instances of the service to the same FT context?

fschuerer:
May be that I misunderstand the CircuitBreaker pattern, so I have created a very simple example and a system test how I use the circuitbreaker: https://github.com/fschuerer/breaker
With payara 5.201 it works an the test is green, but with payara 5.2020.4 the test fails.

jbee:
> With payara 5.201 it works an the test is green, but with payara 5.2020.4 the test fails.

I understand that from your point of view this might look like a defect was introduces. And it very well may be the case - but it may also be the case that the behaviour before was not correct.

I think the way you use the service is not safe. A resource exists once per request, but the `prepareService` call manipulates a service singleton. This should not work or should cause concurrency issues. I can see how it can give you the behaviour you want where the service "switches" context based on the target URI but this is not thread-safe. Now if the service isn't a singleton the concurrency issue is solved but it should also be obvious that with the service being fresh per request a circuit breaker makes no sense as it acts upon experience of the past. 

I can see how custom `hashCode` used to link to a particular FT context gives you another form of usage but I am afraid the interaction of CDI and FT do not allow for such a pattern with their current semantics. I can bring up the question in FT group to see if and how this could be handled but I assume this requires additions to FT spec.

fschuerer:
@jbee thanks for your support. 

Please let me make one final note: I use a `Singleton` scope, because the `CiruitBreaker` don't work for `RequestScoped` (https://github.com/eclipse/microprofile-fault-tolerance/issues/479).
I have changed the sample https://github.com/fschuerer/breaker and removed the `Singelton` scope. Now the example is roughly like our current use case (it acts like a API gateway). Please have a look and I am grateful for your comment.

dmatej:
Just a small note: hash codes are not unique ... they cannot be used as an id. Same applies to System.identityHashCode, see javadocs.

jbee:
@dmatej it cannot be guaranteed in theory as `int` is simply to little of a space to work with but in practice we can assume it is a virtual memory address of the object (virtual as it will not change even if the object is moved in physical memory). This is good enough for our purposes. This https://www.infoworld.com/article/2073618/java-s-system-identityhashcode.html links some resources on the topic.

jbee:
@fschuerer As far as I can tell the example still has a similar issue. You want to use information that first is available in the request to select the FT context to use. As FT annotation work on CDI level the FT context is the CDI scope of the bean that is annotated. A request scoped bean would not make sense even if it worked, as the it would not have a past so the circuit is always in its initial state. If the scope is more durable changing the state of the bean from a request is not thread-safe and would cause some requests being processed with the wrong context as concurrently process requests change the state before it is used to process the FT annotated method. 

I think you have a valid use case in wanting to determine the used FT context from information that are dynamic but I am not aware that FT in combination with CDI would support it. If that worked on older Payara versions this is by accident.

dmatej:
It is not just theoretical. Also the hashCode is not a constant for general objects, see this issue (old but still valid): https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6809470
https://bugs.openjdk.java.net/browse/JDK-4990451
Under some circumstances it might be stable for methods for some time ... but I would still expect it might cause issues in a future ...

jbee:
They commented: 
> Identity hashcode values are based upon heap pointer address. In the provided example you are generating a lot of garbage objects (new Integer(88)). These garbage objects are readily reclaimed and the address space is reused. The collisons result from address space reuse. If the original object remains live (not GCed) then you will not encounter this problem.

So assuming FT annotations sit on beans that are only unload should the application unload we are fine. 

Also we don't really have much alternatives. Only context we have is the instance of an object we are about to call the method on. If you have any better idea to generate a unique ID from this information please share it.

jbee:
@fschuerer I confirmed with FT group that the assumed scope for circuit breaker is indeed a singleton scope. Currently there is no feature in FT and CDI that would support your use case. 

If we want you can make a feature request (create issue) for Payara to support some non standard way to make your use case work.

jbee:
Turns out that FT context should always be _as if_ the annotated class is in application scope, see https://github.com/eclipse/microprofile-fault-tolerance/issues/479

PR #4536, opened 2020-02-27
APPSERV-86 Allow AutoNameInstancesTest to work with non-default domains.

Pandrex247:
# Description
See title - setting a static variable from BeforeClass didn't seem to persist to the actual container, so set a system property instead.

# Testing
Ran test using remote profile against domain1 & production, and ran against server-managed container.


Pandrex247:
Jenkins test please

PR #4541, opened 2020-02-28
APPSERV-88 CUSTCOM-213 generate-encryption-key Command Always Creates Key for Default Domain

Pandrex247:
# Description
This is a bug fix.
The `generate-encryption-key` command will only create the key for the default domain due to the _domain_name_ parameter being commented out on its super class.

A workaround to this is to either make sure there’s only one domain in your domain dir (the default domain is always the solitary domain in the specified domain dir - it doesn't need to be called _domain1_), or to simply copy the generated file from the default domain to where you need it.

# Testing
This also fixes the SfsbEncryptionTest so that it isn't always running against the default domain.

### Testing Performed
Ran the SfsbEncryptionTest using server-managed and server-remote profiles, against domain1 and production domains.

### DIFF
--- a/appserver/tests/payara-samples/test-utils/src/main/java/fish/payara/samples/ServerOperations.java
+++ b/appserver/tests/payara-samples/test-utils/src/main/java/fish/payara/samples/ServerOperations.java
@@ -747,20 +747,7 @@ public class ServerOperations {
             CliCommands.payaraGlassFish("set-hazelcast-configuration", "--encryptdatagrid", "true");
 
             System.out.println("Stopping Server");
-            String domain = System.getProperty("payara.domain.name");
-            if (domain == null) {
-                domain = getPayaraDomainFromServer();
-                if (domain != null && !domain.equals("null")) {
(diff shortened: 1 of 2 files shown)


Pandrex247:
Jenkins test please

appserver/tests/payara-samples/test-utils/src/main/java/fish/payara/samples/ServerOperations.java
```diff
@@ -782,16 +785,19 @@ public static void disableDataGridEncryption() {
             System.out.println("Disabling Data Grid Encryption");
             CliCommands.payaraGlassFish("set-hazelcast-configuration", "--encryptdatagrid", "false");
 
-            String domain = System.getProperty("payara.domain.name", "domain1");
-            if (domain != null) {
+            String domain = System.getProperty("payara.domain.name");
+            if (domain == null) {
```
MarkWareham: This block is a repeat of above. could be put into a method

Pandrex247:
Jenkins test please

PR #4542, opened 2020-02-28
APPSERV-54 Print a Warning if No Encryption Key Present

Pandrex247:
# Description
Prints a warning on the command line and in the admin console when enabling data grid encryption when a key can't be found.

# Testing

From a clean domain, simply run `set-hazelcast-configuration --encryptdatagrid true` and you should get a warning.
You should get a similar warning when attempting to do the same from the admin console: Domain > Data Grid.

### New tests
New asadmin test added to existing `SetHazelcastConfigurationTest` in payara-samples.
Checks that a warning is given when enabling datagrid encryption without generating a key.

# Notes for Reviewers
Touching such a fundamental method in the admin console (`updateEntity`) worries me a bit, as this can easily ripple outwards in unforeseen ways. I've poked around the admin console and haven't managed to make it throw a wobbly but I can't poke everything.

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/RestApiHandlers.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/RestApiHandlers.java
@@ -272,6 +272,7 @@ public class RestApiHandlers {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


Pandrex247:
Jenkins test please

PR #4507, opened 2020-02-18
APPSERV-70 Upgrade Sun Java Streaming XML Parser to 1.0.2

MeroRai:
# Description
This is a component upgrade of the Sun Java Streaming XML Parser to 1.0.2

PR #4545, opened 2020-03-03
APPSERV-70 Remove unnecessary test dependency

Pandrex247:
# Description
This is an improvement.
During our attempts to update this component to 1.0.2, it started throwing errors due to unsupported operations.
Upon further investigation I'm not convinced we need this dependency any more - it's only test scoped and the test works fine without it.

See related PR from Susan: https://github.com/payara/Payara/pull/4507

# Testing
Build Payara Server and don't skip tests - the relevant module is config-api.


Pandrex247:
Jenkins test please

PR #4550, opened 2020-03-04
CUSTCOM-211 Upgraded Trilead library and included ECDSA support

dmatej:
# Important Info

Fixes problems with default keys on Ubuntu.

### Blockers

Depends on following PR or usage of special profile in settings.xml as an alternative:
https://github.com/payara/Payara_PatchedProjects/pull/295

The alternative profile:
```
    <profile>
      <id>ext</id>
      <repositories>
        <repository>
          <id>jenkins</id>
          <url>https://repo.jenkins-ci.org/releases/</url>
          <snapshots><enabled>false</enabled></snapshots>
          <releases><enabled>true</enabled></releases>
        </repository>
      </repositories>
    </profile>
```

# Testing

See https://github.com/payara/Payara/pull/4548
- these pull requests don't depend on each other, BUT with both integrated is possible to change configuration in UI between password and keyfile authentication using ECDSA private key.

### DIFF
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/SecureAdminBootstrapHelper.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/SecureAdminBootstrapHelper.java
@@ -308,8 +308,8 @@ public abstract class SecureAdminBootstrapHelper {
          * @param milliseconds normal Java time (in milliseconds)
          * @return
          */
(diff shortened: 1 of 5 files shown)


dmatej:
Jenkins test please

dmatej:
> I was worried about removal of JSON API in first version, but that's now gone.

Yes, I was thinking about it at the same time. The dependency is brought by another dependency, but finally I decided it should be here too, because it is also directly used by this module.

PR #4566, opened 2020-03-09
CUSTCOM-223 Move log message into if block

Pandrex247:
This is a bug fix

The message stating that encryption is enabled is printed regardless of whether or not it is actually enabled.


jbee:
jenkins test please

PR #4568, opened 2020-03-09
APPSERV-57 Updates to MicroProfile Fault Tolerance 2.1

jbee:
### Summary
A good summary on the spec changes can be found https://download.eclipse.org/microprofile/microprofile-fault-tolerance-2.1/microprofile-fault-tolerance-spec.html#release_notes_21

Changes for Payara FT:
* adds of new exception handling properties to `@Fallback` (`applyOn` and `skipOn`) and `@CircuitBreaker` (`skipOn`)
* adds: CDI `@RequestScope` context is now activated for worker threads working `@Asynchronous` computation
* adds: `@Bulkhead` and `@Asynchronous` annotated methods returning `CompletionStage` now semantically exit the bulkhead when they complete (not when the annotated method returns a value).
* fixes: exception handling corrected to handle both `Exception` and `Error` (missed `Error` before which wasn't spec compliant before but went unnoticed until 2.1 that clarified exception behaviour)
* fixes: `Class` name overrides via configuration now use context `ClassLoader` (failed to load application specific classes before)

### Testing
Adds additional unit test for the new bulkhead semantics where the bulkhead context is exited on completion of the returned `CompletionStage`.

#### Tests Performed
Running TCK 2.1 (needs https://github.com/payara/MicroProfile-TCK-Runners/pull/102) against locally running server:
```bash
cd MicroProfile-Fault-Tolerance/tck-runner
``` 
The important part is that the TCK runner prints API and TCK version 2.1 as shown below:
```bash
     [echo] FT version     2.1
     [echo] FT TCK version 2.1
     [echo] FT suite       tck-suite2.1-stable.xml
```

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/policy/FaultTolerancePolicy.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/policy/FaultTolerancePolicy.java
@@ -559,6 +559,7 @@ public final class FaultTolerancePolicy implements Serializable {
         }
         logger.log(Level.FINER, "Proceeding invocation with bulkhead semantics");
         final boolean async = isAsynchronous();
+        final boolean exitOnCompletion = async && bulkhead.exitOnCompletion;
         final int runCapacity = bulkhead.value;
         final int queueCapacity = async ? bulkhead.waitingTaskQueue : 0;
         AtomicInteger queuingOrRunning = invocation.context.getQueuingOrRunningPopulation();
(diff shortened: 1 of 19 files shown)


jbee:
jenkins test please

jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/policy/FaultTolerancePolicy.java
```diff
@@ -579,15 +601,28 @@ private Object processBulkheadStage(FaultToleranceInvocation invocation) throws
                     try {
                         logger.log(Level.FINE, "Entered bulkhead execution.");
                         // ok, lets run
-                        return proceed(invocation);
+                        Object res = proceed(invocation);
+                        if (!exitOnCompletion) {
+                            return res;
+                        }
+                        return ((CompletionStage<?>) res).whenComplete((value, excetion) -> {
```
Pandrex247:
```suggestion
                        return ((CompletionStage<?>) res).whenComplete((value, exception) -> {
```
Presumably meant to be exception?

jbee:
@Pandrex247 

> Is that a known stability issue? Rejected Execution sounds like the thread pool maxing out (even though it appears there's still some space)?

Yes, Mark had similar experience. The default of 1000 seems bad (for testing at least) as the test want somewhat more. I did run with 2000 and that seemed to solve the issue. So we should either raise the default or configure the tests to use a higher value.

jbee:
jenkins test please

PR #4576, opened 2020-03-17
CLOUD-13 Prevent Hazelcast from being started twice on Payara Micro

Cousjava:
# Description
This is a bug fix

As PayaraMicroImpl#L1060 will start Hazelcast whatever, having it
enabled in the domain.xml means that Hazelcast will be started earlier
then restarted when the bootstrap says to enable Hazelcast.

# Testing
### Testing Performed
Start Payara Micro, check the Hazelcast has only been started once.
Tested with reproducer from https://github.com/payara/Payara/pull/4457

### Test Suites Run
 - JavaEE7-Samples
 - JavaEE8-Samples

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.2

# Reviewer notes

This is the same as #4394, which now has a merge conflict.

PR #4577, opened 2020-03-17
CLOUD-13 Remove command that restarts Hazelcast, so it is only started once

Cousjava:
# Description
This is a bug fix.

This removed the command that will restart Hazelcast, as it is already started during boot.

# Important Info
Alternative to #4576 

# Testing

### Testing Performed
Start Payara Micro, check the Hazelcast has only been started once.
Tested with reproducer from #4457

### Test suites executed
- Java EE7 Samples
- Java EE8 Samples

### Testing Environment
Zulu JDK 1.8_232 on Ubuntu 19.10 with Maven 3.6.2

# Notes for Reviewers


Cousjava:
Jenkins test please

PR #4584, opened 2020-03-20
APPSERV-20 Add Config Source for Payara Variables

Pandrex247:
# Description
This is a feature. 

This PR adds a config source for a new properties file: payara-expression-config.properties. The properties stored in here are converted from their variable into their actual values. For example, a property of `fish.payara.example.alias=${ALIAS=wibbles}` will be converted to the password stored under the alias of _wibbles_. If an alias isn't found, null is returned.

This PR also changes it so that the password alias store will also match against both "wibbles" and "${ALIAS=wibbles}"

# Testing

### New tests
Some unit tests added to check that null is returned from the config source if no matching alias is found, and also that the values are successfully converted for system properties and environment variables.

I also created a small sample app to check that the properties file is read and that password alias store reads both "wibbles" and "${ALIAS=wibbles}". This has been added under `appserver/tests/payara-samples/samples`.

### Test suites executed

- Payara Microprofile TCKs Runner (Config)

### Testing Environment
Windows 10. JDK 8u212.

# Documentation
https://github.com/payara/Payara-Server-Documentation/pull/724


Pandrex247:
Jenkins test please

appserver/tests/payara-samples/samples/microprofile-config-alias-properties/src/test/java/fish/payara/samples/microprofile/config/alias/MicroProfileConfigAliasTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @AfterClass
+    public static void deletePasswordAlias() {
+        CliCommands.payaraGlassFish("delete-password-alias", "wibbles");
+    }
+
+    @Test
+    public void testAliasSubstitution() throws Exception {
+        TextPage page = new WebClient().getPage(url + "ConfigServlet");
+        System.out.println(page.getContent());
```
jbee, line 95: Is this intentionally left `println` ?

appserver/tests/payara-samples/samples/microprofile-config-alias-properties/src/test/java/fish/payara/samples/microprofile/config/alias/MicroProfileConfigAliasTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        Assert.assertTrue(page.getContent().contains("Normal Notation: wobbles"));
+        Assert.assertTrue(page.getContent().contains("Substitution Notation: wobbles"));
+        Assert.assertTrue(page.getContent().contains("Password Alias from File: wobbles"));
+        Assert.assertTrue(page.getContent().contains("System Property Alias from File: Tiddles!"));
+        Assert.assertTrue(page.getContent().contains("System Property Alias from File: Tiddles!"));
+        Assert.assertTrue(page.getContent().contains("Environment Variable Alias referencing System Property Alias from File: Dobbles"));
+        Assert.assertTrue(page.getContent().contains("Environment Variable Alias and System Property Alias from File (same property): Bibbles and Bobbles"));
+    }
```
jbee, line 104: I'd like to see a message in connection with `assertTrue` and `contains`. Maybe its a good idea to create a little helper method that in case of failure has a message like `"Expected contains: Normal Notation: wobbles"` and so forth.

nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java
```diff
@@ -303,5 +297,19 @@ public static String getRealPasswordFromAlias(final String at) throws
         return new String(domainPasswordAliasStore.get(an));
     }
 
+    private static String expandPasswordAlias(String alias) {
```
jbee, line 300: Can't see a caller for this method.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
```diff
@@ -368,6 +373,18 @@ public String getDeployedApplicationProperty(String applicationName, String name
         return result;
     }
 
+    public List<Properties> getDeployedApplicationAliasProperties(String applicationName) {
+        ApplicationInfo info = applicationRegistry.get(applicationName);
+        List<Properties> result = Collections.EMPTY_LIST;
```
jbee, line 378: `Collections.emptyList()` is suggested to be preferred over using `Collections.EMPTY_LIST` directly.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/AliasPropertiesConfigSource.java
```diff
@@ -0,0 +1,107 @@
...
+        return Integer.parseInt(configService.getMPConfig().getAliasPropertiesOrdinality());
+    }
+
+    @Override
+    public Map<String, String> getProperties() {
+        HashMap<String,String> result = new HashMap<>(properties.size());
+
+        for (Object key : properties.keySet()) {
+            String alias = properties.getProperty((String) key);
+            String value = getValue((String) key);
```
jbee, line 76: Something like
```java
return properties.entrySet().stream()
                  .collect(Collectors.toMap(e -> e.getKey().toString(), e -> getValue(e.getValue));
```
would be nice.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
```diff
@@ -300,6 +302,9 @@ Config getNamedConfig(String applicationName) {
             for (Properties props : getDeployedApplicationProperties(appName)) {
                 sources.add(new PropertiesConfigSource(props, appName));
             }
+            for (Properties props : getDeployedApplicationAliasProperties(appName)) {
+                sources.add(new AliasPropertiesConfigSource(props));
```
jbee, line 306: I wonder why `PropertiesConfigSource` gets the `appName` passed while `AliasPropertiesConfigSource` does not.

appserver/tests/payara-samples/samples/microprofile-config-alias-properties/src/test/java/fish/payara/samples/microprofile/config/alias/MicroProfileConfigAliasTest.java
Pandrex247: Yes, it lets you see what the page actually returns without using a debugger. It also means it prints it on the terminal running the test rather than in the server log.
Or are you referring specifically to `print` vs. `println`?

nucleus/admin/config-api/src/main/java/org/glassfish/config/support/TranslatedConfigView.java
Pandrex247: Yep, left behind from a refactor. Thanks :)

appserver/tests/payara-samples/samples/microprofile-config-alias-properties/src/test/java/fish/payara/samples/microprofile/config/alias/MicroProfileConfigAliasTest.java
jbee: Just wanted make sure this is intentionally present.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
Pandrex247: Was copied from another method. I'll make the change there too

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/ConfigProviderResolverImpl.java
Pandrex247: It's not actually used in either, but I left the original alone originally as I thought I might have missed something.
On second look though I can't see anything that uses it since it's private scoped with no getter - I'll remove it

Pandrex247:
Jenkins test please

Pandrex247:
New PR here: https://github.com/payara/Payara/pull/4630

PR #4587, opened 2020-03-25
 CLOUD-59 Add API for adding asadmin commands to run via Micro api 

Cousjava:
# Description
This is a feature.

Allows adding commands to run on Payara Micro with boot via the programmatic API.

# Important Info

# Testing

### Test suites executed
- Java EE7 Samples
- Java EE8 Samples

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 18.04 with Maven 3.6.0

# Documentation

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroBoot.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroBoot.java
@@ -39,6 +39,7 @@
  */
 package fish.payara.micro.boot;
 
+import fish.payara.boot.runtime.BootCommand;
 import fish.payara.micro.BootstrapException;
 import fish.payara.micro.PayaraMicroRuntime;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


Cousjava:
Jenkins test please

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroBoot.java
```diff
@@ -39,6 +39,7 @@
  */
 package fish.payara.micro.boot;
 
+import fish.payara.boot.runtime.BootCommand;
```
pdudits: This class is not part of the API.

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroBoot.java
```diff
@@ -86,6 +87,25 @@
      * @return
      */
     PayaraMicroBoot addRepoUrl(String... URLs);
+    
+    /**
+     * Adds a command to run before booting Payaa Micro.
+     * Not that only a limited set of commands will run here i.e. {@code set}.
+     * @param command Command to add
+     */
+    public void addPreBootCommand(BootCommand command);
```
pdudits: Did you consider callbacks or other means of passing command result back to the caller, as was requested in the issue?

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/PayaraMicroBoot.java
```diff
@@ -86,6 +87,25 @@
...
+     * Adds a command to run after Payara Micro has booted
+     * @param command Command to add
+     */
+    public void addPostBootCommand(BootCommand command);
+    
+    /**
+     * Adds a command to be run after the application is deployed
+     * @param command Command to add
+     */
+    public void addPostDeployCommand(BootCommand command);
```
pdudits: PostDeploy is equivalent `run`, so it likely doesn't need explicit api.

Only if we'd deprecate `run` forms that execute asadmin, because they are mixed with data processing.

Cousjava:
replaced by #4600

PR #4589, opened 2020-03-26
CUSTCOM-166 Make Beans With @RolesPermitted Serializable

MattGill98:
# Description
This is a bug fix. 

Beans with @RolesPermitted could not hold scopes like @ViewScoped or @SessionScoped as the interceptor wasn't serialisable.
# Testing

### New tests

Updated the existing test to use a serializable scope and to make some serial tests to test this


MattGill98:
jenkins test please

MattGill98:
It looks like this change may break across deployment groups. I'm doing some testing to confirm

MattGill98:
This seems to work when deploying the rolesAllowed Payara Samples WAR to a deployment group and visiting each instance with the same session. I'm not currently how we run distributed tests in order to test this directly

MattGill98:
jenkins test please

PR #4593, opened 2020-03-27
CUSTCOM-40 Upgrade to Yasson 1.0.6

MeroRai:
# Description
This is a component upgrade of the Yasson to 1.0.6

# Testing

### Test suites executed

- Quicklook
- Build test

### Testing Environment

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -152,7 +152,7 @@
         <tyrus.version>1.15</tyrus.version>
         <jaxb-extra-osgi.version>2.3.0</jaxb-extra-osgi.version>
         <json.bind-api.version>1.0.2</json.bind-api.version>
-        <yasson.version>1.0.4.payara-p2</yasson.version>
+        <yasson.version>1.0.6</yasson.version>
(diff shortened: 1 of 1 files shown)


MeroRai:
This upgrade introduces a breaking change. The following code will throw a recursive reference exception which is intended/correct behaviour, it was working fine before due to a bug in the implementation: 

```
   @Override
    public void serialize(Entity1 instance, JsonGenerator generator, SerializationContext context) {
        generator.writeStartObject();
        context.serialize(instance.getClass().getName(), instance, generator);
        generator.writeEnd();
    }
```

There’s a custom serializer for Entity1, which asks the runtime to serialize Entity1 in the same context. The runtime inevitably calls the custom serializer again, with the same context, forming infinite recursion.

More info on how to implement a customer serializer can be found on [Jarkata JSON B](http://json-b.net/docs/user-guide.html#serializersdeserializers) user guide.

MeroRai:
jenkins test please

PR #4594, opened 2020-03-27
CUSTCOM-130 Better compatibility of docker images

dmatej:
# Description

* Fixes incompatibility with debian/ubuntu based JDK11 images
* Fixes warning about unsupported JVM option
* Adds features - JVM memory settings for Micro and Server-Full and JVM_ARGS for Micro
* All Dockerfiles use same commands for same things
* URL for gpg can be changed with -D parameter

# Important Info

# Testing

## JDK8
`mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd`

## JDK11

`mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd -Ddocker.payara.tag=5.202-SNAPSHOT-JDK11 -Ddocker.java.repository=adoptopenjdk/openjdk11 -Ddocker.java.tag=jdk-11.0.6_10-debian-slim`

## Payara4

1. Build some Payara4 versions including all required distribution packages
2. This will generate docker images
3. PayaraServerNodeTest will fail, because there is a difference in syntax, but other images should pass their respective tests

`mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd -Dpayara.version=4.1.2.191.14-SNAPSHOT -Ddocker.payara.tag=4.1.2.191.14-SNAPSHOT -Ddocker.payara.rootDirectoryName=payara41 -Ddocker.payara.domainName=payaradomain`

## Changing env variables, extending

```
docker create --name payaramicroxxx --env "MEM_XSS=2m" --env MEM_MAX_RAM_PERCENTAGE=25.0 payara/micro:latest
# replace dcb7e7328f90 with a result of docker create
docker start dcb7e7328f90
docker logs dcb7e7328f90
```

# Notes

* I need a tester with Mac and Windows, all tests were done on Linux, but I know that other systems have limitations on Docker and TestContainers.
* EDIT: passed on  Kubuntu, Windows 10 and Mac

### DIFF
--- a/appserver/extras/docker-images/basic/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/basic/src/main/docker/Dockerfile
@@ -1,0 +1,42 @@
+FROM @docker.java.image@
+
+ENV HOME_DIR=/opt/payara
+ENV PAYARA_DIR=${HOME_DIR}/appserver \
+    SCRIPT_DIR=${HOME_DIR}/scripts \
+    CONFIG_DIR=${HOME_DIR}/config \
+    DEPLOY_DIR=${HOME_DIR}/deployments \
+    PASSWORD_FILE=${HOME_DIR}/passwordFile \
+    ADMIN_USER=admin \
+    ADMIN_PASSWORD=admin \
+    JVM_ARGS="" \
+    MEM_MAX_RAM_PERCENTAGE="70.0" \
+    MEM_XSS="512k"
+ENV PATH="${PATH}:${PAYARA_DIR}/bin"
+
+ARG TINI_VERSION=v0.18.0
+
+# Download tini
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
+    https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /
+
+RUN true \
+    && apt-get update \
+    && apt-get install -y gpg \
+    && rm -rf /var/lib/apt/lists/* \
+    && mkdir -p ${HOME_DIR} \
+    && addgroup --gid 1000 payara \
+    && adduser --system --uid 1000 --no-create-home --shell /bin/bash --home "${HOME_DIR}" --gecos "" --ingroup payara payara \
+    && echo payara:payara | chpasswd \
+    && mkdir -p ${PAYARA_DIR} \
+    && mkdir -p ${DEPLOY_DIR} \
+    && mkdir -p ${CONFIG_DIR} \
+    && mkdir -p ${SCRIPT_DIR} \
+    && chown -R payara:payara ${HOME_DIR} \
+    # Verify tini
+    && gpg --verbose --keyserver @docker.keyserver.url@ --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
+    && gpg --verify /tini.asc \
+    && chmod +x /tini \
+    && true
+
+USER payara
+WORKDIR ${HOME_DIR}
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -29,4 +29,5 @@ WORKDIR ${HOME_DIR}
 COPY --chown=payara:payara maven/artifacts/payara-micro.jar .
 COPY --chown=payara:payara maven/bin/* ${SCRIPT_DIR}/
 
-ENTRYPOINT "${SCRIPT_DIR}/entrypoint.sh"
+ENTRYPOINT ["/bin/sh", "entrypoint.sh"]
+CMD ["--deploymentDir","/opt/payara/deployments"]
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -26,8 +26,8 @@ RUN true \
 USER payara
 WORKDIR ${HOME_DIR}
 
-COPY --chown=payara:payara maven/artifacts/payara-micro.jar .
 COPY --chown=payara:payara maven/bin/* ${SCRIPT_DIR}/
+COPY --chown=payara:payara maven/artifacts/payara-micro.jar .
 
 ENTRYPOINT ["/bin/sh", "entrypoint.sh"]
 CMD ["--deploymentDir","/opt/payara/deployments"]
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -8,6 +8,7 @@ EXPOSE 6900 8080
 ENV PAYARA_HOME=/opt/payara
 ENV HOME_DIR=${PAYARA_HOME}
 ENV PAYARA_DIR=${HOME_DIR} \
+    SCRIPT_DIR=${HOME_DIR} \
     DEPLOY_DIR=/opt/payara/deployments \
     JVM_ARGS="" \
     MEM_MAX_RAM_PERCENTAGE=70.0 \
@@ -25,8 +26,7 @@ RUN true \
 USER payara
 WORKDIR ${HOME_DIR}
 
-COPY maven/artifacts/payara-micro.jar .
+COPY --chown=payara:payara maven/artifacts/payara-micro.jar .
+COPY --chown=payara:payara maven/bin/* ${SCRIPT_DIR}/
 
-# Default command to run
-ENTRYPOINT ["sh", "-c", "java -XX:MaxRAMPercentage=${MEM_MAX_RAM_PERCENTAGE} -Xss${MEM_XSS} -XX:+UseContainerSupport ${JVM_ARGS} -jar payara-micro.jar"]
-CMD ["--deploymentDir", "/opt/payara/deployments"]
+ENTRYPOINT "${SCRIPT_DIR}/entrypoint.sh"
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -9,8 +9,8 @@ ENV PAYARA_HOME=/opt/payara\
 
 # Create and set the Payara user and working directory owned by the new user
 RUN true \
-    && addgroup payara \
-    && adduser -D -h ${PAYARA_HOME} -H -s /bin/bash payara -G payara \
+    && groupadd --gid 1000 payara \
+    && useradd -u 1000 -M -s /bin/bash -d ${PAYARA_HOME} payara -g payara \
     && echo payara:payara | chpasswd \
     && mkdir -p ${DEPLOY_DIR}  \
     && chown -R payara:payara ${PAYARA_HOME} \
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -4,7 +4,9 @@ FROM @docker.java.image@
 EXPOSE 6900 8080
 
 # Configure environment variables
-ENV HOME_DIR=/opt/payara
+# PAYARA_HOME is deprecated - it is here for backward compatibility
+ENV PAYARA_HOME=/opt/payara
+ENV HOME_DIR=${PAYARA_HOME}
 ENV PAYARA_DIR=${HOME_DIR} \
     DEPLOY_DIR=/opt/payara/deployments
 
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -8,7 +8,10 @@ EXPOSE 6900 8080
 ENV PAYARA_HOME=/opt/payara
 ENV HOME_DIR=${PAYARA_HOME}
 ENV PAYARA_DIR=${HOME_DIR} \
-    DEPLOY_DIR=/opt/payara/deployments
+    DEPLOY_DIR=/opt/payara/deployments \
+    JVM_ARGS="" \
+    MEM_MAX_RAM_PERCENTAGE=70.0 \
+    MEM_XSS=512k
 
 # Create and set the Payara user and working directory owned by the new user
 RUN true \
@@ -25,5 +28,5 @@ WORKDIR ${HOME_DIR}
 COPY maven/artifacts/payara-micro.jar .
 
 # Default command to run
-ENTRYPOINT ["java", "-XX:+UseContainerSupport", "-XX:MaxRAMPercentage=90.0", "-jar", "payara-micro.jar"]
+ENTRYPOINT ["sh", "-c", "java -XX:MaxRAMPercentage=${MEM_MAX_RAM_PERCENTAGE} -Xss${MEM_XSS} -XX:+UseContainerSupport ${JVM_ARGS} -jar payara-micro.jar"]
 CMD ["--deploymentDir", "/opt/payara/deployments"]
--- a/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
+++ b/appserver/extras/docker-images/micro/src/main/docker/Dockerfile
@@ -4,20 +4,21 @@ FROM @docker.java.image@
 EXPOSE 6900 8080
 
 # Configure environment variables
-ENV PAYARA_HOME=/opt/payara\
+ENV HOME_DIR=/opt/payara
+ENV PAYARA_DIR=${HOME_DIR} \
     DEPLOY_DIR=/opt/payara/deployments
 
 # Create and set the Payara user and working directory owned by the new user
 RUN true \
-    && groupadd --gid 1000 payara \
-    && useradd -u 1000 -M -s /bin/bash -d ${PAYARA_HOME} payara -g payara \
+    && addgroup --gid 1000 payara \
+    && adduser --system --uid 1000 --no-create-home --shell /bin/bash --home "${HOME_DIR}" --gecos "" --ingroup payara payara \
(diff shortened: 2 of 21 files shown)


appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
```diff
@@ -29,25 +29,25 @@ ENV PATH="${PATH}:${PAYARA_DIR}/bin"
 ARG TINI_VERSION=v0.18.0
 
 # Download tini
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
-    https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc /tini.asc
```
pdudits: Unless there is special reason for that, we should keep single add and thus single intermediate image.

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
```diff
@@ -4,20 +4,21 @@ FROM @docker.java.image@
 EXPOSE 6900 8080
 
 # Configure environment variables
-ENV PAYARA_HOME=/opt/payara\
+ENV HOME_DIR=/opt/payara
```
dmatej: As in other Dockerfiles, despite it uses same directory here

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
```diff
@@ -4,20 +4,21 @@ FROM @docker.java.image@
...
+ENV PAYARA_DIR=${HOME_DIR} \
     DEPLOY_DIR=/opt/payara/deployments
 
 # Create and set the Payara user and working directory owned by the new user
 RUN true \
-    && addgroup payara \
-    && adduser -D -h ${PAYARA_HOME} -H -s /bin/bash payara -G payara \
+    && addgroup --gid 1000 payara \
```
dmatej: Evil block - this works in alpine, debian and ubuntu

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
```diff
@@ -29,25 +29,25 @@ ENV PATH="${PATH}:${PAYARA_DIR}/bin"
...
+    && adduser --system --uid 1000 --no-create-home --shell /bin/bash --home "${HOME_DIR}" --gecos "" --ingroup payara payara \
     && echo payara:payara | chpasswd \
     && mkdir -p ${DEPLOY_DIR} \
     && mkdir -p ${CONFIG_DIR} \
     && mkdir -p ${SCRIPT_DIR} \
     && chown -R payara: ${HOME_DIR} \
     # Verify tini
-    && gpg --batch --keyserver "hkp://ipv4.pool.sks-keyservers.net" --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
-    && gpg --batch --verify /tini.asc /tini \
+    && gpg --verbose --keyserver @docker.keyserver.url@ --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
```
pdudits: Since this is source of majority of build failures, I'd propose specific exit code (until we implement different issue for taking this away completely from main build). 
Like `(gpg --recvkeys... || exit 55)`

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
rdebusscher: The change of the ENV name will affect all users who extend our images and possible require them to adjust their dockerfiles.

dmatej: In this release is much more breaking changes in docker files, so I think it is a good oportunity to synchronize behavior of all images to same principles and provide documentation about how things work.
But then it would be good to use also same directory structure in all three Dockerfiles.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: We already discussed it - exit codes are not much helpful when you don't remember what they mean. So until somebody would provide such code (for cycle?), 
`--verbose` + `-Ddocker.verbose` is more helpful as it directly prints what it tried to do and why it failed.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: We use --no-cache, so layers are not cached and used.
And also docker does not mention this as allowed in documentation, I found only a merged PR where it was implemented somehow. So I would rather avoid it.
The RUN command is different case, it is in all examples and each execution can make many changes.
https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image
https://github.com/fabric8io/docker-maven-plugin/issues/348

dmatej:
> The memory changes is part of QACI-40

There is not PR, memory changes are not even mentioned in description and finally - Patrik already did it in micro here. So I only updated the full server dockerfile with same settings.
Also I think original fraction value 1 was a bad idea, because there is also off-heap memory ... I doubt even about the 90% value ... but we need performance tests for this.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
pdudits: https://docs.docker.com/engine/reference/builder/#add

`ADD [--chown=<user>:<group>] <src>... <dest>`

Note the ellipsis.

dmatej: OK, you win. I don't know why my brain ignored it :D

rdebusscher:
Use an environment variable  to define the MaxRamPercentage as suggested here https://github.com/payara/docker-payaraserver-full/pull/118/files

dmatej:
> Use an environment variable to define the MaxRamPercentage as suggested here https://github.com/payara/docker-payaraserver-full/pull/118/files

This should go to standalone issue and PR, here I only "synchronized" it between containers. At this moment this PR is useful for both Patrik's and my work, other changes should have own issue and also testing. Also I don't think this option is sufficient, there are at least other three extremely important (ie -XX:MaxDirectMemorySize). 
Again, this is for another discussion and testing.

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
dmatej: Seems that other changes are not so much breaking as I thought, so I brought  the PAYARA_HOME back to life here.

poikilotherm:
As the author of the mentioned PR: good to know development moved here. What do you want me to do? This stuff is crucial IMHO

dmatej:
> As the author of the mentioned PR: good to know development moved here. What do you want me to do? This stuff is crucial IMHO

Probably only wait, we did more changes and at this moment we have some internal discussion about it ;)

I expect we will push and merge the final version soon, very soon ;)

poikilotherm:
Great. I'm really eager to release Dataverse images based on Payara 5 :-)

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
```diff
@@ -4,23 +4,29 @@ FROM @docker.java.image@
...
 
 USER payara
-WORKDIR ${PAYARA_HOME}
+WORKDIR ${HOME_DIR}
 
 COPY maven/artifacts/payara-micro.jar .
 
 # Default command to run
-ENTRYPOINT ["java", "-XX:+UseContainerSupport", "-XX:MaxRAMPercentage=90.0", "-jar", "payara-micro.jar"]
+ENTRYPOINT ["sh", "-c", "java -XX:MaxRAMPercentage=${MEM_MAX_RAM_PERCENTAGE} -Xss${MEM_XSS} -XX:+UseContainerSupport ${JVM_ARGS} -jar payara-micro.jar"]
```
poikilotherm, line 31: Just my 2 cents: I added `tini` in the full server image because any shell is not a proper init daemon. You might argue that Docker can use a tini-wrapper, but IMHO it should be done within the image, as there are more container and orchestration environments, which don't wrap it.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
```diff
@@ -35,26 +37,26 @@ ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
...
-    && groupadd --gid 1000 payara \
-    && useradd -u 1000 -M -s /bin/bash -d ${HOME_DIR} payara -g payara \
+    && addgroup --gid 1000 payara \
+    && adduser --system --uid 1000 --no-create-home --shell /bin/bash --home "${HOME_DIR}" --gecos "" --ingroup payara payara \
     && echo payara:payara | chpasswd \
     && mkdir -p ${DEPLOY_DIR} \
     && mkdir -p ${CONFIG_DIR} \
     && mkdir -p ${SCRIPT_DIR} \
     && chown -R payara: ${HOME_DIR} \
     # Verify tini
```
poikilotherm, line 50: If you guys don't want gpg et al in the resulting image, this could as well be done in a multistage build step.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
```diff
@@ -66,14 +68,13 @@ RUN true \
...
-       $(${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} list-jvm-options | grep "Xm[sx]"); \
+       $(${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} list-jvm-options | grep "Xm[sx]\|Xss"); \
        do\
          ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} delete-jvm-options $MEMORY_JVM_OPTION;\
        done \
-    # FIXME: when upgrading this container to Java 10+, this needs to be changed to
-    #        '-XX:+UseContainerSupport' and '-XX:MaxRAMPercentage'
     && ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} create-jvm-options \
-      '-XX\:+UnlockExperimentalVMOptions:-XX\:+IgnoreUnrecognizedVMOptions:-XX\:+UseCGroupMemoryLimitForHeap:-XX\:MaxRAMFraction=1' \
+      '-XX\:+UnlockExperimentalVMOptions:-XX\:+IgnoreUnrecognizedVMOptions:-XX\:+UseCGroupMemoryLimitForHeap:-XX\:+UseContainerSupport:-XX\:MaxRAMPercentage=${ENV=MEM_MAX_RAM_PERCENTAGE}:-Xss${ENV=MEM_XSS}' \
```
poikilotherm, line 76: `-XX\:+UseCGroupMemoryLimitForHeap` is deprecated since Java 10, it has been superseded by `-XX\:+UseContainerSupport`. IMHO it should be removed, together with `-XX\:+UnlockExperimentalVMOptions`.

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
dmatej: We use signed tini in the server-full and I have to admit I don't like it from two reasons:

1. gpg used to sign tini crashes builds from time to time and makes them much slower.
2. yesterday I found that it ignores system signals, I'm not sure if it is caused by tini usage or not. But I think we should target these things in standalone issue and stop extending this, because this was originally targeting only better usability and compatibility.

The shell here is only because environment options are not resolved without it.

Generally I agree that all docker files should use same principles only customized to their own usage.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: Caches are disabled in this build, but this leads me to a question - is it possible to tell docker to use cache only for some stage?
Enabled cache lead few versions back to the state where docker image of 5.194 contained 5.193 inside it.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: You're right, just to document it: https://bugs.openjdk.java.net/browse/JDK-8196595

dmatej:
Jenkins test please

rdebusscher:
The Image for Payara Micro looks good and Application deployment is ok again.

But the `PayaraMicroTest`fails

>ERROR [main] docker[payara/micro:5.202-SNAPSHOT]:472 - Could not start container
com.github.dockerjava.api.exception.InternalServerErrorException: Mounts denied: 
path /var/folders/2p/g987c05s2f1d8tw5pzhr4tt00000gn/T/TestServlet7784738673461879995.war

dmatej:
Nevermind, I used temporary file, and it seems on your computer it cannot be mapped into the container. I will change it to use the target folder instead, I hope it will be enough.

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
dmatej: Now it should be better - I created separate module for the shared basic image. Only the micro has it's own, because it is based on alpine.

dmatej:
I need someone to verify that everything works as expected as I described in the description in other than Linux systems: Mac and Windows.
Currently I believe that @rdebusscher's problem is now fixed, but to believe is not enough.
In last PR I also made build much faster - now three of four images use the same precreated image, only micro is alpine-based to be lean as possible.

dmatej:
Jenkins test please

appserver/extras/docker-images/micro/src/main/docker/Dockerfile
```diff
@@ -4,23 +4,29 @@ FROM @docker.java.image@
...
-    && chown -R payara:payara ${PAYARA_HOME} \
+    && chown -R payara:payara ${HOME_DIR} \
     && true
 
 USER payara
-WORKDIR ${PAYARA_HOME}
+WORKDIR ${HOME_DIR}
 
-COPY maven/artifacts/payara-micro.jar .
+COPY --chown=payara:payara maven/artifacts/payara-micro.jar .
```
rdebusscher, line 29: Revert order of these 2 COPY commands as script content will change less likely as Payara Micro. -> reuse of layers

appserver/extras/docker-images/server-full/src/main/docker/Dockerfile
```diff
@@ -7,77 +7,31 @@ FROM @docker.java.image@
...
-    && gpg --batch --keyserver "hkp://ipv4.pool.sks-keyservers.net" --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 \
-    && gpg --batch --verify /tini.asc /tini \
-    && chmod +x /tini \
-    && true
-
-USER payara
-WORKDIR ${HOME_DIR}
-
-COPY --chown=payara:payara maven/artifacts/payara5 ${PAYARA_DIR}/
+COPY --chown=payara:payara maven/artifacts/@docker.payara.rootDirectoryName@ ${PAYARA_DIR}/
```
rdebusscher, line 16: Revert order of these 2 COPY commands as script content will change less likely as Payara Server. -> reuse of layers

appserver/extras/docker-images/server-node/src/main/docker/Dockerfile
```diff
@@ -14,20 +11,10 @@ ENV PAYARA_PASSWORD_FILE=${PAYARA_PASSWORD_FILE_DIR}/passwordfile.txt
...
-    && groupadd -g 1000 payara \
-    && useradd -u 1000 -M -s /bin/bash -d ${HOME_DIR} payara -g payara \
-    && echo payara:payara | chpasswd \
-    && chown -R payara:payara ${HOME_DIR} \
+    && mkdir -p ${PAYARA_PASSWORD_FILE_DIR} \
     && true
 
-USER payara
-WORKDIR ${HOME_DIR}
+COPY --chown=payara:payara maven/artifacts/@docker.payara.rootDirectoryName@ ${PAYARA_DIR}
```
rdebusscher, line 17: Revert order of these 2 COPY commands as script content will change less likely as Payara Micro. -> reuse of Server

pdudits:
The necessary changes are in https://github.com/dmatej/Payara/pull/3

dmatej:
Jenkins test please

dmatej:
Jenkins test please

appserver/extras/docker-images/basic/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    DEPLOY_DIR=${HOME_DIR}/deployments \
+    PASSWORD_FILE=${HOME_DIR}/passwordFile \
+    ADMIN_USER=admin \
+    ADMIN_PASSWORD=admin \
+    JVM_ARGS="" \
+    MEM_MAX_RAM_PERCENTAGE="70.0" \
+    MEM_XSS="512k"
+ENV PATH="${PATH}:${PAYARA_DIR}/bin"
+
+ARG TINI_VERSION=v0.18.0
```
poikilotherm, line 16: Heads up that tini v0.19 has been released two days ago.
Releases now offer SHA256 sums, thus easing installation A LOT!
https://github.com/krallin/tini/releases/tag/v0.19.0

dmatej: I think we will merge this first and then upgrade this in standalone PR, but thanks for the good news!

poikilotherm:
Congrats getting this huge PR merged :tada: :partying_face: 
Looking forward to build awesome stuff with it!

PR #4597, opened 2020-03-30
CUSTCOM-247 Propagates realm name from WAR module to its application wrapper 

OndroMih:
# Description

This fix allows using a custom realm defined in web.xml in a WAR application for authenticating SOAP requests using the WS policy mechanism. 

Without the fix, the default realm is always used even if a custom realm is defined in the WAR file.

This fixes a customer issue.

### Testing Performed
Manual testing on Payara Server 5.191.7 (Java 8) and 5.201 (Java 11), on Ubuntu Linux.


OndroMih:
> Needs copyright header updating

Hi @MattGill98 , is what I did enough? I only changed the year from 2019 to 2020 for Payara copyright. Am I still missing something?

MattGill98:
jenkins test please

PR #4598, opened 2020-03-30
CUSTCOM-234 JVM options REST management test

MattGill98:
Added a test for the JVM options REST management endpoint, to make sure
that the endpoint continues to work as intended.


MattGill98:
jenkins test please

PR #4599, opened 2020-03-31
CUSTCOM-205 On ZuluJDK 8.44, accessing HTTPS throws NoSuchMethodError exception 

MeroRai:
# Description
This is a bug fix to fix an issue on ZuluJDK 8.44, where accessing HTTPS throws NoSuchMethodError exception.  

# Important Info
To test this issue edit the Grizzly NPN version of the following JVM options: `-Xbootclasspath/p:${com.sun.aas.installRoot}/lib/grizzly-npn-bootstrap-1.8.1.jar` to `1.9.payara-p1` and run https://github.com/summerwind/h2spec with h2spec -t -k -p 8181 on Zulu 8.44. After a request has been made check the log file to see if the NoSuchMethodError exception has been thrown. 

### Dependant PRs 
https://github.com/payara/Payara_PatchedProjects/pull/297
https://github.com/payara/patched-src-grizzly-npn/pull/2

# Testing

### Test suites executed
- Build test
- Quicklook

### Testing Environment

# Documentation
TODO

### DIFF
--- a/nucleus/packager/nucleus-grizzly/pom.xml
+++ b/nucleus/packager/nucleus-grizzly/pom.xml
@@ -103,6 +103,11 @@
                                     <destFileName>grizzly-npn-bootstrap.jar</destFileName>
                                 </artifactItem>
                                 <!-- Other versioned NPN Bootstrap Jars referenced in new domain.xml (since 5.184) -->
+                                  <artifactItem>
+                                    <groupId>org.glassfish.grizzly</groupId>
+                                    <artifactId>grizzly-npn-bootstrap</artifactId>
+                                    <version>1.9.payara-p1</version>
+                                </artifactItem>
                                 <artifactItem>
                                     <groupId>org.glassfish.grizzly</groupId>
                                     <artifactId>grizzly-npn-bootstrap</artifactId>


nucleus/packager/nucleus-grizzly/pom.xml
```diff
@@ -103,6 +103,11 @@
                                     <destFileName>grizzly-npn-bootstrap.jar</destFileName>
                                 </artifactItem>
                                 <!-- Other versioned NPN Bootstrap Jars referenced in new domain.xml (since 5.184) -->
+                                  <artifactItem>
```
MattGill98, line 106: Is this JAR actually activated anywhere?

nucleus/packager/nucleus-grizzly/pom.xml
```diff
@@ -103,6 +103,11 @@
...
+                                    <groupId>org.glassfish.grizzly</groupId>
+                                    <artifactId>grizzly-npn-bootstrap</artifactId>
+                                    <version>1.9.payara-p1</version>
```
smillidge, line 109: Are grizzly npn changes being pushed back upstream to Grizzly. Ideally we would patch here then get it into a Grizzly release ASAP

OndroMih: AFAIK, the latest version of Grizzly NPN contains support for the latest version of OpenJDK, so it's equal to our latest version of Grizzly. However, Grizzly NPN is designed to support only a single version of the JDK security classes so this fix would only be accepted after OpenJDK is updated with the same changes as the latest Zulu version. Azul claim their going to push this change to OpenJDK in the following months.

nucleus/packager/nucleus-grizzly/pom.xml
OndroMih: We agreed with @MeroRai that the JAR file won't be activated automatically and will only be available if users want to use the latest Zulu JDK. They would have to modify JVM options to use this version of the NPN library. This should be documented in the docs.

dmatej:
Jenkins test please

PR #4602, opened 2020-04-02
APPSERV-114 Addresses possible sources of Race Conditions in InvocationManager

jbee:
### Background
Running FT TCK strongly suggested that a race condition in or around the `InvocationManger` leads to inconsistent context when trying to derive the _current_ application like shown below:
```java
 ComponentInvocation current = invocationManager.getCurrentInvocation();
 if (current == null) {
     return invocationManager.peekAppEnvironment().getName();
 }
 String appName = current.getAppName();
 if (appName == null) {
     appName = current.getModuleName();
 }
 if (appName == null) {
     appName = current.getComponentId();
 }
return appName;
```
The issue observed was that different application names were returned in a non deterministic manner for the actual same application processing tests.

### Summary
The most likely cause is related to the frames that push and pop the stack which is peeked by `getCurrentInvocation()`. But it could also be related to application environment stack peeked by `peekAppEnvironment()` for case that not `ComponentInvocation` was on the stack.
The issue can either be within the implementation of the `InvocationMangerImpl` but also on the outside as the  `ComponentInvocation` is essentially managed from the outside by calling `preInvoke` and `postInvoke`.

Changes:
* removes the unnecessary generic of the `InvocationArray` / `InvocationFrames` (renamed) inner class
* replaces use of `Stack` with `ConcurrentLinkedDeque` (`synchronized` vs. CAS based synchronisation)
* replaces `ArrayList` as bases of frames stack with `ConcurrentLinkedDeque` (as this is only used via thread local non synchronised should not be an issue but to be sure and to provide better readability of the base class was changed)
* replaces `ComponentInvocationHandler[]` handlers with a single `ComponentInvocationHandler` that is initialised with `ListComponentInvocationHandler` in case multiple handlers are actually needed. The underlying `List` uses `ArrayList` as it is read-only after being initialised. This allows to simplify the rest of the implementation that only has to assume a single handler.
* uses `ListComponentInvocationHandler` to manage `RegisteredComponentInvocationHandler`. In this case the underlying `List` uses a `CopyOnWriteArrayList` as multiple threads might register handlers concurrently. 
* replaces the `Map` for `RegisteredComponentInvocationHandler` (that are per type) with a thread-safe `ConcurrentHashMap` as multiple threads can register handlers concurrently.
* adds `try-finally` to `preInvoke` to make sure that any exception thrown by handlers does not prevent adding of the invocation instance (I'd say this is the best candidate for the observed issue)

### Testing
The implementation of `InvocationMangerImpl` got covered with unit tests to a coverage > 95%.
This mostly intends to make sure the changes in the implementation do not cause unexpected behaviour, in particular throwing exceptions when it should not. The tests do include a few tests that include multiple threads to verify the thread parent-child stack inheritance behaviour, not to show that the class is thread-safe. To allow better reasoning about thread safety the implementation was cleaned and simplified so the use of collection becomes more clear. 

#### Testing Performed
The unit tests were added mostly to make sure changes did not introduce new bugs like NPEs and alike and that the implementation does behave as expected from reading the code.

In addition I ran the FT TCK as that had shown issues before and since it deploying lots of applications.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
@@ -60,6 +60,7 @@ import org.glassfish.api.StartupRunLevel;
 import org.glassfish.api.admin.ServerEnvironment;
 import org.glassfish.api.event.EventListener;
 import org.glassfish.api.event.Events;
+import org.glassfish.api.invocation.ComponentInvocation;
 import org.glassfish.api.invocation.InvocationManager;
 import org.glassfish.hk2.api.ServiceLocator;
 import org.glassfish.hk2.runlevel.RunLevel;
@@ -418,15 +419,16 @@ public class MetricsService implements EventListener, ConfigListener, Monitoring
     public String getApplicationName() {
         InvocationManager invocationManager = Globals.getDefaultBaseServiceLocator()
                 .getService(InvocationManager.class);
-        if (invocationManager.getCurrentInvocation() == null) {
+        ComponentInvocation current = invocationManager.getCurrentInvocation();
+        if (current == null) {
             return invocationManager.peekAppEnvironment().getName();
         }
-        String appName = invocationManager.getCurrentInvocation().getAppName();
+        String appName = current.getAppName();
         if (appName == null) {
-            appName = invocationManager.getCurrentInvocation().getModuleName();
+            appName = current.getModuleName();
         }
         if (appName == null) {
-            appName = invocationManager.getCurrentInvocation().getComponentId();
+            appName = current.getComponentId();
         }
         return appName;
     }
--- a/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
+++ b/nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2020] [Payara Foundation and/or its affiliates]
 package org.glassfish.api.invocation;
 
 import static java.lang.ThreadLocal.withInitial;
@@ -50,11 +50,11 @@ import java.lang.reflect.Method;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Deque;
-import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Stack;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -72,14 +72,13 @@ public class InvocationManagerImpl implements InvocationManager {
     // The ArrayList contains ComponentInvocation objects which represent
     // the stack of invocations on this thread. Accesses to the ArrayList
     // don't need to be synchronized because each thread has its own ArrayList.
-    private InheritableThreadLocal<InvocationArray<ComponentInvocation>> frames;
+    private final InheritableThreadLocal<InvocationArray<ComponentInvocation>> frames;
 
     private final ThreadLocal<Stack<ApplicationEnvironment>> applicationEnvironments = withInitial(Stack<ApplicationEnvironment>::new);
     private final ThreadLocal<Deque<Method>> webServiceMethods = withInitial(ArrayDeque<Method>::new);
 
-    private Map<ComponentInvocationType, List<RegisteredComponentInvocationHandler>> regCompInvHandlerMap = new HashMap<>();
-
-    private final ComponentInvocationHandler[] invHandlers;
+    private final ConcurrentMap<ComponentInvocationType, ListComponentInvocationHandler> singleTypeHandlers = new ConcurrentHashMap<>();
+    private final ComponentInvocationHandler multiTypeHandler;
 
     public InvocationManagerImpl() {
         this(null);
@@ -87,20 +86,7 @@ public class InvocationManagerImpl implements InvocationManager {
 
     @Inject
     private InvocationManagerImpl(@Optional IterableProvider<ComponentInvocationHandler> handlers) {
-        if (handlers == null) {
-            invHandlers = null;
-        } else {
-            LinkedList<ComponentInvocationHandler> localHandlers = new LinkedList<ComponentInvocationHandler>();
-            for (ComponentInvocationHandler handler : handlers) {
-                localHandlers.add(handler);
-            }
-
-            if (localHandlers.size() > 0) {
-                invHandlers = localHandlers.toArray(new ComponentInvocationHandler[localHandlers.size()]);
-            } else {
-                invHandlers = null;
-            }
-        }
+        this.multiTypeHandler = initInvocationHandlers(handlers);
 
         frames = new InheritableThreadLocal<InvocationArray<ComponentInvocation>>() {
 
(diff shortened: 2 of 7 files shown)


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
```diff
@@ -418,15 +419,16 @@ private void deregisterApplication(String applicationName) {
     public String getApplicationName() {
         InvocationManager invocationManager = Globals.getDefaultBaseServiceLocator()
                 .getService(InvocationManager.class);
-        if (invocationManager.getCurrentInvocation() == null) {
```
jbee, line 421: NB: while the use of thread local should ensure that multiple calls to `getCurrentInvocation()` should return the same instance it is better to not take changes and do it once. Also improves readability.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
...
-    // This TLS variable stores an ArrayList.
-    // The ArrayList contains ComponentInvocation objects which represent
-    // the stack of invocations on this thread. Accesses to the ArrayList
-    // don't need to be synchronized because each thread has its own ArrayList.
-    private InheritableThreadLocal<InvocationArray<ComponentInvocation>> frames;
-
-    private final ThreadLocal<Stack<ApplicationEnvironment>> applicationEnvironments = withInitial(Stack<ApplicationEnvironment>::new);
-    private final ThreadLocal<Deque<Method>> webServiceMethods = withInitial(ArrayDeque<Method>::new);
-
-    private Map<ComponentInvocationType, List<RegisteredComponentInvocationHandler>> regCompInvHandlerMap = new HashMap<>();
```
jbee, line 80: NB. Using `HashMap` here is definitely not thread-safe but should be. But is is hard to see how this can cause the observed issues. However, we do talk about multi-threading so everything is possible :D

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
 @Singleton
 public class InvocationManagerImpl implements InvocationManager {
 
-    // This TLS variable stores an ArrayList.
-    // The ArrayList contains ComponentInvocation objects which represent
-    // the stack of invocations on this thread. Accesses to the ArrayList
-    // don't need to be synchronized because each thread has its own ArrayList.
-    private InheritableThreadLocal<InvocationArray<ComponentInvocation>> frames;
-
-    private final ThreadLocal<Stack<ApplicationEnvironment>> applicationEnvironments = withInitial(Stack<ApplicationEnvironment>::new);
```
jbee, line 77: NB. Replacing `Stack` here is more of a modernisation to avoid `sychronized` in favour of CAS based synchronisation.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
...
-        // If ejb call EJBSecurityManager, for servlet call RealmAdapter
-        ComponentInvocationType invocationType = invocation.getInvocationType();
+        ComponentInvocation prev = frames.peekLast();
+        ComponentInvocationType type = invocation.getInvocationType();
+        ComponentInvocationHandler typeHandler = typeHandlers.get(type);
 
-        if (invHandlers != null) {
-            for (ComponentInvocationHandler handler : invHandlers) {
-                handler.beforePreInvoke(invocationType, previousInvocation, invocation);
+        try {
```
jbee, line 143: NB. Missing `try-finally` here to make sure the `frames.addLast(invocation);` will definitely happen is my best guess on what might have caused the issue observed,

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
...
             }
-
         } finally {
             // pop the stack
-            invocationArray.remove(beforeSize - 1);
+            frames.removeLast();
 
-            if (invHandlers != null) {
-                for (ComponentInvocationHandler handler : invHandlers) {
-                    handler.afterPostInvoke(invocation.getInvocationType(), prevInv, invocation);
```
jbee, line 206: NB. You might have noticed that `invocation` is passed to `afterPostInvoke` instead of `curInv`. I believe this is just a inconsistency in the chosen way to express the logic. At this point `curInv` is (or really should be) `invocation`. To avoid any future confusion I unified it to use `invocation` everywhere as it is done in `preInvoke` and as I think it makes most sense as this is what is passed to the method as "current". It is unfortunate how the API was designed and that it leaves room for inconsistencies like calling `postInvoke` with a different value then `preInvoke`. `postInvoke` really should not have had an argument as the current is that pushed to the stack when invoking `preInvoke`. Or the API should have made clear that the argument is only passed to `postInvoke` to verify that the nesting is correct and that it indeed is the current top of the stack.

jbee:
jenkins test please

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
pdudits: I think it is worth a warning when this assumption doesn't hold and you actually have a good spot since you already jump over the invocation head.

This might disclose bugs we have around nested transactions or cleanups.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
...
             }
         }
 
-        return childInvocationArray;
+        return childFrames;
     }
 
 
-    static class InvocationArray<T extends ComponentInvocation> extends ArrayList<T> {
+    static final class InvocationFrames extends ConcurrentLinkedDeque<ComponentInvocation> {
```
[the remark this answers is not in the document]
jbee: > I'm also afraid that extending rather than delegating makes too many public methods available.

`InvocationFrames` is only used privately. 

Again, done this for peace of mind. Could go to `ArrayDequeue`.

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -68,347 +71,334 @@
...
-        if (setRegCompInvHandlers == null) {
-            setRegCompInvHandlers = new ArrayList<RegisteredComponentInvocationHandler>();
-            regCompInvHandlerMap.put(type, setRegCompInvHandlers);
-        }
-
-        if (setRegCompInvHandlers.size() == 0) {
-            setRegCompInvHandlers.add(handler);
-        }
+        typeHandlers.computeIfAbsent(type,
+                key -> new ListComponentInvocationHandler(new CopyOnWriteArrayList<>())) // OBS! must be thread safe List here
```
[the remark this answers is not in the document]
jbee: No, the `allTypesHandler` is composed once during construction and never changed so it does not need any form of synchronisation while the handlers per type can change while being used so they need it.

jbee:
jenkins test please

jbee:
@pdudits Now uses unsynchronised collections within thread locals. Also added the logging you suggested with uncovered 2 issues in the test setup. :tada:

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -173,7 +177,11 @@ public void setThreadInheritableInvocation(List<? extends ComponentInvocation> p
             throw new InvocationException();
         }
 
-        iter.next(); // the last is the current is "invocation"
+        ComponentInvocation current = iter.next(); // the last is the current is "invocation"
+        if (invocation != current) {
+            LOGGER.log(WARNING, "postInvoke not called with top of the invocation stack. Expected {0} but was: {1}",
+                    new Object[] { current, invocation });
```
pdudits, line 183: Last Nitpick would be, that only `EjbInvocation` provides reasonable `toString()`. It would be great to add one into `WebComponentInvocation` and `ComponentInvocation`.

jbee: Good point, I remember checking it myself but then forgot to do something about it. Too many context switches. Will look into it tomorrow.

jbee:
jenkins test please

PR #4600, opened 2020-04-01
CLOUD-59 Add programattic API for Payara Micro to do pre- and post-boot commands

Cousjava:
# Description
This is a  feature.

This PR adds abilities to add a callback that will execute asadmin commands during pre- or post- boot of Payara Micro.

CommandRunner and CommandResult having been added to the payara-api module.

# Testing

### New tests
Includes new test in payara-samples module

### Testing Performed
New programmatic test using comands

### Test suites executed
- Payara Samples
- Java EE8 Samples

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 18.04 with Maven 3.6.0

# Documentation

PR #4604, opened 2020-04-02
CLOUD-59: Programmatic API for preboot and postboot commands in Micro

pdudits:
# Description
This is a feature. 

Adds methods `setPreBootHandler` and `setPostBootHandler` that offer functionality similar to `--prebootcommandfile` and `--postbootcommandfile` of Payara Micro jar.

The methods accept Consumer of `CommandRunner`, allowing the user to also inspect the output of the commands, and direct their flow.

Two secondary things are addressed with this PR:
* 38a8641 adds flatten plugin to Payara Micro distribution. The installed POM does not contain any dependencies anymore, making it equal with release POM.
* 2abfc10 adds more diagnostic information to failure of Bulkhead tests, as I've experienced some during development.

# Important Info

### Dependant PRs 
Alternative to #4600 

# Testing

### New tests
`PayaraMicroCommandsTest` tests and demonstrated the API.

### Test suites executed

- Payara Samples

### Testing Environment

OpenJ9 1.8.0_242 on Windows 10, Maven 3.6.3

# Documentation
Javadoc comments were added, documentation update pending.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/AdminCommandRunner.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/AdminCommandRunner.java
@@ -42,6 +42,23 @@ package fish.payara.micro.boot;
 
 import fish.payara.micro.ClusterCommandResult;
 
+import java.util.function.Consumer;
+
+/**
+ * Payara Micro interface for running admin commands.
+ * Runner allows to use arbitrary commands as if they were run via {@code asadmin} utility.
+ * <p>While similar to {@code CommandRunner} from API of embedded server, these two runners are kept separate due
+ * to backward compatibility reasons.</p>
+ * @see PayaraMicroBoot#setPreBootHandler(Consumer)
+ * @see PayaraMicroBoot#setPostBootHandler(Consumer)
+ */
 public interface AdminCommandRunner {
+    /**
+     * Run an admin command.
+     * @param command the name of the command
+     * @param args optional arguments to the command. The arguments need to respect escaping rules of {@code asadmin} utility,
+     *             chiefly escaping colons (:) and equal signs (=) in property lists.
+     * @return
+     */
     ClusterCommandResult run(String command, String... args);
 }
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/AdminCommandRunner.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/AdminCommandRunner.java
@@ -1,0 +1,47 @@
+/*
+ *    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) [2020] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ *    The contents of this file are subject to the terms of either the GNU
+ *    General Public License Version 2 only ("GPL") or the Common Development
+ *    and Distribution License("CDDL") (collectively, the "License").  You
+ *    may not use this file except in compliance with the License.  You can
+ *    obtain a copy of the License at
+ *    https://github.com/payara/Payara/blob/master/LICENSE.txt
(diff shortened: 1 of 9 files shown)


pdudits:
Jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/CommandResultWrapper.java
```diff
@@ -0,0 +1,67 @@
...
+ *    only if the new code is made subject to such option by the copyright
+ *    holder.
+ */
+
+package fish.payara.micro.impl;
+
+import fish.payara.micro.ClusterCommandResult;
+import org.glassfish.embeddable.CommandResult;
+
+class CommandResultWrapper implements ClusterCommandResult {
```
jbee, line 46: I'd think of this as an _adapter_.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1077,6 +1056,13 @@ public PayaraMicroRuntime bootStrap() throws BootstrapException {
         }
     }
 
+    private void callHandler(Consumer<AdminCommandRunner> handler) throws GlassFishException {
+        CommandRunner runner = gf.getCommandRunner();
+        if (handler != null) {
+            handler.accept((cmd, args) -> new CommandResultWrapper(runner.run(cmd, args)));
```
jbee, line 1062: Don't know how the dependencies look like but it might be possible to have `CommandRunner` implement `AdminCommandRunner` and have `CommandResult` implement `ClusterCommandResult`.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/test/java/fish/payara/microprofile/faulttolerance/policy/AbstractBulkheadTest.java
```diff
@@ -134,7 +134,7 @@ public void delay(long delayMillis) throws InterruptedException {
     static class InOut {
         static final Object IN = new Object();
 
-        Object result;
+        volatile Object result;
```
jbee, line 137: As discussed I believe the issue here is that the async worker entering and exiting the bulkhead also sets the result in test code after having left the bulkhead. At that point the main thread can see all workers exited the bulkhead and starts to assert but there is no synchronisation forcing it to wait until the worker also has set its result. This part has to be outside of the bulkhead. So a proper fix is to introduce some kind of waiting in the main thread that keeps track of the workers and if they have set the result. This is by no means a request to do such a change in this PR but meant to document the issue. We should address it in a separate PR.

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/AdminCommandRunner.java
```diff
@@ -0,0 +1,47 @@
...
+ *    and therefore, elected the GPL Version 2 license, then the option applies
+ *    only if the new code is made subject to such option by the copyright
+ *    holder.
+ */
+
+package fish.payara.micro.boot;
+
+import fish.payara.micro.ClusterCommandResult;
+
+public interface AdminCommandRunner {
```
jbee, line 45: A javadoc on what this runner does or where its roll is compared to other runners and why this is needed would be nice.

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
pdudits: Unfortunately those are part of Embedded API, and we don't want embedded to be dependant on Micro, or vice versa. Also that embedded API is public API since Glassfish 3 and adding inheritance would break compatibility. Especially problematic one is `CommandResult.ExitStatus` which is an enum.

pdudits:
Jenkins test please

PR #4605, opened 2020-04-03
Increase FaultTolerance async max pool size from 1000 to 2000

jbee:
To avoid TCK runs failing due to pool rejecting execution the default is increased to 2000. The TCK at some point uses 1000 threads in a single test. Local tests with a max of 2000 has shown to avoid this issue.


jbee:
jenkins test please

PR #4606, opened 2020-04-03
MicroProfile FaultTolerance Unit Tests: Wait for caller threads to set their results

jbee:
### Summary
This is a fix for flaky unit tests due to a race condition in the test code. The tests could fail as the assert tries to compare a value that has not yet been set by a async caller thread. After the caller exists the bulkhead the main thread can observe that all callers did exit the bulkhead and continue to assert expectations on the result. The individual results of each caller however might not have been recorded in the test code at this point. This PR addresses this by always recording an entry in either the result map (no exception) or the error map (exception thrown) and waiting until either map has a key for the given thread before it compares the value with the expected one.

### Testing Performed
Running the unit tests of FT module.


jbee:
jenkins test please

pdudits:
Jenkins test please

jbee:
Jenkins test please

PR #4608, opened 2020-04-03
Cleanup Generics in MP Config

jbee:
My original intention was to add some caching to MP config since I became aware how often e.g. MP metrics is resolving values. I started by cleaning up warnings and then decided to stay there for the moment as I have to rethink if and how to best do caching given the semantics we have to deal with. 

This PR mostly adds missing generics and does some other details like using `final` and visibility modifiers.

**Edit**: Had an idea for cache that I add in another PR on top of this one.


jbee:
Jenkins test please

jbee:
Closing in favour of #4614

PR #4610, opened 2020-04-05
Fixes #4596: Honor non-standard realm name for client-cert login

### LINKED ISSUE #4596: Certificate realm not working with WebServices in payara 5.194
# Description #
----------

I'm upgrading a payara-server-full application from 5.182 to 5.194 
The application is deployed as an ear-file and it has one ejb-jar and two web-jars. 

The ejb-jar defines a webs ervice as follows. I leave out unrelated code. 

```
@WebService(...)
@Stateless
@DeclareRoles({"admin"})
@RolesAllowed({"admin"})
public class .... {
protected SessionContext context;
public xxxx() {
        boolean isAdmin = context.isCallerInRole("admin");
        String name = context.getCallerPrincipal().getName();
}
```

In glassfish-ejb-jar.xml the following is set: 

```
<security-role-mapping>
    <role-name>admin</role-name>
    <group-name>admin</group-name>
 </security-role-mapping>
  <ejb>
      <ejb-name>......</ejb-name>
      <webservice-endpoint>
        <port-component-name>.......</port-component-name>
        <endpoint-address-uri>.....</endpoint-address-uri>
        <login-config>
          <auth-method>CLIENT-CERT</auth-method>
          <realm>admin-cert</realm>
        </login-config>
        <transport-guarantee>CONFIDENTIAL</transport-guarantee>
      </webservice-endpoint>
  </ejb>

```
In the prior version of the app, which was based on 5.182, I could assign the group to the certificate realm by way of defining a custom jaas-context in domain.xml. This is broken since 5.184. I've submitted another related issue https://github.com/payara/Payara/issues/4533 recently. 

To amend this, I've put

```
@CertificateAuthenticationMechanismDefinition
@CertificateIdentityStoreDefinition(value = "admin-cert", assignGroups = "admins")
```

at a suitable place in one of the web-jars in order to assign groups to certificate login. This works in all other cases (other REST services) but not for the SOAP WebService above in the ejb-jar. 

If the `@RolesAllowed({"admin"})` above is present as shown, I get a "Failed permission check". If I comment it out, I get `false` for `isAdmin`, but the right name of the authenticated caller for `name`. 

To summarize, client cert authentication itself works for ejb web services, but group/role assigment is obviously broken. Callers are authenticated but not assigned the provided groups/roles. 

BTW: Setting 
`<property name="default-role-mapping" value="true">`
in glassfish-ejb-jar.xml doesn't make a difference. 

# Expected outcome #
----------

The expected outcome is that the caller must be authenticated and assigned the designated group/role.  `isAdmin` should be `true`. 

If this is a bug, is there a workaround? Is there a switch to turn off all the new authentication features since 5.184 for backward compatibility?

If it's not a bug, please let me know what I'm doing wrong.

borisheithecker:
# Description
This is a bug fix for https://github.com/payara/Payara/issues/4596

The standard realm name for client certificate login is "certificate". A non standard name must not be ignored when calling https://github.com/payara/Payara/blob/master/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/WebAndEjbToJaasBridge.java

Here is how to reproduce the bug https://github.com/borisheithecker/payara-clientcert-test2

# Testing

### Testing Performed
No testing yet. 

### Test suites executed
No tests executed.


Pandrex247 (approved):

MeroRai:
jenkins test please

dmatej:
Default realm name "certificate" stopped working after this. I have to add realm name element to the web.xml.

borisheithecker:
Is there a test case for the servlet failures in FISH-867? How can I reproduce it?

MattGill98:
Hi @borisheithecker,

I can't find an exact message for the failed TCK, but I do remember it was a 401 on a couple of Servlet TCK tests. I'm sorry I can't provide more information! You may find this repository useful if you want to reproduce the error: https://github.com/payara/jakartaeetck-runner

PR #4615, opened 2020-04-06
QACI-158 Fixed windows line endings in all bat files

dmatej:
# Description

Fixes problems with changing branches, cherrypicks and rebases


dmatej:
Jenkins test please

MarkWareham:
A find on Payara directrory yields 28 results. Why aren't the other 15 in here?

dmatej:
> Disputed that all .bats have been done. See conversation

I don't see any problem in other than these, so they don't need any change.
Did you find any bat file with incorrect line endings? What command did you use to find it?

dmatej:
I made it a global rule and fixed remaining bat files with this command:
`find . -type f -name "*.bat" -print0 | xargs -0 unix2dos`
I did not change .gitattributes files updated by @pdudits , because they enforce unix line endings in their subdirs, but we don't have such global rule, because it could result in much more changes and cause problems.

dmatej:
Jenkins test please

PR #4616, opened 2020-04-08
CUSTCOM-254 Fixed parsing of Accept header in metrics

dmatej:
# Description

- by default jersey client sends valid header which was causing Internal Server Error:
  `text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2`
- now spaces and missing leading zeroes are tolerated
- reduced copy and paste

# Testing

- behavior of the regular expression was tested locally modifying this simple test (q=0.1 or q= 0.1 vs expected 0.1 AND q=.1 or q = .1 vs expected .1):
```
    @Test
    public void test() {
        Pattern pattern = Pattern.compile("[q\\s]+=\\s*(.+)");
        assertTrue(pattern.matcher("q=0.1").matches());
        final Matcher matcher = pattern.matcher("q = .1");
        matcher.find();
        assertEquals(".1", matcher.group(1));
    }
```

- then I verified that Payara with fixed MetricsResource works in tests under PAYARA-4176 + CUSTCOM-133 (this was a blocker for these tests).

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/rest/MetricsResource.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/rest/MetricsResource.java
@@ -50,6 +50,9 @@ import fish.payara.microprofile.metrics.writer.MetricsWriterImpl;
 import fish.payara.microprofile.metrics.writer.OpenMetricsExporter;
 import java.io.IOException;
 import java.io.Writer;
+import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static fish.payara.microprofile.Constants.EMPTY_STRING;
 import static java.nio.charset.StandardCharsets.UTF_8;
@@ -72,8 +75,9 @@ import org.eclipse.microprofile.metrics.MetricRegistry.Type;
 import org.glassfish.internal.api.Globals;
 
 public class MetricsResource extends HttpServlet {
-
+    private static final Logger LOG = Logger.getLogger(MetricsResource.class.getName());
     private static final String APPLICATION_WILDCARD = "application/*";
+    private static final Pattern PATTERN_Q_PART = Pattern.compile("[q\\s]+=\\s*(.+)");
 
     /**
      * Processes requests for both HTTP <code>GET</code> and <code>OPTIONS</code>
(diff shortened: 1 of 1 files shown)


dmatej:
Jenkins test please

OndroMih:
Looks good.

OndroMih:
I agree with @jbee that it would be nice to add a junit test to the pull request too.

dmatej:
Unit test for pattern (and not the class) is worthless. Useful would be unit tests for the servlet, but then I would need to play with easymock/mockito or even better refactor this class and move these things somewhere else (something like HttpHeadersHelper?).

That's why I did not create a unit test here and I created only that "regexp test" aside locally. There is no risk it would change it's behavior, but the servlet can. And the servlet is still covered by integration tests - those will be merged soon in separate PR (I still have some work on that branch before it could be merged to master branch).

jbee:
@dmatej Remember the beauty of pure functions a.k.a. static methods. You can test them without all the stuff around and you are testing the actual code.

svendiedrichsen:
@jbee I'd rather use the Interfaces from https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html to have testable units.

dmatej:
That's the point - private methods are not testable units. It would make sense to move them to some utility package, BUT then I have a strong feeling we are reinventing the wheel and there is already some open source (probably Apache) implementation (which could replace much more than this).

Then it is safer to have these methods private and replace them in some refactoring with some more mature solution (we don't use apache commons libs now except several old).

jbee:
You can make them less private :D 
Its not that non private pure functions pose any risk. There is no state to protect.
And the library might never come. Until it does having the test is a good way to make sure it does not happen again and you might even test some other cases you can think of.

dmatej:
I have a tendency to start being sarcastic, so I rather stop this conversation - but generally making methods more visible only because of tests is a design antipattern.

Tests must have sense. Do we have any test covering behavior of this servlet? Except what I created [here](https://github.com/dmatej/Payara/blob/PAYARA-4176-TestContainers-prototype/appserver/tests/test-containers/src/test/java/fish/payara/test/containers/tst/security/MetricsSecurityITest.java)?

jbee:
> Do we have any test covering behavior of this servlet?

https://github.com/payara/Payara/tree/master/appserver/payara-appserver-modules/microprofile/metrics/src/test/java/fish/payara/microprofile/metrics/writer

Test that the content of the response itself is correct. The servlet aspects like the headers are not tested anywhere else I am aware of.

PR #4626, opened 2020-04-16
APPSERV-142: Shutdown fault tolerance thread pools on server shutdown

pdudits:
# Description
This is a bug fix. 

New threadpools for Fault tolerance doesn't shut down cleanly while having a permanent cleanup task scheduled.

There is probably more that should be done (i. e. utilizing `PayaraExecutorService` instead), but it fixes the largest trouble at this moment.

# Testing

## Testing Performed
`java -jar payara-micro.jar --warmup` didn't terminate before the fix, it does now.

### Testing Environment
AdoptOpenJDK 14+36 (by mistake) on Windows 10.
Zulu 11.0.6 in Docker


pdudits:
Jenkins test please

PR #4630, opened 2020-04-17
APPSERV-20 Add Config Source for Payara Variables

Pandrex247:
# Description
This is a feature. 

This PR adds a config source for a new properties file: payara-expression-config.properties. The properties stored in here are converted from their variable into their actual values. For example, a property of `fish.payara.example.alias=${ALIAS=wibbles}` will be converted to the password stored under the alias of _wibbles_. If an alias isn't found, null is returned.

This PR also changes it so that the password alias store will also match against both "wibbles" and "${ALIAS=wibbles}"

# Testing

### New tests
Some unit tests added to check that null is returned from the config source if no matching alias is found, and also that the values are successfully converted for system properties and environment variables.

I also created a small sample app to check that the properties file is read and that password alias store reads both "wibbles" and "${ALIAS=wibbles}". This has been added under `appserver/tests/payara-samples/samples`.

### Test suites executed

- Payara Microprofile TCKs Runner (Config)

### Testing Environment
Windows 10. JDK 8u212.

# Documentation
https://github.com/payara/Payara-Server-Documentation/pull/724


Pandrex247:
Requires environment variables adding to Jenkins instance.

Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

PR #4635, opened 2020-04-21
CUSTCOM-266: Fix HK2 Config when Payara Micro is launched via Root Dir Launcher

pdudits:
# Description

SingleModuleHK2Registry assumes that all runtime jars are enumerated by
context class loader, which is only true when Payara Micro's special
classloader is involved. When launched with RootDir launcher, classpath
only contains launcher jar, and the module list need to be read from
its manifest.

# Testing
### Testing Performed
Testing was done by trying to start a JSF application in rootdir mode:

```
java -jar $micro --rootDir root --deploy jsf-app.war --nocluster --warmup
java -jar $micro --rootDir root --outputlauncher
java -jar root/launch-micro.jar --nocluster --warmup
```

Application would not start up in third case, it didn't before, because `FacesContext.getCurrentContext()` would fail.

### Testing Environment
Tested multiple JVMs on Windows, as this relates to root classloader:
* Zulu11.37+17-CA
* Zulu 8.46.0.19-CA-win64
* AdoptOpenJDK (build 11.0.6+10) / openj9-0.18.0
* AdoptOpenJDK (build 1.8.0_242-b08) / openj9-0.18.1

# Notes for Reviewers

`JarManifestModuleRegistry` is a modification of HK2's `SingleModuleRegistry`, therefore the Eclipse license.

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/JarManifestHk2Factory.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/boot/runtime/JarManifestHk2Factory.java
@@ -1,0 +1,97 @@
+/*
+ *    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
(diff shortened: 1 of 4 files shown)


MattGill98:
jenkins test please

PR #4637, opened 2020-04-21
APPSERV-141 Fixes undefined behaviour of PayaraConfig 

jbee:
### Summary
Fixes behaviour for undefined properties of `PayaraConfig` for arrays, lists and sets.
If an undefined property is resolved the `Config` implementation should always throw an `NoSuchElementException`. This behaviour was changed unintentional and incorrectly by #4614 .

### Testing
Added unit tests for undefined property lookup.

#### Testing Performed
Running MP TCKs https://github.com/payara/MicroProfile-TCK-Runners :

```bash
# setup environment for TCKs
export my_int_property=45
export MY_BOOLEAN_PROPERTY=true
export my_string_property=haha
export MY_STRING_PROPERTY=woohoo

# start server
{payara-bin-dir}/asadmin start-domain

# run different MP TCKs
cd {tck--runner-base-dir}/tck-runner/
```
This is used to run MP Config, MP Metrics, MP Rest Client and MP FaultTolerance TCKs.
This will run all TCK with cache disabled (`MP_CONFIG_CACHE_DURATION=0`). This configuration is only needed for the MP Config TCK. Other TCKs should also pass with cache enabled (do `export MP_CONFIG_CACHE_DURATION=30` and restart the server).
Note that both starting the server and running the TCK should be done in same console to avoid having to setup same environment in multiple shells.

I did run both with and without caching enabled (except the MP Config TCK that requires cache to be disabled as otherwise 2 tests will fail)

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/config/src/main/java/fish/payara/microprofile/config/cdi/ConfigPropertyProducer.java
+++ b/appserver/payara-appserver-modules/microprofile/config/src/main/java/fish/payara/microprofile/config/cdi/ConfigPropertyProducer.java
@@ -96,12 +96,13 @@ public class ConfigPropertyProducer {
             result = config.getValue(name, property.defaultValue(),(Class<?>)type);
         } else if ( type instanceof ParameterizedType) {
             ParameterizedType ptype = (ParameterizedType)type;
-            if (List.class.equals(ptype.getRawType())) {
-                result = config.getListValues(name, property.defaultValue(), (Class<?>) ptype.getActualTypeArguments()[0]);
-            } else if (Set.class.equals(ptype.getRawType())) {
-                result = config.getSetValues(name, property.defaultValue(), (Class<?>) ptype.getActualTypeArguments()[0]);
+            Type rawType = ptype.getRawType();
+            if (List.class.equals(rawType)) {
+                result = config.getListValues(name, property.defaultValue(), getElementTypeFrom(ptype));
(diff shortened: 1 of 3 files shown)


appserver/payara-appserver-modules/microprofile/config/src/main/java/fish/payara/microprofile/config/cdi/ConfigPropertyProducer.java
```diff
@@ -111,4 +112,14 @@ public static final Object getGenericProperty(InjectionPoint ip) {
         return result;
     }
 
+    private static Class<?> getElementTypeFrom(ParameterizedType collectionType) {
+        Type elementType = collectionType.getActualTypeArguments()[0];
+        if (!(elementType instanceof Class)) {
+            throw new DeploymentException(
```
jbee, line 118: I was looking into not using `getSetValues` and `getListValues` directly (plain API access) as I realised that these should be implemented by using array access and later wrapping the returned array in a `List` or `Set`. What is unclear is how the default gets past on using the standard API. So I made this cleanup as a preparation and noticed that it was simply assumed that the actual type argument is a `Class` type which it might not be so I added a more descriptive exception instead of the `ClassCastException` that was caused in such a case. As far as I understand it using a non `Class` type is not allowed according to standard but a user might get the idea it is valid and use a type like `List<List<String>>`.

jbee:
jenkins test please

PR #4638, opened 2020-04-22
CUSTCOM-247 tests

OndroMih:
# Description
Adds tests to payara-samples.

# Testing
### New tests
New tests in a new project "jaxws-security":

* ServletEndpointTest.testSoapRequestUserNotAllowedExecution
* ServletEndpointTest.testSoapRequestWithIncorrectCredentials
* ServletEndpointTest.testPermittedSoapRequest 
* ServletEndpointTest.testUnrestrictedSoapRequest
* EjbEndpointTest.testSoapRequestUserNotAllowedExecution (unstable)
* EjbEndpointTest.testSoapRequestWithIncorrectCredentials (unstable)
* EjbEndpointTest.testPermittedSoapRequest (unstable)
* EjbEndpointTest.testUnrestrictedSoapRequest (unstable)

Unstable tests don't pass on master due to another bug which isn't in scope of CUSTCOM-247. it fixes only the Servlet-based WS but there are other issues in EJB-based WS. 

The unstable tests are executed only with `unstable` maven profile.

### Testing Environment

Ubuntu 18.04
Zulu 8.42.0.21 (openjdk 1.8.0_232)
Maven 3.6.0

# Notes for Reviewers

I created a new `Unstable` interface in the `test-utils` project to filter out unstable tests. I also modified `pom.xml` in `payara-samples` parent project to trigger tests in Unstable category only in the `unstable` profile.

PR #4640, opened 2020-04-24
CUSTCOM-247: Fix of switched assertions in tests

OndroMih:
This is just a fix of my previous PR #4638 where I mistakenly switch 2 assertions in 2 unstable tests.


dmatej:
Jenkins test please

PR #4641, opened 2020-04-25
QACI-95 Refactor Payara Samples

dmatej:
This PR replaces https://github.com/payara/Payara/pull/4525 . The main difference is in possibility to execute unstable tests simply by enabling -Punstable profile, so this is still useful for developers despite jenkins will not run them.

- using annotations NotMicroCompatible and Category to respect capabilities
  of the tested impl and dependencies of tests.
- NotMicroCompatible has argument for string explanation
- killed tabs and trimmed whitespaces again
- arquillian update
- disabled trimming of stacktraces in tests (how can you find what is broken?)
- note: the SetHazelcastConfigurationTest crashes on JDK8 with profiles
        unstable+payara-micro-managed. It is caused by the bug in JDK8,
        fixed in JDK11.
- using profile to detect if environment is suitable to run the test
  - added profile auto-skip-test to expression-config-properties which
    automatically skips the test if there are not configured env properties
    in expected way. This test is a candidate for TestContainers as it
    requires external environment configuration.
- module samples-programmatic renamed to samples-micro-programmatic, which is not
  executed if the payara-server-remote profile is activated.
- embedded-vs-jersey was moved to Payara Embedded modules
- ServerLogTest was stabilized to test only single startup, so it avoids messing with executions of all previous tests.

# Testing

### Testing Performed
```
mvn clean install -Ppayara-server-managed,payara-samples -fae -amd -pl :payara-samples
success, 8 min

mvn clean install -Ppayara-micro-managed,payara-samples -fae -amd -pl :payara-samples
success, 2 min

mvn clean install -Ppayara-server-remote,payara-samples -fae -amd -pl :payara-samples
succes, 5 min

mvn clean install -Ppayara-server-managed,payara-samples,unstable -fae -amd -pl :payara-samples
failure, 9 min, 3 unstable modules, 

mvn clean install -Ppayara-micro-managed,payara-samples,unstable -fae -amd -pl :payara-samples
failure, 2 min, 1 unstable module (asadmin)

mvn clean install -Ppayara-server-remote,payara-samples,unstable -fae -amd -pl :payara-samples
failure, 5 min, 3 unstable modules

export Bibbly=Bibbles; export Dibbles="\${Dabbles}";
mvn clean install -Ppayara-server-managed,payara-samples -fae -amd -pl :payara-samples
unset Bibbly; unset Dibbles;
(PayaraExpressionConfigPropertiesTest was executed and build succeeded)

mvn clean install -PBuildExtras -pl :payara-embedded-all -amd
Total time:  02:24 min, one test
```

### Testing Environment

AdoptOpenJDK)(build 1.8.0_212-b03

I tested also 8u252, 8u242 and OpenJDK Runtime Environment (build 11.0.7+10-post-Ubuntu-2ubuntu219.10), there are differences:
* JDK11 does not fail on usage of @Category, it is a bug of JDK8: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8209742 (fixed in JDK11)
* same test (SetHazelcastConfigurationTest) fails on JDK8 when executed in same run with other tests, which leave domain dirty. So it is unstable, but I left it as stable as it has no problems when it runs as first test on clean domain; after it will be stabilized, both problems dismiss.
* on JDK8u232 and later tests using SSL fail due to incompatibility of grizzly with new JDK distributions. With older JDKs some test methods still fail. -> unstable.


dmatej:
Jenkins test please

dmatej:
Jenkins test please

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
```diff
@@ -39,15 +39,22 @@
...
-
-import fish.payara.nucleus.hazelcast.HazelcastRuntimeConfiguration;
+import org.junit.experimental.categories.Category;
 
 /**
  * Verifies the correctness of the {@code SetHazelcastConfiguration} command.
  */
+@Category(Unstable.class)
+// Fails from two reasons:
+// 1) Requires completely new domain. Side effects of other tests break this one.
```
Pandrex247, line 55: The only test I see that requires a clean domain is the dataGridEncryptionWarning test, which surely could also be acheived by making the other test that creates the key clean up after itself properly?

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
```diff
@@ -39,15 +39,22 @@
...
+// 2) On JDK8 fails because of usage of @Category annotation which has problems with
```
Pandrex247, line 56: Why are we introducing things that don't work on what is arguably the primary java version?

appserver/tests/payara-samples/samples/payara-expression-config-properties/pom.xml
```diff
@@ -105,4 +112,36 @@
             </plugin>
         </plugins>
     </build>
-</project>
\ No newline at end of file
+    <profiles>
+        <profile>
+            <id>auto-skip-test</id>
```
Pandrex247, line 117: I'm not fond of this.
A test that is disabled due to not having an environment variable set is never run.

appserver/tests/payara-samples/samples/payara-expression-config-properties/src/test/java/fish/payara/samples/microprofile/config/expression/PayaraExpressionConfigPropertiesTest.java
```diff
@@ -71,14 +70,16 @@
     @Deployment(testable = false)
     public static WebArchive createDeployment() {
         return ShrinkWrap.create(WebArchive.class, "microprofile-config-expression.war")
-                .addPackage("fish.payara.samples.microprofile.config.expression")
-                .addAsManifestResource(new File("src/main/resources/META-INF/payara-expression-config.properties"));
+            .addPackage(ConfigServlet.class.getPackage()) //
```
Pandrex247, line 73: Why less indent? 8 spaces is the default for breaking up long lines.
Also what's with the random empty comments?

appserver/tests/payara-samples/samples/pom.xml
```diff
@@ -15,20 +15,26 @@
     <name>Payara Samples - Profiled Tests</name>
 
     <modules>
+        <!-- This module must run first, because it collides with some change made by other test(s) -->
+        <module>payara-expression-config-properties</module>
```
Pandrex247, line 19: It shouldn't do? Tests should clean up after themselves.

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
jbee: Agree. Eventually we might start using JUnit5 but this should be a planned step.

dmatej: First Categories are used in JUnit4 as a standard for maybe 15 years and they do work. 

Probably this issue is caused by interaction with this in PayaraArquillianTestRunner. But it is worthless to investigate, because the test does not work at all.
And if it would, we would not need the category ;)

image

appserver/tests/payara-samples/samples/payara-expression-config-properties/pom.xml
dmatej: Repeatable tests should run without hacking computer environment. By default it only fails.

appserver/tests/payara-samples/samples/payara-expression-config-properties/src/test/java/fish/payara/samples/microprofile/config/expression/PayaraExpressionConfigPropertiesTest.java
dmatej: It is called a "phantom comment". In some circumstances it is best possible way to protected readable code from automatic reformatting.
I don't see 2 block indent useful and always used 1 block, but yes, in Eclipse default is 2 blocks (8 spaces). Honestly I don't see it as a most important thing worth of discussion.

appserver/tests/payara-samples/samples/pom.xml
dmatej: They reuse the managed/remote domain, so you can never get to the state before all tests. But yes, it seems some (or all?) tests leave some garbage behind them. Btw this one too.

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
dmatej: https://dzone.com/articles/closer-look-junit-categories (2010)

appserver/tests/payara-samples/samples/payara-expression-config-properties/pom.xml
dmatej: I understand that in this case there was no other way possible - I plan to move this test to TC PR including jenkinsfile etc., so final state should be clean next week ;)

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
dmatej: Both should clean up after their run. Or better we should have some architecture in tests so instead of testing "random" features in random order we could join them to some scenarios, use cases, categories/groups (?) and each category would start with a new domain. Then we can have a control over it, but imagine we would have thousand of tests in payara samples - which of those test result states interact with any other? The current state of things is not maintainable in larger scope.

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
jbee: Okay, thought the issue here was related to use of a feature that was introduced later.

dmatej: Maybe we should use categories instead of NotMicroCompatible too, but ... I think there will be more work with Andrew's point around repeatibility of tests. Probably it could spark some more stable (and faster and cheaper) architecture of tests, I will think about it :-)

MarkWareham:
CI won't be ready to handle this before release. Please do not merge until after.

MarkWareham:
I would like to see this merged or closed, whichever is most applicable

dmatej:
I agree, I expect I will get to this before sunday. Or maybe even today if I will finish my cloudy issue soon :-)

dmatej:
Jenkins test please

dmatej:
Jenkins test please

dmatej:
Jenkins test please

dmatej:
Jenkins test please

appserver/extras/embedded/tests/embedded-vs-jersey/pom.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
dmatej, line 1: Moved, because this tests were too far from building extras and payara embedded, which is hidden under a profile. That had two consequences:
- the Payara Samples tested against old version of Payara Embedded 
- sometimes the embedded wasn't built yet -> dependency issues.

appserver/tests/payara-samples/pom.xml
```diff
@@ -86,7 +82,7 @@
             <dependency>
                 <groupId>org.jboss.arquillian</groupId>
                 <artifactId>arquillian-bom</artifactId>
-                <version>1.5.0.Final</version>
+                <version>1.6.0.Final</version>
```
dmatej, line 85: Upgrade!

appserver/tests/payara-samples/pom.xml
```diff
@@ -108,6 +104,16 @@
                 <artifactId>smallrye-config</artifactId>
                 <version>1.3.7</version>
                 <scope>test</scope>
+                <exclusions>
```
dmatej, line 107: This was discovered when we discussed related dependency issues with @MattGill98

appserver/tests/payara-samples/pom.xml
```diff
@@ -238,6 +244,7 @@
                         <systemPropertyVariables>
                             <isUsingMicroProfile>${micro.isActive}</isUsingMicroProfile>
                         </systemPropertyVariables>
+                        <trimStackTrace>false</trimStackTrace>
```
dmatej, line 247: This configuration should be probably default.

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java
```diff
@@ -39,15 +39,22 @@
...
+//    this bug: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8209742 (fixed in JDK11)
```
dmatej, line 57: Probably better link:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-7183985

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHealthCheckServiceConfigurationTest.java
```diff
@@ -83,13 +83,13 @@ public void setUp() {
```
dmatej, line 83: Deleted trailing spaces.

appserver/tests/payara-samples/samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHealthCheckServiceConfigurationTest.java
```diff
@@ -261,17 +261,17 @@ public void hogginThreadsThresholdDynamicAffectsConfigAndService() {
 
     @Test
     public void hogginThreadsRetryCountBelowMinimumCausesError() {
-        CommandResult result = asadmin("set-healthcheck-service-configuration", 
+        CommandResult result = asadmin("set-healthcheck-service-configuration",
                 "--service", "ht",
                 "--enabled", "true",
-                "--hogging-threads-retry-count", "0");
```
dmatej, line 267: Fixed to respect the current implementation.

appserver/tests/payara-samples/samples/ejb-invoker-secure-endpoint/src/test/java/fish/payara/samples/ejb/invoker/security/RemoteBeanBasicAuthTest.java
```diff
@@ -55,7 +56,8 @@
  * BASIC Auth.
  *
  */
-@RunWith(Arquillian.class)
```
dmatej, line 58: I noticed the Jenkins build is much slower now. What is the cause? Is PayaraArquillianTestRunner slower than Arquillian?

appserver/tests/payara-samples/samples/payara-expression-config-properties/pom.xml
dmatej: After this PR will make it to master I will rebase my TC branch and transform this test to TC.

appserver/tests/payara-samples/samples/pom.xml
```diff
@@ -43,25 +49,7 @@
...
-		        <module>openid</module>
-		        <module>jaxws-tracing</module>
-		        <module>asadmin</module>
-		        <module>microprofile-endpoints</module>
-         	</modules>
-		</profile>
-		<profile>
-			<id>unstable-remote</id>
-			<modules>
-                <module>ejb-invoker-secure-endpoint</module>
```
dmatej, line 59: Ok, these tests are the cause of the slowdown. I will take a look once more if they couldn't be faster (unclosed rest clients or what?)

dmatej:
Jenkins test please

appserver/tests/payara-samples/samples/pom.xml
dmatej: They do many restarts, that's why they are so slow. But that's out of the scope of this issue and I'm not sure if it is even resolvable.

dmatej:
Rebased and retested.

dmatej:
Jenkins test please

dmatej:
Jenkins test please

appserver/tests/payara-samples/samples/rest-management/src/test/java/fish/payara/samples/rest/management/ServerLogTest.java
```diff
@@ -39,24 +39,41 @@
...
+import org.junit.Before;
 import org.junit.Test;
 
 /**
  * Tests to run against the server log
  */
 public class ServerLogTest extends RestManagementTest {
 
+    @Before
+    public void resetLogging() throws Exception {
```
dmatej, line 61: This test consumed logs also from previous executions. Some of them logged warnings.

appserver/tests/payara-samples/test-utils/src/main/java/fish/payara/samples/NotMicroCompatible.java
```diff
@@ -48,15 +48,17 @@
...
- * 
+ *
  * @author Cuba Stanley
  */
 @Target({ElementType.METHOD, ElementType.TYPE})
 @Retention(RetentionPolicy.RUNTIME)
 @Inherited
-
-public @interface NotMicroCompatible {}
+public @interface NotMicroCompatible {
```
dmatej, line 61: Added possible explanation why it is not micro compatible.

PR #4646, opened 2020-05-01
APPSERV-157 Fixes log spamming with InvocationManager warning by web-glue

jbee:
### Summary
The `InvocationManager.preInvocation` and `InvocationManager.postInvocation` should get passed the same instance of the `ComponentInvocation` as this is opening and closing the "context" for that element. This semantic was not honoured by 2 callers which did create new instances of the passed `ComponentInvocation` each time any of the two methods were called. 

The reason pre- and post were not called consistently was that the caller themselves were just reacting to events so there was no sequential execution for which pre is called at the beginning and post at the end. 

To simulate a "sequential execution context" consisting of a pair of before and after events the passed invocation is kept in a map so the one created for or by before handler can be used by after handler. As the `WebComponentInvocation` instances themselves only depend on the `WebModule` (in one case) and the `WebModule` + an instance object (in the other case) both of which are not really changing for most of the events the assumption is that it is equally fine to reuse the same `WebComponentInvocation` instance for the combination key of `WebModule.getID()` and the identify hash of the instance object.

### Testing
* Start the server, open admingui
* Check no WARNINGs were logged (and no error occur)

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/web/server/J2EEInstanceListener.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/web/server/J2EEInstanceListener.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-
+// Portions Copyright 2020 Payara Foundation and/or its affiliates
 package com.sun.web.server;
 
 import com.sun.enterprise.container.common.spi.util.InjectionException;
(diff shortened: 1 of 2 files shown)


jbee:
jenkins test please

jbee:
This solution might be problematic because it isn't cleaned up.

jbee:
I'll looked at it for some time now and I can't see a solution that isn't getting more and more complex so my thinking goes towards relaxing the conditions for which a warning is given a bit.

PR #4647, opened 2020-05-01
APPSERV-157 preInvoke/postInvoke ComponentInvocation logging

jbee:
### Summary
Strictly it would be correct to assume that the `ComponentInocation` passed to `preInvoke` should be the same as the one passed to `postInvoke`. This however is difficult to do in some cases (see #4646). Relaxing expectations `equal` is also problematic as `WebComponentInocation#equal` would need to change. To avoid introducing new issues `WebComponentInocation`s will be excluded from the logging. This needs to be done matching the class name as string as the module making the check cannot become dependent on the module defining that class. I figured that checking the simple class name is sufficient here.

I also improved the formatting of the log message and added the strack trace as it is important to find where the inconsistent call is coming from.

After revisiting this and with the greater insight gained over time spent looking at the invocation manager I also decided to use the `current` instance from internal stack (that is the `ComponentInvocation` passed to `preInvoke`) for the `postInvoke` to the listeners. 
The implementation did that in 3/4 of calls before and I replaced that with the argument passed to `postInvoke`, that was used 1/4 as back then I felt this would be more consistent with `preInvoke`. But now I think that "more consistent" means passing the instanced passed to `preInvoke` also to `postInvoke` listeners as originally done for 3/4 of listener notifications.

### Testing
* start server
* open admin-gui
* check that no warning has been logged (related to `ComponentInocation`)

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
@@ -162,6 +162,11 @@ public class WebComponentInvocation extends ComponentInvocation {
     }
 
     @Override
+    public int hashCode() {
+        return instance == null ? container.hashCode() : container.hashCode() ^ instance.hashCode();
+    }
+
+    @Override
     public boolean equals(Object obj) {
         if (!(obj instanceof WebComponentInvocation)) {
             return false;
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
@@ -47,7 +47,6 @@ import javax.servlet.Filter;
 import javax.servlet.Servlet;
 import javax.servlet.SingleThreadModel;
 import java.lang.reflect.Method;
-import java.util.Objects;
 
 public class WebComponentInvocation extends ComponentInvocation {
 
@@ -70,12 +69,12 @@ public class WebComponentInvocation extends ComponentInvocation {
         container = wm;
         this.instance = instance;
         setResourceTableKey(_getResourceTableKey());
-
+        
         moduleName = wm.getModuleName();
         appName = wm.getWebBundleDescriptor().getApplication().getAppName();
         registrationName = wm.getWebBundleDescriptor().getApplication().getRegistrationName();
     }
-
+    
     public WebComponentInvocation(WebModule wm, Object instance, String instanceName) {
       this(wm, instance);
       setInstanceName(instanceName);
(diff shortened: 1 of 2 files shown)


jbee:
jenkins test please

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
```diff
@@ -160,4 +160,18 @@ public boolean equals(Object obj) {
             return eq;
         }
     }
+
+    @Override
+    public int hashCode() {
+        return instance == null ? container.hashCode() : container.hashCode() ^ instance.hashCode();
```
svendiedrichsen, line 166: How about using Objects.hashCode(container, instance)?

jbee: If there are other changes requested I use it for better readability. Just this alone I feel is to small to go through a new round.

dmatej: So I am back here (see comments down there) ... 
We have two implementations of this class
* WebComponentInvocation - now you say two invocations are equal if they are done in same container and same instance. Are you sure it is alright?
* EjbInvocation - this uses Object.equals, so two invocations are equals if they is one invocation

These all and even more are events ... 
image

nucleus/common/glassfish-api/src/main/java/org/glassfish/api/invocation/InvocationManagerImpl.java
```diff
@@ -206,6 +206,10 @@ public void setThreadInheritableInvocation(List<? extends ComponentInvocation> p
         }
     }
 
+    private static boolean isEqual(ComponentInvocation a, ComponentInvocation b) {
+        return a == b || a.getClass() == b.getClass() && a.equals(b);
```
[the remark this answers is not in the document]
jbee: Checking the class is not equal to only doing `a.equals(b)`. Implementations could consider sub types as equal while I do not want that to happen.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebComponentInvocation.java
jbee: There are even more implementations. That is part of the reason for the chosen solution.

Generally it seems reasonable for an invocation to only assume two instances being equal when `a == b` as they basically represent a call stack element. However, callers themselves only reacting to events have trouble to pass the very same instance of the invocation to both pre- and post as those originate from different execution stacks. The point where those are within one method is either too far does not even exist when originating from scheduled executions of working events.

Given this the question becomes: how do we know the pre/post invocation instance represent the same stack element? If `==` is no option `equals` is the next reasonable choice. `WebComponentInvocation` are always constructed from a `WebModule` and possibly an `instance`. Being more strict here is better since behaviour before was `==`.

jbee:
jenkins test please

jbee:
@pdudits do you have any comments?

lprimak:
Since you are looking at this, there is JavaEEContextUtil that does this in a modern way, in case you were unaware

PR #4655, opened 2020-05-06
APPSERV-148 Adds MP Metadata as permanent MC annotation

jbee:
### Summary
Goal was to make the MicroProfile Metadata of MP metrics available as "related data" for the MC series resulting from each MP metric.

The solution uses the existing concept of MC annotations which are key-value pairs associated with a series at a certain point in time of that series. As annotations are limited to a certain capacity and then replace the oldest the concept of annotation needed to be extended so that annotation could be made that would be permanent. Meaning other non-permanent annotations are disposed to make room for newer annotations before any permanent annotation is replaced. This is done by https://github.com/payara/monitoring-console/pull/9 and used in this PR by the `MetricsService` to add the annotation data.

### CI
Needs https://github.com/payara/monitoring-console/pull/9 and https://github.com/payara/monitoring-console/pull/10 merged before jenkins will succeed to build it.

### Testing
For changes in https://github.com/payara/monitoring-console/pull/9 run
```bash 
/monitoring-console/$ mvn clean install
/monitoring-console/$ cp process/target/monitoring-console-process.jar {payara-home}/glassfish/modules/
```
Deploy the webapp war file built at `monitoring-console/webapp/target`.

Check annotations for a MP metric at `http://localhost:8080/monitoring-console-webapp/api/annotations/data/ns:metric%20@:classloader%20LoadedClassesCount/` (assuming app is deployed in context `monitoring-console-webapp`):

```json
[
    {
        "attrs": {
            "Name": "classloader.loadedClasses.count",
            "Type": "gauge",
            "Unit": "none",
            "DisplayName": "Currently Loaded Class Count",
            "Discription": "Displays the number of classes that are currently loaded in the JVM."
        },
        "instance": "server",
        "permanent": true,
        "series": "ns:metric @:classloader LoadedClassesCount",
        "time": 1588756123000,
        "value": 0
    }
]
```
The `attrs` shows the important `Metadata` information from the `Metric` called `classloader.loadedClasses.count` in MP metrics and which results in a series called `ns:metric @:classloader LoadedClassesCount` in MC.

Create and start another instance and refresh the above URL. The result should be like:
```json
[
    {
        "attrs": {
            "Name": "classloader.loadedClasses.count",
            "Type": "gauge",
            "Unit": "none",
            "DisplayName": "Currently Loaded Class Count",
            "Discription": "Displays the number of classes that are currently loaded in the JVM."
        },
        "instance": "server",
        "permanent": true,
        "series": "ns:metric @:classloader LoadedClassesCount",
        "time": 1588767450000,
        "value": 0
    },
    {
        "attrs": {
            "Name": "classloader.loadedClasses.count",
            "Type": "gauge",
            "Unit": "none",
            "DisplayName": "Currently Loaded Class Count",
            "Discription": "Displays the number of classes that are currently loaded in the JVM."
        },
        "instance": "Lovely-Quillback",
        "permanent": true,
        "series": "ns:metric @:classloader LoadedClassesCount",
        "time": 1588767550000,
        "value": 0
    }
]
```
Check that both the `"server"` and the other instance's annotation have a `"permanent"` field being `true`.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricsService.java
@@ -41,6 +41,7 @@ package fish.payara.microprofile.metrics;
 
 import fish.payara.microprofile.metrics.admin.MetricsServiceConfiguration;
 import fish.payara.microprofile.metrics.exception.NoSuchRegistryException;
+import fish.payara.microprofile.metrics.impl.MetricRegistrationListener;
 import fish.payara.microprofile.metrics.impl.MetricRegistryImpl;
 import fish.payara.microprofile.metrics.jmx.MBeanMetadata;
 import fish.payara.microprofile.metrics.jmx.MBeanMetadataConfig;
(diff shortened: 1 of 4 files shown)


jbee:
Waiting for 1.1 release of MC.

jbee:
jenkins test please

Issue #4625, opened 2020-04-16
Payara 5.201 will not start with Amazon Corretto 8.252.09.1 after enable-secure-admin command


vogeldavid:
# Payara 5.201 will not start with Amazon Corretto 8.252.09.1 after enable-secure-admin command #
----------

I can not run Glassfish 5.201 with Amazon Corretto 8.252.09.1 after entering enable-secure-admin command. 

- With openjdk-8u252-b09 or amazon-corretto-8.252 (Amazon Corretto 8.252.09.1) it will not work.
- With zulu-jdk8.0.252, openjdk-8u242 or amazon-corretto-8.242 it works fine.

 

## Expected Outcome

Payara 5.201 starts

## Current Outcome

Payara 5.201 does not start properly

## Steps to reproduce (Only for bug reports) 

Setup:

I have Ubuntu 16.04 lts, then I install openjdk-8u252-b09 and download glassfish server on zip file.

```
$ java -version
openjdk version "1.8.0_252"
OpenJDK Runtime Environment (build 1.8.0_252-b09)
OpenJDK 64-Bit Server VM (build 25.252-b09, mixed mode)
```

`$unzip payara-5.201.zip`

```

$ payara5/bin/asadmin start-domain
Waiting for domain1 to start ......
Successfully started the domain : domain1
domain  Location: /opt/payara/payara5/glassfish/domains/domain1
Log File: /opt/payara/payara5/glassfish/domains/domain1/logs/server.log
Admin Port: 4848
Command start-domain executed successfully.
```

Set password for admin user
```
vogel@birdy:/opt/payara$ payara5/bin/asadmin change-admin-password
Enter admin user name [default: admin]>
Enter the admin password> 
Enter the new admin password> 
Enter the new admin password again> 
Command change-admin-password executed successfully.
```

Enable secure admin:
```
$ payara5/bin/asadmin enable-secure-admin
Enter admin user name> admin
Enter admin password for user "admin"> 
You must restart all running servers for the change in secure admin to take effect.
Command enable-secure-admin executed successfully.
```

Restart server will fail:
```
$ payara5/bin/asadmin restart-domain
NCLS-ADMIN-00010
NCLS-ADMIN-00010
...
```

server.log:
```
[...]

PR #4656, opened 2020-05-06
Remove Grizzly ALPN from 1.8.0u251 and above

MattGill98:
# Description

These versions also contain the ALPN functionality included in Java 9,
so should be treated as such by the domain options.

Fixes https://github.com/payara/Payara/issues/4625

# Testing

### Testing Performed

JDK 1.8u154 SSL connection.


MattGill98:
jenkins test please

PR #4659, opened 2020-05-08
ECOSYS-157 OpenID Connect Caller's name and groups are null

jGauravGupta:
# Description
This is a fix to set the caller's name & groups after the authentication in `OpenIdContext` and error message logged if redirect URL not matched with request URL.

# Testing
### Testing Performed
Manually Tested, Payara-Samples

### DIFF
--- a/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
+++ b/appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
@@ -236,7 +236,7 @@ public class OpenIdAuthenticationMechanism implements HttpAuthenticationMechanis
 
         Optional<OpenIdState> receivedState = OpenIdState.from(request.getParameter(STATE));
         String redirectURI = configuration.buildRedirectURI(request);
-        if (receivedState.isPresent() && redirectURI.contains(request.getServletPath())) {
+        if (receivedState.isPresent()) {
             if (!request.getRequestURL().toString().equals(redirectURI)) {
                 LOGGER.log(INFO, "OpenID Redirect URL {0} not matched with request URL {1}", new Object[]{redirectURI, request.getRequestURL().toString()});
                 return httpContext.notifyContainerAboutLogin(NOT_VALIDATED_RESULT);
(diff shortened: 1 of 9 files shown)


appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdAuthenticationMechanism.java
```diff
@@ -235,21 +236,22 @@ private AuthenticationStatus authenticate(
 
         Optional<OpenIdState> receivedState = OpenIdState.from(request.getParameter(STATE));
         String redirectURI = configuration.buildRedirectURI(request);
-        if (receivedState.isPresent()
-                && request.getRequestURL().toString().equals(redirectURI)) {
-
+        if (receivedState.isPresent() && redirectURI.contains(request.getServletPath())) {
```
NikitaZ, line 239: Hi, what is the reason of adding extra redirectURI.contains(request.getServletPath()) check?
it breaks existing applications, (I just debugged it), 
if redirectURI is "https://localhost:8181/myapp/" and a welcome file is specified, i.e.
    <welcome-file-list>
       <welcome-file>faces/About.xhtml</welcome-file>
    </welcome-file-list>
request.getServletPath() returns "/faces" and this check fails. In case of <welcome-file>index.html</welcome-file> it returns "/index.html". Setting welcome-file to "/" makes authentication work again, but breaks user experience.

NikitaZ: I don't think I want to set redirectURI to https://localhost:8181/myapp/faces/About.xhtml, it breaks the idea of welcome-file, if it changes one also have to change configuration at openID provider (Google Cloud in my case). Or we have to add a servlet for redirect URI and then forward to proper page which again reinvents welcome-file. Please, advice!

jGauravGupta: Thanks Nikita,

I also confirm the issue, if the callback URL is mapped to home. Updated the PR.

jGauravGupta:
jenkins test please

PR #4663, opened 2020-05-11
ECOSYS-150 Fix Hot Deploy codesmell

jGauravGupta:


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
```diff
@@ -1696,7 +1696,7 @@ private WebModule loadWebModule(VirtualServer virtualServer, WebModuleConfig web
                  */
                 webModule.setAvailable(true);
                 return webModule;
-            } else if (appState.get().isActive()) {
+            } else if (appState.filter(ApplicationState::isActive).isPresent()) {
```
svendiedrichsen, line 1699: Shouldn't this be rather .orElse(false). Not that it makes a funtional difference but it looks more explicit, doesn't it?

jGauravGupta:
jenkins test please

PR #4664, opened 2020-05-12
CUSTCOM-237 In Weld, parallel execution mode for async observers doesn't trigger all observers

MeroRai:
# Description
When a CDI event is fired in async mode with Weld-specific option to enable parallel execution of events, not all events are triggered as expected if one of them throws an exception.

# Testing

### Testing Performed
- Build test
- Jakarta TCK

### Testing Environment


pdudits:
Hm. Do we know why Weld is using single security context here, and why it doesn't disassociate it in between calls?

MeroRai:
@pdudits, the issue is two threads are running asynchronously and each of them calls the `associate` method before one of them has called the `disassociate` method. Hence, we get the `Security context is already associated` expectation because the `oldContext` is not removed

MeroRai:
jenkins test please

PR #4666, opened 2020-05-12
APPSERV-124 Monitoring-Console: Adds server side page configuration

jbee:
### Summary
This PR adds the page names (IDs) and values to the server side monitoring console configurations which is used by https://github.com/payara/monitoring-console/pull/11 to provide a fully integrated server side page configuration for MC. Requires https://github.com/payara/monitoring-console/pull/11 and https://github.com/payara/monitoring-console/pull/15 to be merged first and release 1.2 of MC implementation made.

### Testing
Assuming MC changes are released (or build locally) build this PR and install it.

* start server
* enable monitoring console
* open monitoring console webapp in the browser
* clear browsers local storage for the page
* refresh the page, check the role selection popup is shown

mc_dia_role

* select _User_ role
* check that the page synchronisation popup is shown

mc_dia_pages

* with no previous pages shared the list should be empty. press update or cancel.
* change the core page (e.g. add or remove a widget), open settings and expand the _Page_ settings, click _Push_ button.
* open MC in another browser, the page synchronisation now should show the changed and pushed page

Now the basic principle of page synchronisation was tested. There are many possible scenarios from here using the different roles.

As a _Guest_ the idea is that by default all updates shared by other are applied at the beginning so you work with the latest persistent settings. Guests change their local pages but don't share any changes. Buttons to push are not available. Automatic pushes do not occur.

As a _User_ the idea is that by default all updates are applied as long as they do not override a page that has been changed locally. This is reflected in the checkbox being checked or not by default when doing the page synchronisation. Obviously users can chose which pages to override or not. User can also manually push pages to the persistent configuration by using the _Push_ button in the _Page_ _Sync_ setting.

As an _Administrator_ the idea is that the role is used to configure the configuration that acts as base for other users. When pulling changes defaults work similar to users. But any change done to local page is automatically shared (pushed) when the _Page_ _Sync_ _auto_ setting is checked for the modified page (which it is by default). Administrator can also explicitly update already shared paged by using the _Update Remote Pages_ button in the _Global_ _Page Sync_ settings or push an individual page explicitly by using the _Push_ button (same as users).

Note also that once a user preferred a locally changed page over a remotely updated one this choice is considered in later updates by default that page will not be updated as long as the remote page is not updated with an even newer version.

In general "versions" are identical to moment of change. More recent changes are considered newer.  

See also the documentation https://github.com/payara/Payara-Server-Documentation-Antora/pull/6

### DIFF
--- a/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
+++ b/appserver/monitoring-console/core/src/main/java/fish/payara/monitoring/admin/SetMonitoringConsoleConfigurationCommand.java
@@ -115,6 +115,18 @@ public class SetMonitoringConsoleConfigurationCommand implements AdminCommand {
     @Param(optional = true, alias = "remove-watch")
     private String _removeWatch;
 
+    @SuppressWarnings("squid:S116")
+    @Param(optional = true, alias = "add-page-name")
+    private String _addPageName;
+
+    @SuppressWarnings("squid:S116")
+    @Param(optional = true, alias = "add-page-json")
+    private String _addPageJson;
+
+    @SuppressWarnings("squid:S116")
+    @Param(optional = true, alias = "remove-page")
+    private String _removePage;
+
     @Inject
     protected CommandRunner commandRunner;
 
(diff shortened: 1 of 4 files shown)


jbee:
Jenkins test please

PR #4667, opened 2020-05-12
CUSTCOM-262 HZ_LISTENER_PORT property error when setting system properties via the admin console

MeroRai:
# Description
When trying to set the values of one or more system properties for a non-DAS instance (or through its configuration) in the admin console, an error is thrown.

### Testing Performed

- Build test

### Testing Environment

# Documentation

# Notes for Reviewers
**Steps to reproduce:**
1. Start the default Payara Server domain.
2. Launch the admin console. Proceed to create a new configuration called instance-config that copies the default configuration.
3. Head to the System Properties option of this new configuration, add a new property called fish.payara.properties.sample and set its value to example. Hit the Save button, which will cause the error to appear.


MattGill98:
jenkins test please

PR #4692, opened 2020-06-02
QACI-242 Payara Samples Tests Are Attempting to Run Micro Tests for Remote Profile

AlanRoth:
# Description

Quick fix to stop micro tests on running on remote profile

# Testing

### Testing Performed

Tested and passed locally

Ran maven clean install;

```mvn -V -B -fae help:active-profiles clean install -f appserver/tests/payara-samples -Dpayara.version=5.2020.2-SNAPSHOT -Dpayara_domain=test-domain -Dglassfish.home=/home/alanroth/Workspace/Payara/appserver/distributions/payara/target/stage/payara5 -Ppayara-server-remote```

### Testing Environment

Openjdk 11.0.7
Ubuntu 64 bit
Maven 3.6.3

# Notes for Reviewers

This isn't designed to be a permanent solution, origionally planned on having a annotation like 'NotMicroCompatible' but for remote, 'NotRemoteCompatible'

### DIFF
--- a/appserver/tests/payara-samples/samples-programmatic/pom.xml
+++ b/appserver/tests/payara-samples/samples-programmatic/pom.xml
@@ -29,26 +29,4 @@
             <artifactId>junit</artifactId>
         </dependency>
     </dependencies>
-
-    <profiles>
-        <profile>
-            <id>payara-server-remote</id>
-            <properties>
-                <exclude.tests>**/programatic/**.java</exclude.tests>
-            </properties>
-            <build>
-                <plugins>
-                    <plugin>
-                        <artifactId>maven-surefire-plugin</artifactId>
-                        <configuration>
-                            <excludes>
-                                <exclude>${exclude.tests}</exclude>
-                            </excludes>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-    </profiles>
-
 </project>
--- a/appserver/tests/payara-samples/samples-programmatic/pom.xml
+++ b/appserver/tests/payara-samples/samples-programmatic/pom.xml
@@ -29,4 +29,26 @@
             <artifactId>junit</artifactId>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>payara-server-remote</id>
+            <properties>
+                <exclude.tests>**/programatic/**.java</exclude.tests>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>${exclude.tests}</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
 </project>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -36,11 +36,16 @@
     <modules>
         <module>test-utils</module>
         <module>samples</module>
-        <module>samples-programmatic</module>
         <module>classpath</module>
     </modules>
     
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: appserver/tests/payara-samples/samples-programmatic/pom.xml [hunk 1]
```diff
@@ -29,4 +29,26 @@
             <artifactId>junit</artifactId>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>payara-server-remote</id>
```
cubastanley, line 35: Samples is only run against remote so shouldn't need to have a separate profile for this - just in default

appserver/tests/payara-samples/samples-programmatic/pom.xml
```diff
@@ -29,4 +29,26 @@
...
+            <properties>
+                <exclude.tests>**/programatic/**.java</exclude.tests>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>${exclude.tests}</exclude>
```
cubastanley, line 45: Exclude explicitly rather than using a property? Or possibly exclude the module from the list of included modules if that's how the project is set up

appserver/tests/payara-samples/samples-programmatic/pom.xml
AlanRoth: Isn't the default profile managed?

appserver/tests/payara-samples/samples-programmatic/pom.xml
AlanRoth: We want the tests to run when not using remote profile

appserver/tests/payara-samples/samples-programmatic/pom.xml
cubastanley: Maybe in the poms but we only run samples on remote

appserver/tests/payara-samples/samples-programmatic/pom.xml
cubastanley: See above comment

AlanRoth: I could exclude the module instead, but I dont think its neccessery

appserver/tests/payara-samples/samples-programmatic/pom.xml
AlanRoth: Yes, it would be imprudent to assume that will always be the case though

AlanRoth: I completely understand what you mean though 😀

AlanRoth:
> 
> 
> What about
> `mvn clean verify -pl :payara-samples-profiled-tests -Ppayara-server-remote`
> instead of adding a profile where it makes no sense?

The mvn command I used is the one jenkins uses, which is where we had the problems, which is why it might not make much sense. 

```mvn clean verify -pl :payara-samples-profiled-tests -Ppayara-server-remote``` Builds successfully

appserver/tests/payara-samples/samples-programmatic/pom.xml
MarkWareham: going forward it needs to be run over all profiles

dmatej:
It is the reason why those "programmatical" tests are separated from "profiled" tests. Now you created a profile with the same name in "programmatical" build branch where those profiles intentionally don't exist.
The mistake was in jenkinsfile, not in Maven.

cubastanley:
Should we be changing the jenkins command for samples in that case then?

AlanRoth:
From the looks of it, yes

dmatej:
Approving too, despite too late ;)

PR #4693, opened 2020-06-02
APPSERV-152 Asadmin command to restart network listeners

jbee:
### Summary
Adds a command to restart the network listeners.

This uses the `GrizzlyService` to for the restart.

### Testing
The command only has a `target` parameter that by default targets the DAS.
To run the command just use

```
asadmin restart-http-listeners
```

### DIFF
--- a/appserver/web/admin/src/main/java/fish/payara/grizzly/config/admin/cli/RestartHttpListenersCommand.java
+++ b/appserver/web/admin/src/main/java/fish/payara/grizzly/config/admin/cli/RestartHttpListenersCommand.java
@@ -45,10 +45,13 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 
 import org.glassfish.api.ActionReport;
+import org.glassfish.api.ActionReport.ExitCode;
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.AdminCommand;
 import org.glassfish.api.admin.AdminCommandContext;
 import org.glassfish.api.admin.ExecuteOn;
+import org.glassfish.api.admin.FailurePolicy;
+import org.glassfish.api.admin.ParameterMap;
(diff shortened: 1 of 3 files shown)


jbee:
@Pandrex247 Selected you as reviewer as you basically already did review and test the PR.

appserver/web/admin/src/main/java/fish/payara/grizzly/config/admin/cli/RestartHttpListenersCommand.java
```diff
@@ -0,0 +1,141 @@
...
+    private String target;
+
+    @Param(name = "all", optional = true)
+    private Boolean all;
+
+    @Override
+    public void execute(AdminCommandContext context) {
+        ActionReport report = context.getActionReport();
+
+        if (all != null && target == null) {
```
Pandrex247, line 106: A bit of a odd situation, but this would technically still evaluate to true if someone ran:
`asadmin restart-http-listeners --all false`

appserver/web/admin/src/main/java/fish/payara/grizzly/config/admin/cli/RestartHttpListenersCommand.java
```diff
@@ -0,0 +1,141 @@
...
+    @Inject
+    private ServiceLocator locator;
+
+    @Inject
+    private Target targetUtil;
+
+    @Inject
+    private GrizzlyService service;
+
+    @Param(name = "target", optional = true)
```
Pandrex247, line 96:
```suggestion
    @Param(name = "target", optional = true, primary = true)
```
Save people writing out _--target_ - I think it's fairly safe to assume if someone types the below they mean an instance
`asadmin restart-http-listeners Flabbergasted-Blobfish`

jbee:
jenkins test please

PR #4727, opened 2020-06-18
QACI-262 Exclude problematic tests

AlanRoth:
# Description
This is a temporary band-aid for public jenkins failing Quicklook, which passes on private jenkins and locally. This will let some sort of testing to be peformed while a concrete solution is investigated and implemented.

# Testing
### Testing Performed
Ran on zulu-8, openjdk-8, maven 3.5.4, maven 3.6.3, but all resulted in the same problem.

### Testing Environment
Public jenkins
Amazon Linux
Maven 3.5.4/3.5.3
Zulu-8/Openjdk-8

# Notes for Reviewers
This is a temporary solution, quicklook failing is stopping from PRs being tested. This aims to enable some sort of functionality, however crude it may be.


appserver/tests/quicklook/ejb/singleton/build.properties
```diff
@@ -43,7 +43,7 @@ testsuite.name=EJB_Singleton
 testng.test.name=EJB_Singleton
 #possible choices war|ear|jar
 contextroot=singleton
-archive.type=ear
+archive.type=war
```
AlanRoth, line 46: Somewhat unrelated, EJB_Singleton was failing due to the file not being generated, changing to war generates the file and passes the test. Happy to revert if neccessery.

appserver/tests/quicklook/gfproject/db-targets.xml
```diff
@@ -146,16 +146,16 @@
 <target name="sql-common" depends="initprop">
   <echo message="sql.file = ${sql.file}"/>
   <echo message="db.class.path = ${db.class.path}"/>
-  <sql
-    driver="${db.driver}"
+  <sql driver="${db.driver}"
+    encoding="utf-8"
```
AlanRoth, line 150: Figured the main failure may be due to encoding - but no cigar.

AlanRoth:
Skip Quicklook altogether since most tests are being excluded anyway

PR #4730, opened 2020-06-19
QACI-262 Skip Quicklook

AlanRoth:
# Description
Skipping quicklook due to a peculiar problem with Quicklook failing ONLY on public jenkins. Skipping quicklook while investigating the root cause because PRs cannot be tested and merged due to this.

# Important Info
### Blockers
Why is it failing only on Public Jenkins?

# Notes for Reviewers
This is temporary change while the cause is investigated. It will allow PRs to be partially tested.


Pandrex247:
jenkins test please

PR #4732, opened 2020-06-19
QACI-262 - Remove exisiting EE8 .zip log before archiving new log

AlanRoth:
# Description
EE8 failed due to log already existing. Quicklook removed all .zips in its setup which is now commented out.


AlanRoth:
jenkins test please

PR #4735, opened 2020-06-22
Update readme for Payara Community

OndroMih:
# Description
Update readme for Payara Community.

Removed the sections about supported platforms as this is now maintained in Payara docs.


README.md
```diff
@@ -1,82 +1,42 @@
 ![#badassfish](payara-logo-blue.png)
 
-# Payara Server 5
+# Payara Server 5 Community Edition
```
OndroMih, line 3: Maybe we should remove the version number, what do you think?

Pandrex247: I'm happy to leave it.
On the other hand is there a reason we've got both _Payara Server 5 Community Edition_ and _Payara Platform 5 Community Edition_ headings?

OndroMih: The reason is that I didn't find a better solution. The original title was Payara Server but the second half of the page was about Platform (features shared by Server and Micro). So I added a new title for the Platform part. If it's OK to change the original title from Payara Server to Payara Platform then it's enough to have a single Payara Platform title.

OndroMih: Our docs at https://docs.payara.fish/ now say Payara Platform so it should be OK to change the main title to Payara Platform and remove the second title. Or add a separate title for the Payara Server section and also for the Payara Micro section. Then we would have the main title "Payara Platform" and 2 section titles "Payara Server" and "Payara Micro".

OndroMih:
Marketing is going to review this too, so please don't merge before their approval.

OndroMih:
Hi @Pandrex247, the marketing team have approved the current proposal. You can go ahead and merge this.

PR #4752, opened 2020-06-30
QACI-284 Remove expired certificates

AlanRoth:
It is what it is


AlanRoth:
jenkins test please

Pandrex247:
I don't think there are any expired certs?
Nothing comes up in the logs

AlanRoth:
> I don't think there are any expired certs?
> Nothing comes up in the logs

QACI-195 didn't get the expired certs PR

PR #4758, opened 2020-07-02
FISH-161 OpenAPI Class Data Reading using HK2 Class Model API 

jGauravGupta:
**FISH-161 OpenAPI Class Data Reading using HK2 Class Model API** 

## Description
This is a feature that replaces OPEN API class metadata reading from reflection API to HK2 Class Model API (underlying processing with ASM Visitor API)

### Depends on PR
https://github.com/payara/patched-src-hk2/pull/12
https://github.com/payara/patched-src-hk2/pull/13
https://github.com/payara/patched-src-hk2/pull/14
https://github.com/payara/Payara_PatchedProjects/pull/316

### Testing Performed

- MicroProfile OpenAPI TCK
- OpenAPI Impl unit tests
- Manual testing with a sample application

### Testing Environment
Windows 10, JDK 11.0

## Notes for Reviewers
Exceptionally Classloading is needed to invoke the `javax.ws.rs.core.Application#getClasses()`:
https://github.com/jGauravGupta/Payara/blob/f491727a90f87230faa0a45a256b9ebf979dcf82/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java#L274-L275

## Also Fixes
https://payara.atlassian.net/browse/FISH-70
https://payara.atlassian.net/browse/FISH-92
https://payara.atlassian.net/browse/FISH-50


MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/callbacks/CallbackImpl.java
```diff
@@ -39,29 +39,52 @@
...
 
 public class CallbackImpl extends ExtensibleTreeMap<PathItem, Callback> implements Callback {
 
     private static final long serialVersionUID = 5549098533131353142L;
 
     private String ref;
+    
+    private String urlExpression;
+
+    private List<Operation> operations;
```
jbee, line 66: I am surprised to see addition of these fields. Is this adding a feature that wasn't there before?

CC @MattGill98

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/ContentImpl.java
```diff
@@ -88,42 +120,49 @@ public void setMediaTypes(Map<String, MediaType> mediaTypes) {
         putAll(mediaTypes);
     }
 
-    public static void merge(org.eclipse.microprofile.openapi.annotations.media.Content from, Content to,
-            boolean override, Map<String, Schema> currentSchemas) {
+    public static void merge(Content from, Content to,
```
jbee, line 123: If this method only works with `ContentImpl` passed it maybe should demand it in is signature? If `createInstance` would return `ContentImpl` the caller joining the two maybe has all necessary type data?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/SchemaImpl.java
```diff
@@ -108,6 +110,133 @@
...
+        Schema to = null;
+        if (schemaClassName != null
+                && !schemaClassName.equals("java.lang.Void")) {
+            Type schemaType = context.getType(schemaClassName);
+            if (schemaType instanceof ClassModel) {
+                ClassModel schemaClassModel = (ClassModel) schemaType;
+                if (schemaClassModel.isInstanceOf(Schema.class.getName())) {
+                    try {
+                        Class<?> oneOfClass = context.getApplicationClassLoader().loadClass(schemaClassName);
+                        to = (Schema) oneOfClass.newInstance();
```
jbee, line 231: `return` avoids mutable state

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/SchemaImpl.java
```diff
@@ -346,6 +475,9 @@ public void setProperties(Map<String, Schema> properties) {
 
     @Override
     public Schema addProperty(String key, Schema propertiesItem) {
+        if (this.properties == null) {
+            this.properties = new HashMap<>();
```
jbee, line 479: Would it hurt to instead initialise the `properties` field with an empty map?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/AnnotationInfo.java
```diff
@@ -40,53 +40,61 @@
...
 /**
  * Represents the aggregated annotations on a type, its fields and methods including annotations "inherited" from
  * super-classes and implemented interfaces. Should a field or method from a super-class be overridden the
  * {@link Annotation} closest to the represented type (the overriding one) is kept.
  */
-public final class AnnotationInfo<T> {
+public final class AnnotationInfo {
 
-    private static final Map<Class<?>, AnnotationInfo<?>> TYPES = new ConcurrentHashMap<>();
+    private static final Map<ExtensibleType, AnnotationInfo> TYPES = new ConcurrentHashMap<>();
```
jbee, line 64: `ExtensibleType<?>` I think

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/util/ModelUtils.java
```diff
@@ -552,30 +577,44 @@ public static String getResourcePath(GenericDeclaration declaration, Map<String,
...
+                    return normaliseUrl(entry.getKey() + "/" + annotations.getAnnotationValue(Path.class));
                 }
             }
         }
         return null;
     }
 
-    private static String getResourcePath(Method method, Map<String, Set<Class<?>>> resourceMapping) {
-        AnnotationInfo<?> annotations = AnnotationInfo.valueOf(method.getDeclaringClass());
+    public static String getResourcePath(MethodModel method, Map<String, Set<String>> resourceMapping) {
```
jbee, line 602: Moving from `Class` to `String` does take away a lot of the context. Maybe this should be using a wrapper type on `String` to become more typed again.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -141,29 +114,65 @@
...
 
     /**
+     * A list of allowed classes for scanning
+     */
+    private final Set<Type> allowedTypes;
+
+    private final ClassLoader appClassLoader;
+
+    public ApplicationProcessor(ApplicationInfo appInfo) {
+        this.types = new HashSet<>(appInfo.getTypes().getAllTypes());
```
jbee, line 127: Should use the other constructor to set the fields.

jGauravGupta:
Hi @smillidge ,

OpenAPI metadata processing is initiated on the request to `/openapi` endpoint, so the application deployment time is not impacted at all.

See:
https://github.com/payara/Payara/blob/payara-server-5.2020.2/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/rest/app/service/OpenApiResource.java#L84
https://github.com/payara/Payara/blob/payara-server-5.2020.2/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java#L276

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
```diff
@@ -548,6 +548,7 @@ public void actOn(Logger logger) {
             // first, then register the real one
             appRegistry.remove(appName);
             appInfo.setIsJavaEEApp(sortedEngineInfos);
+            appInfo.setTypes(context.getTransientAppMetaData(Types.class.getName(), Types.class));
```
jbee, line 551: This should be lazily done on demand. If OpenAPI is requesting the types we still could compute this data from the sources used here or am I missing something?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/api/visitor/ApiContext.java
```diff
@@ -67,7 +68,19 @@
...
-    boolean isApplicationType(Class<?> type);
-}
\ No newline at end of file
+    boolean isApplicationType(String type);
+
+    /**
+     * @param type any class, not null
+     * @return type, if the give type is a known type in this context, else null
+     */
+    Type getType(String type);
```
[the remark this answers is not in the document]
jGauravGupta: `org.glassfish.hk2.classmodel.reflect.Type` HK2 API wrapped the ASM object and underling all processing is done with the ASM Visitor API by processing the bytecode of classes hence `Class.forName` is never called.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/ApplicationLifecycle.java
jGauravGupta: `org.glassfish.hk2.classmodel.reflect.Types` data was always created and loaded during application deployment and removed after deployments completed. 

Now `Types` info is stored in the `ApplicationInfo` for further processing by other modules like OpneAPI.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OpenAPIImpl.java
```diff
@@ -72,6 +69,41 @@
...
+        List<AnnotationModel> tags = annotation.getValue("tags", List.class);
+        if (tags != null) {
+            for (AnnotationModel tag : tags) {
+                from.getTags().add(TagImpl.createInstance(tag));
+            }
+        }
+        AnnotationModel components = annotation.getValue("components", AnnotationModel.class);
+        if (components != null) {
+            from.setComponents(ComponentsImpl.createInstance(components, context));
+        }
```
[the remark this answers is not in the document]
jGauravGupta: Added util function for List/Map attributes annotation mapping.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/callbacks/CallbackImpl.java
jGauravGupta: Few OpenAPI Annotations and Models class  (e.g Callback, Content, etc) differ in attributes but not the end results.
As earlier merge operation performed from Annotation (from) instance to Model (to), these attributes storage is not required. But with this PR HK2 types converted to Model (from) and merge operation is performed, hence locally these additional attributes are required.

jGauravGupta:
> What about if expected values do not exist? Are we settings null or rather empty string (as usually default in annotations).

if defaultValue not passed to the function `AnnotationModel#getValue` then the annotation type default value is returned.
````
<T> T getValue(String key, Class<T> type, Object defaultValue);
<T> T getValue(String key, Class<T> type);
````
See: https://github.com/payara/patched-src-hk2/blob/2.6.1.payara-maintenance/class-model/src/main/java/org/glassfish/hk2/classmodel/reflect/AnnotationModel.java#L47-L66

> As far as I can tell the way the class processing is hooked into the deployment makes it eagerly evaluated which isn't good. I would think one can make this so that the metadata is computed on-demand when OpenAPI actually needs it.

Yes, HK2 Class Model processing is always hooked with application deployment and used by the annotation processor during application deployment. 
Now, we are also storing the Class Model processed info and OpenAPI processing is performed on-demand on access to `/openapi` endpoint.

MarkWareham:
jenkins test please

jGauravGupta:
jenkins test please

PR #4772, opened 2020-07-07
FISH-7 Adds more MP Metrics Metadata as MC Annotation for Query Pages (Virtual Pages)

jbee:
Payara changes for https://github.com/payara/monitoring-console/pull/18, details can be found there.

In this PR the main goal is to provide two additional attributes in the permanent annotation:

* `BaseUnit`: The unit considered base unit for MP Metrics
* `ScaleToBaseUnit`: Factor to convert the metric unit to its base unit as used by MP Metrics

In order to share the code responsible for unit scale factor is extracted to a utility class.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricUnitsUtils.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricUnitsUtils.java
@@ -47,7 +47,11 @@ import org.eclipse.microprofile.metrics.MetricUnits;
  *
  * @author Jan Bernitt
  */
-public class MetricUnitsUtils {
+public final class MetricUnitsUtils {
+
+    private MetricUnitsUtils() {
+        throw new UnsupportedOperationException("util");
+    }
 
     public static String baseUnit(String unit) {
         switch (unit) {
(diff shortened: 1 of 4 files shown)


jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricUnitsUtils.java
```diff
@@ -0,0 +1,124 @@
...
+
+import org.eclipse.microprofile.metrics.Metadata;
+import org.eclipse.microprofile.metrics.MetricUnits;
+
+/**
+ * Utilities around {@link MetricUnits}.
+ *
+ * @author Jan Bernitt
+ */
+public class MetricUnitsUtils {
```
rdebusscher, line 50: Suggestion: Use final class and private default constructor to have align with utility type usage?

appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/MetricUnitsUtils.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public static Number scaleToBaseUnit(Number value, Metadata metadata) {
+        if (!metadata.getUnit().isPresent()) {
+            return value;
+        }
+        String unit = metadata.getUnit().get();
+        return scaleToBaseUnit(value, unit);
+    }
+
+    public static Number scaleToBaseUnit(Number value, String unit) {
```
rdebusscher, line 87: Suggestion: Unit Test

jbee:
@rdebusscher added tests and made class into a utility class.

jbee:
jenkins test please

PR #4776, opened 2020-07-08
FISH-186 Add Admin Console Integration for Certificate Management

Pandrex247:
##  Description
Community sync PR for https://github.com/payara/Payara-Enterprise/pull/107

Adds admin console integration for the integrated certificate management featureset.
This includes viewing the certificates present in the stores of key & trust stores, and adding/removing entries from the stores.

This integration takes the form of a new tab present under each _instance_ - **not** the config. This is due to keystores potentially being machine specific despite the instances themselves sharing the same config.
The page provides the user with a list of the various store entries (split into keystore and truststore entries), with details on the entry itself, the store it was retrieved from, and where this store is used.

From this table, a user can select a number of entries and remove them by clicking the _Delete_ button.
A user can also add entries to the key or trust stores by clicking _Add_ **on the appropriate table** (e.g. to add to the keystore, click Add on the keystore table.). This opens up a new page, where the user can define the new entry's alias, where the file to add is, and the listener target if desired.

### New Commands
Adds two new extra CLI commands: `list-keystore-entries` and `list-truststore-entries`

This PR also adds remote equivalents of a number of the certificate management CLI commands:
* `AddKeystoreEntryCommand` - remote equivalent of `AddToKeyStoreCommand`
* `AddTruststoreEntryCommand` - remote equivalent of `AddToTrustStoreCommand`
* `RemoveKeystoreEntryCommand` - remote equivalent of `RemoveFromKeyStoreCommand`
* `RemoveTruststoreEntryCommand` - remote equivalent of `RemoveFromTrustStoreCommand`
* `ListCertificatesCommand` - remote equivalent of `ListKeystoreEntriesCommand` **and** `ListTruststoreEntriesCommand`

All of these remote commands are hidden, so as to not cause confusion with their CLI equivalents.

### Notes
Adding entries via the admin console does **not** transfer keys across the wire, the commands are run locally on the target instance, and so the file must be present on the target instance in question. For example, if the DAS is on _machine1_, and an instance is on _machine2_, to add a certificate to the truststore of the instance on _machine2_, the certificate in question must be present on _machine2_.

I rejigged the _certificate-management_ module under extras into a parent module with children:
* `certificate-management-cli` - where the CLI commands live
* `certificate-management-admin` - where the remote admin commands live. Separate module so dependencies aren't being dragged into lib/asadmin unnecessarily
* `certificate-management-console-plugin` - the admin console integration
* `certificate-management-common` - module to share methods across all three modules to prevent unneccessary dependency bleed

Also contains a fix for adding to key/trust stores, where they couldn't add a bundle.

### Blockers 
Jonathan's PR likely needs merging first as I had to move the modules around - it will cause conflicts.
https://github.com/payara/Payara-Enterprise/pull/82

## Testing
Make sure that you test both the _server_ and _instances_ page. They're almost identical so vulnerable to copy-paste errors. 

Generating of certs and cert bundles can be done using the `generate-self-signed-certificate` command, [keytool commands](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/keytool.html), and/or using the openssl commands shown in this blog: https://blog.payara.fish/securing-payara-server-with-custom-ssl-certificate

### Testing Performed
There are likely combinations that I've missed, but this is what I've explictly tested. Behaviour should mirror the expected behaviour of the associated CLI commands: https://docs.payara.fish/enterprise/docs/5.20.0/documentation/payara-server/server-configuration/certificate-management.html

* Adding a cert to server keystore
* Adding a cert bundle to server keystore
* Adding a cert to server listener keystore
* Adding a cert to server listener truststore
* Adding a cert to server listener with custom keystore with non-master password
* Deleting a truststore entry
* Deleting multiple truststore entries
* Adding a cert to an instance keystore
* Removing a cert from an instance truststore

### Testing Environment
Windows 10, Zulu 8u252
OpenSUSE Tumbleweed, IcedTea 8u252

## Documentation
https://github.com/payara/Payara-Enterprise-Documentation/pull/24

## Notes for Reviewers
I had to essentially redo some of the logic around resolving the key and trust stores and their passwords for the remote commands. This was due to NodeRunner mandatorily printing the output to the log if you want to retrieve the output - when dealing with commands like `list-truststore-entries` this would obviously spam the log with all of the truststore entries which isn't exactly desirable.

I'm also not sure if it's just my particular setup of Windows machine calling an instance on a Linux VM, but loading of the certificates seems particularly slow - especially on remote instances. I may have overlooked something.


jbee:
@Pandrex247 did build and run your PR but the tabs for certificate management do not show (worked for Enterprise) - is there any special build flag needed to make it work?

Pandrex247:
Yes - Community doesn't have the packaging modules so it isn't included in the final distribution. The modules themselves also aren't compiled during a standard build.

You have to build the modules from the certificate-management directory itself (or build from the top-level with `-PBuildExtras`). After that you need to drop the _admin_, _console_, and _common_ jars into _payara5/glassfish/modules_, and the _cli_ jar into _payara5/glassfish/lib/asadmin_.

jbee:
jenkins test please again

jbee:
jenkins test please

PR #4764, opened 2020-07-06
FISH-56 OpenAPI document doesn't use @Schema when class is in jar dependency of the project

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>

## Description
This is a bug fix that dynamically picks and scans `@Schema` class, if not found in the list of selected classes for OpenAPI metadata processing but exists in the classpath.

### New tests
Unit  test
https://github.com/payara/Payara/blob/1bcbbd250b8204ac0ffc9cfdcc387b27dc9a3197/appserver/payara-appserver-modules/microprofile/openapi/src/test/java/fish/payara/microprofile/openapi/test/app/application/ExternalSchemaExampleTest.java#L68

### Testing Performed
Unit tests
MP OPEN API TCK Runner

### Testing Environment
Windows 10, JDK 11.0

PR #4777, opened 2020-07-09
FISH-162 OpenAPI Class Data Processing

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta@payara.fish>

## Description
This is a feature that includes refactoring of OpenAPI implementation by scanning OpenAPI metadata per class and support of the merging of multiple OpenAPI documents created via the deployment of more than one application.

### Testing Performed
Unit test
MP OpenAPI TCK Runner
Manually tested with 2 WARs

### Testing Environment
Windows 10, JDK 11.0

### Important Info
Merging document from multiple application works well for list attributes but attribute like 'info' is overridden by the document of last deployed application.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
@@ -196,27 +196,27 @@ public class OpenApiService implements PostConstruct, PreDestroy, EventListener,
     public OpenAPI getDocument() throws OpenAPIBuildException {
         if (mappings.isEmpty() || !isEnabled()) {
             return null;
-        } else if (mappings.size() == 1) {
+        }
+        if (mappings.size() == 1) {
             OpenAPI document = mappings.peekLast().getDocument();
-            if(document == null) {
-               document = mappings.peekLast().buildDocument();
+            if (document == null) {
+                document = mappings.peekLast().buildDocument();
             }
             return document;
-        } else {
-            List<OpenAPI> docs = new ArrayList<>();
-            for (OpenApiMapping mapping : mappings) {
-                if(mapping.getDocument() == null) {
-                    allDocuments = null;
-                    mapping.buildDocument();
-                }
-                docs.add(mapping.getDocument());
-            }
-            if(allDocuments == null) {
-                allDocuments = new OpenAPIImpl();
-                OpenAPIImpl.merge(allDocuments, docs, true);
-            }
-            return allDocuments;
         }
+        List<OpenAPI> docs = new ArrayList<>();
+        for (OpenApiMapping mapping : mappings) {
+            if (mapping.getDocument() == null) {
+                allDocuments = null;
+                mapping.buildDocument();
+            }
+            docs.add(mapping.getDocument());
+        }
+        if (allDocuments == null) {
+            allDocuments = new OpenAPIImpl();
+            OpenAPIImpl.merge(allDocuments, docs, true);
+        }
+        return allDocuments;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
```diff
@@ -174,28 +174,49 @@ public void event(Event<?> event) {
...
     public OpenAPI getDocument() throws OpenAPIBuildException {
         if (mappings.isEmpty() || !isEnabled()) {
             return null;
+        } else if (mappings.size() == 1) {
+            OpenAPI document = mappings.peekLast().getDocument();
+            if(document == null) {
+               document = mappings.peekLast().buildDocument();
+            }
+            return document;
+        } else {
```
jbee, line 205: suggestion: else-if and else not needed when you use return in the if-blocks.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
```diff
@@ -1087,8 +1075,8 @@ private boolean insertObjectReference(ApiContext context, Reference<?> referee,
                     // Create the schema
                     if (context.isAllowedType(referenceClassType)) {
                         visitSchema(schemaAnnotation, referenceClass, context);
-                    } else {
-                        apiWalker.processAnnotations(singleton(referenceClassType), Schema.class, this::visitSchema);
+                    } else if(referenceClassType instanceof ClassModel) {
+                        apiWalker.processAnnotation((ClassModel)referenceClassType, this);
```
jbee, line 1079: Makes me wonder if there is an `else` case now - is it an error? Should we maybe log something?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
```diff
@@ -104,211 +95,161 @@
...
-                    Set<Type> resourceClasses = new HashSet<>();
-                    mapping.put(key, resourceClasses);
-                    try {
-                        Class<?> clazz = appClassLoader.loadClass(classModel.getName());
-                        Application app = (Application) clazz.newInstance();
-                        // Add all classes contained in the application
-                        resourceClasses.addAll(app.getClasses()
-                                .stream()
-                                .map(Class::getName)
-                                .filter(name -> !name.startsWith("org.glassfish.jersey")) // Remove all Jersey providers
```
jbee, line 286: I assume filtering is no longer needed because of the new way we extract class information or it is done elsewhere now?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/processor/ApplicationProcessor.java
jGauravGupta: This comes from PR (https://github.com/payara/Payara/pull/4764) although we now support scanning for the packaged jar that's an optional feature.

jGauravGupta: I will add log statement, thx.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiWalker.java
jGauravGupta: Added here:
https://github.com/payara/Payara/blob/8d65b9ebcbfd989d06862e1661e10d5385695482/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/visitor/OpenApiContext.java#L166

jGauravGupta:
jenkins test please

PR #4780, opened 2020-07-10
QACI-249 Revert "Exclude ServerLogTest"

AlanRoth:
The warnings that failed ServerLogTest should be solved with FISH-31, which has been now merged.

Testing with this PR


AlanRoth:
Jenkins test please

PR #4798, opened 2020-07-20
Fix Cert Management Versions

Pandrex247:
## Description
Cert Management was missed from the version increment, leading to failures when building with the `-PBuildExtras` profile.


Pandrex247:
Jenkins test please

PR #4803, opened 2020-07-22
FISH-289 Create JDK11 Docker images

MattGill98:
The JDK11 Docker images are currently built in a separate Maven lifecycle with a manual system property override. This change causes the Docker Maven plugin to generate a '-jdk11' version of each Docker image in the same lifecycle.

### DIFF
--- a/appserver/extras/docker-images/pom.xml
+++ b/appserver/extras/docker-images/pom.xml
@@ -24,8 +24,8 @@
         <docker.noCache>true</docker.noCache>
 
         <docker.java.repository>azul/zulu-openjdk</docker.java.repository>
-        <docker.java.tag>8u232</docker.java.tag>
-        <docker.java.image>${docker.java.repository}:${docker.java.tag}</docker.java.image>
+        <docker.jdk8.tag>8u232</docker.jdk8.tag>
+        <docker.jdk11.tag>11.0.7</docker.jdk11.tag>
 
         <docker.payara.domainName>production</docker.payara.domainName>
         <docker.payara.rootDirectoryName>payara5</docker.payara.rootDirectoryName>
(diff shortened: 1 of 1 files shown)


appserver/extras/docker-images/pom.xml
```diff
@@ -91,7 +119,25 @@
...
+                                        </assembly>
+                                    </build>
+                                </image>
+                                <image>
+                                    <name>${docker.payara.repository}</name>
+                                    <build>
+                                        <!-- On Windows with default setting ${*}, $PATH would get filtered as well -->
+                                        <filter>@</filter>
+                                        <tags>
+                                            <tag>${docker.payara.tag}-jdk11</tag>
```
MarkWareham, line 136: There's no -jdk8 on the jdk8 image tag. perhaps it should be introduced for consistency?

dmatej:
Also original state allowed to use other images as experimental. Now it would be impossible.
Probably cleaner solution would be to create standalone maven modules - I expect there will be more differences between JDK8 and JDK11 in the future (Payara6 or 7?).

MattGill98:
@MarkWareham the antrun plugin is required to perform the filter as the Docker Maven plugin doesn't support ARG variables in the base image selection (as per the comment). It can't use the current maven property solution as it requires both to be built separately, which would need 2 maven lifecycles to change the base version.

Regarding naming - we've unfortunately already released with the current naming, which illustrates why decisions like this should be strategic, and not ad-hoc.

@dmatej Realistically both don't need testing. The logic being if our Docker image works with Payara - and Payara works with JDK11, the Docker image implicitly works with JDK11. Any problem with the base image would come up in the build.

As Payara should always work with multiple JDKs, there should be no reason for the image to be built differently. If we, in future, decide to change the JDK 11 image with changes to the default we can look at splitting them into different modules. As it stands though, the extra effort of maintaining twice the number of images didn't seem appropriate for the (currently) negligible benefits.

MattGill98:
jenkins test please

dmatej:
They do need testing - before me and Patrik nearly rewrote the Docker files, we distributed docker images which even were not able to start because of incompatible JVM options. Now we are going back again ... :-/

MarkWareham:
@dmatej  what you're talking about is out of scope of this change

dmatej:
> @dmatej what you're talking about is out of scope of this change

It is directly in this scope. Build will fail if you remove purge local docker repo.

dmatej:
EDIT: I wasn't right with the last comment, because we still use same tags for JDK8. But still - it wouldn't be so much work to cover also JDK11 images by tests.

FYI, this will delete all images, containers, etc, then maven will download what is configured to download):
```
docker system prune --all
mvn clean install -PBuildExtras,BuildDockerImages -pl :docker-images -amd
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14:16 min
[INFO] Finished at: 2020-07-22T17:23:20+02:00
[INFO] ------------------------------------------------------------------------
```

PR #4819, opened 2020-08-06
FISH-257 Adds RAG Status metrics and watches for MP Health Checks

jbee:
### Summary
To support the RAG indicator added by https://github.com/payara/monitoring-console/pull/20 each health check is made into a metric series. In addition aggregation series are created for:

* all checks of an app liveness
* all checks of an app readiness
* all checks of an app health
* all checks of an app (overall - that is liveness,  readiness or health)
* all checks of any app liveness
* all checks of any app readiness
* all checks of any app health
* all checks of any app (overall - that is liveness,  readiness or health)

The results of the health checks can carry additional key-value data which in case of a `DOWN` state are added as MC annotation.

Lastly each of the created series is also watched by a watch that is collected for it.

When checks are run they always run in their application class loader context.
They run each 12 sec as this is the interval the metrics are collected.

### Testing
See https://github.com/payara/monitoring-console/pull/20

Due to the newly introduced `Unit` adding the new metrics can be done without the changes of the referenced MC PR but the collection of the source will fail due to unknown unit. TL;DR; We could merge it independently.

### Documentation
Pending...

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
+++ b/appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
@@ -205,7 +205,7 @@ public class HealthCheckService implements EventListener, ConfigListener, Monito
 
     private static void addWatch(MonitoringWatchCollector collector, String appName, String metric) {
         String series = appName == null ? "ns:health " + metric : "ns:health @:" + appName + " " + metric;
-        String watchName = "RAG "+ metric + " [" + (appName == null ? "*" : appName) + "]";
+        String watchName = "RAG "+ metric + (appName == null ? "" : " @" + appName);
         collector.watch(series, watchName, "updown")
             .red(-1L, 5, false, null, 1, false)
             .amber(-1L, 1, false, null, 1, false)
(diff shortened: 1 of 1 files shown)


jbee:
jenkins test please

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -231,6 +370,7 @@ public void registerClassLoader(String appName, ClassLoader classloader) {
                         oldValue.addAll(newValue);
                         return oldValue;
                     });
+            //FIXME most likely this unintentionally changes the fields
```
jbee, line 373: This is a hit left for people looking into FISH-367

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -115,6 +125,8 @@
     private final Map<String, ClassLoader> applicationClassLoaders = new ConcurrentHashMap<>();
     private final List<String> applicationsLoaded = new CopyOnWriteArrayList<>();
 
+    private final AtomicReference<Map<String, Set<String>>> checksCollected = new AtomicReference<>();
```
jbee, line 128: FYI: This state is needed to know when collecting watches which metrics have been collected before.

PR #4834, opened 2020-08-12
FISH-82 Command Line option --warmup results in an Exception when Payara Micro instance stops when Request Tracing is activated

Pandrex247:
## Description
This is a bug fix.

Adds a check to the RequestTraceStoreFactory for whether the ClusteredStore service is actually active.

## Testing

`java -jar payara-micro.jar --enableRequestTracing --warmup --rootdir startup`
You should see no errors about ClusteredStore being started at _RunLevel_ 9 rather than 10.

### Testing Environment
Windows 10, Zulu JDK 8u262


Pandrex247:
Jenkins test please

PR #4835, opened 2020-08-13
FISH-277 Better Monitoring Console Integration into the Admin Console

jbee:
### Summary
This PR adds a couple of small connections to better integrate monitoring console into the admin GUI.

Now by default the monitoring for _Web_ and _HTTP_ is set to _HIGH_ and the monitoring console application is deployed on default **for the `production` domain**
These changes occur in the `domain.xml` template.

In the admin GUI links to the monitoring console are added to the _Common Tasks_ page as well as the _Monitoring Data_ home page. 
The URL is resolved from the application name in two steps: 

* First resolve the application `contextRoot` by application name (new handler method added), 
* secondly use application name and the resolved  `contextRoot` to URLs (HTTP and HTTPS) and use the HTTP one

The links do not show if MC is not deployed.

#### Testing Performed
* Start server `production` domain `asadmin stop-domain production` and open admin GUI
* click _Open Monitoring Console_ button on Common Tasks page and check that a working MC opens in new tab
* navigate to _Monitoring Data_, click the _Open Monitoring Console_ link and check that a working MC opens in new tab
* check that on the start page of MC all widgets show data (HTTP and Web Monitoring need to be enabled for that)
* create another instance, start it, check MC is not deployed on that instance
* use `asadmin set-monitoring-console-configuration --enabled=false` to disable MC, check the links are no longer shown
* use `asadmin set-monitoring-console-configuration --enabled=true` to enable MC, check the links are shown again
* Start server `domain1` domain, check that the links to MC do not show

### Documentation
https://github.com/payara/Payara-Community-Documentation/pull/56

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
@@ -93,7 +93,7 @@ public class ApplicationHandlers {
         Applications applications = Globals.getDefaultBaseServiceLocator().getService(Applications.class);
         String appName = (String) handlerCtx.getInputValue("appName");
         Application app = applications.getApplication(appName);
-        handlerCtx.setOutputValue(NAME_RESULT, app.getContextRoot());
+        handlerCtx.setOutputValue(NAME_RESULT, app == null ? null : app.getContextRoot());
     }
 
     /**
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2020] [Payara Foundation and/or its affiliates]
 /**
  *
  * @author anilam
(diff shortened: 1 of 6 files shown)


appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
```diff
@@ -80,6 +84,17 @@
...
+    @Handler(id = "py.getContextRoot",
+            input = {
+                @HandlerInput(name = "appName", type = String.class, required = true)},
+            output = {
+                @HandlerOutput(name = NAME_RESULT, type = String.class)})
+    public static void getContextRoot(HandlerContext handlerCtx) {
+        Applications applications = Globals.getDefaultBaseServiceLocator().getService(Applications.class);
+        String appName = (String) handlerCtx.getInputValue("appName");
+        Application app = applications.getApplication(appName);
+        handlerCtx.setOutputValue(NAME_RESULT, app.getContextRoot());
```
jbee, line 96: @Pandrex247 @MattGill98 brought up the question of including the virtual server context root here to create working URLs in case the virtual server uses a root path. WDYT?

jbee: Looking at `org.glassfish.admingui.common.handlers.ApplicationHandlers.getTargetURLList(String, String)` I'd think that line 772 takes care of it. So while this resolves the "plain" context root of the application the URLs are still resolved using the `getTargetURLList` handler which calls the referenced code which takes VS into account.

jbee:
jenkins test please

appserver/admingui/common/src/main/java/org/glassfish/admingui/common/handlers/ApplicationHandlers.java
jbee: This is still a relevant topic, I just changed the line for other reasons.

smillidge:
This should be targeted at Enterprise

jbee:
> This should be targeted at Enterprise

FISH-277 says otherwise, @Pandrex247 can you clarify? Way I understand it we bring this to Enterprise by first adding it to Community?

smillidge:
Nope it should be Enterprise

Pandrex247:
Should indeed be on Enterprise - issue was mislabelled

PR #4838, opened 2020-08-14
FISH-89 Address possible NPE during request tracing startup

Pandrex247:
## Description
This is a bug fix.

During the startup of the server with request tracing enabled, it was possible for an NPE to be thrown due to a check of `shouldStartTrace`. This simply enforces that the execution options have default values for everything rather than relying on the `bootstrapRequestTracing` method.

## Testing
Not really feasible due to race condition nature and lack of reproducer.
I cannae see how this could break anything though (famous last words) - what could go wrong?


Pandrex247:
Jenkins test please

PR #4841, opened 2020-08-18
QACI-195 Add MP TCKs to PR tests - Managed doesn't save logs

AlanRoth:
Caused enterprise to fail, missing change in community. Managed doesn't save the domain after testing


AlanRoth:
jenkins test please

AlanRoth:
> Would it not be better, though, to port the fix from community?

This is community

Issue #4350, opened 2019-11-26
JAX-WS 'Test webservice' does not work /CUSTCOM-190


rlfnb:
# Description #
----------

## Expected Outcome

http://localhost:8080/StupidWebService/StupidWebService?Tester
should show a page allowing me to call the web service

## Current Outcome

> Error generating artifacts for the following WSDL http://localhost:8080/StupidWebService/StupidWebService?WSDL
> 
> Possible causes can be invoking https when the application is not configured for security
> 

The root cause might be:

>   Payara Server  5.193.1 #badassfish (275) startup time : Felix (1,987ms), startup services(7,015ms), total(9,002ms)|#]
>   Payara Notification Service bootstrapped with configuration: NotificationExecutionOptions{enabled=true, notifierConfigurationExecutionOptionsList={LOG=LogNotifierConfigurationExecutionOptions{useSeparateLogFile=false} NotifierConfigurationExecutionOptions{notifierType=LOG, enabled=true, noisy=true}}}|#]
>   Network Listener JMS_PROXY_default_JMS_host started in: 6ms - bound to [/0.0.0.0:7676]|#]
>   Skipping registration of inhabitant for service reference [org.osgi.service.metatype.MetaTypeProvider] as the service object could not be obtained.|#]
>   JMXStartupService has started JMXConnector on JMXService URL service:jmx:rmi://0.0.0.0:8686/jndi/rmi://0.0.0.0:8686/jmxrmi|#]
>   Loading application __admingui done in 617 ms|#]
>   Initializing Mojarra 2.3.9.payara-p3 for context ''|#]
>   Loading application [__admingui] at [/]|#]
>   Context path from ServletContext:  differs from path from bundle: /|#]
>   Invoking wsimport with http://localhost:8080/StupidWebService/StupidWebService?WSDL|#]
>   parsing WSDL...
> |#]
>   
> Generating code...
> |#]
>   
> Compiling code...
> |#]
>   /tmp/jax-ws14817042957282565111tester/com/mycompany/jaxrspoc/StupidWebService.java:4: error: package javax.jws does not exist
> |#]
>   import javax.jws.WebMethod;
> |#]
>                   ^
> |#]
>   /tmp/jax-ws14817042957282565111tester/com/mycompany/jaxrspoc/StupidWebService.java:5: error: package javax.jws does not exist
> |#]
>   import javax.jws.WebParam;
> |#]
>                   ^
> |#]
[...]

PR #4837, opened 2020-08-14
FISH-90 Fix incorrect classpath in wscompile script

Pandrex247:
## Description
This is a bug fix.
The wscompile script hadn't been adapted to use the new jakarta jar.
I've also removed a classpath entry which doesn't seem to exist? (aixporting).

## Testing
Ran wscompile script and **didn't** get a lovely CNF exception.

### Testing Environment
Windows 10, JDK 8u262 & JDK 11.0.8

PR #4842, opened 2020-08-19
FISH-25 Fix SOAP Web Service Tester not working correctly on JDK 11

Pandrex247:
## Description
Bug fix.

Fixes the incorrect file in the webservices scripts, and adds the Jakarta JWS API Jar to the import script too as it seems to need it.
The tester servlet doesn't use these scripts however, and instead uses the Java APIs directly. To fix this, I've had to add classpath entries as a system property as it appears to not use the already loaded classes and instead gets the classpath specified during bootstrap (by default `${installRoot}/glassfish/modules/glassfish.jar` which is just the bootstrap jar).

Related to https://github.com/payara/Payara/pull/4837
Fixes https://github.com/payara/Payara/issues/4350

## Testing
Using https://github.com/rlfnb/JaxWSPoC, deploying, and hitting endpoint: http://localhost:8080/StupidWebService/StupidWebService?Tester

Should successfully import and not spew out a load of CNF exceptions.

I also checked from the command line with the following command:
`.\payara5\glassfish\bin\wsimport.bat -d .\testy\ -keep http://localhost:8080/StupidWebService/StupidWebService?WSDL -Xendorsed -target 2.1 -extension`

I've checked using embedded-all and it seemed happy (aside from an unrelated missing _allows access_) to import when I poked the tester endpoint so don't need to change the embedded domain.xml.
```
    public static void main(String[] args) {
        try {
            GlassFishRuntime runtime = GlassFishRuntime.bootstrap();
            GlassFishProperties glassfishProperties = new GlassFishProperties();
            GlassFish glassfish = runtime.newGlassFish(glassfishProperties);
            glassfish.start();
            glassfish.getDeployer().deploy(new File("C:\\Users\\pandr\\Downloads\\JaxRSPoC-1.0-SNAPSHOT.war"));
        } catch (GlassFishException ex) {
            Logger.getLogger(Main.class.getName()).log(Level.SEVERE, null, ex);
        }
    }
```

### Testing Environment
Windows 10, JDK 8u262 and 11.0.8

## Notes for Reviewers
Would appreciate a Linux test to check the classpath xml :)

I think this is also potentially a shakey solution, so an additional pair of eyes looking through Metro would be appreciated but I didn't see a nice point to hook in without a refactor (it very much seemed intentional that it specifically used the classpath system property).

The entry point into Metro is https://github.com/payara/Payara/blob/master/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/monitoring/WebServiceTesterServlet.java#L421

### DIFF
--- a/appserver/admin/gf_template/src/main/resources/config/domain.xml
+++ b/appserver/admin/gf_template/src/main/resources/config/domain.xml
@@ -264,6 +264,7 @@
                 <jvm-options>[|8]-Djava.endorsed.dirs=${com.sun.aas.installRoot}/modules/endorsed${path.separator}${com.sun.aas.installRoot}/lib/endorsed</jvm-options>
                 <jvm-options>[|8]-Djava.ext.dirs=${com.sun.aas.javaRoot}/lib/ext${path.separator}${com.sun.aas.javaRoot}/jre/lib/ext${path.separator}${com.sun.aas.instanceRoot}/lib/ext</jvm-options>
                 <jvm-options>[Azul-1.8.0u222|1.8.0u260]-XX:+UseOpenJSSE</jvm-options>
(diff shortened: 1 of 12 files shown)


Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

Cousjava (approved):

Cousjava:
This issue does not fix FISH-93

jGauravGupta:
jenkins test please

PR #4845, opened 2020-08-20
FISH-388 Fix NPE when printing out active span

Pandrex247:
## Description
This is a bug fix.

Currently if you try to use the _toString_ method on an active span (as in a span that hasn't actually finished yet) you get a NPE due to it not having an end time. This fixes this by simply not attempting to print this detail.

It also calculates a trace duration.

## Testing
Tested using Rudy's app.
You can also see this by evaluating an active span during debugging.


Pandrex247:
Failed to find namespace transformer? Not related at all to this - sounds like a version mismatch from a different PR

Pandrex247:
Jenkins test please

PR #4850, opened 2020-08-24
FISH-477 Deployment group properties are now resolved

Cousjava:
## Description
This is a bug fix

### Testing Performed
Manual tests, see Jira for reproducer

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 20.04 with Maven 3.6.3

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
@@ -37,6 +37,8 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2020] Payara Foundation and/or affiliates
+
 package com.sun.enterprise.v3.server;
 
 import com.sun.enterprise.config.serverbeans.Cluster;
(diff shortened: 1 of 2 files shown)


Cousjava:
Jenkins test please

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/SystemTasksImpl.java
```diff
@@ -173,15 +174,20 @@ private void setSystemPropertiesFromDomainXml() {
...
+        // 4. domain
         // so we need to add System Properties in *reverse order* to get the
         // right precedence.
 
         List<SystemProperty> domainSPList = domain.getSystemProperty();
         List<SystemProperty> configSPList = getConfigSystemProperties();
         Cluster cluster = server.getCluster();
         List<SystemProperty> clusterSPList = null;
-
+        List<DeploymentGroup> depGroups = server.getDeploymentGroup();
```
Pandrex247, line 186: Can this be null?
Having trouble reproducing atm

Cousjava: This cannot be null, but the list may be of length 0.

Cousjava: This cannot be null, but is can be a 0-sized list.

jGauravGupta:
Jenkins test please

Pandrex247 (changes requested):
This gets validated twice, once against the DAS config, and once against the target config - meaning you need to have the property set on both configs for it to pick it up. Ideally this should validate it against just the target config.

Pandrex247 (approved):

Cousjava:
Jenkins test please

PR #4851, opened 2020-08-25
FISH-392 Fix Trace getting started when Request Tracing is Disabled

Pandrex247:
## Description
Bug fix.
OpenTracingIiopServerInterceptor starts a trace even if request tracing is disabled - this is inconsistent with the behaviour of opentracing and request tracing everywhere else.

## Testing
Ran the remote-ejb-tracing sample. Before it didn't enable request tracing and would still pass against a clean domain - with the changes in this PR it should fail against a clean domain so I've added extra setup and teardown steps to the test.


Pandrex247:
Jenkins test please

Cousjava (approved):

jGauravGupta:
jenkins test please

PR #4853, opened 2020-08-25
FISH-389 Add Transaction ID as Span Baggage Item

Pandrex247:
## Description
Improvement.

Adds the transaction ID as a baggage item to the active span.
To accomodate this, the old-style request tracing was changed to be in the OpenTracing style.
The old style of attaching the log details to the currently active thread local trace is still present however for cases where we can't determine the tracer or active span.

### New tests
New test added to existing `remote-ejb-tracing` sample, checking that a transaction ID has been added as a baggage item.

## Documentation
Pending...

### DIFF
--- a/appserver/ejb/ejb-container/pom.xml
+++ b/appserver/ejb/ejb-container/pom.xml
@@ -40,7 +40,7 @@
     holder.
 
 -->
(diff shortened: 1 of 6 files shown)


MattGill98 (approved):

jGauravGupta:
jenkins test please

PR #4868, opened 2020-09-02
FISH-407 List of enabled application targets is not always correct

MeroRai:
## Description
When using the Maven cargo plugin to deploy an application on a deployment group, it is listed as not being enabled, when going to the endpoint of an instance in the deployment group show it is.

## Notes for Reviewers
Detail information on the reproducer is available on [FISH-87](https://payara.atlassian.net/browse/FISH-87).


Cousjava (changes requested):
Copyright notices need updating

Cousjava (approved):
Tested locally and works

jGauravGupta:
jenkins test please

PR #4872, opened 2020-09-04
Disable ServerLogTest until Grizzly upgrade

MattGill98:
Will otherwise fail until Grizzly is upgraded


jGauravGupta (approved):

dmatej:
I added link to this PR to a description of the QACI-406 so I will revert that with the upgrade.

PR #4874, opened 2020-09-07
FISH-426: Store initial request path when accessing protected path with OpenIdMechanism

rdebusscher:
## Description
New Feature: This allows the developer to go back to the original request URL by the user by configure in the Callback servlet the following line:

`response.sendRedirect(request.getSession().getAttribute(OpenIdConstant.ORIGINAL_REQUEST).toString());`

This is only a limited type of support, only 'replaying' a GET request.  Other types like POST request are not fully supported as the body is not stored.

### Testing Performed
Example application

### Testing Environment
Oracle 1.8.0_181 on Mac 10.14.6 with Maven 3.5.4

## Documentation
Documentation PR needs to be created


jGauravGupta:
jenkins test please

jGauravGupta (approved):

PR #4883, opened 2020-09-11
FISH-333 Add Payara HealthCheck Service checkers to MicroProfile Health Readiness Checks

MeroRai:
## Description
Added Payara HealthCheck Service checkers to MicroProfile Health Readiness Checks. The `set-healthcheck-service-configuration` asadmin command has a new parameter `display-on-health-endpoint`, which can be used to expose the Heathcheck Service checker readiness to MicroProfile Health Check endpoint. Below is an example of what the `/health/ready` endpoint returns:

```
{
  "status": "UP",
  "checks": [
    {
      "name": "HEAP",
      "status": "UP",
      "data": {
        "Message": "heap: init: 92.00 Mb, used: 204.93 Mb, committed: 332.00 Mb, max.: 455.50 Mbheap%: 45%",
        "HealthCheckStatus": "GOOD"
      }
    },
    {
      "name": "CPU-USAGE",
      "status": "UP",
      "data": {
        "Message": "CPU%: 6.42, Time CPU used: 5 minutes 11 seconds 774 milliseconds",
        "HealthCheckStatus": "GOOD"
      }
    }
  ]
}
```

### Testing Performed
- Build test
- MicroProfile Health TCKS (TODO)

### Testing Environment

## Documentation
TODO


appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/servlet/HealthCheckServlet.java
```diff
@@ -77,8 +83,28 @@ protected void processRequest(HttpServletRequest request, HttpServletResponse re
             return;
         }
 
-        healthCheckService.performHealthChecks(response, HealthCheckType.fromPath(request.getPathInfo()));
+        for (Checker checker : payaraHealthCheckService.getConfiguration().getCheckerList()) {
```
jbee, line 86: I think this whole loop does not belong in this class. If a `Checker` is configured to also occur in MP HealthCheck it should add a `HealthCheck` implementation to the service which is then processed like all other registered checks. If the  `Checker` configuration changes and the Payara health check no longer is exposed in MP the health check should be de-registered as part of the reaction to the configuration change, for example though listening to configuration change events.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -494,5 +502,25 @@ public UnprocessedChangeEvents changed(PropertyChangeEvent[] events) {
...
         return new UnprocessedChangeEvents(unchangedList);
     }
-
+    
+      /**
+     * Gets the application name from the invocation manager.
+     *
+     * @return The application name
+     */
+    public String getApplicationName() {
```
jbee, line 511: As far as I can see from the changes this method is never used in the service class but only called from the outside and it does not use any of the service specific state. It should not be here but where it is actually used or as it basically is a static method it could just as well be in a utility so it can be reused easier. I am pretty sure this exact method already exists multiple times in other places where the same issue exists. Would be nice if we finally make it a helper instead of duplicating the code again.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
```diff
@@ -98,15 +98,15 @@
     private final AtomicInteger checksDone = new AtomicInteger();
     private final AtomicInteger checksFailed = new AtomicInteger();
     private final AtomicBoolean inProcess = new AtomicBoolean(false);
-    private volatile HealthCheckResultStatus mostRecentCumulativeStatus;
+    private volatile HealthCheckResult mostRecentCumulativeResult;
```
jbee, line 101: The result is not _cumulative_. Its collecting entries in a list. The accumulation of the status of these entries is _cumulative_. So this should just be the `mostRecentResult`.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
```diff
@@ -121,7 +121,11 @@ public final HealthCheckResult doCheck() {
     public abstract O constructOptions(C c);
 
     public HealthCheckResultStatus getMostRecentCumulativeStatus() {
-        return mostRecentCumulativeStatus;
+        return mostRecentCumulativeResult.getCumulativeStatus();
+    }
+    
+     public HealthCheckResult getMostRecentCumulativeResult() {
```
jbee, line 127: and accordingly `getMostRecentResult()`

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
```diff
@@ -140,6 +144,10 @@ public boolean isReady() {
         O options = getOptions();
         return !isInProgress() && options != null && options.isEnabled();
     }
+    
+    public boolean isEnabled(){     
+        return getOptions().isEnabled();
```
jbee, line 149: pretty sure this can cause NPE. See `ìsReady()`. Also maybe `isReady()` can use the `isEnabled()` in its logic.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/checks/PayaraHealthCheck.java
```diff
@@ -0,0 +1,89 @@
...
+    private String name;
+    private String healthStatus;
+    private HealthCheckResult healthCheckResult;
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    public void setHealthStatus(String healthStatus) {
+        this.healthStatus = healthStatus;
```
jbee, line 63: The status is part of the result. If the result is used to update the check the status can be taken from that result and does not have to be set separately.

jbee (changes requested):
I think the way the bridge or adaptor between Payara and MP health checks is done needs to change. Configuration changes to the newly introduced flag should register or un-register the adaptor in MP health so that the checks can be processed as before without the servlet knowing about any special extra thing that needs to be done.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
MeroRai: `getOptions().isEnabled()` is a `boolean`, so it can't be `null`. The `isReady()` method already uses that `isEnabled()` method.

jbee: options can be `null`

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/checks/PayaraHealthCheck.java
```diff
@@ -0,0 +1,87 @@
...
+    private String name;
+    private BaseHealthCheck checker;
+
+    public PayaraHealthCheck(String name, BaseHealthCheck checker) {
+        this.name = name;
+        this.checker = checker;
+    }
+
+    public void setName(String name) {
+        this.name = name;
```
jbee, line 63: Is there a reason to make this modifiable? I would have thought setting name in constructor should be sufficient.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/checks/PayaraHealthCheck.java
```diff
@@ -0,0 +1,87 @@
...
+            responseBuilder.withData("Message", healthCheckResultEntry.getMessage());
+        }
+
+        String healthStatus = checker.getMostRecentCumulativeStatus().name();
+        if (!healthStatus.trim().isEmpty()) {
+            responseBuilder.withData("HealthCheckStatus", healthStatus);
+            if (healthStatus.equals("CRITICAL") || healthStatus.equals("CHECK_ERROR")) {
+                state = false;
+            }
+        }
```
jbee, line 82: Good candidate for a method returning the state.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/HealthCheckExecutionOptions.java
```diff
@@ -23,11 +23,13 @@ and Distribution License("CDDL") (collectively, the "License").  You
     private boolean enabled;
     private long time;
     private TimeUnit unit;
+    private boolean displayOnHealthEndpoint;
```
jbee, line 26: I am not sure about the wording. I think this is a flag to make the Payara Health Check appear on MP Health as well, isn't it?

jbee (commented):
I think using the events is a good solution to link the two health check systems.

Only important comment is the one about NPE, I am pretty sure I have encountered NPEs in the past because options were null at some point before the whole thing is tuned on. So have a look again. Might no longer be the case or apply but in general I'd think the issue should be the same in `isEnabled`.

Otherwise just some comments for consideration.

jbee (approved):
Susan an I discussed the wording and he will clarify that the flag refers to MicroProfile. No further comments.

MeroRai:
jenkins test please

Pandrex247:
Jenkins test please

MeroRai:
Jenkins test please

MeroRai:
Jenkins test please

MeroRai:
Jenkins test please

PR #4884, opened 2020-09-11
FISH-315 Implement New Notifier API

MattGill98:
## Description

This is a major refactor of the notification service. The intention is to allow modules depending solely on the internal-api module to send notifications and implement custom notifiers without either having knowledge of the notification service itself. Extra functionality / modifications include:

- PayaraExecutorService is now used across all notifiers, rather than each notifier implementing their own pooling.
- It's possible to implement dynamic notifiers without domain.xml that are always enabled.
- Implementing custom notifiers no longer requires implementations of: `MessageImpl`, `MessageQueue`, `NotificationEvent`, `NotificationEventFactory`, `NotifierConfigurationExecutionOptions`, `NotifierConfigurationExecutionOptionsFactory`, `NotifierExecutionOptions`, `NotifierExecutionOptionsFactory`.
- New `GlassFishStubBean` allows copying ConfigBeanProxy objects to be referenced statically (respect commands run without the dynamic flag).

### Blockers

Waiting for Payara Samples notifier fixes to work with the new API. Leaving here for initial review and testing anyway.

### Testing Performed

Manual testing of CDI eventbus notifier with the health check and request tracing services.

## Documentation

Not yet, TBD

## Notes for Reviewers

The intention is for the notifiers left out of this update to be moved into a different repository by having them depend on an internal-api artifact. Reviewing should keep this in mind - if it's not properly decoupled and version independent it doesn't work.


jbee (changes requested):
This is a big step in the right direction getting rid of a lot of ceremony while generalising and simplifying the concept and API of notifications. A lot of code could be deleted :+1: . Glad you also updated all the command names to follow the currently used naming conversions with `get-` and `set-`. 

The configuration classes and commands around them to me seem the least mature which is due to the limitation HK2 style configurations. My hopes were that we could make the configuration even more generic so less code would need virtual duplication. I highlight how HK2 config is cause of ceremony and duplication as I think this is equally the case in many other areas of the server and improvements to this concept would greatly improve what can be done and reduce what needs to be written to do so. 

Mostly general, API related comments...

* `Topic<PayaraNotification>` to me should be extracted to an interface so the actual implementation via HK2 topics is hidden to the sender making him not depend on HK2 APIs directly
* I hope `PayaraNotificationFactory` isn't really needed and that the builder and message could be created just using static methods, e.g. `PayaraNotificationBuilder.newBuilder().set. ... .build()`.
* Naming: not sure I am a big fan of prefixing the API classes with `Payara`, it is apparent from the package already but I guess it helps to not confuse old and new API
* currently a sender will need to use the `internal-api` `common` artefact. It might be worth even extracting just the notifier API to a API module on its own. @Pandrex247 WDYT?

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotification.java
```diff
@@ -52,60 +56,88 @@
     private String instanceName;
     private String subject;
     private String message;
+    private Serializable data;
```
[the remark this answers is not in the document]
MattGill98: This is a hangover from before. The API specifies that getData returns a typed version of the string message. I.e. the receiver is supposed to already know what to expect

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotification.java
```diff
@@ -52,60 +56,88 @@
...
+        this.blacklist = blacklist;
+    }
+
+    public Serializable getData() {
+        return data;
+    }
+
+    protected void setData(Serializable data) {
+        this.data = data;
+    }
```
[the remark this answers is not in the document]
MattGill98: It's immutable outside of the builder API with one notable caveat - if the notification is extended. I figured that since it has to be serializable that's not too much of an issue. The use case would be writing your own notification sender with extra field that your own receiver uses

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotification.java
```diff
@@ -36,14 +37,17 @@
...
+import java.util.List;
 
 /**
- * Base class for notification events.
- *
- * @author mertcaliskan
+ * A notification received by notifiers and the notification service
  */
-public abstract class NotificationEvent {
+public class PayaraNotification implements Serializable {
```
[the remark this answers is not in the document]
MattGill98: I don't see any problem with extending it - see my comment below

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotificationFactory.java
```diff
@@ -37,95 +37,51 @@
...
  * @author mertcaliskan
  * @since 4.1.2.171
  */
-@Contract
-public abstract class NotificationEventFactory<E extends NotificationEvent> {
-
-    @Inject
-    NotificationEventFactoryStore store;
+@Service
+public class PayaraNotificationFactory {
```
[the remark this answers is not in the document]
MattGill98: It needs some HK2 services to be injected, and as I'm not a fan of the Globals class usage this seems to be the only clean option unless you can think of another way?

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
```diff
@@ -0,0 +1,119 @@
...
+        final PayaraNotification notification = notificationQueue.peek();
+        try {
+            if (notification != null && isEnabled()) {
+                notifier.handleNotification(notification);
+                notificationQueue.remove();
+            }
+        } catch (Exception ex) {
+            LOGGER.log(Level.WARNING,
+                    format("Notifier %s failed to handle notification \"%s\".", notifierName, notification), ex);
+        }
```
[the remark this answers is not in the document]
MattGill98: It's worth bearing in mind that each thread run loops through every single notifier, so I wasn't sure if trying to flush every single notification every time might have the potential to stall the other notifiers

Cousjava:
This fails to compile on:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project rest-service: Compilation failure: Compilation failure: 
[ERROR] /home/jcoustick/NetBeansProjects/Payara/nucleus/admin/rest/rest-service/src/test/java/fish/payara/admin/rest/utils/RunCommandTest.java:[60,40] package fish.payara.nucleus.notification does not exist
```

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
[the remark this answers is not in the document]
jbee: I understand your concern but I can't see how the notifiers actually would affect each other directly other then competing for CPU in general. If we want to enforce fairness beyond OS multi-threading I suggest to look into a single worker thread multi-dispatch where any message enters a single queue and the worker thread pulls from there and dispatches directly to the `PayaraNotifier`s. So instead of a `NotifierHandler` per type of `PayaraNotifier` there is only the single dispatch for all of them.

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/NotifierManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A contract for the notification service, which allows the internal API to
+ * send events to the notification service without having a direct reference to
+ * it.
+ */
+@Contract
+public interface NotifierManager {
+
+    void reconfigureNotifier(PayaraNotifierConfiguration configuration);
```
jbee, line 52: Looking at this I wonder how or where the target notifier is identified. I assume there is something in the `configuration` that allows this. Maybe the javadoc can explain a but how this method is used and what it does.

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotification.java
```diff
@@ -117,20 +117,20 @@ protected void setMessage(String message) {
...
-    public List<String> getBlacklist() {
-        return blacklist;
+    public List<String> getNotifierBlacklist() {
+        return notifierBlacklist;
     }
 
-    protected void setBlacklist(List<String> blacklist) {
-        this.blacklist = blacklist;
+    protected void setNotifierBlacklist(List<String> blacklist) {
+        this.notifierBlacklist = blacklist;
```
jbee: I see the setters are still in and the builder is mostly a fluent API alias for calling setters on the notification. I really would like to see this using builder with fields for the individual values and a notification with a constructor accepting all the fields called on `build()`.

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
```diff
@@ -285,7 +286,7 @@ public void sendNotification(String name, HealthCheckResult checkResult, Level l
         String subject = "Health Check notification with severity level: " + level.getName();
         String messageFormatted = getMessageFormatted(new Object[]{name, getCumulativeMessages(checkResult.getEntries())});
 
-        List<String> enabledNotifiers = healthCheckService.getEnabledNotifiers();
+        Collection<String> enabledNotifiers = healthCheckService.getEnabledNotifiers();
```
jbee, line 289: Just noticed that factory in line below is also still on. Is there a reason to not have the `newBuilder` be a static method in the builder class?

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotificationService.java
```diff
@@ -190,7 +192,9 @@ public void shutdownNotificationService() {
         if (execution != null) {
             execution.cancel(true);
         }
-        notifiers.forEach(NotifierHandler::destroy);
+        if (enabled) {
```
jbee, line 195: Not so sure about this guard. If `notifiers` isn't empty shouldn't the `destroy` run even when not disabled? What if I do the bootstrap, disable the service and then get here, shouldn't we still `destroy`?

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
```diff
@@ -61,34 +61,75 @@
...
+
     protected void destroy() {
         notifier.destroy();
     }
 
+    @SuppressWarnings("unchecked")
     protected void bootstrap() {
+        // Set the configuration before bootstrapping the notifier
+        if (config != null) {
+            PayaraConfiguredNotifier.class.cast(notifier).setConfiguration(config);
```
jbee, line 126: This is the kind of _set_ that you will not see coming should this not behave as expected and you start looking. I wonder if there is a better way where the config and the notifier interact a bit more obvious with each other.

jbee (commented):
A couple of more comments, mostly details. 

I once more try to convince you to use immutable messages and to create the builder using a static method in the builder class itself (and to get rid of the factory class) :smiley_cat:

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/preliminary/BaseHealthCheck.java
MattGill98: The factory is needed to inject the HK2 properties required in the notification. I've reworked the builder API so it should be more clear now!

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
MattGill98: I know what you mean, but I can't currently think of a better way. Any ideas?

Cousjava:
Fails to build, failing on test with:
```
[ERROR] Tests run: 23, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.453 s <<< FAILURE! - in fish.payara.nucleus.microprofile.config.spi.PayaraConfigTest
[ERROR] convertedValuesAreCached(fish.payara.nucleus.microprofile.config.spi.PayaraConfigTest)  Time elapsed: 0.1 s  <<< FAILURE!
java.lang.AssertionError: Run 8, value is not cached expected:<2> but was:<4>
	at fish.payara.nucleus.microprofile.config.spi.PayaraConfigTest.assertValue(PayaraConfigTest.java:252)
	at fish.payara.nucleus.microprofile.config.spi.PayaraConfigTest.assertCachedValue(PayaraConfigTest.java:243)
	at fish.payara.nucleus.microprofile.config.spi.PayaraConfigTest.convertedValuesAreCached(PayaraConfigTest.java:105)
```

MattGill98:
jenkins test please

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
jbee: No, would need to checkout the codebase and browse around this a little to get a better understanding of the relations involved. But as long as we are aware of this I think we could also just give it some time and at some point in the future a batter way becomes apparent to us - then we just have to also do the change :)

jbee (approved):
Looks good. Have not run any testing as it is my understanding that you mostly wanted conceptual feedback. Correct me if I am wrong here.

Cousjava (changes requested):
Fails for me, with error:
```
GRIZZLY0200: Service exception
        at java.util.Collections$UnmodifiableCollection.<init>(Collections.java:1028)
        at java.util.Collections$UnmodifiableList.<init>(Collections.java:1304)
        at java.util.Collections.unmodifiableList(Collections.java:1289)
        at fish.payara.internal.notification.PayaraNotification.<init>(PayaraNotification.java:77)
        at fish.payara.internal.notification.PayaraNotificationBuilder.build(PayaraNotificationBuilder.java:140)
        at fish.payara.nucleus.requesttracing.RequestTracingService.processTraceEnd(RequestTracingService.java:520)
        at fish.payara.nucleus.requesttracing.RequestTracingService.endTrace(RequestTracingService.java:463)
```
Which occurred by enabled notification, enabling request tracing with timeout of 300 milliseconds, including non-applications, and store longest traces; then click around in the admin console a couple of times.

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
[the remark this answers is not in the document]
jbee: Oh I wasn't looking close enough and assumed from your comment that the notifiers would each be scheduled so that they actually do compete with each other in some form - Now I had a closer look and saw that there is only a single job scheduled that sequentially works the notifiers. So your concern was that working a single notifier in its entirety could delay working the other notifiers later in the sequence.

So what I suggested was more of a round robin. Forget the `NotifierHandler` class exist. Think that in the service you have a list of notifiers and a (generic) queue of messages to deliver. The method called periodically has a loop that runs until the queue is empty. It pulls one message from the queue and forwards it to each notifier that should receive it. This way all notifier will always see the first message before any can see the second which can be said is quite fair. On the downside a notifier that is very slow in its message processing will slow down message processing for all other notifiers. A problem the current loop has as well just in another form. 

Which brings me back to what I thought initially the solution was - that each notifier would have its own queue and its own scheduled task. As this would solve "stalling" messages for other notifiers. Each would in their task method try to send all messages of the queue. Now the notifiers compete for CPU and co but given we want messages going through the system as fast as possible this is a good thing. But we still might delay message processing because of the scheduling. This makes me think: why do we use scheduled tasks when it can be event-reaction that are asynchronously coupled by a queue.

I think what we want is:
* the thread creating a message to be send should not be delayed
* notifiers do not affect each other message delivery
* messages are delivered as promptly as possible
* messages are delivered in order they are handed to the service (for each notifier)

With this in mind I think a queue already decouples a message source from sending the message so the scheduling is just causing unnecessary delay and work. The message processor or processors should just block wait on the queue. A solution that has all the qualities above is a bit more complicated. 
1. Source puts messages in a common queue.
2. a worker block waits on that queue, it takes the message out of the queue and puts it in a queue for each notifier.
3. each notifier specific queue also has a worker task block waiting on the queue that forwards the message to the notifier
4. notifier does its notifier specific thing

Only downside I see here is that this needs n+1 worker threads for n notifier types. But they mostly be waiting.
This can be reduced and made more dynamically scaling by not doing one second level queue for each notifier but a second level queue with work items where a work item is created for each pair of message and target notifier. In that case the sequence is:

1. Source puts messages in a common queue.
2. a worker block waits on that queue, it takes the message out of the first queue, creates a work item each receiving notifier and puts it in the second level queue.
3. another (work stealing) pool of workers is used to process these work items as fast as possible (no scheduling). 

Should there be slow items because of slow notifier implementations the workers in the pool "steal" the work of later items in the queue to some degree. Downside is that order can no longer be guaranteed. Which makes me think that 1 thread per notifier is the best option as usually the number of actually enabled notifiers is low.

MattGill98:
@Cousjava has reported that request tracing sends more notifications than before. We're both investigating

jbee:
> request tracing sends more notifications than before. We're both investigating

I would not be surprised if this is more correct then before - keep an open mind for this possibility ;)

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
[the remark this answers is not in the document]
MattGill98: I've created a thread for each notifier that runs with a 5ms delay but waits on the queue to fill up - I hope this addresses the threading problems!

MattGill98:
jenkins test please

nucleus/payara-modules/notification-core/src/main/java/fish/payara/nucleus/notification/NotifierHandler.java
```diff
@@ -146,17 +162,20 @@ public void accept(final PayaraNotification notification) {
...
-    @Override
-    public void run() {
-        final PayaraNotification notification = notificationQueue.peek();
+    private void run() {
+        assert isEnabled();
         try {
-            if (notification != null && isEnabled()) {
-                notifier.handleNotification(notification);
-                notificationQueue.remove();
+            final PayaraNotification notification = notificationQueue.take();
```
Cousjava, line 168: take() is a blocking operation, and you are calling it every 5 milliseconds, so that will end up with a lot waiting on this if there aren't many notifications being sent, the opposite problem of what was wrong before. I think a better solution would be to flush the queue every 50 milliseconds or something like that.

Cousjava (changes requested):
Saving the status of notifiers in the admin console does not work.

MattGill98:
Nice spot 👍 Just modified the API to help reduce the risk of incorrect extension (as I did). Testing and will push ASAP.

MattGill98:
jenkins test please

Cousjava:
Fails to build:
```
[ERROR] Failed to execute goal on project notification-email: Could not resolve dependencies for project fish.payara.server.internal.packager:notification-email:distribution-fragment:5.2020.5-SNAPSHOT: Could not find artifact fish.payara.server.internal.payara-appserver-modules:notification-email-core:jar:5.2020.5-SNAPSHOT
```

MattGill98:
Which module is that in? I’ve built with BuildExtras and not come across that?

MattGill98:
jenkins test please

Cousjava (approved):
All seems to work for me

MattGill98:
Thanks @jbee and @Cousjava!

PR #4909, opened 2020-09-25
[QACI-445] - Don't fail EE7 if no tests

cubastanley:
Don't fail the EE7 jenkins test if there's no tests to run


cubastanley:
jenkins test please

MarkWareham:
This solve the problem. It only turns a script error into a test failure

PR #4914, opened 2020-09-25
FISH-25 FISH-93 Fix SOAP Web Service Tester not working correctly and wsgen NoClassDefFoundError with JDK 11

Pandrex247:
## Description
Bug fix. Revert of [revert](https://github.com/payara/Payara/pull/4885) of original fixes, plus a change.

### Blockers
None.

### Testing Performed
Using https://github.com/rlfnb/JaxWSPoC, deploying, and hitting endpoint: http://localhost:8080/StupidWebService/StupidWebService?Tester

Should successfully import and not spew out a load of CNF exceptions.

Also ran through the ejb30/lite/overridexml TCK which is what failed last time and it passed - the JDK specific java options present in the domain.xml don't seem to ever get deactivated when using the StaticGlassFishRuntime, so got around this by simply specifying the required JARs. I was always a little unsure of using the java.class.path property so while also fragile this is a nicer solution.

### Testing Environment
Windows 10, Zulu JDK 11.0.8

## Documentation
Nope

## Notes for Reviewers
Narp


Pandrex247:
Jenkins test please

Cousjava (approved):

PR #4918, opened 2020-09-27
FISH-334 Integrate MicroProfile Metrics into Payara Health Service

MeroRai:
## Description
Integrated MicroProfile Metrics into Payara Health service. This provides a way to monitor and log the values of metrics exposed by MicroProfile Metrics.

### Testing Performed
Manually tested by monitoring metrics for different scopes, using MicroProfile Metrics sample application available in payara samples.

## Documentation
TODO


appserver/payara-appserver-modules/healthcheck-metrics/src/main/java/fish/payara/healthcheck/microprofile/metrics/MicroProfileMetricsCheck.java
```diff
@@ -0,0 +1,251 @@
...
+
+    @Override
+    protected String getDescription() {
+        return "healthcheck.description.MPmetrics";
+    }
+
+    @Override
+    protected HealthCheckResult doCheckInternal() {
+        String registryName = options.getMetricsScope();
+        List<String> metricNames = new ArrayList<>(Arrays.asList(options.getMetricName().split(",")));
```
MattGill98, line 105: I'd remove extra whitespace here

appserver/payara-appserver-modules/healthcheck-metrics/src/main/java/fish/payara/healthcheck/microprofile/metrics/MicroProfileMetricsCheck.java
```diff
@@ -0,0 +1,251 @@
...
+        metricNames.removeAll(Arrays.asList("", null));
```
MattGill98, line 106: This is slow, I'd remove it from the string when removing whitespace to save 2 passes

appserver/payara-appserver-modules/healthcheck-metrics/src/main/java/fish/payara/healthcheck/microprofile/metrics/MicroProfileMetricsCheck.java
```diff
@@ -0,0 +1,251 @@
...
+        MetricsService metricsService = Globals.getDefaultBaseServiceLocator().getService(MetricsService.class);
```
MattGill98, line 107: Is there any reason you're injecting the metrics service fresh each time?

appserver/payara-appserver-modules/healthcheck-metrics/src/main/java/fish/payara/healthcheck/microprofile/metrics/MicroProfileMetricsCheck.java
```diff
@@ -0,0 +1,251 @@
...
+                if (metricsInfos != null) {
+                    array.add(metricsInfos);
+                }
+            }
+        }
+        return array;
+    }
+
+    private String getMetricInfos(String metricName, MetricRegistry state) {
+        Map<MetricID, Metric> metricInfos = state.getMetrics().entrySet().stream()
```
MattGill98, line 145: Does this need repeating each loop? Also, users won't be familiar with the dot notation of the metrics - they're mote likely to use underscore notation as it appears in the metrics result

appserver/payara-appserver-modules/healthcheck-metrics/src/main/java/fish/payara/healthcheck/microprofile/metrics/MicroProfileMetricsCheck.java
```diff
@@ -0,0 +1,251 @@
...
+        MetricsService metricsService = Globals.getDefaultBaseServiceLocator().getService(MetricsService.class);
+        if (registryName.equalsIgnoreCase("application")) {
+            registryName = options.getMetricsApplicationName();
+        }
+
+        HealthCheckResult result = new HealthCheckResult();
+        List<String> metrics = collectMetrics(metricsService.getRegistry(registryName), metricNames);
+
+        result.add(new HealthCheckResultEntry(metrics.isEmpty() ? WARNING : GOOD,
+                metrics.isEmpty() ? "The metric you entered doesn't exist under " + registryName : metrics.stream().map(Object::toString).collect(Collectors.joining())));
```
MattGill98, line 116: Can we have a few more variables? I don't find the streaming particularly intuitive here, and the result contains excess whitespace

nucleus/payara-modules/healthcheck-core/src/main/java/fish/payara/nucleus/healthcheck/configuration/MicroProfileMetricsChecker.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Attribute
+    String getMetricsScope();
+    void setMetricsScope(String metricsSource) throws PropertyVetoException;
+    
+    @Attribute
+    String getMetricApplicationName();
+    void setMetricApplicationName(String metricsType) throws PropertyVetoException;
+    
+    @Attribute
```
MattGill98, line 67: This should be an actual list IMO (`@Element("metric")`), to prevent comma separated strings in XML files - which can cause all sorts of issues!

MattGill98 (changes requested):
Broadly good, well done! Just a few comments which shouldn't require too much invasive change.

The get-command doesn't return the metric checker name in the admin console (so the text field is always empty. Check the field name from the additional properties map.

The syntax of the specified metric is weird, I'd expect underscore notation as the metric appears. Also, why am I limited to specifying only app metrics or only base metrics etc (I suspect it's because of the dot notation as above)? Does the admin console provide any sort of standard way for specifying lists? I really personally hate this string list strategy.

Finally, the processing is a little unoptimised from what I can tell. Where the process is currently:

1. Get measured metric names
2. Get all metrics
3. Filter all metrics by name
4. Concatenate result with the value of each remaining metric

I think it should be:
1. Get names of measured metrics
2. Get those individual metric values (or at least break early to prevent processing each metric in the stream)
3. Concatenate the values

In other words, most of this logic can be removed from the main processing loop

MattGill98 (changes requested):
Not happy with the format for specifying metrics, so I've made a PR to Susan's branch which he's agreed to incorporate into this

MeroRai:
jenkins test please

Pandrex247:
Jenkins test please

MeroRai:
Jenkins test please

MattGill98 (approved):

MeroRai:
Jenkins test please

PR #4919, opened 2020-09-28
FISH-356 All Payara Health Checks Define Green State

jbee:
### Summary
This is a tiny improvement that makes sure all of the Payara Health Checks define a Green state.

### Testing
Start server, enable Payara health checks, open MC and navigate to RAG/Traffic Light Status page.
Check that both hogging and stuck threads are shown as green instead of white.


Pandrex247 (approved):

jbee:
jenkins test please

PR #4931, opened 2020-10-01
FISH-474 display vendor if specified in jvm-options

Cousjava:
## Description
This is a bug fix.

This applies both in the admin console and asadmin command.
Also allows for saving of JVM options with vendor info.

### Testing Performed
Run the `list-jvm-options` asadmin command and see that the vendor is listed
Go the the jvm options and add and edit a jvm option that has a vendor as part of the minimum JDK version.

### Testing Environment
Zulu JDK 1.8_262 on Ubuntu 20.04 with Maven 3.6.3


Pandrex247 (approved):

Cousjava:
Jenkins test please

PR #4940, opened 2020-10-09
Attempt at fixing DI TCK Error

MattGill98:
Looks to be caused by the MetricsService being initialised too early,
most likely by the new health checker. This change lazy initialises the
metrics service when a health check is run. Hopefully this will be late
enough that the metrics service has started.


MarkWareham:
Does not fix the failure

PR #4762, opened 2020-07-05
Update Hzl API to 4.1, include tenant control, add APIs for EE context

lprimak:
## Description
Context for this is https://github.com/hazelcast/hazelcast/issues/10875 and all the issues mentioned within this issue as well.
Main feature is that Hazelcast Cache / Map / Queue etc. now work with Java EE constructs, and CDI beans, in Payara Data Grid as well.

### Blockers
https://github.com/hazelcast/hazelcast/pull/17184 needs to be merged and Hz 4.1 released

### New tests
Got https://github.com/payara/Payara/issues/759 / https://github.com/bgjug/jcache-workshop to work in clustered environment using Payara Micro and Server mix.
Most of this PR is removing a whole bunch of Proxy classes that are no longer needed

### Testing Performed
All test suites pass. Manual testing with distributed app mentioned above

### Testing Environment
JDK 8

## Documentation
Bug fixes, features work in more environments

PR #4948, opened 2020-10-13
FISH-40 EJB timers with @Clustered will execute on one and only one instance

### LINKED ISSUE #4572: Making Clustered Singleton working as documented with Payara Micro Cluster /FISH-40
# Description #
----------
I am trying to make use of [`@Clustered`](https://docs.payara.fish/documentation/payara-server/public-api/clustered-singleton.html) annotation with `@EJB` or `@ApplicationScoped` beans but in a vain. Can you please advise on this or update Examples Project so it can be more clear how to implement `@Clustered` feature in a cluster?

My test scenarios are based on the [Payara Clustered Singleton examples](https://github.com/payara/Payara-Examples/tree/master/javaee/clustered-singleton).
For all test scenarios:
1) I only add the class `SchedulerTest` as shown below
2) Build the project
3) Launching it as (on the same machine):
i) `java -jar payara-micro-5.201.jar --deploy d:\_2\Payara-Examples\javaee\clustered-singleton\target\clustered-singleton-1.0-SNAPSHOT.war --name i1 --port 8091`
ii) `java -jar payara-micro-5.201.jar --deploy d:\_2\Payara-Examples\javaee\clustered-singleton\target\clustered-singleton-1.0-SNAPSHOT.war --name i2 --port 8092`

I have prepared several test scenarios, will start with one of them:

## Code
```
@Clustered(callPostConstructOnAttach = false)
@Singleton
public class SchedulerTest implements Serializable {

    private Set<String> _eventHistoryList;

    @PostConstruct
    protected void init() {
        System.out.println(">>> INIT...");
        _eventHistoryList = new HashSet();
    }

    @Schedule(second = "*/5", minute = "*", hour = "*", info = "testSched1", persistent = true)
    public void testSched1() {
        SimpleDateFormat sdf = new SimpleDateFormat("yyyyMMdd-HHmmss");
        String jobId = sdf.format(new Date());
        System.out.println(">> " + getEventHistoryList());
        if (!getEventHistoryList().contains(jobId)) {
            getEventHistoryList().add(jobId);
            System.out.println(">>>>> executing jobId: " + jobId);
        }
    }

    public Set<String> getEventHistoryList() {
        return _eventHistoryList;
    }
}
```

## Expected Outcome

According to the documentation:
> Persistent EJB Timers (which are the default) will work correctly with clustered singletons since they will only be executed in a single node of the cluster

It is expected that method `testSched1()` is to be executed only on one node in the cluster. Because that didn't happen - I have tried to add some manual synchronization, but that didn't help much as well.

## Current Outcome

Timer is executed on both cluster nodes. Also added manual sync doesn't help as well.
test1

## Environment ##

- **Payara Version**: 5.201
- **Edition**: Micro
- **JDK Version**: Zulu 8.44.0.11-CA-win64
- **Operating System**:  Windows 10 x64 and Linux CentOS 7/8

Cousjava:
## Description
This is a bug fix. Fixes #4572 .

### Testing Performed
Use the example application provided with Jira. Start three Payara Micro instance with the app deployed. The timer should start on one of them. Then kill that instance. The timer should restart on one and only one other instance.

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 18.04 with Maven 3.6.3

## Notes for Reviewers
The use of ILock is deprecated and removed in Hazelcast 4. #4762 contains the Hazelcast upgrade and that PR will need to be updated to take account of this one.

### DIFF
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBTimerService.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBTimerService.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2020] [Payara Foundation and/or its affiliates]
 package com.sun.ejb.containers;
 
 import java.io.Serializable;
@@ -941,7 +941,7 @@ public abstract class EJBTimerService {
 
     /**
      * Create automatic non-persistent timers defined by the @Schedule 
-     * annotation on the EJB bean. Recover part is a no-op in thise case.
+     * annotation on the EJB bean. Recover part is a no-op in this case.
      * 
      * @return a Map of created timers, where the key is TimerPrimaryKey 
      * and the value is the Method to be executed by the container when the timer with
@@ -1552,6 +1552,17 @@ public abstract class EJBTimerService {
             }
         }
     }
+    
+    /**
+     * Sets failed times to be rescheduled rather than removed.
+     * This is called by {@link fish.payara.ejb.timer.hazelcast.HazelcastTimerStore},
+     * as when an instance is removed from a cluster it is possible that a delivery
+     * could fail in the time in which an instance is removed from the cluster and before
+     * the timer is recreated.
+     */
+    protected void enableRescheduleTimers() {
+        rescheduleFailedTimer = true;
+    }
 
     /**
      * Checks whether this timer is owned by the server instance in
@@ -1708,7 +1719,7 @@ public abstract class EJBTimerService {
     public static class TimerCache {
 
         // Maps timer id to timer state.
-        private Map timers_;
+        private Map<TimerPrimaryKey, RuntimeTimerState> timers_;
 
         // Map of timer information per container.
         //
@@ -1722,7 +1733,7 @@ public abstract class EJBTimerService {
         // in the case where the same entity bean identity has more
         // than one associated timer.
         
-        private Map containerTimers_;
+        private Map<Long, Object> containerTimers_;
 
         // Map of non-persistent timer id to timer state.
         private Map<TimerPrimaryKey, RuntimeTimerState> nonpersistentTimers_;
@@ -1730,9 +1741,9 @@ public abstract class EJBTimerService {
         public TimerCache() {
             // Create unsynchronized collections.  TimerCache will 
             // provide concurrency control.
-            timers_ = new HashMap();
-            containerTimers_ = new HashMap();
-            nonpersistentTimers_ = new HashMap<TimerPrimaryKey, RuntimeTimerState>();
+            timers_ = new HashMap<>();
+            containerTimers_ = new HashMap<>();
+            nonpersistentTimers_ = new HashMap<>();
         }
 
         public synchronized void addTimer(TimerPrimaryKey timerId, 
@@ -1756,7 +1767,7 @@ public abstract class EJBTimerService {
                     // NOTE : This list *can* contain duplicates, since
                     // the same entity bean can be the timed object for 
                     // multiple timers.
-                    entityBeans = new ArrayList();
+                    entityBeans = new ArrayList<>();
                     containerTimers_.put(containerId, entityBeans);
                 } else {
                     entityBeans = (Collection) containerInfo;
@@ -1764,7 +1775,7 @@ public abstract class EJBTimerService {
                 entityBeans.add(timerState.getTimedObjectPrimaryKey());
             } else {
                 Long timerCount = (containerInfo == null) ? 1 :
-                    ((Long) containerInfo).longValue() + 1;
+                    ((Long) containerInfo) + 1;
                 containerTimers_.put(containerId, timerCount);
             }
 
@@ -1780,8 +1791,7 @@ public abstract class EJBTimerService {
                 logger.log(Level.FINE, "Removing timer {0}", timerId);
             }
 
-            RuntimeTimerState timerState = (RuntimeTimerState) 
-                timers_.remove(timerId);
+            RuntimeTimerState timerState = timers_.remove(timerId);
 
             if( timerState == null) {
                 return;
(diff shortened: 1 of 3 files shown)


Cousjava:
Jenkins test please

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
```diff
@@ -96,6 +101,7 @@ static void init(HazelcastCore core) {
         }
     }
 
+    @SuppressWarnings("LeakingThisInConstructor")
```
MattGill98, line 104: Hmmmm really don't like this. Can the DataGridEjbTimerService register this class when it creates it?

MattGill98 (changes requested):
One change

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
```diff
@@ -1193,4 +1157,37 @@ private boolean restoreEJBTimers() {
...
     }
 
+    @Override
+    public void memberAdded(MemberEvent event) {
+        //do nothing
+    }
+
+    @Override
+    public void memberRemoved(MemberEvent event) {
+        ILock hazelcastLock = Globals.getDefaultBaseServiceLocator().getService(HazelcastCore.class).getInstance().getLock("EJB-TIMER-LOCK");
```
jbee, line 1167: Isn't there a field holding the `HazelcastCore` that was passed to the constructor?

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
```diff
@@ -107,6 +113,8 @@ public HazelcastTimerStore(HazelcastCore core) throws Exception {
         serverName = core.getInstance().getCluster().getLocalMember().getStringAttribute(HazelcastCore.INSTANCE_ATTRIBUTE);
         this.ownerIdOfThisServer_ = serverName;
         this.domainName_ = core.getInstance().getConfig().getGroupConfig().getName();
+        super.enableRescheduleTimers();
+        Globals.getDefaultBaseServiceLocator().getService(PayaraCluster.class).addClusterListener(this);
```
jbee, line 117: Should the cluster be injected instead of using `Globals`?

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
```diff
@@ -1193,4 +1157,37 @@ private boolean restoreEJBTimers() {
...
+            
+            
+            
+            if (!removedTimers.isEmpty()) {
+                logger.log(Level.INFO, "==> Restoring Timers ... ");
+                Collection<HZTimer> restored = _restoreTimers(removedTimers);
+                for (HZTimer timer : restored) {
+                    pkCache.putAsync(timer.getKey().getTimerId(), timer);
+                }
+                logger.log(Level.INFO, "<== ... Timers Restored.");
```
jbee, line 1186: This could become a bit confusing in the logs as it suggest that at this point the timers are restored which might not yet be the case due to `putAsync`.

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
jbee: @MattGill98 I assume you refer to the `Globals` access and registering this as a cluster listener? Can you motivate why you think this should occur in the `DataGridEjbTimerService` for me? I don't see the connection but I also don't know the details of the timer infrastructure.

jbee (commented):
Good work with updating the logging and generics.
The open question seems to be how to register the store as a cluster listener. Looks like this needs discussion. I suspect there are some difficulties that drove the solution via `Globals`?

appserver/payara-appserver-modules/hazelcast-ejb-timer/src/main/java/fish/payara/ejb/timer/hazelcast/HazelcastTimerStore.java
MattGill98: That's the class that initialises this one 😉

Cousjava:
Jenkins test please

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EJBTimerService.java
```diff
@@ -1552,6 +1552,17 @@ protected void cancelTimerSynchronization(EJBContextImpl context_,
...
+    
+    /**
+     * Sets failed times to be rescheduled rather than removed.
+     * This is called by {@link fish.payara.ejb.timer.hazelcast.HazelcastTimerStore},
+     * as when an instance is removed from a cluster it is possible that a delivery
+     * could fail in the time in which an instance is removed from the cluster and before
+     * the timer is recreated.
+     */
+    protected void enableRescheduleTimers() {
+        rescheduleFailedTimer = true;
```
jbee, line 1564: I really wonder if this field shouldn't be thread-safe. This class has a number of fields I'd think should be thread-safe as I would assume that methods are called by different threads. Is there any reasoning why this would not be the case?

Cousjava: This method does not need to be thread-safe as it is only called by the constructor of `HazelcastTimerStore` which will only occur once on startup. The field effectively final, the but I can't make the field final due to the `initProperties()` method also setting this based on the config, which is what `HazelcastTimerStore` overrides to always be true (as that is the point of a clustered timer).

jbee: I see - technically I agree it doesn't need it. But how should someone know looking at the code unless he also knows what you just explained or finds this out by analysing the code. A short comment would help.

Cousjava: There is a comment on the method already

jbee (approved):

MattGill98 (approved):
Nice work!

PR #4950, opened 2020-10-14
FISH-152 Mark Applications as Unavailable as soon as they are Disabled

Pandrex247:
## Description
Improvement.

It was possible to get errors if an application endpoint was accessed during disable/undeployment. This improvement should help to alleviate these by marking the context for an application as unavailable (returning an 503 error when accessed) during the disable process while it is being unloaded and cleared away.

## Important Info
There is possibly a race condition present here due to processing the new _Disable_Start_ event asynchronously, but from all investigations and testing it's not possible for anything bad to happen. If the web container disables the context first, that's the goal we're aiming for and everything is bon. If the app engine and its modules are torn down before the context can be disabled (this _should_ be slower), then it either does nothing or simply logs a message at level "fine" because the application is already gone (meaning the context we wanted to mark as unavailable is too).

### New tests
None due to the sensitive timing nature of this.

### Testing Performed
The general format I used for testing was to spin up a number of terminal windows that constantly poked a deployed application's endpoint, and checking the output and server logs as I disabled or undeployed the app. In Powershell, I used this (adjusting port and endpoint as required):

```
while ($true) { 
   curl http://localhost:8080/endpoint/
}
```

I've tested the following scenarios:

* Wars:
  * Deploying the venerable clusterjsp application to the DAS, and undeploying (target set to _server_ & _domain_)
  * Deploying the clusterjsp application to an instance, and undeploying
  * Deploying the clusterjsp application to the DAS and an instance, and undeploying from both
  * Deploying the clusterjsp application to the DAS and an instance, and undeploying from one
  * Deploying the clusterjsp application to the DAS, and disabling (target set to _server_ & _domain_)
  * Deploying the clusterjsp application to an instance, and disabling
  * Deploying the clusterjsp application to the DAS and an instance, and disabling on both
  * Deploying the clusterjsp application to the DAS and an instance, and disabling on one
  * Deploying the clusterjsp application to the DAS, and disabling
* EJBs:
  * Deploying an EJB with no endpoint to the DAS and an instance, and undeploying from both
* Ears:
  * Deploying the Ear attached to FISH-242 to the DAS (after cherry-picking the associated fix so it actually deploys), and undeploying from both
  * Deploying the Ear attached to FISH-242 to the DAS (after cherry-picking the associated fix so it actually deploys), and disabling on both

### Testing Environment
Windows 10, Zulu JDK 8

## Documentation
N/A

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
@@ -633,13 +633,17 @@ public class WebContainer implements org.glassfish.api.container.Container, Post
             ApplicationInfo applicationInfo = (ApplicationInfo) event.hook();
             String appName = applicationInfo.getName();
             Applications applications = serviceLocator.getService(Applications.class);
+            String contextRoot = applications.getApplication(appName).getContextRoot();
 
-            StringBuilder stringBuilder = new StringBuilder();
-            // The final comma gets stripped out in suspendWebModule
-            getVirtualServers().forEach(virtualServer -> stringBuilder.append(virtualServer.getID()).append(","));
-            // Just provide all virtual hosts, it'll only suspend the app on those where it's deployed
-            suspendWebModule(applications.getApplication(appName).getContextRoot(), appName, stringBuilder.toString(),
-                    true);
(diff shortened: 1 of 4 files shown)


Pandrex247:
Jenkins test please

MattGill98 (approved):
LGTM

PR #4962, opened 2020-10-21
FISH-628 Remove expired certs at point of including from JDK install

MarkWareham:
## Description
This is an improvement. 

The process for removing expired certs is manual and tedious. This prevents that. During domain creation there's existing code to copy existing certs from the JDK. I have adjusted so it will also remove expired certs at this point.

### New tests
Updated existing unit tests. keystore that's created now copies a sample keystore with an expired cert

### Testing Performed
run unit tests

### Test suites executed

Unit tests.

### Testing Environment

Apache Maven 3.6.3, Java version: 11.0.8, vendor: Ubuntu. OS name: "linux", version: "5.4.0-48-generic", arch: "amd64"

## Documentation

none applicable

## Notes for Reviewers

Rather than create highly duplicated method getInvalidCertificateAliases that does the opposite of getValidCertificateAliases I have extracted the functionality to another method and the original code now contains a conditional. The original and new method signatures exist but simply call this extracted method

Test updated. keystore that's created now copies a sample keystore with an expired cert rather than use an empty keystore


Pandrex247:
Jenkins test please

MattGill98:
@MarkWareham feel free to merge this

PR #4966, opened 2020-10-27
FISH-327 Implement MicroProfile Sniffers

MattGill98:
## Description

This is an improvement. MicroProfile servlets and CDI extensions are now hidden behind a sniffer engine. This should mean that they won't startup unnecessarily, and should be a decent performance improvement for applications not utilising any or all of the MP specs.

### Blockers

**Needs testing against large non-MP app.**

### Testing Performed

Tested the test app for FISH-242.
Passes MP TCKs.

## Notes for Reviewers

Should also fix FISH-431.

**Make sure https://github.com/payara/Payara/commit/103806009f126e8e3f48c43c7c5911915cfc210e is still relevant**


MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/config/src/main/java/fish/payara/microprofile/config/ConfigSniffer.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Class<? extends Annotation>[] getAnnotationTypes() {
+        Class<? extends Annotation>[] annotations = new Class[1];
+
+        // Search for ConfigProperty annotation used in the CDI extension
+        annotations[0] = org.eclipse.microprofile.config.inject.ConfigProperty.class;
+
+        return annotations;
```
jbee, line 61: much nicer using an array literal, something like
```java
return { org.eclipse.microprofile.config.inject.ConfigProperty.class }
```
Maybe also use a import for `ConfigProperty`?

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceApplicationContainer.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean resume() throws Exception {
+        return true;
+    }
+
+    @Override
+    public boolean suspend() {
+        return true;
+    }
```
jbee, line 71: Why override the methods identical to how they are implemented in the super class?

jbee (changes requested):
First impressions feedback...

Classes extending `MicroProfileSniffer` have some duplication. Instead of implementing the getters in each one with constants the abstract class could have fields holding the values returned by the getters which are initialised in the constructor.

The classes extending `MicroProfileApplicationContainer` all implement the methods with `return true` while this is already how it is implemented in the abstract class - seems to be a case of unnecessary override.

Not sure why the `CdiExtention` did not get named differently for the different standards while almost all other classes have their standard prefix. I think for consistency and less confusion these should likewise be named e.g. `JwtAuthCdiExtension` and so forth (some actually have a prefix I found).

I am not sure if the idea of a lazy bootstrapping of MP is even compatible with MP standards as standards do require you to throw `DeploymentException`s in certain situations. Would this still work with the indirection added?

Maybe you can add a section to the description explaining how the general concept of sniffer, deployer, container and so forth do their magic? What would happen before, what is happing now in comparison.

MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceApplicationContainer.java
MattGill98: These aren't implemented in the superclass which is why they're required. I'm hesitant to implement them in the super class, as thinking abstractly it doesn't make sense to me to have these as the default implementation. I want to remove the instances of "magic" that make the class harder to understand. I can add some javadocs to explain what these methods are doing?

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceApplicationContainer.java
jbee: Must have become confused by all the new classes. I thought I saw these also being implemented in a base class. So you say these are implemented identical several times for different reasons?

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceApplicationContainer.java
MattGill98: The `start` and `stop` methods are stubs in a couple of cases, but the `resume` and `suspend` methods may be stubs everywhere. I'll implement default `resume` and `suspend` methods.

MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenAPISupplier.java
```diff
@@ -0,0 +1,264 @@
...
+                doc = new FileProcessor(classLoader).process(doc, config);
+                doc = new ApplicationProcessor(
+                        types,
+                        filterTypes(archive, config, types),
+                        classLoader
+                ).process(doc, config);
+                doc = new BaseProcessor(baseURLs).process(doc, config);
+                doc = new FilterProcessor().process(doc, config);
+            } finally {
+                this.document = doc;
```
jbee, line 140: Does this work like as intended should an exception be thrown at some of the steps so that the document contains as much information as possible to that point?

jbee (approved):
It is a lot of ceremony needed here but my best guess is that all of it is required.

Not sure you have commented on one of my previous comments on `DeploymentException` required by MP standards. I assume we would detect usage of MP and bootstrap it which then would cause the exception to be thrown as required?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenAPISupplier.java
MattGill98: My changes don't (see "shouldn't") change any functionality when the sniffer returns `true` for the application. All DeploymentExceptions related to MP deployment should still propagate, this just disables all CDI extensions and processing when no MP or JAX-RS annotations are detected

MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/healthcheck/pom.xml
```diff
@@ -67,6 +67,11 @@
             <groupId>org.eclipse.microprofile.health</groupId>
             <artifactId>microprofile-health-api</artifactId>
         </dependency>
+        <dependency>
```
Cousjava, line 70: Copyright year needs updating

Cousjava (changes requested):

Cousjava (approved):

MattGill98:
The previous commit passed Jenkins and this was just a copyright update, so I'm going to merge it to not hold up CI

PR #4986, opened 2020-11-10
FISH-660 Upgrade MicroProfile Health Implementation to 3.0-RC5

MattGill98:
## Description

This is a feature. Will need upgrading to the final release version when it's release, but for now the release candidate is good enough.

### New tests

TCK updated: https://github.com/payara/MicroProfile-TCK-Runners/pull/122

### Testing Performed

Ran the new TCK against our current implementation

## Documentation

TBW


appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
```diff
@@ -116,16 +115,11 @@
...
-    @Inject
-    private ConfigProviderResolver microConfigResolver;
-
-    private boolean backwardCompEnabled;
+    private MicroprofileHealthCheckConfiguration configuration;
 
     private static final Logger LOG = Logger.getLogger(HealthCheckService.class.getName());
 
-    private final Map<String, Set<HealthCheck>> health = new ConcurrentHashMap<>();
+    private final Map<String, Set<HealthCheck>> health = unmodifiableMap(new HashMap<>());
```
jbee, line 122: Can't wrap my head around this one. How is an empty modifiable map of any use? Why would a concurrent map be wrong in a shared service? How is this only applying to `health` and not equally to `readiness` and `liveness`?
Maybe you can shine some light on this for me.

jbee (commented):
One big question mark where I need some help understanding the change.

appserver/payara-appserver-modules/microprofile/healthcheck/src/main/java/fish/payara/microprofile/healthcheck/HealthCheckService.java
MattGill98: I did it in the interest of "non invasive changes" whilst still trying to make it clear that this isn't modified anymore. Perhaps it would be more readable if I just use an explicit new HashMap each time instead. I'll refactor this

MarkWareham:
jenkins test please

jbee (approved):
Have not tested it yet but looks good to me now.

I checked InSight docs and don't think it needs an update. It was general enough to not be affected by this change.

cubastanley:
Jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
jenkins test please

MattGill98:
Fails because the TCK is randomly downgraded, waiting on https://github.com/payara/MicroProfile-TCK-Runners/pull/125

MattGill98:
jenkins test please

MattGill98:
jenkins test please

PR #5008, opened 2020-11-26
Added proprietary property to disable verification of token's "typ" c…

fturizo:
# FISH-615 Allow JWT verification to skip type validation

## Description
This is a minor improvement to make the MP JWT Authentication module to accept tokens generated without the `typ` claim in its header. Currently, the specification requires this claim to be present at all times, but NOT all modern third-party services generate tokens with this claim, as the corresponding RFC for JWT sets it as optional:

> 5.1.  "typ" (Type) Header Parameter
>
> The "typ" (type) Header Parameter defined by [JWS] and [JWE] is used by JWT applications to declare the media type [IANA.MediaTypes] of this complete JWT.  This is intended for use by the JWT application when values that are not JWTs could also be present in an application data structure that can contain a JWT object; the application can use this value to disambiguate among the different kinds of objects that might be present. 
> ...
> Use of this Header Parameter is OPTIONAL.

This PR introduces a custom property called disable.type.verification that turn’s this validation `ON/OFF`, which can be defined in the custom payara-mp.-jwt.properties file. 

It is important to note that next versions of the specification are already defining "recommended" claims, which means that this behaviour will become standardized in the next iteration of the specification.

### Testing Performed
No tests present as there is no testing infrastructure for the MP JWT Auth module

## Documentation
Documentation PR: https://github.com/payara/Payara-Community-Documentation/pull/115

## Notes for Reviewers
None


Pandrex247:
Jenkins test please

Pandrex247 (approved):
LGTM - not actually tested though

Issue #5000, opened 2020-11-19
NPE when get result from JCache - @CacheResult 


ciment7:
# Description #
----------

Hi,

I`m running war application on payara 5.201 (Jakarta EE 8). Sometimes (usually when the application is under heavy load) I got this exception:

```
caused by: java.lang.NullPointerException
	at java.base/java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
	at com.sun.enterprise.container.common.impl.ComponentEnvManagerImpl.getJndiNameEnvironment(ComponentEnvManagerImpl.java:161)
	at fish.payara.appserver.context.JavaEEContextUtilImpl.createInvocationContext(JavaEEContextUtilImpl.java:222)
	at fish.payara.appserver.context.JavaEEContextUtilImpl.setInstanceComponentId(JavaEEContextUtilImpl.java:190)
	at fish.payara.nucleus.hazelcast.PayaraHazelcastSerializer.read(PayaraHazelcastSerializer.java:73)
	at com.hazelcast.internal.serialization.impl.StreamSerializerAdapter.read(StreamSerializerAdapter.java:48)
	at com.hazelcast.internal.serialization.impl.AbstractSerializationService.toObject(AbstractSerializationService.java:187)
```

This exception I get when I hit cache by JCache API - @CacheResult. I use multiple payara nodes clustered with multicast options.

This is my snippet of code:
```
 @CacheResult
    public MyObject getScheduleById(@CacheKey Long myObjectId) {
        ...
        
    }

```

## Expected Outcome

It should never throw exception :)

## Current Outcome
Less than 1% of all requests finish with exception.

## Steps to reproduce (Only for bug reports) 

When the cache is under heavy load. 

## Environment ##

- **Payara Version**: 5.201
- **Edition**:  Full
- **JDK Version**: java-11-openjdk-11.0.8.10-0.el8_2.x86_64
- **Operating System**: Linux (Centos 8)

AlanRoth:
Hi @ciment7, can you please try to reproduce this issue in the latest version of Payara Community? (5.2020.6) Please also provide a reproducer ideally following the SSCCE rules: http://www.sscce.org/. It will greatly help us to find the cause and fix it.

lprimak:
I think this will be fixed when this PR (or it's subsets) are merged: https://github.com/payara/Payara/pull/4762

ciment7:
[...]

PR #5010, opened 2020-11-27
FISH-790 Refactor, fix bugs in  JavaEE Context Utility

lprimak:
## Description
Refactored Context utilities into immutable, serializable class.
Fixed many bugs along the way.

### Blockers
This PR is a blocker for Hazelcast 4 integration.

### New tests
Added a unit test

### Testing Performed
All test suites pass, tested with Hazelcast 4 branch

### Testing Environment
Java 8, Mac

## Documentation
Documentation added via Internal API javadoc for new EE context API

## Notes for Reviewers
This issue includes APIs necessary for Hazelcast 4 to be much more tightly integrated with Jakarta EE and Payara
when you think `invocation.preInvoke()` this is easy-to-use replacement and so much more!
This refactoring has been in the works for 6+ months, so it's not "new" in any sense of the word


lprimak:
jenkins test

lprimak:
jenkins test

lprimak:
jenkins test

lprimak:
jenkins test

lprimak:
Fixes #5000

api/payara-api/src/main/java/fish/payara/context/ContextProducer.java
```diff
@@ -0,0 +1,171 @@
...
+ * // insure 'saved' is not leaked when application undeployed,
+ * // otherwise use producer.fromComponentId(producer.getInvocationComponentId())
+ * // and in another, non EJB/CDI thread:
+ * try (Context ctx = saved.pushRequestContext()) {
+ *     // runs with EJB / CDI context
+ * }
+ *
+ * @author lprimak
+ */
+public interface ContextProducer {
```
pdudits, line 60: Is this really something we encourage users to use in their applications, and promise to keep API binary compatible?

I think this should have home elsewhere, somewhere under `nucleus/common`.

appserver/web/gf-web-connector/src/main/java/fish/payara/appserver/context/CDIExtension.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.BeforeBeanDiscovery;
+import javax.enterprise.inject.spi.Extension;
+
+/**
+ * adds context producer
+ *
+ * @author lprimak
+ */
+public class CDIExtension implements Extension {
```
pdudits, line 53: Making it user-facing API is apparently your intent here, I'm not convinced that we should.

jbee (commented):
I did a first rough review. I feel I would want way more description of what this PR tries to do, what problems are behind it. As far as I could see this mostly adds a new API and its implementation but isn't used in so many places. Where it is used the main change appears to be to use an instance (something got bound I guess) instead of what corresponds to a generic context. Before I look closer I really would like to get better understanding so please add some paragraphs to the description of the PR that help with that. Thanks.

api/payara-api/src/main/java/fish/payara/context/ContextProducer.java
lprimak: nucleus/common is where I started with this before migrating to Payara-api.
This was designed before MP context propagation, and perhaps we should use that in the future,
but it's not a reason to hold up this PR currently.

nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/JavaEEContextUtil.java
```diff
@@ -47,68 +48,21 @@
...
 @Contract
-public interface JavaEEContextUtil {
-    /**
-     * pushes Java EE invocation context onto the invocation stack
-     * use try-with-resources to pop the context
-     *
-     * @return the new context that was created
-     */
-    Context pushContext();
+public interface JavaEEContextUtil extends ContextProducer {
```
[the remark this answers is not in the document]
lprimak: The internal API is a superset of the external one now, which is correct.

appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionScopedCDIUtil.java
```diff
@@ -219,7 +219,7 @@ public BeanWrapper(Class<?> beanClass) {
             catch(MultiException e) {
                 log(e.getMessage());
             }
-            this.ctxUtil = ctxUtil;
+            this.ctxUtil = ctxUtil.map(JavaEEContextUtil::currentInvocation);
```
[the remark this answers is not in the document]
lprimak: Context Util API changed, this line just tracks that change, it would be a compile error otherwise.
It's self-explanatory for me here and shouldn't require a comment, as it would be like
"tracking API changes"

appserver/web/gf-web-connector/src/main/java/fish/payara/appserver/context/ContextualizerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import java.lang.reflect.Proxy;
+import java.util.stream.Stream;
+import javax.inject.Inject;
+import org.glassfish.internal.api.JavaEEContextUtil;
+import org.jvnet.hk2.annotations.Service;
+
+/**
+ *
+ * @author lprimak
+ */
```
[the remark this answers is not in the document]
lprimak: Javadoc is in the Contextualizer API itself

api/payara-api/src/main/java/fish/payara/context/ContextProducer.java
```diff
@@ -0,0 +1,171 @@
...
+     * such as CDI initializations, extensions start, etc.
+     *
+     * @return true if current invocation exists and is loaded / ready
+     */
+    boolean isInvocationLoaded();
+
+    /**
+     * specific, immutable, thread-safe instance of the context
+     */
+    interface Instance extends Serializable {
```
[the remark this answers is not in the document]
lprimak: When I use this in NetBeans, it always puts in `ContextProducer.Interface` and I would not use Interface as a standalone type, so IMHO this isn't an issue

lprimak:
@jbee Please read the Javadoc in the new public API. This should explain most, if not all of your questinos

api/payara-api/src/main/java/fish/payara/context/ContextProducer.java
pdudits: There's also Java EE Concurrency API, that deals with context propagation. But neither of these APIs expose internals of application servers, like invocation stack or underlying classloaders and give explicit control over them to user application. They only allow propagating current context to async calls.

I don't believe this belongs into application server's public api.

lprimak: Do you have an alternative? How else is Hazelcast callback supposed to deal with EJB/CDI objects?
I don't look at it as 'App server internals' rather than an integration point that let's apps use Jakarta EE.

Other frameworks always touted 'we have a main method' type approach, this gives and entry-way to do something like that on a hybrid basis at least, which is very useful

pdudits: I cannot offer an alternative, because I don't know what the exact requirements are. Does any of the other PRs demonstrate how would user app use these?

Jakarta EE is all about integration, yet so far they seemed to solve this without giving fine-grained control over invocation context to user. I can give you lots of examples that are probably not relevant, but may help understanding the motivation. And then we can look at whether the API could be more opaque, and then look at security implications of this.

Is there anything in JCA's inbound flows (so special kind of message driven bean) that would cover these use cases? Could CDI events be produced by application server instead of user app directly interacting with hazelcast over callback? An interceptor that would get installed around Hazelcast callback by the application server and manage the stack?

pdudits: Ah I now saw the example in the PR description, I can see few alternatives to write it:

lprimak: I edited the top of the PR with an example.
You do raise a good point about security, as this API makes it much easier to switch context into another application running on the same server (if ran without security manager)

pdudits: I found out, that I'm not entirely sure what the code sample is trying to achive.
```

        map.put("hello", "world"); // (1)

        map2.put("bye", listener); // (2)
```
What I think that (1) does is, that when when entry is added to map, it will invoke `@Inject MyListener listener` asynchronoulsy in current context. i.e invocation of Singleton EJB.

For this I can imagine
- Using `javax.enterprise.concurrent.ContextService`, its api is strinkingly similar.
- Set of Hazelcast specific APIs, i. e. CDI extension that would install map listeners and then invoke event observers (`void entryAdded(@Observes @EntryAdded(mapName = "hallo") String entry) {}`)


I'm not sure what (2) is trying to achieve. What should happen when we put managed object into distributed map? How do we manage it serializes properly since it the reference might be just a proxy (to a proxy) of the application class? How do we guarantee that map is updated when objects (or proxies) internal state changes? How do we cope with tha fact, that the grid is usually not running single version of an application and therefore serialization compatibility of the managed object must be retained? What if effects of invoking a method on the objects are local only?

lprimak: Jakarta EE has 'solved' - not really? this problem by mandating that objects can be used only in App server control flow. If we want true Hazelcast integration (which people are asking for), some sort of context API needs to be exposed.

Also, keep this in mind: https://github.com/payara/Payara/pull/5013
I am also working on tenant control for Hazelcast which will make it possible for some Hazelcast objects to automatically propagate Payara app context, but some cannot and still need an API to do this

nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/PayaraHazelcastSerializer.java
```diff
@@ -44,15 +44,24 @@
...
 
 /**
  *
  * @author lprimak
  * @since 4.1.2.173
  */
 public class PayaraHazelcastSerializer implements StreamSerializer<Object> {
+    private final JavaEEContextUtil ctxUtil;
+    private final StreamSerializer<Object> delegate;
+    private static final Logger log = Logger.getLogger(PayaraHazelcastSerializer.class.getName());
```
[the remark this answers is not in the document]
lprimak: fixed

api/payara-api/src/main/java/fish/payara/context/ContextProducer.java
lprimak: > Set of Hazelcast specific APIs, i. e. CDI extension that would install map listeners and then invoke event observers (void entryAdded(@Observes @EntryAdded(mapName = "hallo") String entry) {})

This is clearly out of scope for this PR

> Using javax.enterprise.concurrent.ContextService, its api is strinkingly similar.

Again, this was pre-MP context propagation, and if we are going to use something, it's MP stuff I would think
I thin it would also be out of scope here

lprimak:
jenkins test

lprimak:
@jbee @pdudits Since the public API is such a hot button issue and this PR is meant as a bug fix, I moved it into internal API so none of the above is an issue anymore

lprimak:
jenkins test

jbee:
@lprimak 

> Please read the Javadoc in the new public API. This should explain most, if not all of your questinos

Code tells me what is going on not why. I'd like to know you motivations both towards the problem and the chosen solution so I can understand this change in context.

jbee (approved):
I dont't see any obvious issues with the PR but likewise it is hard to tell how or where it would introduce new issues. This is just an area that is hard to verify other then understanding the consequences of the change and its impact which is a hard nut to crack for this one. I'm quite convinced this is an improvement over the old code but ultimately I feel this is your call @MattGill98 being responsible for community.

pdudits (approved):

PR #5002, opened 2020-11-20
FISH-662 MicroProfile Metrics 3.0 (MP 4.0)

jbee:
### Summary
This PR makes changes required to update implementation of MicroProfile Metrics to version 3.0 and pass its TCK.

Changes include:

* methods in `Metadata` changed (breaking), `Optional`s are now returned by property named methods, e.g. `unit()`, whereas `getUnit()` returns the value or a default value.
* `MetricRegistry` is now an interface
* `MetricRegistry` `getMetadata` is not included in the MP API (did exist in our implementation before)
* `MetricRegistry` got a new `getType()` method returning `base`, `vendor` or `application` 
* `MetricRegistry` got numerous new methods to get or create metrics, one group to support using `MetricID` arguments, another group (`gauge`) to support this operation for `Gauge`s which did not allow it directly before.
* `MetricRegistry` got numerous new methods to access metrics by their `MetricID` (or return null)
* all metrics are now considered _reusable_, therefore the flag got removed from all annotations, metadata and implementation
* global tags are no longer added within `MetricID` class but added explicitly during export
* `@Metric` is no longer used in the role to indicate registration of a metric producing method (in connection with `@Produces`) - it is now only used to `@Inject` a metric. Therefore most CDI fiddling could be removed.
* CDI sterotype handling is explicitly done in `AnnotationReader`
* `SimpleTimer` got a new `getMaxTimeDuration`, `getMinTimeDuration` and `getElapsedTime` methods
* `Timer` got new `getElapsedTime` method
* `Histogram` got new `getSum` method
* methods added to different metrics all added to both JSON and Open Metrics export formats of those metrics

Changes that could also be considered bug fixes:

* OpenMetrics export correctly computes seconds with fractions from `Duration` (would not consider sub seconds time before)
* JSON export does not include type nodes (e.g. `application`) if they have no child nodes 

### Testing
Because of the extended API of `MetricRegistry` many of our unit tests could now use the interface instead of requiring the implementation.

#### Updates to Unit Tests
Existing tests were extended to cover the newly added methods of metrics and the `MetricRegistry`.
Some tests were removed that were testing difference between _reusable_ and _not reusable_ metrics which no longer exists.
New test were added which are based on TCK tests that failed while our unit tests passed indicating untested feature.

#### Testing Performed
Testing was done with MP API and TCK RC-4 as updated for TCK runner in this PR https://github.com/payara/MicroProfile-TCK-Runners/pull/126

Build the community server and started it with:
```bash
[...]

PR #5023, opened 2020-12-02
FISH-659 MicroProfile FaultTolerance 3.0 (MP 4.0)

jbee:
### Summary
The goal of this PR is to make Payara's MicroProfile Fault Tolerance implementation pass the 3.0 TCK.

On the surface the update to FT 3.0 might look like a smaller change as "only" the metrics got changed to using tags.
This is not really the case. With the metrics some semantics got changed and some grey-areas clarified which had a big impact on our implementation since we had understood many of them in a different way before.

One of the essential aspects that changed is what state is picked for a particular FT method.
In the past we would have a state context per instance and method. The FT spec clarified this to be only the method. All instances (within the same application) calling the same method share the FT state, that is the bulkhead and circuit breaker state.

Other change affect the question when a result counts as failure or as success and when these are updated as well as when threads are considered to have "exited" the method. This isn't all due to changes in the spec. I think its a combination of misunderstandings and clarifications. 

Last but not least the TCK got much better testing corner cases which also revealed slight differences in the understanding of how to process certain configurations, one of which is for example how to handle a `@Retry` without and retry attempts.

This PR does also:

* update the InSight metrics collection of FT state
* we now clean up application related state as a reaction to undepoy event
* reorganise the FT state in a maps of maps, 1st level being the app, 2nd the method
* add more unit tests and improve existing tests
* change the way bulkhead state is used: for synchronous bulkhead we only use the counter, for asynchronous bulkhead we also use the queue containing the processing (not waiting) threads. This might appear a little confusing at first.

### Additional Changes to MicroProfile Metrics
Note that this PR builds upon the changes in https://github.com/payara/Payara/pull/5002. To see actual changes within this PR the referenced PR needs to be merged first.

Additional changes are:

* adds `MetricsService` `@Contract` interface (extracted from `class` that got moved to `MetricsServiceImpl`)
* adds `MetricContext` interface as an access API for `base`, `vendor` and `application` type `MetricRegistry` for each application and the "global" server context
* better `toString()` for `Metric` implementations and the `MetricRegistry` implementation (for easy debugging and logging)

The change to a `base`, `vendor` and `application` type `MetricRegistry` per application did simplify the processing that was based on `MetricRegistry` contents. This is most of all writing the metrics (export via REST) and converting metrics to InSight series data.

### Testing
Running the FT TCK 3.0-RC2 against remote Payara.

This required setting (both in the context running the server and the context running the TCK):
```
```
I also tend to disable config caching using (set before server start!)
```
```
To run the TCK I used in `MicroProfile-TCK-Runners/MicroProfile-Fault-Tolerance/tck-runner/`:
```
mvn clean install "-Ppayara-server-remote" "-Dpayara.version=5.2020.7-SNAPSHOT"
```
See https://github.com/payara/MicroProfile-TCK-Runners/pull/130 for my settings

#### CI
Note that FT have a dependency on MP Metrics 3.0. So Metrics 3.0 can only pass all Payara CI testing if FT testing also uses FT 3.0 (which then uses Metrics 3.0) or if FT TCK isn't run until this PR is merged (what we do now).
When both FT and Metrics 3.0 are ready all MP TCKs should pass when using the 3.0 version of FT and Metrics.


MattGill98:
jenkins test please

MattGill98 (approved):

MattGill98:
Thanks @jbee!! 🍾

PR #5025, opened 2020-12-02
FISH-863 Fix MP Sniffer RAR Issue

MattGill98:
Connectors failed as the MP Config Sniffer always activated and was incompatible with connector modules. Now the Config Sniffer is hidden from the user and compatible with every module.

Signed-off-by: Matthew Gill <matthew.gill@live.co.uk>

Is this the correct branch?


MattGill98:
jenkins test please

cubastanley:
This is the correct branch indeed, Just going to run it through internal to see if the problem resolves in the TCK

cubastanley (approved):
Passes as needed, will merge

PR #5035, opened 2020-12-07
FISH-886 Deploy GAV from local repository + some code cleanup

avpinchuk:
## Description

Added ability to deploy GAV from local repository.

***Example:***
`java -jar payara-micro.jar --deployFromGAV org.example:someapp:1.0 --additionalRepository file:///c:/Users/alex/.m2/repository`

### Testing Environment

Zulu JDK 8, Windows 10 Pro

### DIFF
--- a/nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
+++ b/nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2016-2018 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016-2020 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
--- a/nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
+++ b/nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
@@ -44,7 +44,10 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
+import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.AbstractMap;
 import java.util.Base64;
 import java.util.Collection;
@@ -77,7 +80,7 @@ public class GAVConvertor {
      */
     public Map.Entry<String, URL> getArtefactMapEntry(String GAV, List<URL> repositoryURLs) throws MalformedURLException {
         final Map<String, String> GAVMap = splitGAV(GAV);
-        Map.Entry<String, URL> artefactMapEntry = null;
+        Map.Entry<String, URL> artefactMapEntry;
         
         final String relativeURLString = constructRelativeURLString(GAVMap);
         final URL artefactURL = findArtefactURL(repositoryURLs, relativeURLString);
@@ -97,7 +100,7 @@ public class GAVConvertor {
      * the provided GAV
      */
     public Map.Entry<String, URL> getArtefactMapEntry(String GAV, Collection<String> repositoryURLs) throws MalformedURLException {
-        List<URL> repoURLs = new LinkedList<URL>();
+        List<URL> repoURLs = new LinkedList<>();
         for (String url: repositoryURLs) {
             String convertedURL = TranslatedConfigView.expandValue(url);
             if (!convertedURL.endsWith("/")) {
@@ -117,7 +120,7 @@ public class GAVConvertor {
      * the provided GAV as Strings
      */
     private Map<String, String> splitGAV(String GAV) throws MalformedURLException {
-        final String[] splitGAV = GAV.split(",|:");
+        final String[] splitGAV = GAV.split("[,:]");
         final Map<String, String> GAVMap = new HashMap<>();
         try {
             GAVMap.put("groupId", splitGAV[0].replace('.', '/'));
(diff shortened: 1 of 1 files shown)


lprimak (approved):

lprimak:
jenkins test

lprimak (changes requested):
Please:
- Update copyright year
- add an example into the PR on how this can be used (example)

avpinchuk:
- tell me please i have to change the main copyright header or add portions copyright?
- done

lprimak:
Just copyright year to 2020

lprimak (approved):

lprimak:
@svendiedrichsen can you review this please?

lprimak:
jenkins test

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
```diff
@@ -171,18 +173,27 @@ private URL findArtefactURL(List<URL> repositoryURLs, String relativeURLString)
...
-                    String auth = artefactURL.getUserInfo();
-                    if (auth != null) {
-                        String encodedAuth = Base64.getEncoder().encodeToString(auth.getBytes());
-                        httpConnection.setRequestProperty("Authorization", "Basic " + encodedAuth);
+                    if ("file".equalsIgnoreCase(artefactURL.getProtocol())) {
+                        if (Files.exists(Paths.get(artefactURL.toURI()))) {
+                            validURLFound = true;
+                        }
+                    } else {
+                        HttpURLConnection httpConnection = (HttpURLConnection) artefactURL.openConnection();
```
svendiedrichsen, line 181: This makes some assumption about the type of connection but it might be a JarURLConnection. Like this one: jar:http://www.foo.com/bar/baz.jar

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
```diff
@@ -117,7 +120,7 @@
      * the provided GAV as Strings
      */
     private Map<String, String> splitGAV(String GAV) throws MalformedURLException {
-        final String[] splitGAV = GAV.split(",|:");
+        final String[] splitGAV = GAV.split("[,:]");
         final Map<String, String> GAVMap = new HashMap<>();
         try {
             GAVMap.put("groupId", splitGAV[0].replace('.', '/'));
```
svendiedrichsen, line 126: Maven coordinates might also include `packaging` and `classifier`. If the `packaging` coordinate exists you don't have to iterate of the list of possible file extensions. If the `classifier` coordinate you won't be able to find the artefact without it.

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
```diff
@@ -117,7 +120,7 @@
      * the provided GAV as Strings
      */
     private Map<String, String> splitGAV(String GAV) throws MalformedURLException {
-        final String[] splitGAV = GAV.split(",|:");
+        final String[] splitGAV = GAV.split("[,:]");
```
svendiedrichsen, line 123: Isn't a valid GAV only split by `:`? Doesn't it mean when you find a comma that this is really a list of GAV?

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
avpinchuk: I only needed a local repository but jar scheme support also possible

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
avpinchuk: According to documentation https://docs.payara.fish/community/docs/5.2020.7/documentation/payara-micro/deploying/deploy-cmd-line.html#deploying-applications-from-a-maven-repository for each GAV needs its own           --deployFromGAV option

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
avpinchuk: This may may implemented if we stick to some predefined format. In community doc defined _artifactId_, _groupId_ and _version_ only. Link in next answer

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
svendiedrichsen: Yes it is true, but at least uncommon to use a comma nonetheless.

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
svendiedrichsen: Yes, its true according to the doc but shouldn't there be a possibility to use packaging and qualifier?

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
svendiedrichsen: This needs to be decided by someone from Payara.

svendiedrichsen:
> @svendiedrichsen can you review this please?

@lprimak I have done this. Someone from Payara needs to look into this to make some decisions though.

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
Pandrex247: Looking back at the commit history around this, comma was the original delimiter as that's the "default" delimiter for Payara CLI (though we do use different delimiters all over the place), and we added in support for using a colon after to be more in line with the usual Maven notation.

So `fish.payara,guppy,1.0` and `fish.payara:guppy:1.0` are both valid, and if you want to deploy more than one artefact you have to specify the option multiple times.

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
Pandrex247: Not allowing a user to specify packaging and qualifiers is simply a limitation / missing feature of the implementation at the moment - I'd say it's out of scope for this PR though. The use case you described makes sense, I just wouldn't bundle it into this PR to prevent scope creep, particularly since it isn't a trivial 1 line change.

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/GAVConvertor.java
Pandrex247: I'd say that's outside the scope of this PR. 
I'm not immediately sure of the need to handle them either tbh - I think that functionality is covered by the standard --deploy option unless you're talking about using a Jar Url as a "maven" repository in its own right, which sounds mega-niche!

Pandrex247 (approved):

lprimak:
@svendiedrichsen I will take a look in the next few days, thanks!

PR #5036, opened 2020-12-07
QACI-592 Fixed build consistency when using QuickBuild

dmatej:
### Testing Performed
Delete Payara artifacts.
```
rm -rf ~/.m2/repository/fish
docker image rm payara/server-web:5.2020.8-SNAPSHOT-jdk11
docker image rm payara/server-web:5.2020.8-SNAPSHOT
```
Build all artifacts (but don't run Payara Samples)
- to my surprise it passed even with TC8
```
mvn clean install -PQuickBuild,BuildExtras,BuildDockerImages -amd -TC8; # time: 5:53
```
Without "Extras" and payara-web
```
rm -rf ~/.m2/repository/fish/payara/distributions/payara-web* # to be sure that it compiles even without it
mvn clean install -PQuickBuild,BuildDockerImages -amd -TC8; # time: 4:21
```
Build zip files and docker images
-  TC8 - :eyes: Wow!
```
mvn clean install -pl :extras,:docker-images -PQuickBuild,BuildExtras,BuildDockerImages -amd # time: 6:37
mvn clean install -pl :extras,:docker-images -PQuickBuild,BuildExtras,BuildDockerImages -amd -TC8 # time: 3:57
```
Build payara-web and docker images
```
rm -rf ~/.m2/repository/fish/payara/distributions/payara-web*
mvn clean install -pl :payara-web # time: 0:23
mvn clean install -pl :docker-images -PBuildDockerImages -amd # time: 2:30
```

The issue was introduced in commits under CUSTCOM-130

### DIFF
--- a/appserver/extras/docker-images/pom.xml
+++ b/appserver/extras/docker-images/pom.xml
@@ -10,15 +10,6 @@
     <packaging>pom</packaging>
     <name>Payara Docker Images</name>
 
-    <modules>
-        <module>basic</module>
-        <module>server-full</module>
-        <module>server-web</module>
-        <module>server-node</module>
-        <module>micro</module>
-        <module>tests</module>
-    </modules>
-
     <properties>
         <payara.version>${project.version}</payara.version>
         <docker.noCache>true</docker.noCache>
@@ -58,6 +49,35 @@
     </build>
 
     <profiles>
+        <profile>
+            <id>DefaultBuild</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <modules>
+                <module>basic</module>
+                <module>server-full</module>
+                <module>server-web</module>
+                <module>server-node</module>
+                <module>micro</module>
+                <module>tests</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>QuickBuild</id>
+            <activation>
+                <property>
+                    <name>QuickBuild</name>
+                </property>
+            </activation>
+            <modules>
+                <module>basic</module>
+                <module>server-full</module>
+                <module>server-node</module>
+                <module>micro</module>
+                <module>tests</module>
+            </modules>
+        </profile>
         <profile>
             <id>generate-docker-images-only-if-dockerfile-exists</id>
             <activation>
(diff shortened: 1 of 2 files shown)


dmatej:
Jenkins test please

Pandrex247:
Not sure this is really "build consistency", seems more like you're just trying to have the BuildDockerImages profile mirror the QuickBuild behaviour.

As a rundown of the profiles:
QuickBuild disables building of the _payara-ml_, _payara-web_, and _payara-web-ml_ distributions of Payara Server - so just _payara_ and _payara-micro_ get built rather than all 5.
BuildExtras activates the _embedded_ (and also currently the certificate management) modules. There used to be more under extras which we skipped as a part of the normal build (e.g. Micro) so the name of this profile is a bit misleading atm.
BuildDockerImages activates building of the docker image modules. This was kept separate from the BuildExtras profile due to the possibility of not having Docker installed.

To build the docker images you don't need BuildExtras. There is no Docker image that depends on embedded (or cert management).

Pandrex247 (changes requested):
Assuming your intention is to make it so that when QuickBuild is used with BuildDockerImages it skips trying to build the payara-server-web docker image, this doesn't seem to work:

`mvn clean install -PQuickBuild,BuildDockerImages`

```
[ERROR] Failed to execute goal on project docker-images-tests: Could not resolve dependencies for project fish.payara.extras:docker-images-tests:jar:5.2020.8-SNAPSHOT: Could not find artifact fish.payara.extras:server-web-docker-image:pom:5.2020.8-SNAPSHOT
```

dmatej:
I'm not sure now, but I think there was a reason why I always enabled BuildExtras with BuildDockerImages, maybe it is something fixed now. The second comment is certainly my mistake, thanks for the review!

lprimak (approved):

dmatej:
If I would exclude also remaining except payara-web, DockerNodesITest would fail, because it depends on all except web profile.

dmatej:
AHA! The BuildExtras + extras is here to build also zip files, while it expects that everything else was built with QuickBuild enabled where I can use -TC8 so have it asap. So it is just and example, you can combine modules, profiles, as you need. 
The point is that it should pass if it is not some nonsense.

So the only issue is with that dependency, you are right the there is much more possible examples which must pass.

dmatej:
Jenkins test please

Pandrex247 (approved):

PR #5037, opened 2020-12-08
QACI-454 QACI-595 more reliable samples

dmatej:
## Description

- if something is broken, don't log it, but throw exception instead.
- tests should explicitly know what they are doing, not relying
      on ServerOperations/CliCommands magic
- if something depends on server, make a condition, etc., so reader knows
      there is some limitation
- test-utils renamed to samples-test-utils as another module existed with
     the same name
- SecurityUtils - split from ServerOperations to avoid dependency on BC+OmniFaces
      if they are not used at all.
- using own property for remote and managed
- Trimming stacktraces is off everywhere
- Clean the server's directory -> repeatable managed

## Testing

```
mvn clean install -Ppayara-samples,payara-micro-managed -fae -pl :payara-samples -amd
mvn clean install -Ppayara-samples,payara-server-managed -fae -pl :payara-samples -amd
asadmin start-domain
mvn clean install -Ppayara-samples,payara-server-remote -fae -pl :payara-samples -amd
asadmin stop-domain
```

### DIFF
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -287,7 +287,6 @@
                         <systemPropertyVariables>
                             <isUsingMicroProfile>${micro.isActive}</isUsingMicroProfile>
                         </systemPropertyVariables>
-                        <trimStackTrace>false</trimStackTrace>
                     </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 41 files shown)


dmatej:
Jenkins test please

dmatej:
Jenkins test please

lprimak (approved):

PR #5044, opened 2020-12-10
CUSTOM-55 possibly stale log records are propagated

lprimak:
force toString() during the same thread as the logging happens so objects

don't get passed around different threads in logging and have a chance to get
out of sync and be inconsistent

## Description
Pragmatic & minimalist fix to remove race conditions in log records causing stale state to be printed in log message

### Blockers
None

### New tests
Existing tests

### Testing Environment
Jenkins

## Notes for Reviewers
This is a simplest fix for the issue, not touching anything that doesn't directly relate
Choose "hide whitespace changes" from GitHub setup to make the review easier

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
@@ -86,6 +86,29 @@ public class GFLogRecord extends LogRecord {
     }
 
     /**
+     * wrap log record with {@link GFLogRecord} if not already
+     * if setThreadName is true, sets thread name to current
+     *
+     * @param record
+     * @param setThreadName
+     * @return wrapped record
+     */
+    public static GFLogRecord wrap(LogRecord record, boolean setThreadName) {
+        GFLogRecord wrappedRecord;
+        if (record instanceof GFLogRecord) {
+            wrappedRecord = (GFLogRecord)record;
+        } else {
+            wrappedRecord = new GFLogRecord(record);
+        }
+        // Check there is actually a set thread name
+        if (setThreadName && wrappedRecord.getThreadName() == null) {
+            wrappedRecord.setThreadName(Thread.currentThread().getName());
+        }
+
+        return wrappedRecord;
+    }
+
+    /**
      * CUSTOM-55
      * in case of an object passed as a parameter, call it's toString() method
      * to resolve it's values in the current thread, instead of waiting for queues / etc
@@ -96,17 +119,18 @@ public class GFLogRecord extends LogRecord {
      * @param params
      * @return parameter array
      */
-    private Object[] transformParameters(Object[] params) {
+    private static Object[] transformParameters(Object[] params) {
         if (params == null) {
             return null;
         }
         Object[] result = new Object[params.length * 2];
-        for (int stringIndex = 0, originalIndex = params.length; stringIndex < params.length;
-                ++stringIndex, ++originalIndex) {
-            Object param = params[stringIndex];
+        for (int stringParamsIndex = 0, originalParamsIndex = params.length;
+                stringParamsIndex < params.length;
+                ++stringParamsIndex, ++originalParamsIndex) {
+            Object param = params[stringParamsIndex];
             if (param != null) {
-                result[stringIndex] = param.toString();
-                result[originalIndex] = params[stringIndex];
+                result[stringParamsIndex] = param.toString();
+                result[originalParamsIndex] = params[stringParamsIndex];
             }
         }
         return result;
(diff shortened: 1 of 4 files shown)


lprimak:
jenkins test

lprimak:
jenkins test

lprimak:
jenkins test

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
```diff
@@ -84,4 +84,25 @@ public String getThreadName() {
...
+     * @param params
+     * @return parameter array
+     */
+    private Object[] transform(Object[] params) {
+        if (params == null) {
+            return null;
+        }
+        for (int ii = 0; ii < params.length; ++ii) {
+            Object param = params[ii];
+            if (param != null && !param.getClass().equals(String.class)) {
```
sgflt, line 102: What is `param.getClass().equals(String.class)` supposed to do?

lprimak: See it’s already a sting. I could use instanceof String instead I guess

sgflt: I think it is not useful at all. String,toString() returns this, so it is NOOP and the condition just obfuscates code. Or am I missing something?

lprimak: No you are not missing anything. You are right it’s superfluous. I will remove that one

lprimak:
Jenkins test

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
```diff
@@ -84,4 +84,25 @@ public String getThreadName() {
...
+                params[ii] = params[ii].toString();
```
sgflt, line 103: For example UniformLogFormatter expects in some code paths java.util.Collection or java.util.Map. Now paremeters are just strings so it may require some adjustment of formatters. (or different approach)

lprimak: good catch, fixed

lprimak:
jenkins test

lprimak:
jenkins test

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
```diff
@@ -84,4 +84,31 @@ public String getThreadName() {
...
+     *
+     * @param params
+     * @return parameter array
+     */
+    private Object[] transformParameters(Object[] params) {
+        if (params == null) {
+            return null;
+        }
+        Object[] result = new Object[params.length * 2];
+        for (int stringIndex = 0, originalIndex = params.length; stringIndex < params.length;
```
dmatej, line 104: I would probably use paramsIndex instead of originalIndex as a name

dmatej (commented):
I am not sure how this change can affect performance. Also count of parameters doubled.

Also if we would use both SysLogHandler and FileLoggerHandler, GFLogRecord would be created in both, twice. (use case: as a standard log to the server.log, and report severe errors to SysLog)

JSONLogFormatter:
            `        if (parameter instanceof Map) {`
 Did JsonLogFormatIT in Payara Samples pass?

nucleus/common/common-util/src/main/java/com/sun/common/util/logging/GFLogRecord.java
lprimak: Performance is unaffected because toString() gets called eventually anyway.
`JsonLogFormatIT` works because now parameters include both strings and original objects.
All that's is original objects so memory consumption has only tiny increase at the end of the day, just due to object array being bigger

lprimak: > Also if we would use both SysLogHandler and FileLoggerHandler, GFLogRecord would be created in bot

fixed

lprimak:
jenkins test

lprimak:
Performance is unaffected because toString() gets called eventually anyway.
JsonLogFormatIT works because now parameters include both strings and original objects.
All that's is original objects so memory consumption has only tiny increase at the end of the day, just due to object array being bigger

dmatej (approved):
First it doesn't fix CUSTCOM-55, JtaTimeoutLoggingITest is still failing, so this fix didn't help.

When I tried to run VerboseLoggingITest, start of the domain ended in deadlog in fifth iteration, but that is not probably caused by this change as you didn't touch any synchronizations (but ... ie call of a bad toString in different thread than before, or sooner/later can cause that). But I wasn't able to reproduce it again.

Performance is same as of logging in master branch, so this is alright.

So ... despite I still believe it is a step forward, it doesn't help with any of issues on my radar. The bad thing is that it duplicates and conflicts with my branch, but there I can simply ignore your changes.

lprimak:
This fixes “lying logs” only. I’m not quite sure if you are disputing that?

Issue #5027, opened 2020-12-02
Feature Request: MP-JWT public key location respects HTTP cache headers / FISH-868


ghunteranderson:
# Description #
----------
Hello again! A feature request this time. 🙂 

This use case applies to MP-JWT with an HTTP(S) public key location. When SignedJWTIdentityStore.java verifies a token, it makes an HTTP request to the public key endpoint to fetch the public keys. This happens every time that service verifies a token. That can quickly turn into a lot of requests for the same data.

IDPs have an option to provide HTTP cache directives on the public key endpoint in hopes of clients using it. I can see multiple benefits to Payara following the max-age directive:
1. Reduced load on the network
2. Reduced load on the IDP
3. Reduced latency to verify a token. This benefit compounds if multiple upstream service calls are required to complete the user's request.
4. Increased fault tolerance to isolated failed network request.

One edge case that I think this feature would need to handle is if the IDP does not use a long enough delay between when a new public key is published and when the new private key is used to sign tokens. A Payara instance could have a "fresh" copy of the public key cache based on the max-age header but start seeing new key IDs on JWTs. One solution would be to force a cache refresh when JWTs with unrecognized key IDs are received.

An alternative solution with similar benefits: Any public key loaded from `mp.jwt.verify.publickey.location` could be cached with a TTL set by a proprietary config in `payara-mp-jwt.properties`. This has the added benefit of decreasing disk IO load when the public key is on the file system and gives the developer control over how long the cache should be active.

## Expected (Requested) Outcome

Payara's MP-JWT implementation would cache public keys fetched over HTTP until
1. The cache expires
2. A new key ID is introduced that doesn't exist in cache

## Current Outcome

Every request that uses MP-JWT will re-load the public key even if it requires an additional network call.

## Environment ##

- **Payara Version**: 5.2020.6
- **Edition**: any
- **JDK Version**: OpenJDK 11
- **Operating System**: any
- **Database**: n/a

AlanRoth:
Hi @ghunteranderson, I have created a internal ticket `FISH-868`, this feature does seems like a good idea!

It may be a while before our engineering team gets around to implementing this, is there any possibility you could attempt to implement this feature yourself and submit a PR as a community contributor? 

Thank you,
Alan

ghunteranderson:
Hi @AlanRoth!

Thanks for the quick response. I'll see what I can get together this weekend. 🙂

AlanRoth:
Closing the issue as PR has been merged, thank you for your contribution

PR #5062, opened 2020-12-18
FISH-868 JWT public key is cached for a configurable amount of time.

ghunteranderson:
## Description
This is a PR for feature request https://github.com/payara/Payara/issues/5027. With this change, the loaded JWT public key is cached for subsequent requests to use. While the PR is aimed at helping with remote public keys, it handles key caching from any source the same.

I original wrote the feature request with HTTP cache headers in mind but since it is much simpler to have a default cache TTL configurable by `payara-mp-jwt.properties`, I thought that would be a better place to start. A new configuration `publicKey.cache.ttl` sets the TTL in seconds and defaults to 5 minutes (probably want to discuss what the default could be). If set to 0 or less, the cache is disabled and public key loads are no longer synchronized.

Last, I omitted a feature I entertained in https://github.com/payara/Payara/issues/5027 where receiving an unrecognized JWT key ID would hint to the server that the public key cache was stale and needed to be refreshed. If that's something we think would be helpful, I can definitely add it. 

### New tests
I held off on writing tests because I'm not sure how tests are being handled for this module. I do not see any unit test. For what it's worth, I did manually test the feature with embedded and remote public keys.

## Notes for Reviewers
The PR looks a bit bigger than it actually is. With the extra complexity to loading public keys, it seemed like a good time to move that responsibility out of `SignedJWTIdentityStore` and into its own class, `JwtPublicKeyStore`. Most of the "read MP public key" methods were copied verbatim except they now return the public key as a string instead of parsing it first.

I went back and forth on whether to cache the public key before or after parsing. I landed on before because it was the least disruptive to how key IDs, PEMs and JWKS is working right now but I also recognize that each request then has to re-parse the public key. I thought it was an ok tradeoff but I could be easily swayed.

I really wanted to use `org.glassfish.jersey.internal.guava.LoadingCache` but it does not seem to support `expireAfterWrite`. I added a simple double checking synchronized cache wrapper but I'd rather delete it if there's a better option already available in the project.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
+++ b/appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
@@ -1,0 +1,297 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) [2017-2020] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
+ * and Distribution License("CDDL") (collectively, the "License").  You
+ * may not use this file except in compliance with the License.  You can
+ * obtain a copy of the License at
+ * https://github.com/payara/Payara/blob/master/LICENSE.txt
+ * See the License for the specific
+ * language governing permissions and limitations under the License.
+ *
+ * When distributing the software, include this License Header Notice in each
+ * file and include the License file at glassfish/legal/LICENSE.txt.
+ *
+ * GPL Classpath Exception:
+ * The Payara Foundation designates this particular file as subject to the "Classpath"
+ * exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ * file that accompanied this code.
+ *
+ * Modifications:
+ * If applicable, add the following below the License Header, with the fields
+ * enclosed by brackets [] replaced by your own identifying information:
+ * "Portions Copyright [year] [name of copyright owner]"
+ *
+ * Contributor(s):
+ * If you wish your version of this file to be governed by only the CDDL or
+ * only the GPL Version 2, indicate your decision by adding "[Contributor]
+ * elects to include this software in this distribution under the [CDDL or GPL
+ * Version 2] license."  If you don't indicate a single choice of license, a
+ * recipient has the option to distribute your version of this file under
+ * either the CDDL, the GPL Version 2 or to extend the choice of license to
+ * its licensees as provided above.  However, if you add GPL Version 2 code
+ * and therefore, elected the GPL Version 2 license, then the option applies
+ * only if the new code is made subject to such option by the copyright
+ * holder.
+ */
+package fish.payara.microprofile.jwtauth.eesecurity;
+
+import static java.lang.Thread.currentThread;
+import static org.eclipse.microprofile.jwt.config.Names.VERIFIER_PUBLIC_KEY;
+import static org.eclipse.microprofile.jwt.config.Names.VERIFIER_PUBLIC_KEY_LOCATION;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.StringReader;
+import java.math.BigInteger;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.charset.Charset;
+import java.nio.charset.IllegalCharsetNameException;
+import java.security.KeyFactory;
+import java.security.PublicKey;
+import java.security.spec.RSAPublicKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Base64;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import javax.enterprise.inject.spi.DeploymentException;
+import javax.json.Json;
+import javax.json.JsonArray;
+import javax.json.JsonObject;
+import javax.json.JsonReader;
+import javax.json.JsonValue;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.glassfish.grizzly.http.util.ContentType;
+
+class JwtPublicKeyStore {
+    
+    private static final Logger LOGGER = Logger.getLogger(JwtPublicKeyStore.class.getName());
+    private static final String RSA_ALGORITHM = "RSA";
+        
+    private final Config config;
+    private final Supplier<Optional<String>> cache;
+    
+    /**
+     * @param cacheTTL Public key cache TTL in seconds
+     */
+    public JwtPublicKeyStore(Long cacheTTL) {
+        this.config = ConfigProvider.getConfig();
+        if(cacheTTL > 0)
+            cache = new PublicKeyLoadingCache(cacheTTL, this::readRawPublicKey)::get;
+        else
+            cache = this::readRawPublicKey;
+    }
+
+    /**
+     * 
+     * @param keyID The JWT key ID or null if no key ID was provided
+     * @return Public key that can be used to verify the JWT
+     * @throws IllegalStateException if no public key was found
+     */
+    public PublicKey getPublicKey(String keyID) {
+        return cache.get()
+            .map(key -> createPublicKey(key, keyID))
+            .orElseThrow(() -> new IllegalStateException("No PublicKey found"));
+    }
+    
+    private Optional<String> readRawPublicKey() {
+        Optional<String> publicKey = readDefaultPublicKey();
+        
+        if (!publicKey.isPresent()) {
+            publicKey = readMPEmbeddedPublicKey();
+        }
+        if (!publicKey.isPresent()) {
+            publicKey = readMPPublicKeyFromLocation();
+        }
+        return publicKey;
+    }
+    
+    private Optional<String> readDefaultPublicKey() {
+        return readPublicKeyFromLocation("/publicKey.pem");
+    }
+
+    private Optional<String> readMPEmbeddedPublicKey() {
+        return config.getOptionalValue(VERIFIER_PUBLIC_KEY, String.class);
+    }
+
+    private Optional<String> readMPPublicKeyFromLocation() {
+        Optional<String> locationOpt = config.getOptionalValue(VERIFIER_PUBLIC_KEY_LOCATION, String.class);
+
+        if (!locationOpt.isPresent()) {
+            return Optional.empty();
+        }
+
+        String publicKeyLocation = locationOpt.get();
+
+        return readPublicKeyFromLocation(publicKeyLocation);
+    }
+    
+    private Optional<String> readPublicKeyFromLocation(String publicKeyLocation) {
+
+        URL publicKeyURL = currentThread().getContextClassLoader().getResource(publicKeyLocation);
+
+        if (publicKeyURL == null) {
+            try {
+                publicKeyURL = new URL(publicKeyLocation);
+            } catch (MalformedURLException ex) {
+                publicKeyURL = null;
+            }
+        }
+        if (publicKeyURL == null) {
+            return Optional.empty();
+        }
+        
+        try { 
+            return readPublicKeyFromURL(publicKeyURL);
+        } catch(IOException ex) {
+            throw new IllegalStateException("Failed to read public key.", ex);
+        }
+    }
+    
+    private Optional<String> readPublicKeyFromURL(URL publicKeyURL) throws IOException {
+        
+        URLConnection urlConnection = publicKeyURL.openConnection();
+        Charset charset = Charset.defaultCharset();
+        ContentType contentType = ContentType.newContentType(urlConnection.getContentType());
+        if(contentType != null) {
+            String charEncoding = contentType.getCharacterEncoding();
+            if(charEncoding != null) {
+                try {
+                    if (!Charset.isSupported(charEncoding)) {
+                        LOGGER.warning("Charset " + charEncoding + " for remote public key not supported, using default charset instead");
+                    } else {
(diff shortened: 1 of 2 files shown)


AlanRoth:
jenkins test please

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+                    } else {
+                        charset = Charset.forName(contentType.getCharacterEncoding());
+                    }
+                }catch (IllegalCharsetNameException ex){
+                    LOGGER.severe("Charset " + ex.getCharsetName() + " for remote public key not support, Cause: " + ex.getMessage());
+                }
+            }
+        }
+        try (InputStream inputStream = urlConnection.getInputStream();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, charset))){
```
svendiedrichsen, line 185: Does it make sense to read the HTTP cache headers here and use an eventually provided TTL for caching? Just as an override to the configured one.

ghunteranderson: I like that. Assuming we keep the TTL configurable in `payara-mp-jwt.properties` does this priority sound ok for determining the effective TTL? The HTTP headers would be preferred unless the developer overrides it in the configuration.

1. TTL specified in `payara-mp-jwt.properties`. Disabled if set to 0.
2. Cache headers from an `HttpURLConnection`
3. Standard default of n minutes

svendiedrichsen: I wonder if its better to switch priorities 1 and 2. Thus honouring the HTTP header TTL the most although the configured TTL is set to 0. This allows the source of the public key to define the TTL which IMHO is the owner of the public key. Only if the source does not define any TTL the client/user of the public key may have it own rules for caching.

svendiedrichsen:
There ist a `org.glassfish.jersey.internal.guava.CacheBuilder` which allows to create `LocalCache` instances and allows to use `CacheLoader` instances provided. Maybe this is something worth looking at.

ghunteranderson:
@svendiedrichsen Thank you! I thought I checked those cache classes for a "expire after write" option and didn't see one. I might have just missed it. I'll check again to make sure.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
ghunteranderson: My only hesitation with switching 1 and 2 was that the developer is powerless to override the cache headers. I'd hope that the HTTP cache headers are reasonable but we can't guarantee that will be the case. Perhaps unreasonable cache headers isn't a problem Payara should be trying to solve. I'm torn.

I'll get to work on this with the approach you suggested. I need to think through a good way to change the TTL at key load time. If we change our minds on priority, it should be an easy switch 🙂 .

Thanks for the feedback!

svendiedrichsen: Maybe just avoid unreasonable TTL header values which are too short lived or too long. Whatever this is. ;)

ghunteranderson: I suppose "unreasonable" is pretty arbitrary 😅. I'll begin making those changes as soon as I can. I won't be as available with the holidays approaching but I should be able to push an update before the year ends.

svendiedrichsen:
As for the `CacheBuilder` from Jersey framework. it soesn't seem to be fully developed. I couldn't find `expireAfterWriteNanos` being used at all. Maybe it isn't a good idea to use Jersey internal classes after all.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
svendiedrichsen: Sure. Happy holidays!

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+        
+    private final Config config;
+    private final Supplier<Optional<String>> cache;
+    
+    /**
+     * @param cacheTTL Public key cache TTL in seconds
+     */
+    public JwtPublicKeyStore(Long cacheTTL) {
+        this.config = ConfigProvider.getConfig();
+        if(cacheTTL > 0)
```
fturizo, line 93: Add the brackets to these statements for consistent code.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+        
+        public PublicKeyLoadingCache(long ttl, Supplier<Optional<String>> loadingFunction) {
+            this.ttl = ttl;
+            this.loadingFunction = loadingFunction;
+            this.lastUpdated = 0;
+        }
+        
+        public Optional<String> get() {
+            long now = System.currentTimeMillis();
+            if(now - lastUpdated > ttl)
```
fturizo, line 282: Missing brackets.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.glassfish.grizzly.http.util.ContentType;
+
+class JwtPublicKeyStore {
+    
+    private static final Logger LOGGER = Logger.getLogger(JwtPublicKeyStore.class.getName());
+    private static final String RSA_ALGORITHM = "RSA";
+        
+    private final Config config;
+    private final Supplier<Optional<String>> cache;
```
fturizo, line 86: It is a bit confusing that this variable is named `cache` since it is not technically a cache. Better to name it to `cacheSupplier`.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+
+        throw new IllegalStateException("No matching JWK for KeyID.");
+    }
+    
+    private static class PublicKeyLoadingCache {
+        
+        private final long ttl;
+        private final Supplier<Optional<String>> loadingFunction;
```
fturizo, line 269: Better to rename this class field to `keySupplier` instead to keep the consistency of the class.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
```diff
@@ -0,0 +1,297 @@
...
+        private long lastUpdated;
+        private Optional<String> publicKey;
+        
+        
+        public PublicKeyLoadingCache(long ttl, Supplier<Optional<String>> loadingFunction) {
+            this.ttl = ttl;
+            this.loadingFunction = loadingFunction;
+            this.lastUpdated = 0;
```
fturizo, line 277: This statement is redundant since the value has been already initialized to `0`.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/SignedJWTIdentityStore.java
```diff
@@ -179,141 +151,13 @@ public CredentialValidationResult validate(SignedJWTCredential signedJWTCredenti
...
-             BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, charset))){
-            String keyContents = reader.lines().collect(Collectors.joining(System.lineSeparator()));
-            return createPublicKey(keyContents, keyID);
-        }
+    
+    private Long readPublicKeyCacheTTL(Optional<Properties> properties) {
+        return properties
+        		.map(props -> props.getProperty("publicKey.cache.ttl"))
+        		.map(Long::valueOf)
+        		.orElseGet( () -> TimeUnit.SECONDS.convert(5, TimeUnit.MINUTES));
```
fturizo, line 159: Shouldn't this value be in milliseconds? The code in the `PublicKeyLoadingCache` class treats the TTL as milliseconds on its calculations for the `refresh` method. Either convert the value to milliseconds here or specify that the `publicKey.cache.ttl` property needs to be described in milliseconds.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
fturizo: @ghunteranderson, will you make the changes to consider the TTL in the cache headers of the response then?

fturizo (changes requested):
Changes needed, with clarification over the scope of HTTP cache headers consideration for the TTL of the public key.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
ghunteranderson: Hello @fturizo. Thanks for the feedback 🙂  I'll make these changes including the HTTP one. It's taken me a little bit longer than I expect to get back to this.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/SignedJWTIdentityStore.java
ghunteranderson: I'm all over the place on this one. Good catch! Will fix.

ghunteranderson:
jenkins test please

lprimak:
jenkins test

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/SignedJWTIdentityStore.java
```diff
@@ -179,141 +151,14 @@ public CredentialValidationResult validate(SignedJWTCredential signedJWTCredenti
...
-            String keyContents = reader.lines().collect(Collectors.joining(System.lineSeparator()));
-            return createPublicKey(keyContents, keyID);
-        }
+    
+    private Duration readPublicKeyCacheTTL(Optional<Properties> properties) {
+        return properties
+        		.map(props -> props.getProperty("publicKey.cache.ttl"))
+        		.map(Long::valueOf)
+        		.map(Duration::ofMillis)
+        		.orElseGet( () -> Duration.ofMinutes(5));
```
ghunteranderson, line 160: @fturizo To fix the time unit discrepancies, I change the type from `long` to `java.time.Duration`. 

I left the unit of `publicKey.cache.ttl` as milliseconds but I could see an argument for using seconds instead. Let me know if you have any thoughts on that.

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/JwtPublicKeyStore.java
ghunteranderson: I've added support for reading the max-age header and using that as an override for the cache TTL. It's ready for review whenever you are.

svendiedrichsen:
@ghunteranderson Wouldn't it be nice to have some unit tests?

appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/SignedJWTIdentityStore.java
fturizo: As long as it is consistent, you can keep the unit in milliseconds.

fturizo:
@ghunteranderson, before merging the PR, would you be kind enough to document this improvement and submit a PR to the official Payara Community Documentation repository: https://github.com/payara/Payara-Community-Documentation? Let us know if you have any questions

fturizo:
@ghunteranderson, to help you merge this contribution, I've documented it on the Payara Community Edition repository here: https://github.com/payara/Payara-Community-Documentation/pull/127. Once this PR is approved and merged, I'll proceed to merge and close this issue.

fturizo (approved):

ghunteranderson:
Thanks for the help on this @fturizo! Finding time has gotten a little more difficult. I really appreciate it 🙂

PR #5066, opened 2020-12-21
FISH-659: Fault Tolerance 3.0 full pass

pdudits:
## Description
Small refactorings in Fault Tolerance implementation to make all TCK test pass.

The core change was to return back to standard managed thread pools. These pools propagate application context correctly so the issue with having incorrect application name on invocation stack no longer applies.

However the need for application name was greatly reduced in next optimization, where actual class and method reference are used as keys into state map, further stabilizing the tests.

Also discovered and fixed is memory leak in `FaultTolerancePolicy`'s cache where entires were not deleted upon undeployment, which was troublesome especially for run of TCK. With these two changes together in became apparent that cleanup of method data can be removed, as data has relatively small footprint and keeping them removes confusion on what state of the system is.

### New tests

Full FT 3.0 TCK passes:

payara/MicroProfile-TCK-Runners#133

### Testing Environment

```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Java version: 1.8.0_275, vendor: Azul Systems, Inc.
```

## Documentation

Documentation will require update, as all threadpool and cleanup related parameters are not applied anymore.

## Notes for Reviewers

/cc @jbee, if you have bit of time, check if I didn't misunderstood any of the original intentions of the code.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
@@ -199,7 +199,7 @@ public class FaultToleranceServiceImpl
         if (event.is(Deployment.APPLICATION_UNLOADED)) {
             ApplicationInfo info = (ApplicationInfo) event.hook();
             deregisterApplication(info.getName());
-            FaultTolerancePolicy.clean();
+            FaultTolerancePolicy.clean(info.getAppClassLoader());
         } else if (event.is(EventTypes.SERVER_SHUTDOWN)) {
             if (asyncExecutorService != null) {
                 asyncExecutorService.shutdownNow();
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
@@ -53,19 +53,12 @@ import fish.payara.monitoring.collect.MonitoringDataSource;
 import fish.payara.notification.requesttracing.RequestTraceSpan;
 import fish.payara.nucleus.requesttracing.RequestTracingService;
 
-import static java.lang.Integer.parseInt;
-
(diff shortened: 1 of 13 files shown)


appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/FaultToleranceServiceImpl.java
```diff
@@ -119,87 +115,28 @@
...
     }
 
     @Override
     public void event(Event<?> event) {
         if (event.is(Deployment.APPLICATION_UNLOADED)) {
             ApplicationInfo info = (ApplicationInfo) event.hook();
-            deregisterApplication(info.getName());
-            FaultTolerancePolicy.clean();
+            deregisterApplication(info);
+            FaultTolerancePolicy.clean(info.getAppClassLoader());
```
jbee, line 139: Nice idea.

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/MethodKey.java
```diff
@@ -0,0 +1,64 @@
...
+        StringBuilder idBuilder = new StringBuilder();
+        idBuilder.append(targetClass.getName()).append('.').append(method.getName());
+        if (method.getParameterCount() > 0) {
+            idBuilder.append('(');
+            for (Class<?> param : method.getParameterTypes()) {
+                idBuilder.append(param.getName()).append(' ');
+            }
+            idBuilder.append(')');
+        }
+        methodId = idBuilder.toString();
```
jbee, line 56: I assume you are happy to have this run in parallel threads in worst case as the result will be the same.

jbee (approved):
Looks like a pair of fresh eyes is what that needed.
Good work @pdudits !

MattGill98 (approved):

MattGill98:
jenkins test please

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/service/MethodKey.java
pdudits: Yes, that was my thoughts.

MattGill98:
jenkins test please

pdudits:
jenkins test please

PR #5068, opened 2020-12-23
FISH-659 Fix MP FT executor shutdown

MattGill98:
Custom thread pools were swapped back to managed executors in the FT
implementation, so the shutdown methods would fail as the executors are
managed so don't require shutdown.

For reproducing the current error, shutdown the server and check the logs.


MattGill98:
jenkins test please

Pandrex247 (approved):
Haven't followed it all the way through to find how the executors are normally shut down, but they presumably _are_ so this should be fine.

The alternative would be to perform the `shutdownNow` on the executor itself rather than on the adapter received by `(ManagedExecutorService) context.lookup(config.getManagedExecutorService());` (or to add said behaviour to the adapter)

PR #5074, opened 2020-12-29
[FISH-759] sort instances in admin console

### LINKED ISSUE #4795: Sort the list of nodes under "instances" and "configurations" / FISH-759
# Description #
----------

In the UI, sort the list of entries under "Instances" and "configurations"
## Expected Outcome

The list of instances or configurations should be sorted by name.
When there are a large number of instances, it becomes time consuming scrolling through the list looking for the right name.

## Current Outcome
Currently, the list is in order of creation (probably the order in which they appear in the domain.xml)
## Environment ##

- **Payara Version**: 5.2020.2
- **Edition**: All
- **JDK Version**: Oracle 11.0.7
- **Operating System**: Windows Server 2016
- **Database**: Oracle 11g/19c, SQL Server 2016

sgflt:
Fixes #4795

### DIFF
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
@@ -265,6 +265,7 @@ public class ListInstancesCommand implements AdminCommand {
         Properties extraProps = new Properties();
         List instanceList = new ArrayList();
 
+        infos.sort(Comparator.comparing(InstanceInfo::getName));
         for (InstanceInfo ii : infos) {
             String name = ii.getName();
             String value = (ii.isRunning()) ? InstanceState.StateType.RUNNING.getDescription()
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
@@ -120,7 +120,7 @@ public class ListInstancesCommand implements AdminCommand {
     @Param(optional = true, primary = true, defaultValue = "domain")
     String whichTarget = "";
 
-    private List<InstanceInfo> infos = new LinkedList<InstanceInfo>();
+    private List<InstanceInfo> infos = new LinkedList<>();
     private List<Server> serverList;
     private ActionReport report;
     private ActionReport.MessagePart top = null;
@@ -200,7 +200,7 @@ public class ListInstancesCommand implements AdminCommand {
 
                 sb.append(name);
                 top.addProperty(name, "");
-                HashMap<String, Object> insDetails = new HashMap<String, Object>();
+                HashMap<String, Object> insDetails = new HashMap<>();
                 insDetails.put("name", name);
                 instanceList.add(insDetails);
             }
@@ -280,7 +280,7 @@ public class ListInstancesCommand implements AdminCommand {
                 }
             }
 
-            HashMap<String, Object> insDetails = new HashMap<String, Object>();
+            HashMap<String, Object> insDetails = new HashMap<>();
             insDetails.put("name", name);
             insDetails.put("status", value);
             insDetails.put("deploymentgroup", ii.getDeploymentGroups().trim());
@@ -318,7 +318,7 @@ public class ListInstancesCommand implements AdminCommand {
             return getServersForNodeOrConfig();
         }
         else if (rc.isServer()) {
-            List<Server> l = new LinkedList<Server>();
+            List<Server> l = new LinkedList<>();
             l.add((Server) rc);
             return l;
         }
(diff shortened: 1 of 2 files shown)


lprimak:
Please remove `final` in your changes. That just adds clutter to the code and doesn't really do anything

lprimak (changes requested):
please remove all the finals. otherwise looks good to me

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
```diff
@@ -214,32 +214,32 @@ private boolean notDas(String name) {
         return !SystemPropertyConstants.DAS_SERVER_NAME.equals(name);
     }
 
-    private void yesStatus(List<Server> serverList, int timeoutInMsec, Logger logger) {
+    private void yesStatus(final List<Server> serverList, final int timeoutInMsec, final Logger logger) {
         // Gather a list of InstanceInfo -- one per instance in domain.xml
-        RemoteInstanceCommandHelper helper = new RemoteInstanceCommandHelper(habitat);
+        final RemoteInstanceCommandHelper helper = new RemoteInstanceCommandHelper(habitat);
```
lprimak, line 219: final

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
```diff
@@ -214,32 +214,32 @@ private boolean notDas(String name) {
...
 
-        for (Server server : serverList) {
-            boolean clustered = server.getCluster() != null;
-            int port = helper.getAdminPort(server);
-            String host = server.getAdminHost();
+        for (final Server server : serverList) {
```
lprimak, line 221: final

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
sgflt: > Please remove `final` in your changes. That just adds clutter to the code and doesn't really do anything

I have to disagree with following reasons:

1. it should help to distinguish which variables are variable and
2. final variables does the code flow pretty staight forward. I mean if variable is final, there are no side effects and I can concentrate to pitfals of reused variables (wich may be really nasty). For example trying to find where regis variable comes consumes a lot of brain power (which have to be gathered from chocolate or coffe): https://github.com/payara/Payara/blob/75cc47823f5f9d6f401fa84fcdbaef8bbd55a310/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java#L3624 
3. final variables reduces scope where variable might be set and improves learning curve of code for contributors (by reducing possibilities where variable can be changed)
4. non final variables are like exclamation mark and acts like point of interest in case of hunting bugs
5. in old codebase final reveals scoping problems (

```
before (C89 declaration style):
String x = null;
while (...) {
  x = "some String";
  ...
}

after (Java style):
while (...) {
  final String x = "some String";
  ...
}
```

I know some time ago there was a discussion, but I do not know the final decision.

For reference there is PR in commons-lang3 with the similar idiom. https://github.com/apache/commons-lang/pull/687/commits/37c22318bc3e33c3df3a60afb5aac68659eaeadc

Another insteresting references
- https://softwareengineering.stackexchange.com/questions/98691/excessive-use-final-keyword-in-java
- https://stackoverflow.com/questions/137868/using-the-final-modifier-whenever-applicable-in-java

So if there is consensus on that the final is just garbage, I can discard the last commit.

Personally I would preffer to use final at least in methods that are longer than usual and contains if constructs. In case final keyword is not wanted in all possible places.

sgflt: Bu you are right in code review, it should be in different PR.

lprimak (approved):

lprimak:
jenkins test

lprimak:
I did like the diamond operator changes, can you leave them in this PR (just remove the final stuff)
thank you!

sgflt:
Already moved all code style changes into separate PR. This now contains just sorting one liners. :)

lprimak:
If you could add back the diamond interface fixes but not the finals in this PR I appreciate it

sgflt:
Okay, but it is hard to keep other things like `(notRunning > 0) ? false :true` unchanged. I will try to stay minimalistic and won't refactor everything. Maybe later.

lprimak:
> Okay, but it is hard to keep other things like (notRunning > 0) ? false :true unchanged. I will try to stay minimalistic and 

Please feel free to refactor but don't try to introduce new code style that doesn't already exist (such as finals everywhere)
Thank you!

lprimak (approved):

lprimak:
jenkins test

nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/ListInstancesCommand.java
```diff
@@ -263,8 +263,9 @@ private void yesStatus(List<Server> serverList, int timeoutInMsec, Logger logger
         }
 
         Properties extraProps = new Properties();
-        List instanceList = new ArrayList();
+        List<Map<String, Object>> instanceList = new ArrayList<>();
 
+        infos.sort(Comparator.comparing(InstanceInfo::getName));
```
dmatej, line 268: Ha!

dmatej (approved):

PR #5090, opened 2021-01-12
FISH-892 Clear MP config cache after startup so extesions can be picked up

Cousjava:
## Description
This is a bug fix.

### Testing Performed
Used reproducer in original Jira 

### Testing Environment
Zulu JDK 1.8_272 on Ubuntu 20.10 with Maven 3.6.3


Cousjava:
Jenkins test please

MattGill98 (approved):

PR #5092, opened 2021-01-13
FISH-877 Add config ordinal to cloud config sources

Cousjava:
## Description
This is a bug fix

### New tests
New test add at https://github.com/payara/Payara/compare/master...Cousjava:FISH-877-mp-config-ordinals-cloud?expand=1#diff-04fc586ba2c85eee6bd5376c25be567f24a810aba8d3b61b686ddb8064cfe54fR122

### Testing Performed
Using the reproducer from FISH-892, also added a config property with the same key but a different value to the domain config properties. Used curl on the endpoint to get the cloud value. Changed the ordinal of cloud config source to be less then that of domain and ran curl again, this time got the value from the domain config.

### Testing Environment
Zulu JDK 1.8_272 on Ubuntu 20.10 with Maven 3.6.3


Cousjava:
Jenkins test please

MattGill98 (approved):

MeroRai (approved):

PR #5098, opened 2021-01-21
FISH-990 OpenTracing Active Span is NULL when retrieved in EJB tracer on remote execution

Pandrex247:
## Description
This is a bug fix.

There were certain scenarios where it was possible for the span propagated from a remote EJB client to not get picked up by the server-side Tracer. This was traced back to the getTracer method - the server-side tracer wasn't using the scope manager of the RMI tracer. There was essentially a race condition present in the creation of Tracers - everything worked _as long as_ the RMI tracer was initialised first _and_ had an active span.

I've changed it so that we now use a shared scope manager across all tracers. The scope manager is focussed on maintaining the thread local active scope, so I don't _think_ this will cause any issues with span scopes bleeding across the different tracers when they shouldn't be.

### Blockers
None

### New tests
None.

### Testing Performed
Turned on Request Tracing, and configured it to print out to the log for any trace over 30 microseconds.

Ran the [remote-ejb-tracing](https://github.com/payara/Payara/tree/master/appserver/tests/payara-samples/samples/remote-ejb-tracing) sample, [my old microprofile on micro example](https://github.com/Pandrex247/Payara-Examples/tree/MicroProfileOnMicro/microprofile/microprofile-on-micro), and the project attached to the Jira.

### Testing Environment
Windows 10, JDK 8

## Documentation
N/A

## Notes for Reviewers
I highly recommend you try using your own examples to trigger the traces. Tracing is quite delicate and touches many different parts of code, so poking it in many different ways would be appreciated.

Be aware that since these changes are to the internal tracer, the MicroProfile OpenTracing TCK is of limited use as a testing mechanism here (since it uses the MockTracer).

### DIFF
--- a/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
+++ b/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
@@ -77,6 +77,7 @@ public class OpenTracingService implements EventListener {
 
     // The tracer instances
     private static final Map<String, Tracer> tracers = new ConcurrentHashMap<>();
+    private static final ScopeManager scopeManager = new fish.payara.opentracing.ScopeManager();
     
     // The name of the Corba RMI Tracer
     public static final String PAYARA_CORBA_RMI_TRACER_NAME = "__PAYARA_CORBA_RMI";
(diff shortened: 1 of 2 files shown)


lprimak:
Not sure if it conflicts with my fix that happened during samples refactoring otherwise looks good

lprimak (approved):

lprimak:
Jenkins test

PR #5104, opened 2021-01-29
FISH-788: DirConfigSource should not log severe messages if unused.

poikilotherm:
## Description
During QA after merging #5007, @lprimak stumbled over a SEVERE
log message during deploys, triggered by the non-existing directory
of DirConfigSource. The directory config value has had a default
before, also the docs say otherwise.

The level issue has been fixed by checking only default config value
being used, resulting in an INFO level message. The logic has been
refactored not to use an exception, but Optional API.

### Blockers
None.

### New tests
Batteries included.

### Testing Performed
Usual unit tests.

### Testing Environment
- AdoptOpenJDK (build 11.0.9+11)
- Linux, Fedora 33
- Maven 3.6.3

## Documentation
- Docs will be fixed to include the formerly hidden default value, see https://github.com/payara/Payara-Community-Documentation/pull/126

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/DirConfigSource.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/DirConfigSource.java
@@ -56,14 +56,19 @@ import java.nio.file.WatchKey;
 import java.nio.file.WatchService;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static java.util.Collections.unmodifiableMap;
 import static java.util.concurrent.TimeUnit.SECONDS;
-import static java.util.logging.Level.INFO;
+import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.SEVERE;
 import static java.util.logging.Level.WARNING;
 import static java.util.stream.Collectors.toMap;
@@ -290,9 +295,9 @@ public class DirConfigSource extends PayaraConfigSource implements ConfigSource 
         
         // Log that the configured directory is not resolving.
         Level lvl = SEVERE;
-        // Reduce log level to info if default setting, as the admin might simply not use this functionality.
+        // Reduce log level to fine if default setting, as the admin might simply not use this functionality.
         if(path.equals(DEFAULT_DIR))
-            lvl = INFO;
+            lvl = FINE;
         logger.log(lvl, "Given MPCONFIG directory '" + path + "' is no directory, cannot be read or has a leading dot.");
         return Optional.empty();
     }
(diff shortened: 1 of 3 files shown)


pdudits:
Jenkins test please

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/DirConfigSource.java
```diff
@@ -269,22 +270,31 @@ public String getName() {
...
+                return Optional.of(candidate);
             }
         }
-        throw new IOException("Given MPCONFIG directory '"+path+"' is no directory, cannot be read or has a leading dot.");
+        
+        // Log that the configured directory is not resolving.
+        Level lvl = SEVERE;
+        // Reduce log level to info if default setting, as the admin might simply not use this functionality.
+        if(path.equals(DEFAULT_DIR))
+            lvl = INFO;
```
pdudits, line 295: I'd go even to `FINE` for this. Since this is default and directory is not present, there's no need for distraction in logs.

pdudits (approved):
Tested with apps that use and don't use MP Config.

* [x] INFO error message when default directory doesn't exist -- I'd go for `FINE` though
* [x] no message when directory exists
* [x] error when non-default directory doesn't exist

poikilotherm:
My idea with using info: it would be easier to find out if it's not working as intended, in case someone actually wants to use it with the default value. Otherwise there won't be a message and people need to figure out what logging to tune.

I'll do as you folks handle this usually, please guide me 🤗

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/DirConfigSource.java
poikilotherm: My idea with using info: it would be easier to find out if it's not working as intended, in case someone actually wants to use it with the default value. Otherwise there won't be a message and people need to figure out what logging to tune.

I'll do as you folks handle this usually, please guide me 🤗

pdudits: There is already so much log entries during startup, we try not to introduce new ones unless they're necessary.

The default relative directory was also a bit of unfinished feature -- it wasn't documented and both docs and admin UI say it wants a full path. And I think it serves better when it is a full path as well. So I wouldn't advertise it too much by means of a log message at every deployment.

I think the best solution would be to remove default value, but that would break existing installations, therefore this one is the best way out we've got.

poikilotherm: :ok: 

Side note: the docs are updated in https://github.com/payara/Payara-Community-Documentation/pull/126 and now document the default value plus the option to go with relative path. One day there might be other location candidates where a relative path would be nice to use (that's why it's a list in `findDir()`)

poikilotherm: Done within 6921cab

pdudits:
Jenkins test please

lprimak:
jenkins test

lprimak:
the tests failed due to a known bug (which I am fixing)

lprimak (approved):

PR #4989, opened 2020-11-11
FISH-658 MicroProfile Config API 2.0

MattGill98:
Signed-off-by: Matthew Gill <matthew.gill@live.co.uk>

## Description

This is a first attempt at Config 2.0 support. It can't be tested yet, as it depends on the other MP specs. Each of the specs currently require Config <2.0, so this PR would break that. I'm leaving it as a draft PR until it's ready to be tested and refactored.

### Blockers

**Definitely MP Metrics and Fault Tolerance, and maybe each of the other MP APIs**

### New tests

New TCK: https://github.com/payara/MicroProfile-TCK-Runners/pull/124

### Testing Performed

**None yet, cannot merge**

PR #5105, opened 2021-01-29
FISH-882 Finalise MP Artifact Versions

MattGill98:
Blocked on Config upgrade: https://github.com/payara/Payara/pull/4989

TCK runner versions will need finalising first, which in turn depends on: https://github.com/payara/MicroProfile-TCK-Runners/pull/134 and https://github.com/payara/MicroProfile-TCK-Runners/pull/124


cubastanley (approved):
Should there be a version update for config and rest client or is that in a separate PR?

MattGill98:
jenkins test please

PR #5110, opened 2021-02-01
add_adfs_compitblity

### LINKED ISSUE #5083: [payara-api - openid connect] Not compatible to Active Directory Federation Service
# Description #
----------
The openid implementation of the payara security library cannot handle adfs. this is because adfs needs extra parameter to get all claims.

## Expected Outcome

The openid request to adfs returns all claims requested by scopes and extra claims from the web api definition in adfs.

## Current Outcome

adfs returns 401 error because it expect an extra parameter named **resource**. After i set the paramter resource in the definitions extra parameter to **resource=urn:microsoft:userinfo** it returns the claim sub from userinfo endpoint and inside the Identity Token there are some basic user info like upn. but it respects only the scope **openid** all other scopes are not respected.

### Prepare adfs
Install adfs and create server application group with webapi. Make sure the webapi has same id like server application. In the webapi prpoerties add some claims in the issuance transform rules.

### In the example project set this parameter

Set **redirectURI,clientSecret,clientId,providerURI** properties to fit your adfs installation. then set scope **openid allatclaims** and responseType to **code**.

## Context (Optional)

I tryed to set the resource parameter to the webapi identifier for example **c45d1e0b-0f18-4a92-ab3c-a32b12075a8b** or **https://web1.test.local/TestItsNg** but hen always i see with wireshark in the /userinfo endpoint request the http return error coe 401 with message 

> Bearer error="invalid_token", error_description="MSIS9921: Received invalid UserInfo request. Audience 'microsoft:identityserver:c45d1e0b-0f18-4a92-ab3c-a32b12075a8b' in the access token is not same as the identifier of the UserInfo relying party trust 'urn:microsoft:userinfo'."

When i set extra parameter to urn:microsoft:userinfo the i got back the sub claim from userinfo endpoint and inside identity token there are some basic info about user.

After a while i find out that the adfs reource parameter has also o be dded to the token endpoint request there is no resource parameter added.
>HTML Form URL Encoded: application/x-www-form-urlencoded
>   Form item: "client_id" = "c45d1e0b-0f18-4a92-ab3c-a32b12075a8b"
>    Form item: "client_secret" = "yhHfNywTBU9gnCp8dEHfgdFwin1x5XStQJEvCpdG"
>    Form item: "grant_type" = "authorization_code"
>    Form item: "code" = "mDq0_AH7B0Sjtyg89YuIkg.aWyGs-uv2AgFABog0Mik6rhtB7o.KKe5DS1mkj9g-Cc5IT445nISrLzk4H0U5q6bmE4STM2eMncsxMMmXY1IaUz80DSmhrEN_jk7XUIWgP83ryjBjN3IZTTzSlpdSWLBuMndMxMTjviI_QY0SQ3QeFpahCWWocN9rwU3O3BtG7L8G2kIovxuHlabYREO9w6RiYH
>    Form item: "redirect_uri" = "https://web1.test.local:8181/TestItsNg/callback"

So this means that you should either add the extra parameters also to the token request, implement new exra parameter only for token request or make it possible to deactivate the userinfo endpoint call.

## Environment ##

- **Payara Version**: 5.2020.7
- **Edition**: Full
- **JDK Version**: Tested with 8 and 11
- **Database**: not needed

markusg80:
# Add adfs function to openid connect api fixes #5083

## Description
This fix add the ability to use payara openid connect connector to get claims from Active Directory Federation Service 

### Testing Performed
Test to authenticate with adfs in my test environment. **Successfully returned claims if defined in the Web-API**.
All the test are performed with **Windows Server 2019**. It **should work also with Server 2016**, but i can't confirm this.
With **Windows Server 2012 R2 or before it will not work**.

### Testing Environment
Windows Server 2019, OpenJDK 8

## Documentation
i will complete documentation after review is done

## Notes for Reviewers
Please take a look into the comment inside the code. i've tryed to declare with code comments why i'm doing such strange things. Long story short: **Its because Microsoft isn't able to respect standards**


lprimak:
jenkins test

markusg80:
It seams to fail because of build order! In the code there is the RESOURCE field

lprimak:
Hi, @markusg80 
If you think the build order is wrong, please fix it.

markusg80:
Sorry but here i need your help. I don’t know the internas of the application. Can u take a look? I think the code is ok because i use always existing system

lprimak:
Were you able to compile the code on your own machine at least?

nbactions-release-profile.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lprimak, line 1: This file needs to be removed

api/pom.xml
```diff
@@ -53,7 +53,6 @@
     <description>Aggregator Module for Artefacts that exposes the public API of Payara Application Server</description>
 
     <modules>
-        <module>payara-api</module>
         <module>payara-bom</module>
```
lprimak, line 56: This change needs to be reverted

lprimak (changes requested):
See comments about the code. There are zero unit or integration tests.
We use Arquillian for integration tests.

api/pom.xml
markusg80: Ok thanks. This make sense. So i will revert this change this evening.

lprimak:
jenkins test

api/payara-api/src/main/java/fish/payara/security/annotations/AdfsAuthenticationDefinition.java
```diff
@@ -0,0 +1,377 @@
+/*
```
lprimak, line 1: This file is a copy-paste of another one. This needs to be inherited from a common class

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/OpenIdIdentityStore.java
```diff
@@ -108,8 +111,44 @@ public CredentialValidationResult validate(OpenIdCredential credential) {
...
+                Map<String, String> convertedTokenMap = new HashMap<>();
+                accesTokenClaims.entrySet().stream()
+                        .forEach(claim -> {
+                            if (claim.getValue() instanceof String) {
+                                convertedTokenMap.put(claim.getKey(), (String) claim.getValue());
+                            }
+                        });
+                
+                // convert AccesTokenClaims Map to map<String, String> and build json object
+                Map<String, Object> identityTokenClaims = credential.getIdentityToken().getClaims();
```
lprimak, line 132: This needs to be refactored. copy-paste of a previous block

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/adfs/AdfsOpenIdExtension.java
```diff
@@ -0,0 +1,242 @@
+/*
```
lprimak, line 1: Another copy-paste file, needs a complete refactor

lprimak (changes requested):
Unfortunately, this code is mostly copy/paste of Gaurav's work. This cannot be accepted in it's current form. Sorry.

The code needs to be refactored so it's not full of code duplication smells

markusg80:
@lprimak 
AdfsAuthenticationDefinition is realy copy past except of resource field. But i doing so because GoogleAuthenticationDefinition and AzureAuthenticationDefinition are excatly same copy of OpenIdAuthenticationDefinition. So i think of implement the OpenIdAuthenticationDefinition but i decided to not doing so because existing system was copy/past. Sorry

With AdfsOpenIdExtension its same. When you take a look on AzureOpenIdExtension and GoggleOpenIdExtension then this two are mostly identical. And in this case i have to make the question: Why reinvent the wheel.

With the double code block this is really bad programming from my side nad i will fix this now. I also will think about to find better solution for the other two points mentioned from you.

lprimak:
Yes, "the rule of three" applies here.
Two copies *may* (but probably not) be acceptable, but certainly not three :)

lprimak:
Perhaps @jGauravGupta can help with this too

markusg80:
So now i already fixed the double code block from IdentityStore. I also inform myself if i can extend an @Interface. But its not possible at least without using Spring AnnotationUtils. But it not make sense to include new library to only save one class. Maybe @jGauravGupta really can help.

lprimak:
maybe take a look at `@stereotype` for ideas

markusg80:
@Stereotype needs first another interface / class with @OpenIdAuthenticationDefinition Annotation. So i dont see ere big advantages. @jGauravGupta can help. With @Stereotype i have to change existing classes and design.

lprimak:
jenkins test

markusg80:
Maybe to solve this situation i could remove the AdfsAuthenticationDefinition and integrate this in OpenIdAuthenticationDefinition. Maybe i add teo field for example adfsEndpoint = boolean and then resource field. If resource field isn’t set client id will be used as resource field when adfsEndpoint is true. What you zhink about it?

lprimak:
> Maybe to solve this situation i could remove the AdfsAuthenticationDefinition and integrate this in 

That may work. What I am watching out for is code duplication and other code smells.
Also, the "Adfs" name doesn't mean anything to someone (like me) not intimately familiar with what this is supposed to do. 
Can you come up with a better name?

appserver/payara-appserver-modules/security-openid/pom.xml
```diff
@@ -163,6 +163,16 @@
                         </configuration>
                     </execution>
                 </executions>
+                <configuration>
+                    <instructions>
+                        <Export-Package>
+                                    fish.payara.security.openid.azure,
```
lprimak, line 169: What are the security implications of exposing these packages?

appserver/payara-appserver-modules/security-openid/src/main/java/fish/payara/security/openid/adfs/AdfsOpenIdAuthenticationMechanism.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package fish.payara.security.openid.adfs;
+
+import fish.payara.security.annotations.AdfsAuthenticationDefinition;
+import fish.payara.security.openid.OpenIdAuthenticationMechanism;
+import static fish.payara.security.openid.adfs.AdfsOpenIdExtension.toOpenIdAuthDefinition;
+import javax.enterprise.inject.Typed;
+
+/**
+ * The  Active Directory Federation Service AuthenticationMechanism used to authenticate users using the
```
lprimak, line 48: Also copy/paste

markusg80:
@lprimak ok i will integrate this in the main OpenaidAuthenticationDefinition. Regarding to replace the name: This is not ease, because the full name is Active Directory Federation Service. So i stated this in the comments because the full name is far to long. And to use only Federation Service is misleading too. So i will think about to improve also this point.

lprimak:
Also, there are no unit or integration tests of this.
You can use mockito for unit tests are arquillian for integration tests.
For example of integration tests, see `appserver/tests/payara-samples/samples`
It's critical to have tests for a feature such as this, because it will inevitably break and regress with time
if tests are not there.

lprimak:
> Regarding to replace the name: 

Suggestion: ADFederation or some derivative of that. "AD" is recognizable in the industry

AlanRoth:
Hi Markus, 

Thank you for submitting a PR, however, I will be closing this PR as it has been decided that we are not confident accepting your code into our codebase. We still welcome you to submit PRs in the future.

I apologize for any inconvenience caused and I thank you for your understanding,
Alan

markusg80:
@AlanRoth 
I really can not understand this decision. I and @lprimak was close to coming to a solution. Adfs is widely used and it will be bad if payara api lack of support this. It will also help if we can disable the userinfo endpoint call for openid.

PR #5115, opened 2021-02-04
FISH-984 Max Wait Time isn't respected when the JDBC pool is locked for a long time

jGauravGupta:
## Description
This is a bug fix to cancel the connection pool manager retry task on completion max wait timeout.

### Testing Performed
Add the following configuration to domain.xml:
````
    <jdbc-connection-pool connection-creation-retry-interval-in-seconds="20" max-wait-time-in-millis="10000" datasource-classname="org.apache.derby.jdbc.ClientDataSource" connection-creation-retry-attempts="3" name="SamplePool" res-type="javax.sql.DataSource">
      <property name="User" value="app"></property>
      <property name="DatabaseName" value="sample"></property>
      <property name="serverName" value="mylocalhost"></property>
      <property name="PortNumber" value="1527"></property>
      <property name="URL" value="jdbc:derby://mylocalhost:1527/sample"></property>
      <property name="Password" value="app"></property>
    </jdbc-connection-pool>
    <jdbc-resource pool-name="SamplePool" jndi-name="jdbc/sample"></jdbc-resource>
````
You may also add/update the pool info via admin console. And make sure database service is shutdown.
Now run the following source snippet and the retuned time should be equivalent to Max wait time (+Δ).
````
    @Resource(lookup = "jdbc/sample")
    private DataSource testDS;

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    public String ping() {
        long start = System.currentTimeMillis();
        StringBuilder contents = new StringBuilder("Connection valid: ");
        try (Connection conn = testDS.getConnection()) {
            contents.append(conn.isValid(30));
        } catch (SQLException ex) {
            contents.append("Failed to get connection.");
            Logger.getLogger(PingEndpoint.class.getName()).log(Level.SEVERE, contents.toString(), ex);
        }
        contents.append((System.currentTimeMillis() - start));
        return contents.toString();
    }
````

Alternate solution: https://github.com/payara/Payara/pull/5118

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2019] Payara Foundation and/or affiliates
+// Portions Copyright [2019-2021] Payara Foundation and/or affiliates
 
 package com.sun.enterprise.resource.pool;
 
@@ -67,6 +67,13 @@ import javax.resource.ResourceException;
 import javax.resource.spi.RetryableUnavailableException;
 import javax.transaction.Transaction;
 import java.util.*;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.glassfish.api.admin.ServerEnvironment;
@@ -85,6 +92,7 @@ public class ConnectionPool implements ResourcePool, ConnectionLeakListener,
             StringManager.getManager(ConnectionPool.class);
     protected final static Logger _logger = LogDomains.getLogger(ConnectionPool.class,LogDomains.RSR_LOGGER);
 
+    private ExecutorService execService = Executors.newCachedThreadPool();
 
     //pool life-cycle config properties
     protected int maxPoolSize;          // Max size of the pool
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by OndroMih** [comment]
I humbly believe that this solution doesn't solve the core issue. Although it passes your test case and the request is timed out properly, the request leaves a thread that remains waiting in the call to internalGetResource until the connection pool is unblocked. Eventually, the executor service can become blocked because all threads will be blocked like this.

The timeout also has to be accounted with at the deeper levels, in particular in the `createSingleResource` to break from the loop if timeout is reached. This can be done either by propagating the `startTime` through several methods to the `createSingleResource` or store its value into a thread-local field so that the `createSingleResource` method can retrieve it from the `this` object. Then `createSingleResource` would continue with `Thread.sleep()` only if it would sleep for a shorter time than the remaining time, otherwise it would jump out of the loop.

For example:

```
private final ThreadLocal<Long> getResourceStartTime = new new ThreadLocal<>();
```
In `getResource()`:
```
long startTime = System.currentTimeMillis();  // this line already exists in the code
this.getResourceStartTime.set(startTime);            // add this line
```
In `createSingleResource()`:
```
while (true) {
    try {
...
resourceHandle = resourceAllocator.createResource();
...
    } catch (Exception ex) {
         ...
         long elapsedWaitTime = System.currentTimeMills() - this.getResourceStartTime.get();
         if ( elapsedWaitTime + this.conCreationRetryInterval_ >= this.maxWaitTime ) {
              break;
         }
         try {
                    Thread.sleep(conCreationRetryInterval_);
            ...
        }
   }
}
```

jGauravGupta:
Hi @OndroMih ,

As discussed over Teams this PR also fixes the core issue as If timeout happens then[ cancel(true)](https://github.com/payara/Payara/blob/e64fbbe54cb87950fb2953bc1236dc689d559b46/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java#L405) function is called which actually interrupts the poll manager Thread.sleep() and [throws PollingException](https://github.com/payara/Payara/blob/e64fbbe54cb87950fb2953bc1236dc689d559b46/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java#L973) which eventually breaks from the loop.

The solution suggested by you is also good. I have created a PR for that: https://github.com/payara/Payara/pull/5118

jGauravGupta:
jenkins test please

PR #5118, opened 2021-02-05
FISH-984 Max Wait Time isn't respected when the JDBC pool is locked for a long time 

jGauravGupta:
## Description
This is a bug fix to cancel the connection pool manager retry task on completion max wait timeout.

### Testing Performed
Add the following configuration to domain.xml:
````
    <jdbc-connection-pool connection-creation-retry-interval-in-seconds="20" max-wait-time-in-millis="10000" datasource-classname="org.apache.derby.jdbc.ClientDataSource" connection-creation-retry-attempts="3" name="SamplePool" res-type="javax.sql.DataSource">
      <property name="User" value="app"></property>
      <property name="DatabaseName" value="sample"></property>
      <property name="serverName" value="mylocalhost"></property>
      <property name="PortNumber" value="1527"></property>
      <property name="URL" value="jdbc:derby://mylocalhost:1527/sample"></property>
      <property name="Password" value="app"></property>
    </jdbc-connection-pool>
    <jdbc-resource pool-name="SamplePool" jndi-name="jdbc/sample"></jdbc-resource>
````
You may also add/update the pool info via admin console. And make sure database service is shutdown.
Now run the following source snippet and the retuned time should be equivalent to Max wait time (+Δ).
````
    @Resource(lookup = "jdbc/sample")
    private DataSource testDS;

    @GET
    @Produces(MediaType.TEXT_PLAIN)
    public String ping() {
        long start = System.currentTimeMillis();
        StringBuilder contents = new StringBuilder("Connection valid: ");
        try (Connection conn = testDS.getConnection()) {
            contents.append(conn.isValid(30));
        } catch (SQLException ex) {
            contents.append("Failed to get connection.");
            Logger.getLogger(PingEndpoint.class.getName()).log(Level.SEVERE, contents.toString(), ex);
        }
        contents.append((System.currentTimeMillis() - start));
        return contents.toString();
    }
````

Alternate solution: https://github.com/payara/Payara/pull/5115

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/ConnectionPool.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2019] Payara Foundation and/or affiliates
+// Portions Copyright [2019-2021] Payara Foundation and/or affiliates
 
 package com.sun.enterprise.resource.pool;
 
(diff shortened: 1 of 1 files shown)


jGauravGupta:
jenkins test please

Cousjava (approved):

PR #5122, opened 2021-02-05
FISH-1061 Fix NullPointer when Configuring a Notifier against a non-DAS Instance or Config

Pandrex247:
## Description
You would get an NPE when configuring notifiers for anything other than the server-config, due to the default config tags not being present in the default domain.xml. This change makes it so default config tags will be created if not already present.

### Blockers
None

### New tests
Nope

### Testing Performed
Start a fresh domain:

* Attempt to configure a notifier against the server config: .\appserver\distributions\payara\target\stage\payara5\bin\asadmin set-cdieventbus-notifier-configuration --enabled false --target default-config
* Attempt to configure a notifier against the default config: .\appserver\distributions\payara\target\stage\payara5\bin\asadmin set-cdieventbus-notifier-configuration --enabled false --target default-config
* Attempt to configure a notifier against a standalone instance using Admin Console

User should receive no errors and see that config changes taking effect in the domain config

### Testing Environment
Windows 10, JDK 8u282

## Documentation
Nada

### DIFF
--- a/nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/admin/BaseSetNotifierConfigurationCommand.java
+++ b/nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/admin/BaseSetNotifierConfigurationCommand.java
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (c) [2016-2020] Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016-2021 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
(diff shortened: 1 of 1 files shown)


nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/admin/BaseSetNotifierConfigurationCommand.java
```diff
@@ -111,13 +110,42 @@ public void execute(final AdminCommandContext context) {
...
+        }
+
         try {
-            ConfigSupport.apply(new SingleConfigCode<C>() {
-                public Object run(C cProxy) throws PropertyVetoException, TransactionFailure {
-                    applyValues(cProxy);
-                    actionReport.setActionExitCode(ActionReport.ExitCode.SUCCESS);
-                    return cProxy;
-                }
+            ConfigSupport.apply(cProxy -> {
```
Pandrex247, line 145: No change in config here, just using a lambda

Cousjava (approved):

Pandrex247:
Jenkins test please

PR #5131, opened 2021-02-17
FISH-1069 Rationalise Fault Tolerance Configuration

MattGill98:
The Fault Tolerance configuration was a bit of a mess, with the correct
configuration parameters not being used, and the incorrect ones being
reported.

This commit makes the managed executor names the configurable options,
and removes the dead configuration options.

As an added bonus, the set command now validates the executor names.
This should work as expected in the admin console and from the command
line.

Docs: https://github.com/payara/Payara-Community-Documentation/pull/135


MattGill98:
jenkins test please

MeroRai (approved):
Great work @MattGill98. The validation to the set command is an excellent addition.

PR #5161, opened 2021-03-12
FISH-1186: Support server-node on k8s

rdebusscher:
## Description
Fix value for DOCKER_CONTAINER_ID within entrypoint.sh of server-node image so it has meaningful value within k8s.

### Testing Performed
Tested on Docker Desktop on Mac with k8s  1.19

### DIFF
--- a/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
+++ b/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
@@ -2,6 +2,10 @@
 set -e
 
 DOCKER_CONTAINER_ID="$(cat /proc/self/cgroup | grep :/docker/  | sed s/\\//\\n/g | tail -1)"
+if [ -z "${DOCKER_CONTAINER_ID}" ]; then
+    echo "Fallback for Docker Container ID"
+    DOCKER_CONTAINER_ID=$HOSTNAME
+fi
 ASADMIN="${PAYARA_DIR}/bin/asadmin"
 echo "Docker Container ID is: ${DOCKER_CONTAINER_ID}"
 
--- a/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
+++ b/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
@@ -152,7 +152,7 @@ function createNewInstance {
         fi
     fi
 
-    if [ -z "${DOCKER_CONTAINER_ID}" ]; then
+    if [ ! -z "${DOCKER_CONTAINER_ID}" ]; then
        # Register Docker container ID to DAS
        echo "Setting Docker Container ID for instance ${PAYARA_INSTANCE_NAME}: ${DOCKER_CONTAINER_ID}"
        ASADMIN_COMMAND="${ASADMIN} -I false -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} _set-docker-container-id --instance ${PAYARA_INSTANCE_NAME} --id ${DOCKER_CONTAINER_ID}"
--- a/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
+++ b/appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
@@ -2,10 +2,7 @@
 set -e
 
 DOCKER_CONTAINER_ID="$(cat /proc/self/cgroup | grep :/docker/  | sed s/\\//\\n/g | tail -1)"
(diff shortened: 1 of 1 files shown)


Pandrex247 (commented):
Would this work?
The container ID is registered against a Payara Server instance, and is intended to be a unique relationship. It's used when talking to the Docker REST API to perform operations (start, stop etc.), setting it to something like "localhost" seems wrong.

rdebusscher:
@Pandrex247 $HOSTNAME within Kubernetes is the pod name, so it is never localhost.

rdebusscher:
```
>kubectl get pods
NAME                            READY   STATUS    RESTARTS   AGE
payara-domain-7986d45f6-dgdfb   1/1     Running   0          115s
payara-node-54b465c945-ncrd2    1/1     Running   0          11s
payara-node-54b465c945-qdvf9    1/1     Running   0          11s

```

```
>kubectl logs payara-node-54b465c945-ncrd2
Fallback for Docker Container ID
Docker Container ID is: payara-node-54b465c945-ncrd2
No Docker container IP override given, setting to first result from 'hostname -I'
Hostname is 10.1.0.97
Docker Container IP is: 10.1.0.97
No Instance name given.
No node name given.
WARNING: Could not find a matching Docker Node: Creating a temporary node specific to this container - cleanup of this container cannot be done by Payara Server
Creating a temporary node with an autogenerated name.
/opt/payara/appserver/bin/asadmin -I false -T -a -H 10.1.0.95 -p 4848 -W /opt/payara/passwords/passwordfile.txt _create-node-temp --nodehost 10.1.0.97
Running command create-local-instance:
/opt/payara/appserver/bin/asadmin -I false -T -a -H 10.1.0.95 -p 4848 -W /opt/payara/passwords/passwordfile.txt create-local-instance --node Delightful-Hake --config demo-config --dockernode true --ip 10.1.0.97 --deploymentgroup demo
Setting Docker Container ID for instance Silly-Paddlefish: payara-node-54b465c945-ncrd2
/opt/payara/appserver/bin/asadmin -I false -H 10.1.0.95 -p 4848 -W /opt/payara/passwords/passwordfile.txt _set-docker-container-id --instance Silly-Paddlefish --id payara-node-54b465c945-ncrd2
Command _set-docker-container-id executed successfully.
Starting instance Silly-Paddlefish
```

Pandrex247:
OK I see what you're doing, but I'm not sure the fallback is necessary.

Since the Docker container ID is only used when talking to the Docker REST API, and from a 30 sec scan in Payara this only happens if you're specifically using a Docker node (as in a managed node) and not a Temp (unmanaged) node like you are here, I think you could just skip the `_set-docker-container-id` command.

That is unless you have some other reason for wanting the container ID.
I assume the reason you're doing this change is because you get a command exception from not having a container ID.

Pandrex247:
It's a bit more effort, but strikes me as the "more correct" solution.

dmatej:
It still makes sense to me to have same id in docker and hostname, when I was writing those tests it might be easier to see what is what. But I'm not k8s expert, what @pdudits  thinks?
Btw those tests passed.

pdudits:
I on @Pandrex247's side. The pod has no access to Docker REST API, therefore it shoudn't get configured.

rdebusscher:
Moved the if around __set_container_id_ asadmin command.  value is not used in the Temporary Node scenario.

Tested out with example and DeploymEnt Group is correctly created in this scenario.

rdebusscher:
jenkins test please

appserver/extras/docker-images/server-node/src/main/docker/bin/entrypoint.sh
```diff
@@ -151,11 +152,13 @@ function createNewInstance {
         fi
     fi
 
-    # Register Docker container ID to DAS
-    echo "Setting Docker Container ID for instance ${PAYARA_INSTANCE_NAME}: ${DOCKER_CONTAINER_ID}"
-    ASADMIN_COMMAND="${ASADMIN} -I false -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} _set-docker-container-id --instance ${PAYARA_INSTANCE_NAME} --id ${DOCKER_CONTAINER_ID}"
-    echo "${ASADMIN_COMMAND}"
-    ${ASADMIN_COMMAND}
+    if [ -z "${DOCKER_CONTAINER_ID}" ]; then
```
Pandrex247, line 155: Isn't this backwards? You're setting the container ID if it isn't present, rather than if it is.

rdebusscher: oops, how did that copy and paste went wrong? Thx @Pandrex247

Pandrex247 (approved):

cubastanley:
Jenkins test please

PR #5178, opened 2021-03-30
FISH-1208 Fix Failing Tests on Windows

Pandrex247:
The tests in `nucleus/payara-modules/nucleus-microprofile/config-service` fail because:
* Use of File.separator in regex, "\\" isn't valid - changed so on Windows it replaces "\\\\\\\\"
* Symbolic Links aren't enabled by default on Windows - changed so that the test gets skipped if the permissions aren't granted


Pandrex247:
Jenkins test please

lprimak (approved):

PR #5181, opened 2021-04-01
FISH-987 Extra Logging and Option to Disable Evaluating Class References

Pandrex247:
## Description
Feature.

Port from Payara 4.

> Payara Server 4 introduces a breaking change in EL (at least when it's used in a JSP). An EL expression in JSP which is evaluated correctly on GF 3 might confuse Payara Server 4 and 5 to thing that the expression contains a reference to an imported class and attempts to load that class. This ends in ClassNotFoundException. This happens every time the expression is evaluated. If there's a higher load on the server and the EL needs to be evaluated very often, the Classloader spends a significant amount of time searching for the class. Classloader puts a lock on the class and all the other threads that are evaluating the same expression have to wait for the lock. This can lead to too many threads blocked for a long time.
> 
> To fix this, we need to introduce a configuration option (system property) that reverts the behavior to be backwards compatible with GF 3. In other words, the EL engine should skip loading the imported classes if this option is enabled.

Changes are here: 
* https://github.com/payara/patched-src-el-ri/pull/3
* https://github.com/payara/patched-src-jsp-api/pull/3

### Blockers
None.

### New tests
None.

### Testing Performed
Pending...

## Documentation
Pending...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,8 +143,8 @@
         <jakarta.validation.version>2.0.2</jakarta.validation.version>
         <hibernate.validator.version>6.1.5.Final</hibernate.validator.version>
         <hibernate.validator-cdi.version>6.1.5.Final</hibernate.validator-cdi.version>
-        <jakarta.el.version>3.0.3.payara-p3</jakarta.el.version>
-        <jakarta.el-api.version>3.0.3</jakarta.el-api.version>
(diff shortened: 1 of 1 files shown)


Pandrex247:
Jenkins test please

lprimak (approved):

PR #5188, opened 2021-04-08
FISH-1293 Disassociate ClusteredStore from tenants

lprimak:
## Description
Bug. 
Tenant control associates all Hazelcast objects to their calling applications by default.
However, internal maps that belong to `ClusteredStore` should not do this.
This fix prevents it from associating thus preventing blocking by tenant control
when application is undeployed

- pushing empty context in ClusteredStore to prevent association of Hazelcast maps to tenant applications

  that may be later undeployed and would leave maps orphan, which in turn will lead
  to Hazelcast blocking calls to those maps
- minor cleanup

### Testing Performed
Ran Payara Samples twice, which is how I discovered this bug in the first place

## Review Notes
** Please use "ignore whitespace changes" to review,
otherwise it looks like changes don't make sense


lprimak:
jenkins test

dmatej (approved):

PR #5200, opened 2021-04-19
FISH-1315 Improve loading time for REST application when there are many password aliases #5177

### LINKED ISSUE #5177: Bug Report: Load of REST application is very slow (120 seconds) when there are many password aliases created / FISH-1315
# Description #
----------
I've several applications with REST services. The load of each application takes more than 120 seconds.
I've seen that the problem is that I've more than 800 password aliases created. If I empty domain-password file then the application loads in less than 1 second.

## Steps to reproduce ## 
I've attached a zip file than includes: test.war, test.java and a domain-password file with 800 aliases.
[test.zip](https://github.com/payara/Payara/files/6220346/test.zip)

1. Create and start test domain (`asadmin create-domain test; asadmin start-domain test`)
2. Deploy sample application (`asadmin deploy test.war`). It takes only **544ms**.
```
[2021-03-29T09:44:46.228+0200] [Payara 5.2021.1] [INFORMACIÓN] [AS-WEB-GLUE-00172] [javax.enterprise.web] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617003886228] [levelValue: 800] [[

  Loading application [test] at [/test]]]

[2021-03-29T09:44:46.334+0200] [Payara 5.2021.1] [INFORMACIÓN] [] [javax.enterprise.system.core] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617003886334] [levelValue: 800] [[
  test was successfully deployed in 544 milliseconds.]]
```
3. Create 800 password aliases (takes many hours) or copy the attached domain-password file
```
#!/bin/bash
PWFILE=/var/tmp/aspwfile.txt
echo "AS_ADMIN_ALIASPASSWORD=test" > $PWFILE
for ALIAS in {1..800}
do
   asadmin --passwordfile $PWFILE create-password-alias alias_${ALIAS}
done
rm $PWFILE
```
4. Redeploy test application (`asadmin deploy --force test.war`). It takes **133.186ms**.
```
[2021-03-29T09:49:57.832+0200] [Payara 5.2021.1] [INFORMACIÓN] [AS-WEB-GLUE-00172] [javax.enterprise.web] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617004197832] [levelValue: 800] [[
  Loading application [test] at [/test]]]

[2021-03-29T09:49:57.902+0200] [Payara 5.2021.1] [INFORMACIÓN] [] [javax.enterprise.system.core] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617004197902] [levelValue: 800] [[
  test was successfully deployed in 133.186 milliseconds.]]
```

The problem is when the application is loading. If I divide the deploy in several phases, the loading takes **2m12.324s**.

```
$ asadmin undeploy test
$ asadmin deploy --target domain test.war

[2021-03-29T09:50:41.870+0200] [Payara 5.2021.1] [INFORMACIÓN] [] [javax.enterprise.system.core] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617004241870] [levelValue: 800] [[
  test was successfully deployed in 142 milliseconds.]]

$ time asadmin create-application-ref --target server test
[2021-03-29T09:53:23.369+0200] [Payara 5.2021.1] [INFORMACIÓN] [AS-WEB-GLUE-00172] [javax.enterprise.web] [tid: _ThreadID=181 _ThreadName=admin-thread-pool::admin-listener(2)] [timeMillis: 1617004403369] [levelValue: 800] [[
  Loading application [test] at [/test]]]

 real    2m12.324s
 user    0m3.295s
 sys     0m0.483s
```
## Environment ##

- **Distribution**: Server Full Profile (5.2021.1)
- **JDK Version**: OpenJDK 11
- **Operating System**: CentOS 7

AngelTG2:
ISSUE 5177 - Load of REST application is very slow (120 seconds) when there are many password aliases created

## Description
This is a fix for the issue #5177
Load of REST application is very slow (120 seconds) when there are many password aliases created.

In this situation (800 password aliases), the cause of this slowness is that Microprofile OpenAPI reads the Microprofile Config when register an application in OpenAPI. PasswordAlias is a config source and it's necessary to read the keystore that is a taxing process if there are many password aliases.

One possible solution to resolve this issue is disable OpenAPI, but disabling OpenAPI seems that only affect to publish /openapi url. OpenAPI service is started even if OpenAPI is disabled and always register the application causing the slowness.

In this PR I've changed the registerApp method of OpenService class to prepare the document only if OpenAPI is enabled. With this modification, instead of taking 130 seconds to load an application it only takes less than 6.

Instead of modifying OpenAPIService class, the OpenApiApplicationContainer class could have been modified to avoid calling OpenApiService, but I thought it could have other side effects.

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
@@ -55,6 +55,9 @@ import java.util.logging.Logger;
 import org.glassfish.config.support.TranslatedConfigView;
 import org.glassfish.internal.api.Globals;
 
+import java.util.Set;
+import java.util.HashSet;
+
 /**
  *
  * @author steve
@@ -85,6 +88,20 @@ public class PasswordAliasConfigSource extends PayaraConfigSource {
         }
         return properties;
     }
+
+    @Override
+    public Set<String> getPropertyNames() {
+        Set<String> propertyNames = new HashSet<>();
+        if (store != null) {
+            Iterator<String> keys = store.keys();
+            while (keys.hasNext()){
+                String key = keys.next();
+                propertyNames.add(key);
+            }
+        }
+        return propertyNames;
+    }
+
 
     @Override
     public String getValue(String name) {
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2018-2020 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018-2021 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
@@ -47,7 +47,6 @@ import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -65,11 +64,11 @@ import java.util.HashSet;
 public class PasswordAliasConfigSource extends PayaraConfigSource {
 
     private final DomainScopedPasswordAliasStore store;
-    
+
     public PasswordAliasConfigSource() {
         store = Globals.getDefaultHabitat().getService(DomainScopedPasswordAliasStore.class);
     }
-    
+
     @Override
     public int getOrdinal() {
         return Integer.parseInt(configService.getMPConfig().getPasswordOrdinality());
@@ -79,26 +78,14 @@ public class PasswordAliasConfigSource extends PayaraConfigSource {
     @Override
     public Map<String, String> getProperties() {
         Map<String,String> properties = new HashMap<>();
-        if (store != null) {
-            Iterator<String> keys = store.keys();
-            while (keys.hasNext()){
-                String key = keys.next();
-                properties.put(key, new String(store.get(key)));
-            }
-        }
+        store.keys().forEachRemaining(key -> properties.put(key, new String(store.get(key))));
         return properties;
     }
 
     @Override
     public Set<String> getPropertyNames() {
         Set<String> propertyNames = new HashSet<>();
-        if (store != null) {
-            Iterator<String> keys = store.keys();
-            while (keys.hasNext()){
-                String key = keys.next();
-                propertyNames.add(key);
-            }
-        }
+        store.keys().forEachRemaining(propertyNames::add);
         return propertyNames;
     }
 
@@ -136,5 +123,4 @@ public class PasswordAliasConfigSource extends PayaraConfigSource {
     public String getName() {
         return "Password Alias";
     }
-    
 }
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
@@ -56,6 +56,7 @@ import org.glassfish.internal.api.Globals;
 
 import java.util.Set;
 import java.util.HashSet;
+import java.util.Objects;
 
 /**
  *
@@ -92,9 +93,7 @@ public class PasswordAliasConfigSource extends PayaraConfigSource {
 
     @Override
     public String getValue(String name) {
-        if (name == null || store == null) {
-            return null;
-        }
+        Objects.requireNonNull(name, "Name perameter cannot be null");
 
         String value = null;
 
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/OpenApiService.java
@@ -106,29 +106,40 @@ public class OpenApiService {
         this.withCorsHeaders = withCorsHeaders;
     }
 
-	public void registerApp(String applicationId, DeploymentContext ctx) {
+    public void registerApp(String applicationId, DeploymentContext ctx) {
         final WebBundleDescriptorImpl descriptor = ctx.getModuleMetaData(WebBundleDescriptorImpl.class);
         final String contextRoot = descriptor.getContextRoot();
         final ReadableArchive archive = ctx.getSource();
         final ClassLoader classLoader = ctx.getClassLoader();
-        documents.put(applicationId, new OpenAPISupplier(applicationId, contextRoot, archive, classLoader));
-        cachedResult = null;
-	}
 
-	public void deregisterApp(String applicationId) {
-        documents.remove(applicationId);
+        if (this.enabled) {
+            documents.put(applicationId, new OpenAPISupplier(applicationId, contextRoot, archive, classLoader));
+        }
         cachedResult = null;
-	}
+    }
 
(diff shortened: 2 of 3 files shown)


AlanRoth:
jenkins test please

smillidge:
This is not really a comment on this PR as I understand what it is achieving but the fact that OpenAPI is loading all ConfigProperties at boot is a bug in itself as this could have a big performance impact on many ConfigSources and some people will want both OpenAPI and a slow config source or a source with many keys.

AngelTG2:
@smillidge, I think you are right.

I was trying to find a solution to my problem, but probably some people need both OpenAPI and slow config sources.

It seems that OpenAPIConfiguration only needs several properties (mp.openapi.model.reader, mp.openapi.filter, mp.openapi.scan.lib, mp.openapi.scan.disable, mp.openapi.scan.packages, mp.openapi.scan.classes, mp.openapi.scan.exclude.packages, mp.openapi.scan.exclude.classes, mp.openapi.servers, mp.openapi.servers.path.*, mp.openapi.servers.operation.* and mp.openapi.schema.*)

I'm newbie in Microprofile and OpenAPI, so I ignore if these properties can be stored in a concrete config source or can be stored anywhere. If it can be stored anywhere,  I can't think of any way to extract those properties without first reading all the config sources.

smillidge:
If OpenAPI is just looking for specific properties. That isn't of itself a bug, if it is loading all properties that would be. The Password Alias config source probably needs a review to ensure it is performant.

AngelTG2:
I've reviewed the password alias config source:

- The slowest part of password aliases processing is getting the value. Getting  the alias name is very fast.

- OpenApiConfiguration constructor calls method getPropertyNames() of PayaraConfig class
https://github.com/payara/Payara/blob/f3d70f861dd6f517eaa5b33e9e75823062891a47/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/config/OpenApiConfiguration.java#L101

- getPropertyNames() in PayaraConfig calls getProperties().keySet() of PasswordAliasConfigSource instead configSource.getPropertyNames forcing all the properties (property names and values) to be read and slowing the process
https://github.com/payara/Payara/blob/f3d70f861dd6f517eaa5b33e9e75823062891a47/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java#L195

- PayaraConfigSource implements the method getPropertyNames as "return getProperties().keySet()", so if instead calling getProperties().keySet() from PayaraConfig the method _getPropertyNames()_ is called, the result will be same.
https://github.com/payara/Payara/blob/f3d70f861dd6f517eaa5b33e9e75823062891a47/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PayaraConfigSource.java#L77

- If method getPropertyNames() is overwritten in PasswordAliasConfigSource with the following code, then all run faster. I've checked and the test application is loaded in 2 seconds with OpenAPI enabled.
```
    @Override
    public Set<String> getPropertyNames() {
        Set<String> propertyNames = new HashSet<>();
        if (store != null) {
            Iterator<String> keys = store.keys();
            while (keys.hasNext()){
                String key = keys.next();
                propertyNames.add(key);
            }
        }
        return propertyNames;
    }

**Comment by smillidge** [comment]
@AngelTG2 thanks for the detailed analysis. This sounds similar to the problem described here https://github.com/payara/Payara/issues/5187? Do you want to make a PR for the change you mention as it sounds reasonable?

**Comment by AngelTG2** [comment]
@smillidge, I think the problem is similar to #5187 
I've make a new commit (fixes #5177 v2).
I've revert the changes in OpenApiService.java but I have corrected the indentation problems

**Comment by AlanRoth** [comment]
jenkins test please

File: nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java [hunk 1]
```diff
@@ -86,6 +89,20 @@ public int getOrdinal() {
         return properties;
     }
 
+    @Override
+    public Set<String> getPropertyNames() {
+        Set<String> propertyNames = new HashSet<>();
+        if (store != null) {
```
lprimak, line 95: Unless I am missing something, this null check is unnecessary, please remove it

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
```diff
@@ -86,6 +89,20 @@ public int getOrdinal() {
...
+            Iterator<String> keys = store.keys();
```
lprimak, line 96: This can be converted into Java 8 Stream construct instead of using raw iterators.
Code smell

lprimak (changes requested):

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
smillidge: Call me old fashioned I still love raw iterators

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
AngelTG2: I can be wrong, but if I remove this line and store is null, the line 96 will throw a NullPointerException

lprimak: store is an HK2 service singleton. It should never be null

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
lprimak: Yea you are old fashioned. So am I :)

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
AngelTG2: Ok. I'll remove this line. The same ocurrs in   getProperties() method (line 79)
https://github.com/payara/Payara/blob/19cec930b10c65ee02c7c2e000d17f5dec9bf5de/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java#L79
And also in getValue() method (line 91)
https://github.com/payara/Payara/blob/19cec930b10c65ee02c7c2e000d17f5dec9bf5de/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java#L91
Do I remove these lines too?

lprimak: Replace with `Objects.requireNotNul(name, "name should not be null)`
to do argument checking.

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/source/PasswordAliasConfigSource.java
AngelTG2: Just to make sure
Do you refer to
```
store.keys().forEachRemaining((key) -> propertyNames.add(key));
```
or 
```
Stream<String> stream = StreamSupport.stream(Spliterators.spliteratorUnknownSize(store.keys(),Spliterator.ORDERED), false);
stream.forEach(key -> propertyNames.add(key));
```

getproperties() method also use raw iterator. Do I change this too?

lprimak:
@AngelTG2 I figured it would be easier to change myself than to explain :) Let me know if you are ok with this.

lprimak:
jenkins test

lprimak (approved):

AngelTG2:
@lprimak. Perfect. I also think it will be faster that way. Thank you

PR #5213, opened 2021-04-23
FISH-1287 Admin console retrieves info about remote instances much faster

Pandrex247:
## Description

Port of PR @OndroMih made to Enterprise repo: https://github.com/payara/Payara-Enterprise/pull/354

Admin Console no longer hangs on every page if a remote instance responds slowly. Admin Console hangs for a much shorter time when displaying details about a remote instance. The `list-instances` command has shorter default timeout and therefore also doesn't hang for too long by default.

After the first commit, Admin Console doesn't ask for status of instances (which requires a REST call) when it only needs the name of the instance. This is also to display the instance names in the sidebar tree which is on every page. This was slowing down loading Admin Console pages significantly if remote instances were slow to respond.

After the second commit:

 - The list-instances command has shorter default timeout and it's configurable globally by a system property or env var. In case more time is needed, the timeout can be increased for the whole DAS.
 - This shorter timeout is also passed to the subsequent admin REST call so that both timeouts are in synch (and the REST call timeouts faster than usually).
 - The REST calls now respect a system property and env var also for connection timeout, not only for read timeout.
 - Debug info is only printed for running instances, this saves a useless REST call that always fails if an instance isn't running.

### New tests
No new tests added.

### Testing Performed
Manual testing, with breakpoints to make sure the behavior is expected. I simulated slow remote instances using a firewall rule to block and unblock communication between DAS and the remote instance. However, I also observed that the communication was very slow even with unblocked communication but it later improved so I had to simulate it using the firewall to finish debugging and testing.

### Test suites executed
No tests suites executed. The change is very local and doesn't impact functionality of Payara Server, only improves performance.

### Testing Environment

* Zulu 8.42.0.21-CA-linux64 (OpenJDK 1.8.0_232) to build
* AdoptOpenJDK 11.0.9.1 to run
* Ubuntu 20.04
* Maven 3.6.3

## Documentation

Nothing to document.

### DIFF
--- a/appserver/admingui/cluster/src/main/resources/standalone/standaloneInstanceGeneral.jsf
+++ b/appserver/admingui/cluster/src/main/resources/standalone/standaloneInstanceGeneral.jsf
@@ -37,7 +37,7 @@
     and therefore, elected the GPL Version 2 license, then the option applies
     only if the new code is made subject to such option by the copyright
     holder.
-    Portions Copyright [2018] [nPayara Foundation and/or its affiliates]
+    Portions Copyright [2018-2021] [Payara Foundation and/or its affiliates]
 -->
 
 <!initPage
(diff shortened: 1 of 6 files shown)


Pandrex247:
Jenkins test please

PR #5215, opened 2021-04-26
FISH-1323 Zulu JDK Update

dmatej:
## Description

See https://docs.azul.com/zulu/release-notes/april-2021.html

* Changes broke some tests after the JDK update, so before it would affect anyone else or Jenkins, I fixed tests already in master branch here.
* most important changes: the TLS 1.0 and TLS 1.1 are disabled by default, quite large set of ciphers is out too.

## Tests

PrintCertificateCommandTest
CertificateRealmITest (in #5019) also reproduced the issue.
Payara default keystores seems not affected.

## Note

We should discuss possible consequences before the upgrade. This is not just about docker images, users will upgrade their JDK distributions soon too.

### DIFF
--- a/appserver/extras/docker-images/pom.xml
+++ b/appserver/extras/docker-images/pom.xml
@@ -15,8 +15,8 @@
         <docker.noCache>true</docker.noCache>
 
         <docker.java.repository>azul/zulu-openjdk</docker.java.repository>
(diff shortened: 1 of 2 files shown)


lprimak (approved):

lprimak:
Jenkins test

lprimak:
jenkins test

PR #5217, opened 2021-04-26
[FISH-1328] Fixed Request Tracing NPE in Concurrent

lprimak:
Self explanatory

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2019] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.concurrent.runtime;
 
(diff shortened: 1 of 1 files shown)


lprimak:
jenkins test

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -239,12 +239,16 @@ private void startConcurrentContextSpan(ComponentInvocation invocation, Invocati
...
         if (handle.getSpanContextMap() != null) {
+            @SuppressWarnings("unchecked")
             SpanContext spanContext = tracer.extract(Format.Builtin.TEXT_MAP, new MapToTextMap(handle.getSpanContextMap()));
             builder.asChildOf(spanContext);
 
             // Check for the presence of a propagated parent operation name
             try {
-                String operationName = ((RequestTraceSpanContext) spanContext).getBaggageItems().get("operation.name");
+                String operationName = null;
+                if (spanContext != null) {
```
Pandrex247, line 249: Shouldn't this check be further up? Before `builder.asChildOf`? Otherwise you're registering this new Span as a child of `null`

In fact, this scenario seems odd to me.
All Spans should have a SpanContext and an operation name - if we're getting a null on either of them despite the fact that `handle.getSpanContextMap() != null` returned true it seems to me that there's a badly made Span somewhere, or that `extract` has broken

lprimak: There is definitely badly made span somewhere. However I am unable to reliably reproduce the source fault to be able to investigate. 
this fix is so I’m bad span doesn’t cause cascade failures.

lprimak: > Shouldn't this check be further up? Before builder.asChildOf? 

Not sure. Here the "old" code checks for `operation.name != null` and the new code
checks for `operation.name` that doesn't exist. Not much difference in functionality as far as I 
can see fo I think this fix is pragmatic and correct.

lprimak:
jenkins test

Pandrex247 (approved):

PR #5243, opened 2021-05-16
FISH-1375 Improve remote archive deployment

avpinchuk:
Improve remote archive deployment

## Description

***Bug fixes***
* `deploy-remote-archive` command: NPE when `additionalRepositories` option is missing for GAV archive
* `deploy-remote-archive` command: Maven Central is not defined as default repository when deploy GAV archive
* `deploy-remote-archive` command: Application name and context is not set from URI when RAR archive deploy from HTTP(S)
* `deploy-remote-archive` command: Add appropriate name to GAV archive

***Enhancements***
* `deploy-remote-archive` command: Add Basic Auth support
* `deploy-remote-archive` command: Make HTTP(S) scheme check case insensitive

***Improvements***
* Switch GAV converter to use URI instead of URL thus eliminate unnecessary conversion
* Check each additional repository only once
* Check Maven Central in last order
* `Payara Micro`: Print deployed applications summary in order as they deployed
* Refactor and eliminate code duplicates

### Testing Performed

Did some manual testing

### Testing Environment

Windows 10 Pro, Zulu Community 1.8.0_292, Maven 3.6.3

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -166,7 +166,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
     private boolean outputLauncher;
     private File copyDirectory;
     private Properties userSystemProperties;
-    private List<String> repositoryURIs;
+    private final List<String> repositoryURIs;
     private final short defaultHttpPort = 8080;
     private final short defaultHttpsPort = 8181;
     private final BootCommands preBootCommands;
@@ -212,7 +212,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
      * --help Shows this message and exits\n
      * @throws BootstrapException If there is a problem booting the server
      */
-    public static void main(String args[]) throws Exception {
+    public static void main(String[] args) throws Exception {
         create(args);
     }
 
@@ -716,9 +716,6 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
     public PayaraMicroImpl addRepoUrl(String... URLs) {
         //if (runtime != null) {
         checkNotRunning();
-        if (repositoryURIs == null) {
-            repositoryURIs = new LinkedList<>();
-        }
         repositoryURIs.addAll(Arrays.asList(URLs));
         return this;
     }
@@ -1111,6 +1108,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
 
     private PayaraMicroImpl() {
         // Initialise a random instance name
+        repositoryURIs = new LinkedList<>();
         preBootCommands = new BootCommands();
         postBootCommands = new BootCommands();
         postDeployCommands = new BootCommands();
@@ -1243,9 +1241,6 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
                     enableHealthCheck = Boolean.parseBoolean(value);
                     break;
                 case additionalrepository:
-                    if (repositoryURIs == null) {
-                        repositoryURIs = new LinkedList<>();
-                    }
                     repositoryURIs.add(value);
                     break;
                 case outputuberjar:
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -164,8 +164,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
     private boolean outputLauncher;
     private File copyDirectory;
     private Properties userSystemProperties;
-    private List<String> repositoryURLs;
-    private final String defaultMavenRepository = "https://repo.maven.apache.org/maven2/";
+    private List<String> repositoryURIs;
     private final short defaultHttpPort = 8080;
     private final short defaultHttpsPort = 8181;
     private final BootCommands preBootCommands;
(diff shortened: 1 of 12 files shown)


nucleus/deployment/common/src/main/java/fish/payara/deployment/util/URIUtils.java
```diff
@@ -80,4 +105,46 @@ public static HttpURLConnection openHttpConnection(URI uri) throws IOException {
...
+     * @param uri the URI to test.
+     * @return {@code true} if a resource exists; {@code false} otherwise.
+     * @throws IOException if an I/O error occurs.
+     * @throws ClassCastException if the {@code uri} is not a file and does not represent HTTP(S) resource.
+     */
+    public static boolean exists(URI uri) throws IOException {
+        if (hasFileScheme(uri)) {
+            return Files.exists(Paths.get(uri));
+        }
+        HttpURLConnection connection = openHttpConnection(uri);
```
dmatej, line 136: All connections and resources should be properly closed. See javadoc.

> Calling the close() methods on the InputStream or OutputStream of an HttpURLConnection after a request may free network resources associated with this instance but has no effect on any shared persistent connection. Calling the disconnect() method may close the underlying socket if a persistent connection is otherwise idle at that time.

lprimak: using try-with-resources

lprimak:
jenkins test

nucleus/deployment/common/src/main/java/fish/payara/deployment/util/URIUtils.java
avpinchuk: I slightly misunderstand. Should I call `getInputStream()` on `HttpURLConnection` before call `getResponseCode()`?

dmatej: I was thinking rather about the HttpURLConnection and when it is disconnected. It is not Closeable. And yes, all streams should be closed when you don't need them, which seems alright here.

I see even original code wasn't disconnecting, interesting ... i wonder if it had some consequences ... 
Perhaps this is useful explanation: https://stackoverflow.com/a/10117254/375449
Also ...
https://stackoverflow.com/a/40123634/375449
https://docs.oracle.com/javase/tutorial/networking/urls/connecting.html (btw I dislike examples which don't disconnect/close in finally).

I tried to find some usages of the disconnect method ...
image

Generally it seems that the behavior depends on concrete implementation and JVM options (see also KeepAliveCache).

avpinchuk: A you suggesting to disable persistent connections?

avpinchuk: For Open JDK and clones, calling `disconnect()` does not close underlying shared socket for HEAD request, with or without closing inputstream. As I suggesting, socket returns to connection pool for further reuse. 
For GET request, `disconnect()` close underlying socket if and only if `close()` on `InputStream` **was not** called.
Should we close shared socket after getting *each* remote archive?

avpinchuk: I propose to replace return in this method with 
```java
if (connection.getResponseCode() == HttpURLConnection.HTTP_OK) {
    return true;
}
connection.disconnect();
return false;
```
This should be enough because `getInputStream()` method returns an empty no-op stream in both OpenJDK and OpenJ9

AlanRoth:
jenkins test please

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -164,8 +166,7 @@
     private boolean outputLauncher;
     private File copyDirectory;
     private Properties userSystemProperties;
-    private List<String> repositoryURLs;
-    private final String defaultMavenRepository = "https://repo.maven.apache.org/maven2/";
+    private List<String> repositoryURIs;
```
lprimak, line 169: Recipe for an NPE. Should be initialized right here and made final

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -715,7 +716,10 @@ private void validateRuntimeOption(RUNTIME_OPTION option, String optionValue) {
     public PayaraMicroImpl addRepoUrl(String... URLs) {
         //if (runtime != null) {
         checkNotRunning();
-        repositoryURLs.addAll(Arrays.asList(URLs));
+        if (repositoryURIs == null) {
```
lprimak, line 719: remove this if block. Should be initialized inline at the top

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
```diff
@@ -1241,7 +1243,10 @@ private void scanArgs(String[] args) {
                     enableHealthCheck = Boolean.parseBoolean(value);
                     break;
                 case additionalrepository:
-                    repositoryURLs.add(value);
+                    if (repositoryURIs == null) {
```
lprimak, line 1246: Same here. Recipe for NPE. Remove the if block

lprimak (changes requested):
Overall looks good. Just some minor changes

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
avpinchuk: Maybe better in constructor?

lprimak: I wouldn't create a constructor just for this, no :)

avpinchuk: Private constructor already found

lprimak: oh yeah all the way down there... never mind :)

lprimak (approved):

lprimak:
jenkins test

PR #5244, opened 2021-05-17
FISH-1327 Fix unable to disable Wrapping of JDBC Objects

JamesHillyard:
## Description
Fixed Wrapping JDBC Objects unable to be disabled if a SQL Trace Listener was present but it was impossible to have no listener. The validation preventing wrapping being disabled while a SQL Trace Listener was present has been removed along with the assertion testing this.

### Testing Performed
Manually testing the expected behaviour on the ticket with the updated changes. Confirmed the Unit Tests in the module still all pass. Ran the EE7 Samples and confirmed all tests still pass with wrapping on and off.

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 20.04.2 with Maven 3.6.3

### DIFF
--- a/appserver/jdbc/admin/src/main/java/org/glassfish/jdbc/admin/cli/JDBCConnectionPoolManager.java
+++ b/appserver/jdbc/admin/src/main/java/org/glassfish/jdbc/admin/cli/JDBCConnectionPoolManager.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2020] Payara Foundation and/or affiliates
+// Portions Copyright [2020-2021] Payara Foundation and/or affiliates
 
 package org.glassfish.jdbc.admin.cli;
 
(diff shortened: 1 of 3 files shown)


JamesHillyard:
Jenkins test please

Pandrex247 (changes requested):
The [JDBCConnectionPoolManager](https://github.com/payara/Payara/blob/master/appserver/jdbc/admin/src/main/java/org/glassfish/jdbc/admin/cli/JDBCConnectionPoolManager.java) class probably also needs amending.

I'd revert the changes brought in there when the default was changed to the _SilentSqlTraceListener_ class:
* `private String sqltracelisteners = "fish.payara.jdbc.SilentSqlTraceListener";` should become `private String sqltracelisteners = null;`
* `
        // Can't be set to null as the default value is now the SilentSqlTraceListener class, which requires statement 
        // wrapping to be enabled
        sqltracelisteners = (String) attrList.get(SQL_TRACE_LISTENERS);
        if (sqltracelisteners == null) {
            sqltracelisteners = "";
        }
        `
 should be changed back to just
        ```
        sqltracelisteners = (String) attrList.get(SQL_TRACE_LISTENERS);
        ```

Pandrex247:
An alternative fix would be to track down the validation that mandates that you can't disable wrapping while an SQL Trace Listener is configured and remove or loosen it.

JamesHillyard:
Jenkins test please

Pandrex247 (approved):

Pandrex247 (changes requested):
Copyright needs updating on each file you've touched for 2021

JamesHillyard:
Jenkins test please

Pandrex247 (approved):

Issue #5024, opened 2020-12-02
Make MPCONFIG usable in *all* DataSourceDefinition properties / FISH-862


poikilotherm:
Using a data source definition in `web.xml` or `@DataSourceDefinition` does not allow to use `${MPCONFIG=...}` (or `${ENV}`) for Payara specific options ("fish.payara.xxx") or integer values used i. e. for `maxIdleTime`.

This should be changed to allow for more flexible configuration and fine-tuning in deployments.

## Expected Outcome

Successfull deployment with JDBC pools configured via MicroProfile Config API for any kind of setting. Usually non-portable, but most options aren't portable anyway.

Provide an option to set the integer based pool settings (`minPoolSize, maxPoolSize`, ...) from a Payara specific properties setting using a String and thus enable variable substitution. (e. g. `fish.payara.max-pool-size`, ...)

## Current Outcome

When using a data source like 
```
@DataSourceDefinition(
    name = "java:app/jdbc/dataverse",
    className = "org.postgresql.ds.PGConnectionPoolDataSource",
    user = "${MPCONFIG=dataverse.db.user}",
    password = "${MPCONFIG=dataverse.db.password}",
    url = "jdbc:postgresql://${MPCONFIG=dataverse.db.host}:${MPCONFIG=dataverse.db.port}/${MPCONFIG=dataverse.db.name}",
    minPoolSize = 10,
    maxPoolSize = 100,
    maxIdleTime = 300,
    properties = {
        "fish.payara.pool-resize-quantity=${MPCONFIG=dataverse.db.pool-resize-quantity}",
        "fish.payara.is-connection-validation-required=${MPCONFIG=dataverse.db.is-connection-validation-required}",
        "fish.payara.connection-validation-method=${MPCONFIG=dataverse.db.connection-validation-method}",
        "fish.payara.validation-table-name=${MPCONFIG=dataverse.db.validation-table-name}"
    }
)
```
this will fail to deploy with quite generic errors:
```
[#|2020-12-02T12:43:58.635+0000|WARNING|Payara 5.2020.6|javax.enterprise.resource.resourceadapter.org.glassfish.jdbc.deployer|_ThreadID=1;_ThreadName=main;_TimeMillis=1606913038635;_LevelValue=900;|
  error.creating.jdbc.pool|#]
```

## Context (Optional)

The idea is to enable Dataverse admins to configure certain things like monitoring for slow queries etc by configuring simple settings via MicroProfile Config API.

Digging through

[...]

PR #5272, opened 2021-05-26
FISH-862 : Make MPCONFIG usable in all DataSourceDefinition properties

kalinchan:
## Description

Community Enhancement Request. With this PR you should be able to use ${MPCONFIG =...} for Payara specific options ("fish.payara.xxx") inside a Data Source Definition (both web.xml and @DataSourceDefinition) 
resolves https://github.com/payara/Payara/issues/5024
### Testing Performed
Manual Testing. Created a reproducer for the issue using ${MPCONFIG=...} for Payara specific options.
Checked if invalid values would raise errors and valid values would allow the application to be deployed successfully 

### Testing Environment

OpenJDK 1.8.0_292 on Ubuntu 20.04.2 LTS with Maven 3.6.3


lprimak (approved):

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbc/deployer/DataSourceDefinitionDeployer.java
```diff
@@ -1060,7 +1061,7 @@ public String getPropertyValue(String name, String defaultValue) {
             String value = null;
             value = (String) desc.getProperties().get(name);
             if (value != null) {
-                return value;
+                return TranslatedConfigView.expandValue(value);
```
Pandrex247, line 1064: Shouldn't this also be applied to the `getProperty(String name)` and `getPropertyValue(String name)` methods?

Pandrex247 (commented):
Query - see comment.
You've only applied expansion to one of three very similar methods.

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbc/deployer/DataSourceDefinitionDeployer.java
kalinchan: The handling of getting the DataSourceDefinition properties are handled here https://github.com/payara/Payara/blob/745aa8290080e627015011070cb98a16a94414e8/appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbc/deployer/DataSourceDefinitionDeployer.java#L575. If you look at all the payara related properties, they all use the getPropertyValue(String name, String defaultValue) method

PR #5284, opened 2021-06-03
FISH-1167 - Fix Inconsistent Synchronization in Password Alias Alterations for Remote Instances

JamesHillyard:
## Description
Reverted the changed made in ZDEP-1987, fixing the inconsistent behaviour described in FISH-1167 as the alias is now made everywhere. The bug which ZDEP-1987 fixed has been replaced by changing the aliasNameNew text property from the Admin Console UI to 'id'. This means the aliasname isn't lost later on due to it having a variable name that wasn't expected as the REST interface expects 'id'.

### Testing Performed
All steps to reproduce in both FISH-1167 and ZDEP-1987 have been tested and work correctly. All unit tests within the modules passed.

### Testing Environment
Zulu JDK 1.8_212 on Ubuntu 20.04.2 with Maven 3.6.3

### DIFF
--- a/appserver/admingui/common/src/main/resources/appServer/pswdAliasAttr.inc
+++ b/appserver/admingui/common/src/main/resources/appServer/pswdAliasAttr.inc
@@ -38,6 +38,7 @@
     only if the new code is made subject to such option by the copyright
     holder.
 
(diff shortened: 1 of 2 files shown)


JamesHillyard:
jenkins test please

lprimak (approved):

PR #5315, opened 2021-06-22
FISH-758 Fault Tolerance Annotations Not Applied to Rest Client Interfaces

Pandrex247:
## Description
Bug fix.

The MicroProfile Rest Client spec has an untested requirement that FT annotations get applied to Rest Client interfaces.
https://github.com/eclipse/microprofile-rest-client/blob/master/spec/src/main/asciidoc/integration.asciidoc#microprofile-fault-tolerance

Weld wasn't picking up our synthetic "not quite a stereotype" Fault Tolerance interceptor as an interceptor for the individual FT annotations, as it attempts to match annotations to interceptors via this [method](https://github.com/weld/core/blob/3.1.4.Final/impl/src/main/java/org/jboss/weld/resolution/TypeSafeInterceptorResolver.java#L44) - `FaultTolerance` only matches `FaultTolerance`, not `Retry` and only `Retry`.

The way around this was to go back to the old style of having an interceptor for each individual FT annotation, which becomes a problem when it comes to attempting to resolve `Fallback` as you can't write an interceptor for this in the usual way (marking a class with `Fallback` & `Interceptor` and adding an `AroundInvoke` method). 
Alternatively, we could do as we've done here and register the interceptor programmatically as an interceptor for the FT annotations.

### Blockers
None.

### New tests
IT added to Payara Samples.

### Testing Performed
Ran the FT TCK.
Ran [Rudy's example test](https://payara.atlassian.net/browse/FISH-758?focusedCommentId=51437) which demonstrates the `Retry` annotation working on a Rest Client interface.

### Testing Environment
Windows 10, JDK 8
WSL OpenSUSE Leap 15.2, JDK 8

## Documentation
N/A

## Notes for Reviewers
I feel like this block is both over and under-engineered and I'm missing a much simpler way of determining the priority of each interceptor - this solution seems fragile.
https://github.com/payara/Payara/compare/master...Pandrex247:FISH-758?expand=1#diff-d1250f7d0ca1e02fa4c81c96aa7ef9d2c7045b6e4f7bdab566ea282ad3c60940R149-R248


pdudits (changes requested):
Whether you keep the binary search or go for simpler approach, I think that ordering part still needs a unit test.

Pandrex247:
Jenkins test please

pdudits (approved):

Issue #4825, opened 2020-08-08
Applications in DEPLOY_DIR aren't deployed if a custom read-only POSTBOOT_COMMAND file is provided // FISH-371


viktorcitaku:
# Description #
----------

Based on the Docker Hub documentation of the [Payara Server](https://hub.docker.com/r/payara/server-full), there is an environment variable defined named DEPLOY_DIR which says: `The directory where applications are searched for in generate_deploy_commands.sh script.`. However, it doesn't work as it's expected.

The example that I am running is located [here](https://github.com/viktorcitaku/modular-jee).

## Expected Outcome

I expect my WAR application to be deployed.

## Current Outcome

The application is not deployed.

## Steps to reproduce (Only for bug reports) 

1. `git clone git@github.com:viktorcitaku/modular-jee.git`
2. `mvn clean install`
3. `docker-compose up --build`

## Context (Optional)

I did a workaround where I deploy my WAR via the `post-boot-commands.asadmin` script.

The potential issue might be related to the following log entries, which appear at the beginning when the Payara Server container starts.
```
[Entrypoint] running /opt/payara/scripts/init_1_generate_deploy_commands.sh
touch: cannot touch '/opt/payara/config/post-boot-commands.asadmin': Permission denied
post boot commands already deploys /opt/payara/deployments/web-module-1.0.0-SNAPSHOT.war

[Entrypoint] ignoring /opt/payara/scripts/init.d/*

touch: cannot touch '/opt/payara/config/post-boot-commands.asadmin': Permission denied
```

## Environment ##

- **Payara Version**: 5.2020.3
- **Edition**: Full
- **JDK Version**: zulu-8-amd64 (from official Payara Docker image (latest))
- **Operating System**: from official Payara Docker image (latest)
- **Database**: Not relevant

OndroMih:
Hi @viktorcitaku,

I've tested your reproducer and I see what you see. However, the Docker container doesn't ignore the `$DEPLOY_DIR` variable. The problem is that you overwrite the `$POSTBOOT_COMMANDS` file and this file is created under the `root` user. The default `payara` user can't modify it in order to add deploy commands for applications in the `$DEPLOY_DIR` directory.

To fix this, you only need modify the `COPY` command to create the file under the `payara` user. This works in Docker 17.09+ :

```
COPY --chown=payara ./scripts/post-boot-commands.asadmin ${POSTBOOT_COMMANDS}
```

If you use an older Docker version, you need to run the following:

```
[...]

PR #5328, opened 2021-07-02
FISH-371 Improve handling of custom POSTBOOT and PREBOOT file in Docker container

kalinchan:
## Description

resolves https://github.com/payara/Payara/issues/4825

Added new env variable POSTBOOT_COMMANDS_FINAL and PREBOOT_COMMANDS_FINAL as suggested

Changed keyserver as previous one wasn't working 

### Testing Performed

Manual testing
### Testing Environment

openjdk version "1.8.0_292"
ubunutu 20.04
maven 3.6.3

### DIFF
--- a/appserver/extras/docker-images/basic/pom.xml
+++ b/appserver/extras/docker-images/basic/pom.xml
@@ -13,6 +13,6 @@
     <properties>
         <docker.noCache>false</docker.noCache>
         <docker.payara.repository>payara/basic</docker.payara.repository>
-        <docker.keyserver.url>hkp://ipv4.pool.sks-keyservers.net</docker.keyserver.url>
+        <docker.keyserver.url>keyserver.ubuntu.com</docker.keyserver.url>
     </properties>
 </project>
--- a/appserver/extras/docker-images/server-full/src/main/docker/bin/startInForeground.sh
+++ b/appserver/extras/docker-images/server-full/src/main/docker/bin/startInForeground.sh
@@ -16,6 +16,7 @@
 #   - $ADMIN_USER - the username to use for the asadmin utility.
 #   - $PASSWORD_FILE - the password file to use for the asadmin utility.
 #   - $PREBOOT_COMMANDS - the pre boot command file.
+#   - $PREBOOT_COMMANDS_FINAL - copy of the pre boot command file.
 #   - $POSTBOOT_COMMANDS - the post boot command file.
 #   - $POSTBOOT_COMMANDS_FINAL - copy of the post boot command file.
 #   - $DOMAIN_NAME - the name of the domain to start.
@@ -30,6 +31,7 @@
 if [ -z $ADMIN_USER ]; then echo "Variable ADMIN_USER is not set."; exit 1; fi
 if [ -z $PASSWORD_FILE ]; then echo "Variable PASSWORD_FILE is not set."; exit 1; fi
 if [ -z $PREBOOT_COMMANDS ]; then echo "Variable PREBOOT_COMMANDS is not set."; exit 1; fi
+if [ -z $PREBOOT_COMMANDS_FINAL ]; then echo "Variable PREBOOT_COMMANDS_FINAL is not set."; exit 1; fi
 if [ -z $POSTBOOT_COMMANDS ]; then echo "Variable POSTBOOT_COMMANDS is not set."; exit 1; fi
 if [ -z $POSTBOOT_COMMANDS_FINAL ]; then echo "Variable POSTBOOT_COMMANDS_FINAL is not set."; exit 1; fi
 if [ -z $DOMAIN_NAME ]; then echo "Variable DOMAIN_NAME is not set."; exit 1; fi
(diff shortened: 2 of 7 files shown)


kalinchan:
jenkins test

appserver/extras/docker-images/basic/pom.xml
```diff
@@ -13,6 +13,6 @@
     <properties>
         <docker.noCache>false</docker.noCache>
         <docker.payara.repository>payara/basic</docker.payara.repository>
-        <docker.keyserver.url>hkp://ipv4.pool.sks-keyservers.net</docker.keyserver.url>
+        <docker.keyserver.url>keyserver.ubuntu.com</docker.keyserver.url>
```
lprimak, line 16: Isn't this a bad thing? Looks like we are dependent on one server instead of a pool of servers.
I would keep using the pool instead of a single server

kalinchan: It wouldn't build with that keyserver, so I changed it. Unless you have a know of a different pool of servers?

lprimak: How come it built before? There is something missing in your explanation :)

kalinchan: I'm not sure, but if you go to sks-keyservers.net it says:
 This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.

Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

It's probably to do with the DNS records

lprimak: My internet search revealed this:

hkp://pgp.mit.edu:11371

https://www.reddit.com/r/archlinux/comments/o5rcs6/psa_you_need_to_update_your_keyserver/

Pandrex247: Is that any different to just using the ubuntu keyserver? That doesn't look to be a pool either

dmatej: FYI: https://gist.github.com/rjhansen/67ab921ffb4084c865b3618d6955275f
And possible solution: https://github.com/tianon/gosu/issues/39#issuecomment-376028649

lprimak: @Pandrex247 yea you are right. I thought for some reason hkp was a pooled protocol.

lprimak (approved):

appserver/extras/docker-images/server-full/src/main/docker/bin/startInForeground.sh
```diff
@@ -41,8 +43,9 @@ if [ -z $DOMAIN_NAME ]; then echo "Variable DOMAIN_NAME is not set."; exit 1; fi
 # Create pre and post boot command files if they don't exist
 touch $POSTBOOT_COMMANDS
```
OndroMih, line 44: This isn't necessary since POSTOOT_COMMANDS file is now optional and isn't used to start the server.
```suggestion
```

appserver/extras/docker-images/server-web/src/main/docker/bin/init_1_generate_deploy_commands.sh
```diff
@@ -28,10 +29,16 @@
 if [ -z $DEPLOY_DIR ]; then echo "Variable DEPLOY_DIR is not set."; exit 1; fi
 if [ -z $PREBOOT_COMMANDS ]; then echo "Variable PREBOOT_COMMANDS is not set."; exit 1; fi
 if [ -z $POSTBOOT_COMMANDS ]; then echo "Variable POSTBOOT_COMMANDS is not set."; exit 1; fi
+if [ -z $POSTBOOT_COMMANDS_FINAL ]; then echo "Variable POSTBOOT_COMMANDS_FINAL is not set."; exit 1; fi
 
 # Create pre and post boot command files if they don't exist
 touch $POSTBOOT_COMMANDS
```
OndroMih, line 35: As stated above, this isn't necessary now and can be removed.
```suggestion
```

appserver/extras/docker-images/server-web/src/main/docker/bin/startInForeground.sh
```diff
@@ -41,8 +43,9 @@ if [ -z $DOMAIN_NAME ]; then echo "Variable DOMAIN_NAME is not set."; exit 1; fi
 # Create pre and post boot command files if they don't exist
 touch $POSTBOOT_COMMANDS
```
OndroMih, line 44: Again, not necessary and can be removed
```suggestion
```

OndroMih (changes requested):
I know that FISH-371 was only about POSTBOOT command file but can you also do the same for PREBOOT command file so that both behave in the same way?

kalinchan:
jenkins test

rdebusscher:
@OndroMih The Docker scripts doesn't touches the preboot script of the user, so it does not make much sense to copy it.

But changes are already made so probably not worth reverting it.

kalinchan:
jenkins test

lprimak (approved):

OndroMih:
Hi @rdebusscher , there's one case I hit very often, when the Docker container changed user's postboot command file. If the postboot command variable was changed to point to a file in a volume. Then it pointed to a file that exists on the host filesystem. When the container is started 2 and more times, each time it adds the generate commands to user's custom postboot file. That's all possible and supported because POSTBOOT_COMMANDS variable is officially documented and can be changed to point to a different location.

Here's an example how the container can be started to do it:

```
docker run -p 8080:8080 -e POSTBOOT_COMMANDS=/config/postboot -v /my/config:/config payara/server-full
```

and with `/my/config/postboot` file in the host system.

It's always safer to copy the file in case it points to a location outside of the docker virtual filesystem.

OndroMih (approved):

PR #5334, opened 2021-07-08
FISH-1514 Glassfish deployment descriptor should have higher precedence by default as the WebLogic descriptor

jGauravGupta:
## Description
This is a improvement PR to re-prioritize the deployment descriptor with the optional system properties to customize configuration precedence of DD.

#### Removed property
 `gfdd.over.wlsdd`  -  When this property is set to true, GF DD will have higher precedence over WLS DD.
#### New property
`wlsdd.precedence`  -    When this property is set to true, WLS DD will have higher precedence over other DD.
`gfdd.precedence`  -     When this property is set to true, GF DD will have higher precedence over other DD. 

### Testing Performed
Manually deploying 'svc_scripting.ear'

## Documentation
TODO

### DIFF
--- a/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/DOLUtils.java
+++ b/appserver/deployment/dol/src/main/java/com/sun/enterprise/deployment/util/DOLUtils.java
@@ -150,10 +150,15 @@ public class DOLUtils {
     @LogMessageInfo(message = "DOLUtils: converting EJB to web bundle id {0}.", level="FINER")
     private static final String CONVERT_EJB_TO_WEB_ID = "AS-DEPLOYMENT-00017";
 
-    /** The system property to control the precedence between GF DD
-    // and WLS DD when they are both present. When this property is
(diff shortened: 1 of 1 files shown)


OndroMih:
Since the order of precedence can already be modified with `gfdd.over.wlsdd` and `ignore.wlsdd` properties, I think we don't need this PR in Payara 5. Maybe it's better to delay it until Payara 6.

OndroMih (changes requested):
I would wait with merging this PR until we have an agreement whether we want to add it to Payara 5.

AlanRoth:
Hi @OndroMih and @jGauravGupta, I spoke with @eliotmartin and we have agreed since a property already exists there is no point in adding the new properties. The only task to do is to document `gfdd.over.wlsdd`. 

Apologies about the confusion @jGauravGupta, but I think `gfdd.over.wlsdd` already does what the customer wants, and I will close this PR.

Cheers,
Alan

Issue #5111, opened 2021-02-02
Parameter Serialization error in open api document for query parameter - FISH-1068


LarsBrenker:
# Description #
----------

I have a GET resource that contains a query parameter. 

```java
public Response getValue(@Parameter(description = "the value", required = true, example = "8967598475", schema = @Schema(type = SchemaType.STRING)) @QueryParam("value") String value) {
...
}
```

The generated open api document contains the style: simple in the description of the path in the parameters section.

```yaml
/v1/getValue:
    get:
      summary: Returns a test value
      description: Returns a test value
      parameters:
      - name: value
        in: query
        description: the value
        required: true
        style: simple
        schema:
          type: string
        example: "8967598475"

```

This leads to validation errors.

## Expected Outcome

According to the Swagger specification (https://swagger.io/docs/specification/serialization/), simple is not allowed for query parameters.

And org.eclipse.microprofile.openapi.annotations.parameters.Parameter describes for style attribute:
_Describes how the parameter value will be serialized depending on the type of the parameter value.
Default values (based on value of in): for query - form; for path - simple; for header - simple; for cookie - form._

So I would have expected value "form" for the style parameter value because i use QueryParam Annotation

## Current Outcome

If you specify the style parameter in @Parameter , it will be set correctly. 
```java
@Parameter(description = "the value", style = ParameterStyle.FORM ..)
```

If you do not specify it, you will always set "simple" as a value. No matter if query or path parameter

### Samples

if you need a reproducer, let me know.

## Environment ##

- **Payara Version**: 5.2020.6
- **Edition**: 
- **JDK Version**:

MeroRai:
Hi @LarsBrenker,

I was able to reproduce the issue in our current release of Payara community release, which is 5.2020.7. I have now escalated it to our engineering team and created an internal issue FISH-1068 to track it. Thank you for your assist.

aubi:
Hello Lars,
I reviewed your idea and I agree. Implemented, currently under review.
Thank you and I'm looking forward to your next issues! :-)
Petr

aubi:
Merged, will be part of 5.2021.5.

LarsBrenker:
Error has been fixed. Can be closed

PR #5335, opened 2021-07-08
FISH-1068 in generated openapi omit default style for parameter

aubi:
## Description
Fix for https://github.com/payara/Payara/issues/5111
The solution is simple -- don't fill style parameter as the value "simple" is incorrect for some usages (e.g. query of header parameters). The standard has meaningful defaults, so it is not necessary to calculate the value based on usage.

### Testing Performed
manually - in the generated openapi is missing "style: simple" in the parameter

I also tried openapi-generator-cli and there is no problem without style.


aubi:
Jenkins test

aubi:
Jenkins test

kalinchan (approved):

MarkWareham (approved):
LGTM

PR #5336, opened 2021-07-09
FISH-786 FISH-1521 Upgrade security connectors to 2.1.0

OndroMih:
### New tests
There are new tests in Payara Enterprise repository but I didn't have time to copy them here. If I have time today, I'll add them but don't expect it.

## Documentation

To be done


OndroMih:
Jenkins test

eliotmartin (approved):

PR #5345, opened 2021-07-14
FISH-1518 Add Exception Name to FaultTolerance Message.

Pandrex247:
## Description
Workaround fix for the TCK, which currently mandates and checks that specifically this error gets thrown (and not any other kind of Deployment/Definition exception). The "proper" fix for this is liable to be a pain and have ramifications since it entails changing the way the deploy command works and checking for serialisation issues when using the DeploymentException class (as well as then reflecting these changes in Arquillian). 

This workaround fix simply adds the exception name to the message so that it can be read and rethrown as it later in Arquillian. This is done because Weld wraps the FaultToleranceException in a generic CDI definition exception, which is then rethrown as a DeploymentException which strips away all throwables/exceptions to avoid serialisation issues (so the identifying detail of it being a FT exception gets lost).

### Blockers
None.

### New tests
None here, but Fault Tolerance TCK tests currently covered as unit tests are being re-enabled here: https://github.com/payara/MicroProfile-TCK-Runners/pull/151

### Testing Performed
Ran the Fault Tolerance TCK with all tests against remote profile.

### Testing Environment
Windows, JDK 8.

## Documentation
N/A

## Notes for Reviewers
If someone thinks the "proper" fix is less work than I'm envisioning we can throw this away.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/policy/Policy.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/policy/Policy.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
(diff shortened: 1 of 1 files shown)


Pandrex247:
Jenkins test please

aubi (approved):

PR #5338, opened 2021-07-09
FISH-1521 Tests for fix of mapping of groups from an OIDC provider

OndroMih:
## Description

Tests to cover FISH-1521, which was fixed in the connectors repository.

Builds on top of https://github.com/payara/Payara/pull/5337.

PR #5350, opened 2021-07-15
FISH-786 FISH-1521 FISh-385 Revert of external security connectors integration

OndroMih:
After replacing the security connectors implementation with integration of external security connectors, there are failures in Jakarta EE TCK.

Reverting the following PRs related or dependent on the integration of external security connectors:

* https://github.com/payara/Payara/pull/5265
* https://github.com/payara/Payara/pull/5338

This also reverts a few other PRs that only increased the version of the connectors dependency as the new versions of connectors were released.

Leaving the update of the MicroProfile dependency from version 4.0 to 4.0.1, because that doesn't break anything and fixes a potential issue with dependencies.


OndroMih:
This should be created against the 5.2021.5 branch.

PR #5362, opened 2021-07-29
FISH-5654 - Fixed JSON Logging not working with JDK11

JamesHillyard:
## Description
This is a bug fix that causes the intended behaviour on JDK11 when setting the server log properties to JSON which also fixes the failing tests.

### Testing Performed
Ensured SortedLoggingPropertiesTest and JSONLogFormatIT now pass on JDK11. Manually tested once JSON formatting is selected the logs can be seen from the admin console and the logging.properties file is in the order set out in FISH-1509.

### Testing Environment
Windows 10 Pro, Maven 3.6.3. Tested with both JDK8 and Zulu JDK11.

## Documentation
No documentation required.


JamesHillyard:
jenkins test please

Pandrex247 (approved):

PR #5369, opened 2021-08-04
FISH-81 Fix Incorrect error "No valid EE environment for injection" for CDI reported for events

jGauravGupta:
## Description
This is a bug fix that causes the incorrect error message in the log on firing the async CDI events.

### Testing Performed
Manually

## Documentation
Not required

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExecutorServicesImpl.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExecutorServicesImpl.java
@@ -72,14 +72,17 @@ public class ExecutorServicesImpl extends AbstractExecutorServices implements Ex
     private final ContextualTimerExecutor timerExecutor;
     private final ManagedExecutorService executor;
     private final org.glassfish.concurrent.config.ManagedExecutorService config;
-    
+
+    private static final String DEFAULT_MANAGED_SCHEDULED_EXECUTOR_SERVICE = "java:comp/DefaultManagedScheduledExecutorService";
+    private static final String DEFAULT_MANAGED_EXECUTOR_SERVICE = "java:comp/DefaultManagedExecutorService";
+    private static final String DEFAULT_MANAGED_EXECUTOR_SERVICE_PHYS = "concurrent/__defaultManagedExecutorService";
+
     public ExecutorServicesImpl() throws NamingException {
         InitialContext ctx = new InitialContext();
-        executor = (ManagedExecutorService) ctx.lookup("java:comp/DefaultManagedExecutorService");
+        executor = (ManagedExecutorService) ctx.lookup(DEFAULT_MANAGED_EXECUTOR_SERVICE);
         taskExecutor = new ContextualTaskExecutor();
         timerExecutor = new ContextualTimerExecutor();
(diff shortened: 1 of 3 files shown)


appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExecutorServicesImpl.java
```diff
@@ -65,12 +70,16 @@
...
     
-    public ExecutorServicesImpl(PayaraExecutorService service) {
-        executor = service;
+    public ExecutorServicesImpl() throws NamingException {
+        InitialContext ctx = new InitialContext();
+        executor = (ManagedExecutorService) ctx.lookup("java:comp/DefaultManagedExecutorService");
         taskExecutor = new ContextualTaskExecutor();
         timerExecutor = new ContextualTimerExecutor();
+        this.config = Globals.getDefaultHabitat()
+                .getService(org.glassfish.concurrent.config.ManagedExecutorService.class);
```
pdudits, line 82: You may have multiple services created explicitly, so this might have different behaviour when domain has some executor services created under explicit names.

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExecutorServicesImpl.java
```diff
@@ -89,7 +98,7 @@ public void cleanup() {
 
     @Override
     protected int getThreadPoolSize() {
-        return executor.getExecutorThreadPoolSize();
+        return Integer.parseInt(this.config.getMaximumPoolSize());
```
pdudits, line 101: Javadoc says it is allowed to return `-1` for don't know / unlimited (which is the default anyway).

I'm not sure what this actually optimizes (didn't check Weld), but I feel it is the safer form, and we may skip accessing domain config.

pdudits (approved):

Pandrex247:
Jenkins test please

appserver/web/weld-integration/src/main/java/org/glassfish/weld/services/ExecutorServicesImpl.java
jGauravGupta: As per validation constraints, its min value is `0` and the default value is `Integer.MAX_VALUE`
https://github.com/payara/Payara/blob/master/appserver/concurrent/concurrent-connector/src/main/java/org/glassfish/concurrent/config/ManagedExecutorService.java#L86

Thread pool size is used by Weld to create min tasks in the task factory :
https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/executor/IterativeWorkerTaskFactory.java#L48

pdudits: So it is only used to set bounds lower than available processors, which is irrelevant in default cause. 

I'm worried how does this behave when one creates an additional non-default managed executor. Which one is picked by the config lookup (or would it fail)?

Pandrex247:
Missing this release

pdudits (approved):

jGauravGupta:
jenkins test please

Issue #5363, opened 2021-07-30
Bug Report: payara-micro-5.2021.5 fails to deploy WAR with "ServletConfig has not been initialized" Exception/FISH-5660


AndrewG10i:
# Description #
----------

The same app can be deployed fine on payara-micro-5.2021.4 but fails to get deployed on payara-micro-5.2021.5 with the following exception:
```
[2021-07-30T10:58:46.945+0800] [] [←[1;91mSEVERE←[0m] [] [←[1;94mfish.payara.micro.PayaraMicro←[0m] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1627613926945] [levelValue: 1000] [[

java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at fish.payara.micro.boot.loader.MainMethodRunner.run(MainMethodRunner.java:50)
        at fish.payara.micro.boot.loader.Launcher.launch(Launcher.java:114)
        at fish.payara.micro.boot.loader.Launcher.launch(Launcher.java:73)
        at fish.payara.micro.boot.PayaraMicroLauncher.create(PayaraMicroLauncher.java:88)
        at fish.payara.micro.boot.PayaraMicroLauncher.main(PayaraMicroLauncher.java:72)
        at fish.payara.micro.PayaraMicro.main(PayaraMicro.java:456)
Caused by: fish.payara.micro.BootstrapException: ServletConfig has not been initialized
        at fish.payara.micro.impl.PayaraMicroImpl.bootStrap(PayaraMicroImpl.java:1070)
        at fish.payara.micro.impl.PayaraMicroImpl.create(PayaraMicroImpl.java:230)
        at fish.payara.micro.impl.PayaraMicroImpl.main(PayaraMicroImpl.java:217)
        ... 10 more
Caused by: java.lang.IllegalStateException: ServletConfig has not been initialized
        at javax.servlet.GenericServlet.getServletContext(GenericServlet.java:159)
        at org.glassfish.jersey.servlet.ServletContainer.getServletContext(ServletContainer.java:457)
[...]

PR #5390, opened 2021-08-20
FISH-5660 Micro fails to deploy WAR with "ServletConfig has not been initialized"

jGauravGupta:
## Description
This PR fixes the dynamic registration of the Open API endpoint via prioritizing initialization of OpenApiApplication's `org.glassfish.jersey.servlet.ServletContainer` on application deployment.

A bug fix addressing community bug report: #5363.

## Important Info
OpenAPI document processing is done on the first call to `/openapi` endpoint.

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/rest/init/OpenApiServletContainerInitializer.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/rest/init/OpenApiServletContainerInitializer.java
@@ -87,8 +87,6 @@ public class OpenApiServletContainerInitializer implements ServletContainerIniti
         }
 
         // Start the OpenAPI application
-        new JerseyServletContainerInitializer().onStartup(new HashSet<>(asList(OpenApiApplication.class)), ctx);
-
         ServletContainer servletContainer = new ServletContainer(new OpenApiApplication());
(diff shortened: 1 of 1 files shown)


jGauravGupta:
jenkins test please

JamesHillyard (approved):

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/rest/init/OpenApiServletContainerInitializer.java
```diff
@@ -91,6 +91,7 @@ public void onStartup(Set<Class<?>> c, ServletContext ctx) throws ServletExcepti
```
pdudits, line 91: I believe the initializer is not doing anything now, since you're instantiating the servlet container in lines below. Could you try if it works without it?

jGauravGupta:
jenkins test please

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/rest/init/OpenApiServletContainerInitializer.java
jGauravGupta: Thanks @pdudits , Also as of now @ApplicationPath annotation is removed from `fish.payara.microprofile.openapi.rest.app.OpenApiApplication` so the initializer call is redundant now.

pdudits (approved):

PR #5396, opened 2021-08-31
Fish 5702 improper limitation pathname restricted directory

breakponchito:
## Description

Adding fix to solve problem to access Restricted Directory when using context root as /

### New tests

Added new test: StandardContextValveTest.java
### Testing Performed

I already tested this on local environment. Check the following comment from jira to verify results:
https://payara.atlassian.net/browse/FISH-5702?focusedCommentId=57244

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
@@ -280,7 +280,7 @@ final class StandardContextValve
      * @param path to be normalized
      * @return normalized path or null
      */
-    private static String normalize(String path) {
+    protected String normalize(String path) {
         if (path == null) {
             return null;
         }
@@ -309,6 +309,10 @@ final class StandardContextValve
             rv = rv.substring(0, index2) + rv.substring(idx + 3);
         }
 
+        // Resolve occurrences of "/./" example if the path looks like /app/./some/./something/./my.jsp
+        //then after processing should look like /app/some/something/my.jsp
+        rv = evaluateNormalizedPathWithSinglePoint(rv);
+
         //if the path don't start with / then include it
         if(!rv.startsWith("/")) {
             rv = "/" + rv;
@@ -316,6 +320,23 @@ final class StandardContextValve
 
         // Return the normalized path that we have completed
         return rv;
+    }
+
+    /**
+     * this method helps to evaluate the element "/./" on the path
+     * @param path to be normalized
+     * @return normalized path
+     */
+    private String evaluateNormalizedPathWithSinglePoint(String path) {
+        // Resolve occurrences of "/./" in the normalized path
+        while (true) {
+            int idx = path.indexOf("/./");
+            if (idx < 0) {
+                break;
+            }
+            path = path.substring(0, idx) + path.substring(idx + 2);
+        }
+        return path;
     }
 
     private Wrapper preInvoke(Request request, Response response) {
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
@@ -79,7 +79,7 @@ import org.glassfish.grizzly.utils.Charsets;
  * @author Craig R. McClanahan
  * @version $Revision: 1.19 $ $Date: 2007/05/05 05:31:54 $
  */
-
+// Portions Copyright [2021] [Payara Foundation and/or its affiliates]
 final class StandardContextValve
     extends ValveBase {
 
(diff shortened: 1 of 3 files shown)


appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
```diff
@@ -325,7 +325,8 @@ private Wrapper preInvoke(Request request, Response response) {
                     || (requestPath.toUpperCase().startsWith("/META-INF/", 0))
                     || (requestPath.equalsIgnoreCase("/META-INF"))
                     || (requestPath.toUpperCase().startsWith("/WEB-INF/", 0))
-                    || (requestPath.equalsIgnoreCase("/WEB-INF"))) {
+                    || (requestPath.equalsIgnoreCase("/WEB-INF"))
```
smillidge, line 328: Shouldn't the method normalize above this line have converted to the path to /WEB-INF? What is the output of normalize?

OndroMih: I agree. A better solution would be to change the `normalize` method. If the `CoyoteAdapter` class removes the leading `/`, then the `normalize` method should add it at the beginning. The comment on the normalize method also says it normalizes both . and .., but the code only addresses .. and not the . case. The . case seems to be already handled in the `CoyoteAdapter` before this method is called but I believe that the . case should also be addressed here, as the comment says, just in case.

OndroMih (changes requested):

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
Pandrex247: I agree that this doesn't seem complete.
If nothing else, this PR as-is doesn't prevent a user going in to the META-INF directory either since a similar curl to `curl -i --path-as-is http://localhost:8080/.//META-INF/MANIFEST.MF` gets normalised to `META-INF/MANIFEST.MF`, which doesn't match any of those _if_ statements (similar case to what you were trying to catch with your extra line here for WEB-INF)

Pandrex247 (changes requested):

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
breakponchito: > Shouldn't the method normalize above this line have converted to the path to /WEB-INF? What is the output of normalize?

after reviewing in debug mode the CoyoteAdapter is removing the leading / . That is why when this code executes the conditions don't evaluate the final string. For now the approach to improve the normalize method to complete and normalize the URL makes sense. I'm going the changes to improve the normalize method and complete the url with the starting /

breakponchito: > I agree. A better solution would be to change the `normalize` method. If the `CoyoteAdapter` class removes the leading `/`, then the `normalize` method should add it at the beginning. The comment on the normalize method also says it normalizes both . and .., but the code only addresses .. and not the . case. The . case seems to be already handled in the `CoyoteAdapter` before this method is called but I believe that the . case should also be addressed here, as the comment says, just in case.

After reviewing the evaluation of the “/../” and “/./” it seems like the process from the CoyoteAdapter is doing most of the work to remove and let the path normalized. Now I added the case to consider the evaluation of the "/./" when I have a path with the following structure:  /app/./some/./something/./my.jsp
then after evaluating the result is:
/app/some/something/my.jsp

image

let me know if you have questions

OndroMih (approved):
Looks good now. I just suggested renaming of a method to clarify what it does and make the test more readable.

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContextValve.java
```diff
@@ -79,7 +79,7 @@
  * @author Craig R. McClanahan
  * @version $Revision: 1.19 $ $Date: 2007/05/05 05:31:54 $
  */
-
+// Portions Copyright [2021] [Payara Foundation and/or its affiliates]
```
Pandrex247, line 82: There's already a portions copyright block above on line 58 - you can just change the 2019 to 2021 there.

Pandrex247 (approved):
One non-blocking comment

appserver/web/web-core/src/test/java/org/apache/catalina/core/StandardContextValveTest.java
```diff
@@ -0,0 +1,116 @@
...
+        String result = standardContextValve.normalize(path1);
+
+        assertEquals("/my.jsp", result);
+
+        result = standardContextValve.normalize(path2);
+
+        assertEquals("/app/some/something/my.jsp", result);
+    }
+
+    protected void verifyResults(int pipelineResult, int times, HttpRequest httpRequest, HttpResponse httpResponse,
```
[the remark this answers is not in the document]
breakponchito: already done

breakponchito:
Jenkins test

PR #5400, opened 2021-09-03
FISH-146 DAS warning for changing admin port for a running instance

JamesHillyard:
## Description
This is an improvement making it clearer on the DAS about changing the ASADMIN_LISTNER_PORT property while an instance is running. 

When changing ASADMIN_LISTENER_PORT on a running instance, it's still listening on the old admin port. Then DAS thinks the instance isn't running and can't access it or stop it. This is intended behaviour but the user should have sufficient warning.

image
_This shows what Instances -> Properties -> System Properties now looks like_

### Testing Performed
Manually tested the warning text is showed in the appropriate locations.

### Testing Environment
Windows 10 Pro, Maven 3.6.3, JDK 8

## Notes for Reviewers
The left padding being missing for both the warning text and page title is known. See JIRA for more information.


JamesHillyard:
Jenkins test please

appserver/admingui/cluster/src/main/resources/org/glassfish/cluster/admingui/Strings.properties
```diff
@@ -216,6 +216,7 @@ standaloneInstancePropertiesPageTitleHelp=
 standalone.ApplicationsTitle=Applications
 standalone.ApplicationsTitleHelp=Deploy, undeploy, enable, disable, and remove selected applications associated with the instance. Removing the application only removes the application from this instance. The application is not undeployed.
 standalone.InstanceName=Instance Name:
+standalone.asadminListenerPortWarning=Warning: ASADMIN_LISTENER_PORT can not be changed dynamically. You should stop the running instance before saving, otherwise DAS won't be able to stop the running instance and it may need to be forcibly stopped.
```
OndroMih, line 219: This message will always be displayed on page load even if the property isn't changed. I would change the text to Note: instead of Warning: so that users aren't worried that something is wrong.
```suggestion
standalone.asadminListenerPortWarning=Note: ASADMIN_LISTENER_PORT can not be changed dynamically. You should stop the running instance before saving, otherwise DAS won't be able to stop the running instance and it may need to be forcibly stopped.
```

JamesHillyard:
Jenkins test please

OndroMih (approved):

PR #5401, opened 2021-09-06
Increment version numbers 5.2021.8

cubastanley:
Title.


appserver/extras/payara-micro/payara-micro-boot/src/main/resources/MICRO-INF/domain/branding/glassfish-version.properties
```diff
@@ -3,7 +3,7 @@ brief_product_name=@@@PRODUCT@@@
 abbrev_product_name=@@@PRODUCT@@@
 major_version=5
 minor_version=2021
-update_version=7
+update_version=8
```
Pandrex247, line 6:
```suggestion
update_version=8-SNAPSHOT
```

Pandrex247 (changes requested):
Micro missing snapshot, and no update to `appserver/packager/appserver-base/src/main/docs/README.txt`?

appserver/extras/payara-micro/payara-micro-boot/src/main/resources/MICRO-INF/domain/branding/glassfish-version.properties
cubastanley: The scripts aren't designed for snapshots in the glassfish-version.properties file, it's been working like this for as long as I can remember

PR #5410, opened 2021-09-14
FISH-5636 Remote EJB Tracing Improvements

Pandrex247:
## Description

An attempt to improve the throughput of the OpenTracing service, particularly with regards to remote EJBs.
The main change is the un-synchronising of the `getTracer` method - creation of tracers is still synchronised but is now split off into another method and utilises double-checked locking to prevent conflicts. While double-checked locking can be naughty, in this case it's being done on a map itself, so the issue of partially constructed references not returning null Shouldn't™ occur.

My results from performance testing were largely inconclusive since the realm of variance is pretty large - this is working at the micro and millisecond timeframe, any slight deviation outside of my control (e.g. OS runs a background process) will affect the result. In any case, I shall list the reasoning of how each change should help below.

* `getTracer` split into two methods: `getTracer` and `createTracer`
  * `getTracer` is now unsynchronised, meaning there should be less contention waiting for locks in normal operation since only `createTracer` is synchronised (tracers only get created once)
* `StandardWrapper` no longer does two calls for `getTracer` back-to-back (less of an impact now that `getTracer` is no longer synced but still saves performing the get operation on a ConcurrentHashMap twice)
* Lookup of `RequestTracingService` moved from `SpanBuilder` constructor to `OpenTracingService` (and then passed to `Tracer` as a constructor parameter). The intent here is to make it so an HK2 lookup doesn't need to be performed each time a span is created, and is instead done once.  

This PR also includes some cleanup of deprecated or ugly code:
* MockTracer removed from code base - this is now handled using service loader
  * Supplemental - I noticed the export for `log` was missing and so I added it in
* `GlobalTracer.register(Tracer)` is deprecated, so has been replaced with `GlobalTracer.registerIfAbsent(Callable)`
  * A lot of the null checks being done here were redundant, since they are handled by the factory. In any case, a try-catch has been added to the factory around the create method since that can (but Shouldn't™) be thrown by `GlobalTracer.registerIfAbsent(Callable)`
* Removed `Events` from constructor of `RequestTraceStoreFactory` - not used.
* `DAY` variable removed from `RequestTracingService` - not used.
* Static `ScopeManager` moved from `OpenTracingService` to `Tracer` - it isn't used anywhere else, and our `Tracer` impl doesn't use any other kind of `ScopeManager`

### Dependant PRs
MicroProfile OpenTracing TCK Runner update: https://github.com/payara/MicroProfile-TCK-Runners/pull/153

### Blockers
None

### New tests
None - with the level of sensitivity being worked on here (milliseconds) performance tests are not suitable for day-to-day testing.

### Testing Performed
Ran my own test using JMeter numerous times against various points on the branch and compared against master - results largely inconclusive. Added my own logging around every call to `getTracer` to try and get a sense of the response time of synchronised vs. un-synchronised (inconclusive). 

Ran a test created by @fturizo and inspected using JProfiler numerous times against various points on the branch and compared against master to see if any discernible difference could be gained (inconclusive).

Ran MicroProfile OpenTracing TCK to check everything still works.

### Test suites executed
MicroProfile OpenTracing

### Testing Environment
Windows 10, JDK8.

## Documentation
N/A

## Notes for Reviewers
Please keep performance in mind when reviewing, pointing out stuff I've missed or things I may have misunderstood.

### DIFF
--- a/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
+++ b/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
@@ -99,7 +99,7 @@ public class OpenTracingService implements EventListener {
         // registered to that application (if there is one)
         if (event.is(Deployment.APPLICATION_LOADED)) {
             ApplicationInfo info = (ApplicationInfo) event.hook();
-            createAndReturnTracer(info.getName());
+            createTracer(info.getName());
         }
 
         if (event.is(Deployment.APPLICATION_UNLOADED)) {
@@ -124,13 +124,13 @@ public class OpenTracingService implements EventListener {
 
         // If there isn't a tracer for the application, create one
         if (tracer == null) {
-            tracer = createAndReturnTracer(applicationName);
+            tracer = createTracer(applicationName);
         }
 
         return tracer;
     }
 
-    private synchronized Tracer createAndReturnTracer(String applicationName) {
+    private synchronized Tracer createTracer(String applicationName) {
         // Does this NEED to be synchronised? Tracers don't store state, and Scopes are ThreadLocal
 
         // Double-checked locking - potentially naughty
--- a/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
+++ b/nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
@@ -74,7 +74,6 @@ public class OpenTracingService implements EventListener {
 
     // The tracer instances
     private static final Map<String, Tracer> tracers = new ConcurrentHashMap<>();
-    private static final ScopeManager scopeManager = new fish.payara.opentracing.ScopeManager();
     
     // The name of the Corba RMI Tracer
     public static final String PAYARA_CORBA_RMI_TRACER_NAME = "__PAYARA_CORBA_RMI";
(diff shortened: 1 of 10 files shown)


nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
```diff
@@ -123,28 +129,39 @@ public synchronized Tracer getTracer(String applicationName) {
...
 
-            if (Boolean.getBoolean("USE_OPENTRACING_MOCK_TRACER")) {
-                tracer = new MockTracer(new ThreadLocalScopeManager(), MockTracer.Propagator.TEXT_MAP);
-            } else if (tracer == null) {
-                tracer = new fish.payara.opentracing.tracer.Tracer(applicationName, scopeManager);
+    private synchronized Tracer createTracer(String applicationName) {
+        // Does this NEED to be synchronised? Tracers don't store state, and Scopes are ThreadLocal
+
+        // Double-checked locking - potentially naughty
+        Tracer tracer = tracers.get(applicationName);
```
svendiedrichsen, line 142: How about ConcurrentHashMap.computeIfAbsent and put the following code into the compute function.

nucleus/payara-modules/opentracing-adapter/src/main/java/fish/payara/opentracing/OpenTracingService.java
```diff
@@ -154,15 +171,20 @@ public synchronized Tracer getTracer(String applicationName) {
...
-                RequestTracingService.class);
-
-        if (requestTracingService != null) {
+        if (getRequestTracingService() != null) {
             return requestTracingService.isRequestTracingEnabled();
         }
         return false;
     }
 
+    public RequestTracingService getRequestTracingService() {
```
svendiedrichsen, line 180: How about returning an Optional<RequestTracingService>?

Pandrex247: I don't think returning an Optional really adds much here.
The method is intended as a poke to HK2 if required rather than any kind of true getter (public scope currently present is wrong - should be private).

jGauravGupta (approved):

Pandrex247:
Jenkins test please

PR #5411, opened 2021-09-14
FISH-5736 Fix Unknown Exclude Field Provided Warning on Startup

Pandrex247:
## Description
Title.

Handle the empty string which is default so that the log message doesn't get thrown on startup.

### Blockers
None

### New tests
None

### Testing Performed
Started server - no error.
Added an exclude field - no error and field excluded.

### Testing Environment
Windows 10, JDK8

## Documentation
N/A

## Notes for Reviewers
None


Pandrex247:
Jenkins test please

JamesHillyard (approved):

Issue #5418, opened 2021-09-17
Bug Report:  Environment variables are not read in OpenIdAuthenticationDefinition/FISH-5748


LarsBrenker:
# Description #
----------

Since Payara 5.2021.6, environment variables are no longer resolved in the `OpenIdAuthenticationDefinition` Annotation. 

## Expected Outcome ##

As before, environment variables should be resolved.

## Current Outcome ##

![grafik](https://user-images.githubusercontent.com/57911568/133775045-eb61d2ee-5f96-4d8c-bd1b-5c6bcf069614.png)

## Steps to reproduce ## 

```java
@ApplicationScoped
@OpenIdAuthenticationDefinition(
        clientSecret = "${ENV=OIDC_CLIENTSECRET}",
        clientId = "${ENV=OIDC_CLIENTID}"
)
@DeclareRoles({"xxx", "yyy"})
public class ApplicationConfig {

}
```

This has led to the URL shown above since Payara 5.2021.6

This can be remedied by first setting the values ​​in the Microprofile Config.

```
payara.security.openid.clientId=${OIDC_CLIENTID}
payara.security.openid.clientSecret=${OIDC_CLIENTSECRET}
```

## Environment ##
Payara Micro 5.2021.6 and Payara Micro 5.2021.6

shub8968:
Hi @LarsBrenker,

Would you be able to provide the proper sample reproducer app?

OndroMih:
I can reproduce with the sample app here https://github.com/payara/Payara/pull/5422. This app doesn't refer to an environment variable but when I change the annotation in SecuredPage.java to refer to an environment variable as in the example above, I see the same behavior as reported:

```
@OpenIdAuthenticationDefinition(
        providerURI = "http://localhost:8080/openid-server/webresources/oidc-provider",
        clientId = "${ENV=OIDC_CLIENTID}",
        clientSecret = CLIENT_SECRET_VALUE,
        redirectURI = "${baseURL}/Callback"
)
```

While the `OIDC_CLIENTID` environment variable is set to `sample_client_id` before Payara Server starts.

I don't see this with an older version. It's most probably related to the integrating of the security connectors done in FISH-786: https://github.com/payara/Payara/pull/5373

LarsBrenker:
Hi @OndroMih, thanks for the reproducer. It can be. If https://github.com/payara/Payara/pull/5373 was merged on August 6th, this fits in with my observation that the problem exists since Payara 5.2021.6.

[...]

PR #5422, opened 2021-09-22
OpenID Sample app can be deployed manually

OndroMih:
## Description

The OpenID sample project was only usable for running the test suite but the built artifact wouldn't deploy. I've refactored it so that the tests should run as they did before but the build now produces 2 applications (client and server) and these can be deployed manually to test this sample application also manually.

I also added the README file to describe how to do it.

I used this app to reproduce https://github.com/payara/Payara/issues/5418.


OndroMih:
Jenkins test please

Pandrex247 (approved):

breakponchito (approved):
LGTM

PR #5431, opened 2021-10-04
FISH-5734 fix recursive call to map.compute

aubi:
## Description
Occasionally, application crashes with exception from ConcurrentHashMap during microprofile configuration processing:
java.lang.IllegalStateException: Recursive update

### Testing Performed
Deploy reproducer (attached to the JIRA ticket) to root, restart server, try to access http://localhost:8080/ShowConfigValues

### Testing Environment
Linux, OpenJDK 8

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
@@ -53,7 +53,6 @@ import java.lang.reflect.Array;
 
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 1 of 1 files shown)


aubi:
Jenkins test please

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
```diff
@@ -225,11 +224,13 @@ protected ConfigValueImpl getConfigValue(String propertyName, String cacheKey, L
         final String entryKey = cacheKey + (defaultValue != null ? ":" + defaultValue : "")  + ":" + (entryTTL / 1000) + "s";
         final long now = currentTimeMillis();
 
+        boolean isExpansionEnabled = isExpansionEnabled(propertyName);
+
         return cachedValuesByProperty.compute(entryKey, (key, entry) -> {
```
pdudits, line 229: I don't think we can keep this call to compute. Because when expansion *is* enabled, other entries still need to get resolved during computation and break the contract of `compute` method.

aubi:
Jenkins test please

pdudits (approved):

nucleus/payara-modules/nucleus-microprofile/config-service/src/main/java/fish/payara/nucleus/microprofile/config/spi/PayaraConfig.java
aubi: You are right, searchConfigSources() could cause another recursion, e.g. breaking compute contract. Fixed in new version.

PR #5435, opened 2021-10-04
FISH-5768 Clustered CDI Events not Received if Sender Application Deployed on Receiver Instance

Pandrex247:
## Description
Bug Fix.

Clustered CDI Events do not get resolved correctly if both the sender application and receiver application are deployed to the same instance. The cause is our old favourite: class loaders.

When Weld comes to resolve the observer methods for the event, it fails to do a `.equals` comparison on the event class due to it having been loaded by the sender class loader rather than the receiver.

The fix here partially reverts https://github.com/payara/Payara/pull/5010 - notably the bit in `PayaraValueHolder` where the error was tracked down to.
This does seem to have a knock-on effect when resolving qualifiers - where _Sometimes_ it seems to not have the correct class loader set and will hit `java.lang.IllegalArgumentException: interface xyz is not visible from class loader`. The behaviour here however is to just ignore the error and move on, and from my testing this doesn't seem to negatively impact the expected behaviour (the correct number of events are still received and their observing methods invoked whether the error appears or not). Given this, I've added an extra clause to the log statement to use Fine rather than Info, similar to https://github.com/payara/Payara/blob/master/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java#L173 (which likely gets hit afterwards on this particular thread anyway)

### Blockers
None

### New tests
Pending...

### Testing Performed
Pending...

Start two instances, and deploy the receiver and sender applications (Payara-Examples/payara-micro/clustered-cdi-events) to both instances. 

### Testing Environment
Windows 10 - JDK 8

## Documentation
N/A

## Notes for Reviewers
Pending...

### DIFF
--- a/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java
+++ b/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java
@@ -170,7 +170,7 @@ public class ClusteredCDIEventBusImpl implements CDIEventListener, ClusteredCDIE
                     bm.fireEvent(eventPayload,annotations);
                 } catch (IOException | ClassNotFoundException ex) {
                     Logger.getLogger(ClusteredCDIEventBusImpl.class.getName())
-                            .log(ex.getCause() instanceof IllegalStateException? Level.FINE : Level.INFO,
+                            .log(ex.getCause() instanceof IllegalStateException ? Level.FINE : Level.INFO,
                                     "Received Event but could not process it", ex);
                 }
             });
--- a/appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraClusteredCDIEventImpl.java
+++ b/appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraClusteredCDIEventImpl.java
@@ -175,7 +175,7 @@ public class PayaraClusteredCDIEventImpl implements PayaraClusteredCDIEvent {
                             Logger.getLogger(PayaraClusteredCDIEventImpl.class.getName()).log(
                                     ex instanceof IllegalArgumentException &&
                                             ex.getMessage().contains("is not visible from class loader") ? Level.FINE : Level.INFO,
-                                    "Problem determining the qualifier type of an Event ignoring", ex);
+                                    "Problem determining the qualifier type of an Event, ignoring", ex);
                         }
                     }
                 }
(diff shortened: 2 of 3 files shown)


Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

pdudits (approved):

lprimak:
I am quite sure if this isn't going to work. I put this in originally in response to another bug. However, I am not quite sure which one it is but I distinctly remember the original code was necessary

What that piece of code does is find the appropriate application class loader for the component id of the sender and sets it for the duration of deserialization, which is necessary to avoid `ClassNotFoundException`

lprimak:
I remember the issue... it was spamming the logs with CNFs every time the message was sent.

> The problem we faced was that it didn't find appropriate application class loader for receiving application when (another instance of) sending application was also deployed. Receiver's classloader is already set as the thread context classloader earlier in the process and that's the only one that makes sense for reception of clustered event as that's the only one the application can process.

I think I see a cleaner solution that won't break things...

- Try deserializing with existing CL context
- If it doesn't work, use existing code to set the sender's class loader and try again
- only if all fails, print CNF

I believe that will save the original intent and fix the current issue.

WDYT?

Pandrex247:
> I remember the issue... it was spamming the logs with CNFs every time the message was sent.
However, I don't think this was the only issue, there are others if I remember correctly...

Can you think what they might be?

PayaraValueHolder is only used by the clustered event bus, and I'm not getting CNFEs unnecessarily logged.
They're caught [here](https://github.com/payara/Payara/blob/master/appserver/payara-appserver-modules/payara-micro-cdi/src/main/java/fish/payara/micro/cdi/extension/ClusteredCDIEventBusImpl.java#L172): 

```
                } catch (IOException | ClassNotFoundException ex) {
                    Logger.getLogger(ClusteredCDIEventBusImpl.class.getName())
                            .log(ex.getCause() instanceof IllegalStateException? Level.FINE : Level.INFO,
                                    "Received Event but could not process it", ex);
```

I'll need to double-check if they're thrown without this change.

The only other case I think you might be referring to is when qualifiers start to get involved, but this change doesn't affect that (error gets logged and ignored): https://github.com/payara/Payara/blob/8e05ed98e9e40011ce356c8d47c522f76d769232/appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/PayaraClusteredCDIEventImpl.java#L175

I originally had a description of the issue in the PR description but removed it because the problem was found to happen even before Hazelcast 4 was introduced (5.2020.7 was the last community release before Hazelcast 4 was introduced if you want to compare yourself).

lprimak:
The issue is when CNF is thrown, it's unclear whether the app is not deployed or something else went wrong.
Original code just throws CNF whenever it can't deserialize, which would happen in all instances in the cluster when
just the sender is deployed and sending messages, which is not good.

PR #5455, opened 2021-10-15
Fix CompressionTest - disable cache

aubi:
## Description
The CompressionTest occasionally fails. The reason is the cache of the HTTP client. This change disables the cache, so all request are executed and can return different sizes depending on the compression.

### Testing Performed
Payara Samples tests

### Testing Environment
Debian Linux, OpenJDK 8


aubi:
Jenkins tests please

PR #5460, opened 2021-10-19
FISH-1323 upgrade docker jdk to 8u302 and 11.0.12

aubi:
## Description
upgrade docker jdk to 8u302 and 11.0.12
Disable TLS 1.0 and 1.1

### Blockers
Is removal of TLS 1.0 and 1.1 blocker for anything?

### Testing Performed
I ran all tests in project, Payara Samples

### Testing Environment
Zulu JDK 1.8_302 on Debian Linux


rdebusscher:
>Is removal of TLS 1.0 and 1.1 blocker for anything?

When developers use a socket within code and define explicitly the TLS versions 1.0 and 1.1.  Not when using a connection with https protocol.

PR #5477, opened 2021-10-29
FISH-5801: Custom Vendor MP Metrics Using Placeholders Require 'tags'

rdebusscher:
## Description
Fixes Exception when not using tags in custom Vendor metrics having a Placeholder.

### New tests
new Unit test MBeanMetadataTest

### Testing Environment
Zulu 8.52.0.23-CA-macosx (build 1.8.0_282-b08) on Mac 11.8 with Maven 3.8.2

## Documentation
Not needed, Bugfix

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadata.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadata.java
@@ -93,7 +93,7 @@ public class MBeanMetadata implements Metadata {
     private Boolean valid;
 
     @XmlElementWrapper(name = "tags", nillable = true)
-    @XmlElement(name="tag")
+    @XmlElement(name = "tag")
     private List<XmlTag> tags;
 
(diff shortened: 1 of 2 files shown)


appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MBeanMetadata.java
```diff
@@ -215,9 +215,12 @@ private boolean validateMetadata() {
                     validationResult = false;
                 } else if (metadata.getMBean().contains(keyword)) {
                     boolean tagSpecifier = false;
-                    for (XmlTag tag: tags) {
-                        if (tag.getValue().contains(keyword)) {
-                            tagSpecifier = true;
+                    if (tags != null) {
```
aubi, line 218: tags can never be null, was fixed by FISH-5658

aubi (approved):
I would prefer to remove tags!=null.
Otherwise, LGTM

rdebusscher:
jenkins test please

Pandrex247 (approved):

PR #5497, opened 2021-11-11
PAYARA-5496  fixing RwLockDatastructure locking bug (also FISH-985)

### LINKED ISSUE #5496: Bug Report:  RWLockDataStructure may leave a read lock /FISH-5858
# Description #
----------

The class RWLockDataStructure has a flawed lock handling which may result in a read lock not being unlocked. This freezes any thread trying to get a write lock.

https://github.com/payara/Payara/blob/master/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/datastructure/RWLockDataStructure.java

The method removeAll() clears the resources list outside of write lock (line 204).  Which seems redundant anyway.

```
    public void removeAll() {
        writeLock.lock();
        try {
            Iterator it = resources.iterator();
            while (it.hasNext()) {
                handler.deleteResource((ResourceHandle) it.next());
                it.remove();
            }
        } finally {
            writeLock.unlock();
        }
        resources.clear(); // redundant and removes eventually freshly created resources without properly deleting
    }
```

If this happens while some other thread calls the method getResource() (at line 118) then the resulting IndexOutOfBoundsException prevents unlocking the currently held read lock. The code is lacking some finally block to unlock the readlock.

```
    public ResourceHandle getResource() {
        readLock.lock();
        for(int i=0; i<resources.size(); i++){
            ResourceHandle h = resources.get(i); // this will cause an exception if the code above has cleared the list
            if (!h.isBusy()) {
                readLock.unlock();
                writeLock.lock();
                try {
                    if (!h.isBusy()) {
                        h.setBusy(true);
                        return h;
                    } else {
                        readLock.lock();
                        continue;
                    }
                }finally {
                    writeLock.unlock();
                }
            } else {
                continue;
            }
        }
        readLock.unlock(); // this should be in a finally block to prevent lost read locks
        return null;
    }
```

## Expected Outcome ##

All aquired locks will always be unlocked.

## Current Outcome ##

If the aquired read lock is not returned in case an exception is thrown at line 118 the RWLockDataStructure stays read locked and the server starts to block all threads trying the aquire a write lock.

## Steps to reproduce ## 

As always with race conditions it is not that simple to predictably provoke this situation. But the code is simple enough to spot the problem.

## Environment ##

- **Distribution**: 5.2020.5 Full
- **JDK Version**: OpenJDK 11
- **Operating System**: Linux

svendiedrichsen:
## Description
This fixes a lost read lock in RWLockDataStructure in case removeAll() is called while other threads call getResource().
fixes #5496

### New tests
No new tests created

### Testing Performed
Testing could be done by simulation heavy db access load.

## Notes for Reviewers
There were notes about some coarse lock handling in method addResource(...) which as been adapted according to lock handling in getResource() method

### DIFF
--- a/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/datastructure/RWLockDataStructure.java
+++ b/appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/datastructure/RWLockDataStructure.java
@@ -47,9 +47,7 @@ import com.sun.enterprise.resource.allocator.ResourceAllocator;
 import com.sun.enterprise.resource.pool.ResourceHandler;
 import com.sun.logging.LogDomains;
 
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Supplier;
@@ -65,7 +63,9 @@ public class RWLockDataStructure implements DataStructure {
     private final ResourceHandler handler;
     private int maxSize;
 
-    private final List<ResourceHandle> resources;
+    private final List<ResourceHandle> allResources;
+    private final Deque<ResourceHandle> freeResources;
+
     private final ReentrantReadWriteLock reentrantLock = new ReentrantReadWriteLock();
     private final ReentrantReadWriteLock.ReadLock readLock = reentrantLock.readLock();
     private final ReentrantReadWriteLock.WriteLock writeLock = reentrantLock.writeLock();
@@ -73,7 +73,8 @@ public class RWLockDataStructure implements DataStructure {
     protected static final Logger _logger = LogDomains.getLogger(RWLockDataStructure.class,LogDomains.RSR_LOGGER);
 
     public RWLockDataStructure(int maxSize, ResourceHandler handler) {
-        resources = new ArrayList<>(Math.min(maxSize, 1000));
+        allResources = new ArrayList<>(Math.min(maxSize, 1000));
+        freeResources = new ArrayDeque<>(Math.min(maxSize, 1000));
         this.maxSize = maxSize;
         this.handler = handler;
         if(_logger.isLoggable(Level.FINEST)) {
@@ -103,13 +104,17 @@ public class RWLockDataStructure implements DataStructure {
     }
 
     private boolean isNotFull() {
-        return doLockSecured(() -> resources.size() < maxSize, readLock);
+        return doLockSecured(() -> allResources.size() < maxSize, readLock);
     }
 
     private boolean addResourceInternal(final ResourceHandle handle) {
         return doLockSecured(() -> {
-            if (resources.size() < maxSize) {
-                return resources.add(handle);
+            if (allResources.size() < maxSize) {
+                boolean added = allResources.add(handle);
+                if (added) {
+                    freeResources.offerLast(handle);
+                }
+                return added;
             }
             return false;
         }, writeLock);
@@ -119,36 +124,26 @@ public class RWLockDataStructure implements DataStructure {
      * {@inheritDoc}
      */
     public ResourceHandle getResource() {
-        ResourceHandle result = null;
-        readLock.lock();
-        try {
-            for(int i = 0; result == null && i < resources.size(); i++){
-                ResourceHandle h = resources.get(i);
-                if (!h.isBusy()) {
-                    readLock.unlock();
-                    writeLock.lock();
-                    try {
-                        if (!h.isBusy()) {
-                            h.setBusy(true);
-                            result = h;
-                        }
-                    } finally {
-                        readLock.lock();
-                        writeLock.unlock();
-                    }
-                }
+        return doLockSecured(() -> {
+            final ResourceHandle resourceHandle = freeResources.pollFirst();
+            if (resourceHandle != null) {
+                resourceHandle.setBusy(true);
             }
-        } finally {
-            readLock.unlock();
-        }
-        return result;
+            return resourceHandle;
+        }, writeLock);
     }
 
     /**
      * {@inheritDoc}
      */
     public void removeResource(ResourceHandle resource) {
-        boolean removed = doLockSecured(() -> resources.remove(resource), writeLock);
+        boolean removed = doLockSecured(() -> {
+            final boolean removedResource = allResources.remove(resource);
+            if (removedResource) {
+                freeResources.remove(resource);
+            }
+            return removedResource;
+        }, writeLock);
         if (removed) {
             handler.deleteResource(resource);
         }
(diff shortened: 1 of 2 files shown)


OndroMih (approved):
Looks good. I don't see why it wouldn't work, it actually looks much better than before. I think that it also improves performance when the connection pool is configured to create multiple connections at once because the pool no longer needs to wait until the whole batch is created but connections can be assigned to other threads even before the whole batch is finished.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/datastructure/RWLockDataStructure.java
```diff
@@ -59,52 +62,52 @@
...
     /**
      * {@inheritDoc}
      */
     public int addResource(ResourceAllocator allocator, int count) throws PoolingException {
         int numResAdded = 0;
-        writeLock.lock();
-        //for now, coarser lock. finer lock needs "resources.size() < maxSize()" once more.
+        readLock.lock();
         try {
             for (int i = 0; i < count && resources.size() < maxSize; i++) {
```
sgflt, line 91: Nitpicking: Is following construction possible?

```java
            for (int i = 0; i < count && doLockSecured(this::isNotFull, readLock); i++) {
                writeLock.lock();
                try {
                    if (isNotFull()) {
                        ResourceHandle handle = handler.createResource(allocator);
                        resources.add(handle);
                        numResAdded++;
                    }                  
                } catch (Exception e) {
                    PoolingException pe = new PoolingException(e.getMessage());
                    pe.initCause(e);
                    throw pe;
                } finally {
                    writeLock.unlock();
                }
             }
```

At first glance it improves writelock downgrade, which is in this case possible unnecessary. Am I right?

sgflt: 2nd version:

```java
 for (int i = 0; i < count && doLockSecured(this::isNotFull, readLock); i++) {
         numResourcesAdded += doLockSecured(this::addResourceInternal, writeLock);
 }
```

svendiedrichsen: In IMHO it would be best not to create the resource within the write lock. But rather create it before without lock then try to add it write locked and if this does not succeed destroy it again.
```
 for (int i = 0; i < count && doLockSecured(this::isNotFull, readLock); i++) {
         final  ResourceHandle handle = handler.createResource(allocator);
         boolean added = doLockSecured(() -> addResourceInternal(handle), writeLock);
         if (added) {
             numResourcesAdded++;
         } else {
             handler.removeResource(handle);
         }
 }
``` 
This way the DS is not locked during connection time resulting in higher throughput overall.

fturizo: I agree with @svendiedrichsen, as in this way, the resource's creation outside the write lock's scope will lead to better performance, especially when a batch of connections is being requested by the associated pool to each resource.

svendiedrichsen:
@fturizo @sgflt I have adapted the code of the `addResource()` method according to the resolved discussion outcome.

fturizo:
jenkins test please

svendiedrichsen:
@fturizo @aubi I have refactored the `getResource()` to use a separate Deque for managment of free resources.

aubi:
Here is my reproducer - simple REST app with access to a database (db tasks, username+password tasks, table task with 1 row)
[PerformanceReproducer.zip](https://github.com/payara/Payara/files/7579412/PerformanceReproducer.zip)
Commands to setup Payara:
```
./asadmin start-domain
./asadmin add-library /usr/share/java/postgresql.jar
./asadmin create-jdbc-connection-pool --datasourceclassname org.postgresql.ds.PGSimpleDataSource --driverclassname org.postgresql.Driver --restype javax.sql.DataSource --property user=
tasks:password=tasks:URL="jdbc\:postgresql\://localhost/tasks" tasksPool
./asadmin ping-connection-pool tasksPool
./asadmin create-jdbc-resource --connectionpoolid tasksPool jdbc/tasks
./asadmin set configs.config.server-config.thread-pools.thread-pool.http-thread-pool.max-thread-pool-size=50
./asadmin stop-domain
```
Application testing many accesses (current settings total 1.000 tasks in 100 threads, 10 requests/task):
[PerformanceReproducerTester.zip](https://github.com/payara/Payara/files/7579413/PerformanceReproducerTester.zip)

aubi:
Commits to 87ec3c68fb85fd1961b84ca0206673e759809e5e (doing the lock downgrade...):
Fine, I totally agree, code is much simpler and cleaner.

Commits to 1433844d72c14c2055af1583879436e7f3ee0a01 (refactoring code according...):
This speeds start up, but some resources are created and immediately removed. I can imagine, that this may cause problems, e.g. when only limited number of connections are allowed to a database -- some requests may fail.

I made a modification, when only one thread is allowed to add resource. It is a bit slower (not significantly), doesn't waste resources: https://github.com/aubi/Payara/tree/single-thread-addresource
I prefer this version.

Changes from 958fef8ba8312fe2c404b493c2393452336d6a05 (improving on handling free...):
I'm not sure the freeResources help. It removes for-cycle, but doesn't improve performance (by my measurements) and need more careful review.

svendiedrichsen:
@aubi The freeResources Deque alos improves on the way connections are taken and used. Working with the list it distibutes the usage more over the whole list of connections which prevents idle connection reclaim.

aubi:
I rebased my change (single thread addResource) to the latest commits. Before going final, it will require a bit code cleanup (removing the additional deleteResource). https://github.com/aubi/Payara/commit/89ea7d71b230b84ac3df6836849c7f1d1a175d02
Commits in this PR are good except the speculative allocation. I accept this PR with the modification.

pdudits:
I offer solution to making speculative allocation non-speculative.

Every thread that executes `addResource` is guaranteed that when `canGrow` returns true the pool would not exceed its maxSize.

Before check every thread would increment `inflightCreates` and decrement it after it finishes. Oh I need to fix the boundary, it should now read `<=`. I'll be back :)

svendiedrichsen:
@pdudits Very nice solution. I shall remember this one.

svendiedrichsen:
@pdudits After some time thinking about it I think you should either do the inflight increase within the loop or increase by `count`. Otherwise two threads coming in with `count` > 1 could grow the pool beyond it maxSize.

pdudits:
@svendiedrichsen welcome to the club! We're just in chat with @aubi and found the same thing and then every other workaround has some other issue

The sad truth is, that `count` is always `1`, but we want to keep this API, so I'll try another approach with a semaphore.

pdudits:
I couldn't, there were large consistency gaps and they are very hard to explain in written form (I should learn TLA+ to be able to express these)

let's say that remaining capacity is 1 and two threads want to grow:
```
T1                  T2
---------------------------
increment      increment
canGrow        canGrow
  == false        == false
```

No connection will be added even when there's capacity.

Increase/decrease by 1 has the same issue, checks for spare capacity can happen before decrease happened. We would need to decrease within read lock and that is starting to get too clever. Semaphore offers solution that is simpler to verify.

svendiedrichsen:
@pdudits After trying to think this through using a counter I couldn't myself. The seamphore is a good solution. But then we have to consider changing the MaxSize of the pool during runtime.

svendiedrichsen:
@pdudits BTW couldn't you have used the counter like you have the Semaphore? Like counting down from maxSize?

pdudits:
> But then we have to consider changing the MaxSize of the pool during runtime.

Oh, I thought I've seen maxSize as final. This will make things awkward, as semaphore doesn't like arbitrarily changing state

> couldn't you have used the counter like you have used the Semaphore

Yes it looks that semaphore is implemented via compare-and-swap operations itself. It just looks more explicit, but we'll see if I can accomodate for maxSize changing.

svendiedrichsen:
@pdudits IMHO a semaphore looks like a little bit much for the problem to solve. You are not synchonizing threads at all but only counting down.

pdudits:
@svendiedrichsen You're right, and here we've got third attempt which uses `AtomicInteger` again.

appserver/connectors/connectors-runtime/src/main/java/com/sun/enterprise/resource/pool/datastructure/RWLockDataStructure.java
```diff
@@ -102,16 +102,22 @@ public int addResource(final ResourceAllocator allocator, int count) throws Pool
...
                 throw new PoolingException(e.getMessage(), e);
             }
         }
         return numResAdded;
     }
 
     private boolean canGrow() {
-        return growPermit.tryAcquire();
+        int capacity = remainingCapacity.getAndUpdate((x) -> x > 0 ? x - 1 : 0);
+        return capacity > 0;
```
svendiedrichsen, line 116: This should be `!= 0`, shouldn't it?

svendiedrichsen:
@pdudits This should do it. Nice job.

pdudits:
Jenkins test please

aubi (approved):
The code looks good, I tried both performance and correctness testing.

PR #5514, opened 2021-12-02
FISH-26 Skip Init on Servlets Annotated with '@RegisterRestClient'

Pandrex247:
## Description
Bug fix.

If we don't skip trying to initialise Rest Client "servlets" a knock-on effect will occur during validation where a clash will be detected with the implementation endpoint (as in the endpoint that the Rest Client interface actually maps to).

### Blockers
None

### New tests
New sample added that checks that an app with both actual endpoint and Rest Client interface bundled within it can deploy successfully.

### Testing Performed
Ran new sample.
Ran MicroProfile Rest Client TCK.

### Testing Environment
Windows 11, Zulu 8u312

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java
@@ -98,6 +98,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 
 import javax.management.MBeanRegistrationException;
 import javax.management.MalformedObjectNameException;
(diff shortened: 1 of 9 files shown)


Pandrex247:
Jenkins test please

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java
```diff
@@ -189,6 +190,7 @@
 import org.apache.naming.resources.Resource;
 import org.apache.naming.resources.WARDirContext;
 import org.apache.naming.resources.WebDirContext;
+import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
```
fturizo, line 193: Not necessarily a complaint, but rather a concern: From a dependency management aspect: Does it make sense to make this core component of the server depend on one of the MicroProfile specs?

fturizo: Could a solution where a whitelist-type of file with a list of "skippable" interfaces/annotations that is parsed be a good replacement? Apologies if this doesn't make sense since I'm not that familiar with all of the server's dependency management tools

fturizo (commented):
Looks OK, just one major concern

Pandrex247:
Jenkins test please

appserver/web/web-core/src/main/java/org/apache/catalina/core/StandardContext.java
Pandrex247: Discussed with @aubi and we agree that yes pulling in the MicroProfile API dependency isn't ideal, the actual workaround for this would be quite a bit of effort. It would involve essentially as you say making a whitelist/blacklist extension point (e.g. via service loader or similar pattern) which we could use to cull servlets we don't want to initialise.

Pandrex247:
Decision made - investigate better solution.

PR #5527, opened 2021-12-07
FISH-5878 Add Certificate Alias Property Names Into Payara API as Constants

kalinchan:
## Description

Added certificate alias properties as constants to Payara API
### Blockers

N/A
### New tests

N/A
### Testing Performed

Build Payara and checked to see if `CertificateAlias.JAX_RS_CLIENT` and `CertificateAlias.REST_CLIENT` were available from API Dependency 
### Testing Environment

maven 3.6.3 openjdk version "1.8.0_302" ubuntu 20.04
## Documentation

n/a
## Notes for Reviewers

n/a


kalinchan:
jenkins test please

api/payara-api/src/main/java/fish/payara/security/client/CertificateAlias.java
```diff
@@ -0,0 +1,58 @@
...
+ *  holder.
+ */
+package fish.payara.security.client;
+
+/**
+ * Holds the constants for JAX-RS Client and MP Client Certificate Alias (FISH-230)
+ * @author Kalin Chan
+ * @since 5.2022.1
+ */
+public final class CertificateAlias {
```
breakponchito, line 47: I saw that we have the following file used for the constants related to the certificate alias implementation:
https://github.com/payara/Payara/blob/master/appserver/payara-appserver-modules/microprofile/rest-client-ssl/src/main/java/fish/payara/microprofile/jaxrs/client/ssl/PayaraConstants.java

I think that probably you can use this instead of create a new class with constants related to the certificate alias

api/payara-api/src/main/java/fish/payara/security/client/CertificateAlias.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Specifies the JAX-RS property name
+     */
+    public static final String JAX_RS_CLIENT = "fish.payara.jaxrs.client.certificate.alias";
```
breakponchito, line 52: the name of the variable is very generic, probably you can change to be more explanatory

api/payara-api/src/main/java/fish/payara/security/client/CertificateAlias.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Specifies the Rest Client property name.
+     */
+    public static final String REST_CLIENT = "fish.payara.rest.client.certificate.alias";
```
breakponchito, line 57: same here

api/payara-api/src/main/java/fish/payara/security/client/CertificateAlias.java
kalinchan: This class isn't available in the API so I don't believe that is possible?

breakponchito: ok if that is the case, I think this is good.

api/payara-api/src/main/java/fish/payara/security/client/CertificateAlias.java
kalinchan: changes have been made

breakponchito (approved):
LGTM

kalinchan:
jenkins test please

Pandrex247 (changes requested):
~The properties are wrong?~
You're missing a property: PAYARA_MP_CONFIG_CLIENT_CERTIFICATE_ALIAS 

Also we should really not have duplicate properties if we can help it:
https://github.com/payara/Payara/blob/payara-server-5.2021.9/appserver/payara-appserver-modules/microprofile/rest-client-ssl/src/main/java/fish/payara/microprofile/jaxrs/client/ssl/PayaraConstants.java
https://github.com/payara/ecosystem-rest-ssl-configuration/blob/master/src/main/java/fish/payara/ecosystem/jaxrs/client/ssl/PayaraConstantsExtension.java

kalinchan:
@Pandrex247 The properties are correct, they are the same as in the documentation. I believe the one mentioned in https://github.com/payara/Payara/blob/payara-server-5.2021.9/appserver/payara-appserver-modules/microprofile/rest-client-ssl/src/main/java/fish/payara/microprofile/jaxrs/client/ssl/PayaraConstants.java are for adding it to the Microprofile Config? And I also don't believe this file is included in the Payara API so it is not accessible when the API is added as a dependency.

api/payara-api/src/main/java/fish/payara/security/client/PayaraConstants.java
```diff
@@ -0,0 +1,58 @@
...
+ *  either the CDDL, the GPL Version 2 or to extend the choice of license to
+ *  its licensees as provided above.  However, if you add GPL Version 2 code
+ *  and therefore, elected the GPL Version 2 license, then the option applies
+ *  only if the new code is made subject to such option by the copyright
+ *  holder.
+ */
+package fish.payara.security.client;
+
+/**
+ * Holds the constants for JAX-RS Client and MP Client Certificate Alias (FISH-230)
```
Pandrex247, line 43: Only meaningful to us, not as Javadoc
```suggestion
 * Holds the constants for JAX-RS Client and MP Client Certificate Alias
```

Pandrex247:
I'm not suggesting including the PayaraConstants file in the API, I'm suggesting that you drop the duplicate property from the PayaraConstants file and have rest-client-ssl use the property from the API, rather than relying on us to keep the two in sync.

kalinchan:
I have made the requested changes. I will release a new version of ecosystem-rest-client-ssl once Payara 5.2022.1 is released. https://github.com/payara/ecosystem-rest-ssl-configuration/pull/1

Pandrex247 (approved):

kalinchan:
jenkins test please

PR #5528, opened 2021-12-07
FISH-5883 Security Exception When a Remote EJB Is Called While Authenticated Using The OIDC Security Connector

Pandrex247:
## Description
Bug Fix.

Payara expects to be able to propagate a CSIV2 IdentityToken to the remote EJB, and so needs to map the `CallerPrincipal` (wrapped within a `DistinguishedPrincipalCredential`) it has gained from the OIDC provider to a Distinguished Name, Principal Name, or Certificate Chain (or simply pass an anoynmous token).

This PR makes the server now map `DistinguishedPrincipalCredential`s as a GSSAPI principal name (Generic Security Service Application Program Interface).

### Blockers
None

### New tests
None - this is a tricky test to automate (multiple instances, applications deployed to one and not the other etc.).

### Testing Performed
Executed the reproducer attached to the issue.

### Testing Environment
Windows 11, JDK 11.0.13

## Documentation
N/A

## Notes for Reviewers
Slight variation on this solution available here, which tries to avoid the creation of a GSSUPName class [here](https://github.com/Pandrex247/Payara/commit/fb7daa039621c82bfe0ebdc0205f19ab4967cfd3)


Pandrex247:
Jenkins test please

breakponchito (approved):
LGTM

Issue #5219, opened 2021-04-27
Enhancement: Update LogFilterForInstance to work with Docker instances so that log files can be accessed through the Admin Console


natehaus:
# Description #
----------

Currently, I cannot access the logs for my Docker instances on other servers from the Admin Console. Initially, I thought this might be a bug as trying to access it brought up a `NullPointerException` in the DAS server.log. However, upon inspecting the code, I found that there was no check for Docker instances, leading to a variable set to null being incorrectly used, which caused the exception. See Context for more information on the specifics.

I am requesting that an implementation of same log viewer used for SSH and DCOM instances on the Admin console also be made available for Docker instances.

## Expected Outcome

This should work in the exact same way as it does for other types of instances. At the very least, this will require adding code for Docker instances to com.enterprise.server.logging.backend.LogFilterForInstance.downloadGivenInstanceFile(), but there may be other issues that pop up once that is fixed.

## Current Outcome

Currently, it is even more inconvenient to go and manually find the log for Docker instances than it is for other types of instances, as you have to run `sudo docker exec -it <container-name> /bin/bash` on the server to be able to access the container's file space, and as far as I can tell, the only command you can use to read the file it is `tail`. So it is certainly less than ideal in its current state. Additionally, it is certainly nice to have the logs displayed in a more human-readable format, so that would also be improved as well.

## Alternatives

Currently, the only way I can think to make this easier than the method described above to manually open the log file is to write a custom application that allows me to access the logs through a web service, but I believe this should be an integrated feature just as it is for the other instance types.

## Context

Here are the specific files where the enhancement needs to be added:
[com.enterprise.server.logging.backend.LogFilter](https://github.com/payara/Payara/blob/master/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilter.java#L425)
[com.enterprise.server.logging.backend.LogFilterForInstance](https://github.com/payara/Payara/blob/70a16f300ac07faf8e61620192db094cea812678/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java#L73)

At line 425 in the `getLogRecordsUsingQuery()` function is where the `NullPointerException` occurs: `File loggingFileExists = new File(instanceLogFile.getAbsolutePath());`

[...]

PR #5544, opened 2021-12-22
FISH-5799: access docker instances logs via docker rest

aubi:
## Description
Docker instance wasn't able to see logs from inside the docker. With this PR the log is available in the detail of the instance. Both View Log Files and View Raw Log work as well as http://localhost:4848/management/domain/view-log address.

### Blockers
Current version uses "guess" during information acquisition. This needs improvement.

### Testing Performed
Create new Docker node with this settings:
Type `DOCKER`
Node Host: `localhost`
Node Directory: `/opt/payara/appserver/glassfish/nodes`
Installation Directory: `/opt/payara/appserver`

Docker Image: `payara/server-node:5.2021.9`
Docker Port: `2375`
(Payra wrongly enters 2376, although TLS if off)
Docker Password File:  your path to password file

The password file can look like this:
`AS_ADMIN_SSHPASSWORD='payara'`

Create new instance, select the new docker node.

Start the instance

Select the newly created instance, click View Log Files and View Raw Log buttons.

### Testing Environment
Linux, Docker version 20.10.11+

## Documentation
Probably not needed, this functionality was missing.

### DIFF
--- a/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
+++ b/nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
- * Portions Copyright [2017-2018] Payara Foundation and/or affiliates
+ * Portions Copyright [2017-2021] Payara Foundation and/or affiliates
  */
 
 package com.sun.enterprise.server.logging.logviewer.backend;
@@ -53,11 +53,17 @@ import com.trilead.ssh2.SCPClient;
 import com.trilead.ssh2.SFTPv3DirectoryEntry;
 import com.trilead.ssh2.SFTPv3FileAttributes;
 
+import fish.payara.enterprise.server.logging.TrivialTarInputStream;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Logger;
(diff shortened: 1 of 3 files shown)


nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
```diff
@@ -178,14 +184,60 @@ public File downloadGivenInstanceLogFile(ServiceLocator habitat, Server targetSe
...
+            String dockerContainerId = targetServer.getDockerContainerId();
+            String nodeHost = node.getNodeHost();
+            int nodeDockerPort = Integer.valueOf(node.getDockerPort());
+            try {
+                if (dockerContainerId == null && node.getType().equals("TEMP")) {
+                    // FIXME: it would be easies, if the TEMP server had docker container id set!
+                    // temporary docker node, we need to access the original node, cut the added generated name
+                    String originalInstanceName = instanceName.substring(0, instanceName.lastIndexOf('-'));
+                    Server originalInstance = domain.getServerNamed(originalInstanceName);
+                    dockerContainerId = originalInstance.getDockerContainerId();
```
breakponchito, line 200: What if the originalInstance is null? If that is the case this will throw NullPointerException

aubi: This whole part is wrong. We need to get the information -- docker host and port -- from the instance/node, not by searching other instance. So we need to fix creation of the running docker instance+node, fill the information and use it here.

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
```diff
@@ -178,14 +184,60 @@ public File downloadGivenInstanceLogFile(ServiceLocator habitat, Server targetSe
...
+                throw new IOException("Unable to download instance log file from DCOM Instance Node", ex);
             }
 
+        } else if (node.getType().equals("DOCKER") || node.getType().equals("TEMP")) {
+            String dockerContainerId = targetServer.getDockerContainerId();
+            String nodeHost = node.getNodeHost();
+            int nodeDockerPort = Integer.valueOf(node.getDockerPort());
+            try {
+                if (dockerContainerId == null && node.getType().equals("TEMP")) {
+                    // FIXME: it would be easies, if the TEMP server had docker container id set!
```
Pandrex247, line 196: An instance on a TEMP node _should_ have it set?

From the docker-node entrypoint script, it should get set when creating the instance:
```
function createNewInstance {
...
    if [ ! -z "${DOCKER_CONTAINER_ID}" ]; then
       # Register Docker container ID to DAS
       echo "Setting Docker Container ID for instance ${PAYARA_INSTANCE_NAME}: ${DOCKER_CONTAINER_ID}"
       ASADMIN_COMMAND="${ASADMIN} -I false -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} _set-docker-container-id --instance ${PAYARA_INSTANCE_NAME} --id ${DOCKER_CONTAINER_ID}"
       echo "${ASADMIN_COMMAND}"
       ${ASADMIN_COMMAND}
    fi
```

The script even makes efforts to ensure that it at least has _something_ set right at the start to avoid failing that if check:

```
DOCKER_CONTAINER_ID="$(cat /proc/self/cgroup | grep :/docker/  | sed s/\\//\\n/g | tail -1)"
```

It also later on makes efforts to check that the container ID matches:

```
        # Check if an instance with this name is actually registered
        echo "Checking if an instance with name ${PAYARA_INSTANCE_NAME} has been registered with the DAS"
        ...

        if [ ! -z "${INSTANCE_EXISTS}" ]; then
            # Check if Docker container ID registered against the instance name is the same
            echo "Found an instance with name ${PAYARA_INSTANCE_NAME} registered to the DAS, checking if registered Docker Container ID matches this container's ID"
            ASADMIN_COMMAND="${ASADMIN} -I false -t -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} _get-docker-container-id --instance ${PAYARA_INSTANCE_NAME}"
            echo "${ASADMIN_COMMAND}"
            REGISTERED_DOCKER_CONTAINER_ID="$(${ASADMIN_COMMAND})" || true

            if [ ! -z "${REGISTERED_DOCKER_CONTAINER_ID}" ]; then
                # If they're the same, simply create the folders, otherwise create and register a new instance
                echo "Registered Docker Container ID is: ${REGISTERED_DOCKER_CONTAINER_ID}"
                if [ "${REGISTERED_DOCKER_CONTAINER_ID}" == "${DOCKER_CONTAINER_ID}" ]; then
                    echo "Docker Container IDs match, creating local instance filesystem: "
                    ASADMIN_COMMAND="${ASADMIN} -I false -T -H ${PAYARA_DAS_HOST} -p ${PAYARA_DAS_PORT} -W ${PAYARA_PASSWORD_FILE} _create-instance-filesystem --node ${PAYARA_NODE_NAME} --dockernode true ${PAYARA_INSTANCE_NAME}"
                    ${ASADMIN_COMMAND}
                else
                    echo "Docker Container IDs do not match, creating a new instance."
                    createNewInstance
                fi
            else
                echo "Could not retrieve registered Docker Container ID, creating a new instance"
                createNewInstance
            fi
        else
            createNewInstance
        fi
```

So I think the only way it couldn't be set is if you're not using the script? In which case I'd just exit out and reach for my "stop holding it wrong" bat

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
```diff
@@ -178,14 +184,60 @@ public File downloadGivenInstanceLogFile(ServiceLocator habitat, Server targetSe
...
+                    // temporary docker node, we need to access the original node, cut the added generated name
+                    String originalInstanceName = instanceName.substring(0, instanceName.lastIndexOf('-'));
```
Pandrex247, line 198: I don't think this is safe
**All** generated names have a hyphen in them, not just when name resolution to a conflict has occurred.

nucleus/core/logging/src/main/java/com/sun/enterprise/server/logging/logviewer/backend/LogFilterForInstance.java
```diff
@@ -178,14 +184,60 @@ public File downloadGivenInstanceLogFile(ServiceLocator habitat, Server targetSe
...
+                TrivialTarInputStream tarStream = new TrivialTarInputStream(downloadFile);
+                Path instanceLogPath = Paths.get(logFileDirectoryOnServer.getAbsolutePath(), logFileName);
+                //Path instanceLogPathTar = Paths.get(logFileDirectoryOnServer.getAbsolutePath() + ".tar", logFileName);
+                Files.copy(tarStream, instanceLogPath, StandardCopyOption.REPLACE_EXISTING);
+//                Files.copy(downloadFile, instanceLogPath, StandardCopyOption.REPLACE_EXISTING);
+                instanceLogFile = instanceLogPath.toFile();
+            } catch (IOException ex) {
+//            } catch (IOException | ArchiveException ex) {
+                throw new IOException("Unable to download file from docker node at " + nodeHost + ":" + nodeDockerPort + ", instance " + instanceName + ", container " + dockerContainerId, ex);
+            }
```
MarkWareham, line 236: definitely increasing complexity with this addition :wink:

Pandrex247:
This was for https://github.com/payara/Payara/issues/5219

PR #5548, opened 2022-01-04
FISH-5879 The 'stop-deployment-group --kill=true' command does not work with custom Node Directory

JamesHillyard:
## Description
A bug fix where using a custom node directory would result in `Node directory ... does not exist or is not a valid node directory` as it would always look at the default directory when executing `asadmin stop-local-instance` as part of the `asadmin stop-deployment-group` command.

### Testing Performed
Manually tested with reproducer on the ticket using: New node in custom directory, default node and New node in default directory.

### Testing Environment

## Documentation
None.

## Notes for Reviewers
None.


JamesHillyard:
Jenkins test please

Pandrex247 (approved):
LGTM - tested on Windows with two nodes and two instances

PR #5560, opened 2022-01-13
FISH-5922 Refactor usage of or Export sun.security Packages

Pandrex247:
## Description
Title.

This PR drops support for one of our JACC providers as was done in GlassFish upstream. 
We provide two JACC providers, an in-memory one and a file-based one. We previously made efforts to move away from the file-based one during a streamlining of our security modules undertaken by Arjan, and as far as I can tell it isn't used anywhere by default - you have to manually configure the server to use it.

If we wish to maintain support for it, it will require an extensive rewrite to move away from the private Sun packages.

This also removes the `EnhancerClassloader`, also as was done upstream.

### Blockers
None

### New tests
None

### Testing Performed
Compiled server on JDK 11 - no complaints of sun.security packages

### Testing Environment
Windows 11, Zulu JDK 11

## Documentation
N/A

## Notes for Reviewers
None


breakponchito (approved):
LGTM

PR #5584, opened 2022-01-26
FISH-5937 Crashes during deploy of app with HealthCheck

JamesHillyard:
## Description
Every webapp with healthcheck is periodically called every 12 seconds. This call may come when the application is not ready such as mid-deployment. This stops HealthCheck calls on apps which aren't ready.

### Blockers
None.

### New tests
None

### Testing Performed
Manually tested the warnings outputted to the server logs no longer appear with the reproducer attached.

### Testing Environment

## Documentation
None

## Notes for Reviewers
FISH-5937 ticket mentions other errors, these are caused by stopping and starting the server as the reproducer says. These are not HealthCheck related and are not related to this ticket and fix.


JamesHillyard:
Jenkins test please

aubi (approved):
LGTM

Issue #5322, opened 2021-06-30
Bug Report: cannot apply default values in @DataSourceDefinition URL via variable expansion/FISH-5676


poikilotherm:
# Description #
----------

With IQSS/dataverse#7701 we moved to Payara 5.2021.4 and hoped for the default values for our `@DataSourceDefinition`.
While testing the deployment, issues where experienced. If you do not set the variable (via system property or env var), the default value is getting applied, but data source allocation & deployment fails.

## Expected Outcome ##

When using a `@DataSourceDefinition` with
```
url = "jdbc:postgresql://${MPCONFIG=db.host:localhost}:${MPCONFIG=db.port:5432}/${ENV=DB_NAME:dvn}"
```
and none of these settings around, the allocation and deployment should work fine with the defaults given.

## Current Outcome ##

For unknown reasons, this fails with
- `RAR5103 : Error setting java bean value : [...]]]` (where "..." is the first default value used from the URL string)
- `Exception : javax.resource.ResourceException: Access denied to execute this method : setURL`
- due to illegal access at:
  ```
  java.lang.reflect.InvocationTargetException
      [...]
      com.sun.gjc.common.DataSourceObjectBuilder.constructDataSourceObject(DataSourceObjectBuilder.java:118)
   ```

This is failing with both `${MPCONFIG}` and `${ENV}`, but again: if you make the referenced keys available, they are getting used flawless.

## Steps to reproduce ## 

1. Goto https://github.com/poikilotherm/reproducer-payara-mpconfig-seturl and clone
2. Use `mvn -Parq-payara-managed verify` to execute integration test using Arquillian Payara Managed.  
   (Shoutout and kudos to @lprimak for [the handy extension](https://github.com/payara/ecosystem-arquillian-connectors/tree/master/environment-setup))
3. Logs can be found at `target/payara5/glassfish/domains/domain1/logs/server.log`
4. Play around by setting vars in `src/main/resources/system-properties.properties` and `src/main/resources/environment-variables.properties`

Note that I also set `fish.payara.log-jdbc-calls` within the DSD to a nonsense default value. This is not taken into account. No idea if this is due to the failing connection when setting the properties or if non-included #5272 (3 days after Payara 5.2021.4) is causing this to be ignored.

## Environment ##

- **Distribution**: Server Full
- **JDK Version**: 11, OpenJDK & Adopt
- **Operating System**: Linux

shub8968:
Hi, @poikilotherm,

[...]

PR #5606, opened 2022-02-08
FISH-5676 Apply Default Values in @DataSourceDefinition URL via Variable Expansion - P6

kalinchan:
## Description

Strings with multiple default values in the DataSourceDefinition URL would change the entire string to the first default found if no value was found via env properties or mp-config properties.
resolves https://github.com/payara/Payara/issues/5322
### Blockers

N/A
### New tests

Created test that uses multiple env properties with defaults
### Testing Performed

Tested reproducer
### Testing Environment

openjdk version "11.0.11" 2021-04-20 LTS
Apache Maven 3.6.3
ubuntu 20.04
## Documentation

N/A


Pandrex247 (changes requested):
See comment on `master` PR.

Pandrex247 (approved):

PR #5608, opened 2022-02-09
Increment version numbers 6.2022.1.alpha3

cubastanley:
Title


Pandrex247 (approved):
The script remains flawed in that it's not removing the SNAPSHOT from the Payara Micro glassfish-version.properties on the release commit, but otherwise LGTM

PR #5609, opened 2022-02-10
FISH-5927 Document --installdir Command and Installation Directory for Docker Nodes

JamesHillyard:
## Description
An improvement to the DockerInstallDirHelp label. It's intentional behaviour the Installation Directory has no effect for docker nodes, that is defined in the docker image itself not the server. Update the label so this is clear.

### Blockers
None

### New tests
None

### Testing Performed
Built the server, updated label is shown in the admin console for DOCKER Node Types.

### Testing Environment

## Documentation
The documentation surrounding this is already correct

## Notes for Reviewers
None


breakponchito (approved):
LGTM

appserver/admingui/cluster/src/main/resources/org/glassfish/cluster/admingui/Strings.properties
```diff
@@ -263,7 +263,7 @@ node.EditPageTitle=Edit Node
 node.EditPageTitleHelp=Edit properties for the selected node.
 node.InstallDir=Installation Directory:
 node.InstallDirHelp=The full path to the parent of the base installation directory of the Payara Server software on the host, for example, /export/payara41.
-node.DockerInstallDirHelp=The full path to the parent of the base installation directory of the Payara Server software in the Docker container. Defaults to /opt/payara/payara5
+node.DockerInstallDirHelp=The full path to the parent of the base installation directory of the Payara Server software in the Docker container. This option has no effect on docker nodes as the installation directory is determined by the docker image and not the server. Defaults to /opt/payara/payara5
```
Pandrex247, line 266: This isn't true, the option does do something.
It describes `The full path to the parent of the base installation directory of the Payara Server software in the Docker container`

I think the only thing that needs updating is that unless you're using a custom Docker image this can be left as-is

Pandrex247 (changes requested):
This isn't true

JamesHillyard:
Jenkins test please

PR #5619, opened 2022-02-18
FISH-6025: Impose strict ordering of asynchronous callbacks

pdudits:
## Description

Callbacks, especially dataAvailable and allDataRead are delivered in quick succession. Before, passing of these callbacks to application was scheduled as two independent tasks, which resulted in ordering of the callbacks to be undefined, driven by OS scheduling.

Now we always deliver `onDataAvailable` before` onAllDataRead` and that before `onError`.

Similar ordering is imposed for `WriteListener`

### Testing Performed

This resolved problems with gRPC adapter, where out-of-order processing of async events would result in abrupt closes of connections.

### Testing Environment

```
Java version: 1.8.0_302, vendor: Azul Systems, Inc., runtime: C:\Program Files\Zulu\zulu-8\jre
```


pdudits:
Jenkins test please

breakponchito (approved):
LGTM

PR #5621, opened 2022-02-21
FISH-5939 Application Redeployment Breaks virtual server When Using it as Default Module

arieki:
## Description
Whenever an application that is being used as the default web module of a virtual server gets redeployed, the virtual server will stop working and not correctly process requests sent to it.

### Blockers
N/A

### New tests
N/A

### Testing Performed
The server must be tested remotely as the error will only trigger when accessing the virtual server through its hostname. A suggested environment to test this is AWS.
It can be tested locally by changing the domain URL from "localhost" to your local hostname.
Built Payara, Start server, deploy apps, setup default web module on a virtual server, redeploy the application

### Testing Environment
Windows 11, RHEL8.4.0, JDK8, Maven 3.8.4

## Notes for Reviewers
None

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
@@ -624,6 +624,10 @@ public class WebContainer implements org.glassfish.api.container.Container, Post
             ApplicationInfo applicationInfo = (ApplicationInfo) event.hook();
             if (applicationInfo.getSource().getArchiveMetaData("commandparams", DeployCommandParameters.class).isRedeploy()) {
                 for (VirtualServer vs : getVirtualServers()) {
+                    if (ADMIN_VS.equals(vs.getName())) {
+                        continue;
+                    }
+
                     if (vs.getDefaultWebModuleID().equals(applicationInfo.getName())) {
                         try {
                             updateHost(vs.getBean());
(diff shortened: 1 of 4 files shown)


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
```diff
@@ -618,6 +620,27 @@ public void event(Event<?> event) {
...
+            if (applicationInfo.getSource().getArchiveMetaData("commandparams", DeployCommandParameters.class).isRedeploy()) {
+                for (VirtualServer vs : getVirtualServers()) {
+                    if (ADMIN_VS.equals(vs.getName())) {
+                        continue;
+                    }
+
+                    if (vs.getDefaultWebModuleID().equals(applicationInfo.getName())) {
+                        try {
+                            // since this event is asynchronous, wait application module fully loaded,
+                            Thread.sleep(500);
```
Pandrex247, line 634: Not keen on this: it kind of defeats the purpose of using the event listener - if you're reacting to an event by waiting, why not react to a different event?

If the application isn't fully loaded at this point and we need it to be, I'd suggest we need to use a different event.
There's an `APPLICATION_LOADED` or `APPLICATION_STARTED` event we could use which seems more in line with what you're trying to do here.

Pandrex247 (changes requested):
You need to update the "Portions Copyright" blurb at the top.
Needs updating from `2016-2020` to `2016-2022`.

Also additional comments

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
arieki: As our discussion, I have created a new event to notify the deployment command is finished and listen to it to resolve the missing application module while updating the host.

arieki:
jenkins test please

Pandrex247 (approved):

PR #5643, opened 2022-03-08
fix dependency in glassfish quicklook bundle build

aubi:
## Description
Fix build, non-existing org.glassfish.jsftemplating:jsftemplating-dynafaces:jar:0.1 back to com.sun.jsftemplating.


aubi:
will be fixed with other issues as a part of FISH-1522

PR #5666, opened 2022-03-21
FISH-5987 Remove TLS 1.0,1.1 SSL2 and SSL3

kalinchan:
## Description

Since TLS 1.0 and 1.1 are disabled on JDK 11, this has been removed 
### Blockers

N/A
### New tests

N/A
### Testing Performed

Built Payara and checked if options for tls 1.0 and 1.1 were there and executed `set configs.config.server-config.network-config.protocols.protocol.${protocol-name}.ssl.tls-enabled=[true/false]`
### Testing Environment

ubuntu 20.04 maven 3.6.3 openjdk 11.0.11 2021-04-20 LTS
## Documentation

[Here](https://github.com/payara/Payara-Community-Documentation/pull/299)


appserver/admingui/common/src/main/resources/shared/sslPrepare.inc
```diff
@@ -61,8 +61,6 @@ if (! #{requestScope.exist}){
       -->
     createMap(result="#{pageSession.valueMap}");
     mapPut(map="#{pageSession.valueMap}" key="ssl3Enabled" value="false");
```
Pandrex247, line 63: We should increase the scope of this to remove SSL3 too - I didn't realise it was even still there.
I'll update the Jira

Pandrex247 (changes requested):

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -184,17 +169,13 @@ public IIOPSSLSocketFactory() {
                     if (outboundSsl != null) {
                         boolean ssl2Enabled = Boolean.valueOf(outboundSsl.getSsl2Enabled());
```
kalinchan, line 170: I assume this should also cover ssl2? @Pandrex247

Pandrex247: Definitely - that's ancient 😂
Didn't realise that was even an option - it's not something you can select in the Admin Console

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IIOPSSLSocketFactory.java
```diff
@@ -240,43 +204,13 @@ private SSLInfo getDefaultSslInfo() throws Exception {
...
-            boolean ssl3Enabled, String ssl3TlsCiphers,
-            boolean tlsEnabled, boolean tlsEnabled11,
-            boolean tlsEnabled12, boolean tlsEnabled13) throws Exception {
+    private SSLInfo init(String alias, boolean tlsEnabled12, boolean tlsEnabled13) throws Exception {
 
         String protocol;
         if (tlsEnabled13) {
             protocol = TLS13;
-        } else if (tlsEnabled12) {
+        } else {
```
Pandrex247, line 212: I wonder if this should default to 1.3, 1.2, or throw an error?
Need to debug with a reproducer to understand this and why it would originally default to SSL even though that wasn't an option.

Pandrex247 (changes requested):
You've broken the SSL config page:
Configs > server-config > ORB > IIOP Listeners > SSL > SSL
Disable TLS 1.2, enable TLS 1.3, click save - error.

kalinchan:
Removal of the SSL3 Ciphers caused the SSL config page to break, therefore I left this in

Pandrex247 (changes requested):
Numerous methods and references to SSL2 and SSL3 missed in AppClientSSL, config/SSL, ServerTags, CreateSSL, CipherInfo, SSLClientConfigurator, SSLParams, SSLConfigurator, dom/Ssl, and SecureAdminCommand

Pandrex247 (changes requested):
Still missed references in help text e.g. --securityenabled help text in create-http-listener
SSLv2 and SSLv3 still available as enums in SslType

Pandrex247 (changes requested):

Pandrex247 (approved):

PR #5681, opened 2022-03-31
Fish-6136: add support for context service tag

breakponchito:
Add support for context service tag
## Description

Adding support  to use ContextService from xml tag configuration
### Testing Performed

Samples added and TCK executed 
### Testing Environment

Windows 11, Azul JDK 11
## Notes for Reviewers

Samples need to be improved because the result to execute the IntContextProvider is some cases return 0


aubi (approved):
LGTM

PR #5682, opened 2022-03-31
FISH 6061 - Upgrade Jakarta JSON Processing to 2.1.0 (P6)

arieki:
## Description

Upgrading the JSON processing to support Jakarta EE 10. 

### Blockers

N/A

### New tests

[jakartaee-10-tck-runners](https://github.com/aubi/jakartaee-10-tck-runners)

### Testing Performed

Execute one of the JSONP TCK test cases in [jsonp-tck-runner module test](https://github.com/aubi/jakartaee-10-tck-runners/tree/main/jsonp-tck) which is the JsonPluggability test. TCK run against the remote Payara Server.

### Testing Environment

Windows 11, OpenJDK11, Maven 3.6.3

## Documentation

None

## Notes for Reviewers

None


api/payara-bom/pom.xml
```diff
@@ -695,15 +695,15 @@
...
                 <artifactId>jakarta.json</artifactId>
-                <version>${jsonp.version}</version>
+                <version>${parsson.version}</version>
             </dependency>
 
             <dependency>
                 <groupId>org.glassfish</groupId>
                 <artifactId>jsonp-jaxrs</artifactId>
-                <version>${jsonp.version}</version>
+                <version>${jsonp-jaxrs.version}</version>
```
pdudits, line 706: This one is from old API. I think it should be replaced with

````
    <groupId>org.eclipse.parsson</groupId>
    <artifactId>parsson-media</artifactId>
````

For that we also need to upgrade to Jersey 3.1, that depends on same artifact (3.1.0-M5 was tagged recently).

File: appserver/pom.xml [hunk 2]
```diff
@@ -192,16 +192,16 @@
...
+                                <groupId>org.eclipse.parsson</groupId>
                                 <artifactId>jakarta.json</artifactId>
-                                <version>${jsonp.version}</version>
+                                <version>${parsson.version}</version>
                             </artifact>
                             <nonFinal>false</nonFinal>
                             <jarType>impl</jarType>
-                            <specVersion>${jsonp.version}</specVersion>
-                            <implVersion>${jsonp.version}</implVersion>
+                            <specVersion>${parsson.version}</specVersion>
```
pdudits, line 201: spec is likely `jsonp-api.version`

api/payara-bom/pom.xml
pdudits: Ah no, new jersey already requires Servlet 6.0, which we don't have, so that will not be a solution.

Rather we need to work around the dependency of `jersey-media-json-processing` (https://github.com/eclipse-ee4j/jersey/blob/3.x/media/json-processing/pom.xml)

On the place where we add this dependency, exclude jsop-jaxrs and add parsson-media dependency.

pdudits: It appears that I was too optimistic about possible replacement here -- we cannot get parsson package in until we also upgrade Jersey, which will happen later.

I then suggest just to call this property something like `transitional.jsonp-jaxrs.version` so that we know that is something that needs to change before we finish Jakarta EE 10 migration.

appserver/pom.xml
arieki: new PR requested

pdudits (approved):

PR #5694, opened 2022-04-06
FISH-5955 Support lib/ext libraries on JDK 11+

jGauravGupta:
## Description
Support lib/ext libraries on JDK 11+ 

### Testing Performed
EclipseLink sample app with Oracle DB

### Testing Environment
Windows 10, JDK 11

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
- // Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
+ // Portions Copyright [2016-2022] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.v3.server;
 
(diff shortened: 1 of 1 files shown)


pdudits:
Depends on #5649

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/server/CommonClassLoaderServiceImpl.java
```diff
@@ -146,6 +146,11 @@ private void createCommonClassLoader() {
             Collections.addAll(cpElements,
                     domainLib.listFiles(new JarFileFilter()));
         }
+        final File domainLibExt = new File(domainDir, "lib/ext/"); // NOI18N
+        if (domainLibExt.isDirectory()) {
+            Collections.addAll(cpElements,
+                    domainLibExt.listFiles(new JarFileFilter()));
+        }
```
pdudits, line 153: Since the specific functioncality we depend on depends on #5649 (amazing, a permutation, of this pr ID #5694 ! :)) This is likely not the way to go, it would create two classloader hierarchies with incompatible classes. The requirement to extend system classloader still holds, otherwise eclipselink oracle extension will not work.

If I understood right, your problem was that these classes weren't found in application under such circumstances, which is possible because 

I think the solution is to change `com.sun.enterprise.v3.server.APIClassLoaderServiceImpl#getExtensionClassLoader` instead. I believe under these circumstances we should go for system classloader as parent of API classloader, or the best approach might be to target system classloader of Felix (bundle id 0), which will even respect boot delegation properties.

jGauravGupta:
Thanks @pdudits for help, closing this PR.

PR #5707, opened 2022-04-12
FISH-5955 FISH-5948 Support for lib/ext for Java 9+

pdudits:
## Description

This puts `${domainDir{/lib/ext/*` onto system classpath (argument `-cp` to launched server process). System classloader is then used as base for both Felix OSGi runtime, as well as application's API Classloader.

It was verified that this change doesn't alter the set of classes accessible to applications, anything on system classpath (glassfish.jar and its dependencies resolved via `Class-Path` manifest attribute) are already available over API Classloader, because it allows loading much more than just API classes.

Alternative solution to #5694, includes #5649.

Classpath is only extended when system property `java.ext.dirs` is not defined, which is true on any JDK >= 9 as that would cause fatal JVM error.

### Testing Performed

* Reproducer for FISH-5946 against Oracle 12.
* Verified that cp is not altered for Java 8

### DIFF
--- a/nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
+++ b/nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
@@ -113,10 +113,11 @@ org.osgi.framework.bootdelegation=${eclipselink.bootdelegation}, \
                                   org.netbeans.lib.profiler, org.netbeans.lib.profiler.*
 
 # The OSGi R4.2 spec says boot delegation uses the boot class loader by default. We need
-# to configure it to use the framework class loader because that class loader is
-# configured with extra classes like jdk tools.jar, derby jars, etc. that must be
-# made available in GlassFish to work.
-org.osgi.framework.bundle.parent=framework
+# to configure it to use the application class loader by default so we have access to
+# classes and resources on the system class path. From those classes and resources, only those 
+# that reside in packages listed in bootdelegation will be exposed, others will not be visible by
+# OSGi bundles.
+org.osgi.framework.bundle.parent=app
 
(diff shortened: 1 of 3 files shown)


pdudits:
Jenkins test please

OndroMih (approved):
Looks good to me.

jGauravGupta:
Jenkins test please

nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
```diff
@@ -102,7 +102,7 @@ extra-system-packages=${jre-${java.specification.version}}, org.glassfish.embedd
 # Although Eclipselink imports these packages, in typical GlassFish installation,
 # Oracle JDBC driver may not be available as a bundle, so we ask user to install it in
 # java.ext.dirs and the bootdelegation helps there.
-eclipselink.bootdelegation=oracle.sql, oracle.sql.*
+eclipselink.bootdelegation=oracle.sql, oracle.sql.*,oracle.jdbc, oracle.jdbc.*
```
OndroMih, line 105: I don't think this is needed now that we set org.osgi.framework.bundle.parent=app, EclipseLink and Oracle worked for the sample app even without it. But it causes no harm so we can leave it there.

jGauravGupta:
Jenkins test please

nucleus/osgi-platforms/felix/src/main/resources/config/osgi.properties
pdudits: It was necessarry, eclipselink had trouble making connection at some different spot otherwise

PR #5721, opened 2022-04-22
FISH-6216 set core pool size in managedexecutorservice to same value as max

aubi:
## Description
The pool doesn't start any task with default core pool size = 0
Setting core to max makes the MES behaving correctly.

### Testing Performed
Fixes 3 Concurrency TCK failures.

### Testing Environment
Linux, OpenJDK


breakponchito (approved):
LGTM

PR #5726, opened 2022-04-26
FISH-6193 FISH-6160 Reimplement 'LoggerBase' as an 'org.apache.juli.logging.Log' Implementation

Pandrex247:
## Description
Reimplements the `LoggerBase` class in line with the new Tomcat JULI logging, since the `Logger` interface no longer exists.

### Blockers
None

### New tests
None

### Testing Performed
N/A

### Testing Environment
N/A

## Documentation
N/A

## Notes for Reviewers
I upped the "version" String from 1.0 to 2.0 given the rewrite.

`PropertyChangeSupport` removed since nothing listens for these events and it doesn't react to any.

Debating whether or not this class is even worth having any more and if we should just extend from `DirectJDKLog`.


breakponchito (approved):
LGTM

pdudits:
We need a reminder issue to actually have catalina use this (I think you mention ServiceLoader mechanism?)

PR #5729, opened 2022-04-27
Migrate Valves to new catalina APIs

pdudits:
## Description
Single sign on was reimplemented in a way that is compatible with current Catalina APIs.

Two tricks were employed:
* For the fields no longer passed from AuthenticatorBase we rely on the fact, that SingleSignOn is installed onto VirtualHost, whereas Authenticators are on Container level. SingleSignOn's `invoke` method is therefore invoked before any of `SigleSignOn` methods, like `register`. We can therefore keep data from current request in a threadlocal.
* Any access to `cache` fields were overriden so that we can employ relevant subclasses of `SingleSignOnEntry`.

## Notes for Reviewers

Unfortunately Git doesn't seee how AbstractSingleSignOn is rename of GlassfishSingleSignOn.


Pandrex247 (approved):
Missing copyright updates for
* Constants
* PESSOFactory
* SSOFactory
* CombinedAccessLogFormatter
* CombinedAccessLogFormatterImpl
* HASingleSignOn

PR #5739, opened 2022-05-04
Fish 6216 Fix unspecified standard contexts, move the to Remaining

aubi:
## Description
Fixes handling of standard contexts (Security, Application) when not specified (e.g. when loading from xml file). They are configured to the context given by Remaining. If Remaining is not specified, they go to propagated.

Edit: REVERTED - This change stops inheriting JNDI contexts from calling thread when creating new thread from ManagedExecutorService.

### Testing Performed
Currently it decreased number of failing tests from 16 to 11.

### Testing Environment
Linux, OpenJDK

## Notes for Reviewers
Requires current version of wip-payara Referential implementation.

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -106,10 +106,10 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
 
     // Predefined handlers for context propagation
     // TODO: replace with ConcurrentRuntime.CONTEXT_INFO_* ?
-    public static final String CONTEXT_TYPE_CLASSLOADING = "CLASSLOADING";
-    public static final String CONTEXT_TYPE_SECURITY = "SECURITY";
-    public static final String CONTEXT_TYPE_NAMING = "NAMING";
-    public static final String CONTEXT_TYPE_WORKAREA = "WORKAREA";
+    public static final String CONTEXT_TYPE_CLASSLOADING = "CLASSLOADING"; // Conccurency 3.0: N/A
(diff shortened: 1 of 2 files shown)


appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -184,35 +205,37 @@ public ContextHandle saveContext(ContextService contextService, Map<String, Stri
...
+        verifyProviders(contextClear);
+        verifyProviders(contextUnchanged);
 //                }
 //            }
 //        }
 
         ComponentInvocation currentInvocation = invocationManager.getCurrentInvocation();
-        if (currentInvocation != null && !contextUnchanged.contains(CONTEXT_TYPE_NAMING)) {
-            savedInvocation = createComponentInvocation(currentInvocation);
+//        if (currentInvocation != null && !contextUnchanged.contains(CONTEXT_TYPE_NAMING)) {
```
breakponchito, line 229: Should we need to maintain the commented code?

breakponchito (approved):
LGTM

smillidge:
Just going on the title but doesn't this defeat the point of MES?

lprimak:
I suspect this is going to break a lot of things... just a hunch though

aubi:
I have the same (bad) feeling. I expected, that our bigger tests will reveal, how much it affects them.

I wrote a mail to the cu mailing list about the expectations. The answer: "It looks like a mistake in the test case where it is accidentally testing the ‘unchanged’ setting as though it were ‘cleared’."

So I will revert this change.

lprimak:
Also not sure if you are aware of https://github.com/payara/Payara/blob/master/nucleus/common/internal-api/src/main/java/org/glassfish/internal/api/JavaEEContextUtil.java API wrapper

aubi:
@lprimak Interesting. Concurrency uses (historically) ContextSetupProviderImpl and ContextProxyInvocationHandler to handle context.
The change in the InvocationManagerImp was just to solve the empty jndi context.

PR #5747, opened 2022-05-11
FISH-6277: Jakarta EE9 shim

pdudits:
## Description

This improves transition process for major version upgrades, where some parts of the server may still depend on previous version. 
In limited version the shim will also remain in use for Microprofile APIs, which do depend on EE9 versions and keep this dependency at least until MP 7.

Compared to previous trick that required repackaging jakarta API bundles, this is external to them and is thus easier to manage.

## Important Info

The shim is likely important for anyone trying to integrate any of major upgrades, such as CDI, JAX-WS, JAXB.


breakponchito (approved):
LGTM

aubi:
Jenkins test please

PR #5765, opened 2022-05-20
FISH-6310 Reimplement J2EEInstanceListener as a Valve

Pandrex247:
## Description
Reimplements most of the functionality the `J2EEInstanceListener` class used to, namely:
* Creating a `WebComponentInvocation` for the request
* Creating a transaction around the request in the `JavaEETransactionManager` and clearing up after it has been serviced
* Setting the `AppServSecurityContext` with the `Principal` obtained from the `Request` and clearing up after it has been serviced

Potentially important reimplementation details:
* As a `Valve` instead of an `InstanceListener`, we're no longer doing all of the above for each "sub-invocation" of the request (before & after each `Filter#doFilter`, before & after each `Servlet#service`, etc.).
* I've dropped the check for the `getUnwrappedCoyoteRequest` if the `Request` is a subclass of `RequestFacade`. I didn't find any case where this would be the case (unless I was being blind), and this permission appears to be undocumented to the point I'm not sure what it was ever doing - I think it's specifically to do with `ProgrammaticLogin` which is a GlassFish feature (it's **not** the programmatic login feature of Servlets - I checked) and the `SecurityManager` is _potentially_ on its way out anyway.
* We probably need to review if we still need the `AppServSecurityContext` stuff or if there's a better way to link this in, though trying to figure that out dry without compiling and running the server probably isn't worth the effort.
* There were some fixes previously put in (by @pdudits) around checking if we were failing out to prevent us calling `InvocationManager#postInvoke(ComponentInvocation)` twice or more - I don't think it's relevant any more since we're not iterating over each `Filter` any more

### Blockers
Nothing on this PR, but obviously the server doesn't compile.

### New tests
None

### Testing Performed
None - server doesn't compile on this branch at the moment. When it's running we probably need to run it through the TCK and test the `ProgrammaticLogin` feature with the security manager enabled (potentially using the EJB over HTTP feature?)

### Testing Environment
N/A

## Documentation
N/A

## Notes for Reviewers
See above.

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
@@ -234,8 +234,6 @@ public class WebModule extends PwcWebModule implements Context {
     public WebModule(ServiceLocator services) {
         super();
         this.services = services;
-        webModuleValve = new WebModuleValve(this);
-        getPipeline().addValve(webModuleValve);
     }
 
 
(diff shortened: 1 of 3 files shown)


pdudits (approved):
LGTM.

I believe you're right about duplicate prevention -- we were double-popping the invocation stack when there was an error during handling of error. But now that there's only single thing in the stack, we're likely ok.

PR #5772, opened 2022-05-26
FISH-6309 InstanceManager Implementation

Pandrex247:
## Description
Adds an implementation of the Tomcat Catalina `InstanceManager` to use instead of the `DefaultInstanceManager`, the intent being that we want to retain integration with the Payara `InjectionManager` and `InvocationManager`.

### Blockers
None.

### New tests
None

### Testing Performed
Branch doesn't compile so none.

### Testing Environment
N/A

## Documentation
N/A

## Notes for Reviewers
Without the ability to debug it I'm a bit unsure of the behaviour to take with the `newInstance(Object object)` method - when the object has already been instantiated outside of the instance manager. I took inspiration from how the `InjectionManagerimpl#createManagedObject(Object object)` method works when it's not creating a managed bean via `ManagedBeanManagerImpl` and instead creating one via simple `clazz.getConstructor().newInstance()`.

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainer.java
@@ -724,85 +724,6 @@ public class WebContainer implements org.glassfish.api.container.Container, Post
     }
 
     /**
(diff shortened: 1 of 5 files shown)


aubi (approved):
LGTM

PR #5788, opened 2022-06-03
add exclusion of testng from concurrent-api

aubi:
## Description
Remove testng dependency from concurrent-api in Payara. It was obviously let in compile scope by mistake.


aubi:
Jenkins test please

pdudits (approved):

PR #5807, opened 2022-06-14
FISH-6360 Payara Does Not Status Code Set In 'WebApplicationException'

Pandrex247:
## Description
Fix for Core Profile TCK test `CustomJsonbSerializationIT.testMissingClientSerializationWithCDIProvider`.
This test throws a `WebApplicationException` with a response status code of 400 (Bad Request) if it fails to deserialize, and the test expects to receive said status code.

There's likely some underlying bugs happening within Jersey / Yasson itself as there are numerous calls to `ResponseWriter#getResponseContext` that fail and throw container exceptions as it tries to map the error (doesn't happen on Jersey 3.0.x), but given that Jersey is deferring the processing of it to the container (Payara) I believe this is the actual fix. There's a couple of other places where the error code is explicitly set to HTTP 500 (e.g. `StandardWrapperValve#exception`) which was done in GlassFish a while ago all as a part of a fix to the same issue, but this is the end point where the final response code is set.

Also includes Yasson RC2 since we should be using it :)

### Blockers
None

### New tests
None - tested via Core Profile TCK.

### Testing Performed
Ran Core Profile TCK - passed.

### Testing Environment
Windows 10, JDK 11.

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/valves/ErrorReportValve.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/valves/ErrorReportValve.java
@@ -55,16 +55,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2019-2021] Payara Foundation and/or affiliates
+// Portions Copyright [2019-2022] Payara Foundation and/or affiliates
 
 package org.apache.catalina.valves;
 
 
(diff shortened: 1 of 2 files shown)


Pandrex247:
Jenkins test please

breakponchito (commented):
LGTM

PR #5814, opened 2022-06-17
FISH-5722 Correct DTD

Pandrex247:
Just popped into my head a couple of hours after reviewing that I didn't check this.


Pandrex247:
Jenkins test please

aubi (approved):
Thanks for noticing it!

PR #5829, opened 2022-06-29
FISH-6355: disabling trace method from http OPTIONS call

breakponchito:
Disabling TRACE method from http OPTIONS call
## Description

This is a fix for a bug that returns the TRACE method available from an http OPTIONS call when the following property is set to false:

- configs.config.server-config.network-config.protocols.protocol.http-listener-{listener number}.http.trace-enabled


breakponchito:
Jenkins test please

PR #5832, opened 2022-06-30
FISH-6352 File locks prevent undeployment on Windows

jGauravGupta:
## Description
Undeployment of an application takes a long time on Payara6 on Windows.

## Fix
This issue is caused by the NPE exception thrown in the stop operation of WebappClassloader which results in an uncleaned classloader and locking in Windows.

NPE exception is caused during cleaning the cache of BeanELResolver.properties field via Reflection API which is now changed from static to non-static, and that already fixes the cache issue as per the following ticket details.

[BeanELResolver needs cleanup method · Issue #171 · jakartaee/expression-language](https://github.com/jakartaee/expression-language/issues/171) 
[Fix #171. Make cache per instance rather than per class · jakartaee/expression-language@3ec9258](https://github.com/jakartaee/expression-language/commit/3ec9258dd20ebedb76e1a72a63c69bf71b14775b)

### Testing Performed
Java EE 7 Samples

### DIFF
--- a/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
+++ b/appserver/web/war-util/src/main/java/org/glassfish/web/loader/WebappClassLoader.java
@@ -1999,10 +1999,6 @@ public class WebappClassLoader
             return;
         }
 
-        // START GlassFish Issue 587
-        purgeELBeanClasses();
-        // END GlassFish Issue 587
-
         /*
          * Clearing references should be done before setting started to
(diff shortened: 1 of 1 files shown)


jGauravGupta:
jenkins test please

lprimak:
Won’t this restore the class loader leak this code was trying to prevent?

MarkWareham (approved):

jGauravGupta:
jenkins test please

PR #5894, opened 2022-08-11
FISH-6399 Implement New Payara 6 Version Text in Header of Payara Server 6

jGauravGupta:
This PR adds Payara Major version to the top tool bar in the Web Admin console.
image

### DIFF
--- a/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/GuiUtil.java
+++ b/appserver/admingui/common/src/main/java/org/glassfish/admingui/common/util/GuiUtil.java
@@ -38,7 +38,7 @@
  * holder.
  */
 
-// Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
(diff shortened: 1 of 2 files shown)


Pandrex247 (commented):
Is there a reason we're only listing the major version?
I could find this info out by simply clicking on "About..."

Pandrex247:
I query the point of this. Personally I think this is both redundant and ugly 😂

diturner:
Ideally if the UI had a footer this kind of info would sit in there, unfortunately it doesn't so this is the compromise. 

Knowing how tricky this can be to edit I feel adding a whole footer for this would probably be uncalled for.

diturner (approved):

jGauravGupta:
Hi @Pandrex247 , Initially discussion started with the updated Payara 6 logo (shows 6 version number) but later on agreed to add the version number to the top of the admin console page and keep the logo as Payara 5.

jGauravGupta:
jenkins test please

PR #5899, opened 2022-08-18
FISH-6415 Unexpected Error When Starting Instance Hosted On Remote SSH Node on Windows OS Via Cygwin

Pandrex247:
## Description
Recreation of https://github.com/payara/Payara/pull/5895

An instance hosted on a remote SSH node in a Windows environment would quickly return saying that the command had failed, but when refreshing the instance page the instance would show as running.

This seems to come about due to a hard-coded 3 second wait, at which point the SSH connection is terminated. This PR ups the wait to the default of 120 seconds, which can be overidden using a new system property `fish.payara.node.ssh.timeout`. This means:

* Commands which take less than 120 (or user-defined x) seconds to complete should still return within that time - the SSH connection won't just sit there for 120  (or user-defined x) seconds
* Commands which take longer than 120 (or user-defined x) seconds will show as failed, though if the command timeout itself is configured to be longer the command will still continue on the remote instance until it completes or times out.

### Blockers
None

### New tests
A set of three unit tests to check:
* Default is used with no property set
* Custom value is used with custom value set
* Default is used if invalid custom value set

#### Test Default
Spun up two Windows VMs in AWS
Installed Cygwin and configured SSH between the two VMs
Installed server to both
Started Das on VM 1 and added the following JVM option so it doesn't just use "localhost" as the hostname: `-Dcom.sun.aas.hostName=$IP_ADDRESS`
Restarted Das
Created Remote SSH Node on VM2 from DAS (using password auth)
Created an instance on the remote node using DAS
Started instance on remote node using DAS
Instance started without timeout

#### Test Custom
Spun up two Windows VMs in AWS
Installed Cygwin and configured SSH between the two VMs
Installed server to both
Started Das on VM 1 and added the following JVM option so it doesn't just use "localhost" as the hostname: `-Dcom.sun.aas.hostName=$IP_ADDRESS`
Restarted Das
Created Remote SSH Node on VM2 from DAS (using password auth)
Created an instance on the remote node using DAS
Set system property in server-config to 2 seconds
Started instance on remote node using DAS
Command timed out and reported failure
Instance continued to start

#### Test Invalid Custom
Spun up two Windows VMs in AWS
Installed Cygwin and configured SSH between the two VMs
Installed server to both
Started Das on VM 1 and added the following JVM option so it doesn't just use "localhost" as the hostname: `-Dcom.sun.aas.hostName=$IP_ADDRESS`
Restarted Das
Created Remote SSH Node on VM2 from DAS (using password auth)
Created an instance on the remote node using DAS
Set system property in server-config to "asdsa" seconds
Started instance on remote node using DAS
Instance started without timeout
Server log reported warning of invalid property value being ignored

### Testing Environment
Windows 10, JDK 8

## Documentation
https://github.com/payara/Payara-Documentation/pull/71

## Notes for Reviewers
None

### DIFF
--- a/nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/SSHLauncher.java
+++ b/nucleus/cluster/ssh/src/main/java/org/glassfish/cluster/ssh/launcher/SSHLauncher.java
@@ -1112,9 +1112,12 @@ public class SSHLauncher {
     private int getTimeout() {
         int timeout = DEFAULT_TIMEOUT_MSEC;
         try {
-            timeout = Integer.parseInt(System.getProperty(TIMEOUT_PROPERTY));
+            String timeoutPropertyValue = System.getProperty(TIMEOUT_PROPERTY);
+            if (StringUtils.ok(timeoutPropertyValue)) {
+                timeout = Integer.parseInt(timeoutPropertyValue);
+            }
         } catch (NumberFormatException numberFormatException) {
-            logger.log(Level.INFO, "Value of {0} does not appear to be a valid integer, defaulting to {1}ms: {2}",
+            logger.log(Level.WARNING, "Value of {0} does not appear to be a valid integer, defaulting to {1}ms: {2}",
                     new Object[]{TIMEOUT_PROPERTY, DEFAULT_TIMEOUT_MSEC, numberFormatException});
         }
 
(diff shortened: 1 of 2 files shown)


Pandrex247:
Jenkins test please

breakponchito (approved):
LGTM

PR #5926, opened 2022-09-06
SERV-847 Use GitHub Forms

JamesHillyard:
GitHub has recently introduced the concept of GitHub Issue Forms [(More details here)](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms). It providing a form to fill out, instead of a markdown template which doesn't have to be followed.

I have mostly copied the wording from our current template, but now would be a good time to make some changes to it if you notice anything

**Example repository to test is here:** [GitHub - JamesHillyard/IssueExample](https://github.com/JamesHillyard/IssueExample)


kalinchan (commented):
Perhaps add an input field for the Payara Version and maybe a note somewhere for them to test the bug on the latest version of Payara first. 
Other than that, LGTM

fturizo (approved):
As previously agreed, PR looks good.

Issue #5679, opened 2022-03-29
Bug Report:  Microprofile Interceptors not invoked if the EJB is a @Stateless bean


ismailsalloo:
# Description #
----------

We've recently upgraded our Payara Edition which now includes Microprofile 4.1 but our method that is annotated with `@CircuitBreaker` is not intercepted anymore. Our class is annotated with `@Stateless` and if we change this annotation to for example `@RequestScoped`, the interceptor kicks in as expected

## Expected Outcome ##

`@CircuitBreaker` Interceptor is invoked.

## Current Outcome ##

`@CircuitBreaker` Interceptor is ignored

## Steps to reproduce ## 

Apply `@Stateless` annotation to any bean that has a Microprofile FT interceptor and test if the interceptor still works

## Environment ##

- **Distribution**: payara-server-5.2021.5
- **JDK Version**: 11
- **Operating System**: Linux

shub8968:
Hi @ismailsalloo,

Would you be able to provide us with a sample reproducer app?

Shubham

ismailsalloo:
@shub8968 I've created a sample application here -> https://github.com/ismailsalloo/payara-sample

shub8968:
Hi @ismailsalloo,

I am unable to verify it. I tried running the Test Class and both the test got passed.

ismailsalloo:
> Hi @ismailsalloo,
> 
> I am unable to verify it. I tried running the Test Class and both the test got passed.

@shub8968 I've updated the tests to take out the expected exception. You will now notice the failure. Sorry about that

shub8968:
Hi @ismailsalloo,

I am able to reproduce this issue. I have raised an internal issue with JIRA `FISH-6238`. Thanks. It may take a while to fix the issue. You are free to submit a PR to fix this issue.

PR #5933, opened 2022-09-09
FISH-6238 Microprofile Interceptors [@Fallback@/CircuitBreaker] are not getting invoked if the EJB is a @Stateless Bean

arieki:
## Description

This is a fix for https://github.com/payara/Payara/issues/5679 issue.
EJB life-cycle failed to recognize programmatic interceptor regarding fault-tolerance interceptor that is using around invoke interception type

### Blockers

N/A

### New tests

None

### Testing Performed

+ Build payara server with build extras profile `-PBuildExtras`
+ Checkout this reproducer https://github.com/ismailsalloo/payara-sample
+ Change `fish.payara.extras:payara-embedded-all` payara-sample's dependency version to fixed payara which is `5.2022.4-SNAPSHOT`
example:
```
<dependency>
  <groupId>fish.payara.extras</groupId>
  <artifactId>payara-embedded-all</artifactId>
  <version>5.2022.4-SNAPSHOT</version>
  <scope>test</scope>
</dependency>
```
+ Run `mvn clean test` 

### Testing Environment

Windows 11, OpenJDK1.8, Maven 3.8.4

## Documentation

None

## Notes for Reviewers

None

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/cdi/FaultToleranceInterceptor.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/cdi/FaultToleranceInterceptor.java
@@ -51,6 +51,7 @@ import javax.enterprise.context.control.RequestContextController;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.inject.Inject;
+import javax.interceptor.AroundInvoke;
 import javax.interceptor.InvocationContext;
 import java.io.Serializable;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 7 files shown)


arieki:
jenkins test please

arieki:
jenkins test please

pdudits (changes requested):
This doesn't fix invocation of programmatic interceptors on EJBs.

It only fixes invocation of programmatic interceptors that happen to have method called `intercept(InvocationContext)`, and their `Interceptor.intercept` definition calls this method without further logic.

`FaultoreanceExtension.ProgrammaticInterceptor` happens to be such.

We instead should be able to define descriptor is a way, that `Interceptor.intercept(type, instance, ctx` would be called rather than looking for an annotated method.

arieki:
jenkins test please

arieki:
> This doesn't fix invocation of programmatic interceptors on EJBs.
> 
> It only fixes invocation of programmatic interceptors that happen to have method called `intercept(InvocationContext)`, and their `Interceptor.intercept` definition calls this method without further logic.
> 
> `FaultoreanceExtension.ProgrammaticInterceptor` happens to be such.
> 
> We instead should be able to define descriptor is a way, that `Interceptor.intercept(type, instance, ctx` would be called rather than looking for an annotated method.

As discussed, I am using `aroundInvoke` annotation as a workaround solution since it needs deeper analysis on how to combine CDI interceptor that implemented programmatically with EJB interceptor

arieki:
jenkins test please

pdudits (approved):

arieki:
jenkins test please

arieki:
jenkins test please

Issue #5901, opened 2022-08-18
Commands in Postboot File Fail / FISH-6501


elcodedocle:
# Description #
----------

A boot script (pre or post) containing these lines line does not fail on 5.2022.2 but it does on 5.2022.3:
```
deploy ./clusterjsp.war
set-monitoring-configuration --dynamic=true --monitoringserviceenabled=true --amxenabled=true --jmxlogenabled=false --jmxlogfrequency=15 --jmxlogfrequencyunit=SECONDS --target=server-config
```

## Expected Outcome ##

set-monitoring-configuration and deploy commands executed successfuly

## Current Outcome ##

We get the following stacktrace for each command instead:

```
[#|2022-08-18T14:15:20.466+0000|WARNING|Payara 5.2022.3|fish.payara.admin.monitor.cli.SetMonitoringConfiguration|_ThreadID=38;_ThreadName=RunLevelControllerThread-1660832119061;_TimeMillis=1660832120466;_LevelValue=900;|
  Failed to execute the command|#]

[#|2022-08-18T14:15:20.473+0000|WARNING|Payara 5.2022.3|fish.payara.boot.runtime.BootCommand|_ThreadID=38;_ThreadName=RunLevelControllerThread-1660832119061;_TimeMillis=1660832120473;_LevelValue=900;|
  Boot Command set-monitoring-configuration failed PlainTextActionReporterFAILUREA MultiException has 3 exceptions.  They are:
1. java.lang.IllegalStateException: Service fish.payara.jmx.monitoring.JMXMonitoringService was started at level 9 but it has a run level of 10.  The full descriptor is SystemDescriptor(
        implementation=fish.payara.jmx.monitoring.JMXMonitoringService
        name=payara-monitoring
        contracts={fish.payara.jmx.monitoring.JMXMonitoringService}
        scope=org.glassfish.hk2.runlevel.RunLevel
        qualifiers={}
        descriptorType=CLASS
        descriptorVisibility=NORMAL
        metadata=runLevelValue={10},Bundle-SymbolicName={fish.payara.server.internal.payara-appserver-modules.jmx-monitoring},Bundle-Version={5.2022.3}
        rank=0
        loader=OsgiPopulatorPostProcessor.HK2Loader(OSGiModuleImpl:: Bundle = [fish.payara.server.internal.payara-appserver-modules.jmx-monitoring [280]], State = [READY],1039077107)
        proxiable=null
        proxyForSameScope=null
        analysisName=null
        id=1465
[...]

PR #5936, opened 2022-09-12
FISH-6501 Fix BootCommandService Runlevel

aubi:
## Description
Fixes the issue with using DEPLOY command in postbootcommand file.

Set the BootCommandService to 10. Return HazelcastCore and ClusteredStore back to level 10 as they don't need to be started earlier. And add an explicit dependency from ApplicationLoaderService for cases, that postbootcommanfile contains setup for already deployed apps.
Add check during application loading service, if it was already loaded via postbootcommandfile.

### Testing Performed
You need some JakartaEE 9 app.

Create file postboot.asadmin with content:
`deploy /path/to/yourapp.war`

Run this command from Payara:
`/asadmin start-domain --postbootcommandfile /path/to/postboot.asadmin`

Expected output: deployed app (it asks for password when opened).

### Testing Environment
OpenJDK, Linux

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/bootstrap/BootCommandService.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/bootstrap/BootCommandService.java
@@ -55,7 +55,7 @@ import java.util.logging.Logger;
 import static java.util.logging.Level.SEVERE;
 
 @Service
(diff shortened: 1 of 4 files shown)


aubi:
Jenkins test please

JamesHillyard:
Fixes https://github.com/payara/Payara/issues/5901

PR #5945, opened 2022-09-20
FISH-6501 Fix App Loading After Reboot

aubi:
## Description
The previous changes FISH-6501 introduced a problem with app loading during restart. This fix corrects the way, how is already loaded app checked.

An application can be loaded via postbootcommand file, e.g. ApplicationLoaderService will try to load this application later again. The former test was using isLoaded flag, but it is set for the application during server reboot. This new fix uses isRunning.

### Testing Performed
Download the jolokia war file from the ticket. The test can be done with clusterjsp.war as well.

Create file postboot.asadmin with content:
`deploy /path/to/jolokia-war-1.7.1.war`

Run this command from Payara:
`/asadmin start-domain --postbootcommandfile /path/to/postboot.asadmin`

Expected output: deployed app (it asks for password when opened).

Stop the server and start again without parameters.

Expected output: app still works.

It is possible to retry the same with CDIReproducer (simple app with CDI) and MyEar-1.0-SNAPSHOT.ear (ear consisting from two war files), which I added to the Jira ticket.

### Testing Environment
OpenJDK, Linux


aubi:
Jenkins test please

pdudits (approved):

PR #5952, opened 2022-09-21
FISH-5827 Stuck Thread count as MicroProfile Metric Gauge

jGauravGupta:
## Description
This PR adds HealthCheck stats (Stuck thread count) to MicroProfile Metrics.

## Important Info
`StuckThreadsHealthCheck` required by `MetricsServiceImpl`.
`MetricsServiceImpl` required by `FaultToleranceServiceImpl` and `MicroProfileMetricsChecker`.
All of these services started at `@RunLevel(StartupRunLevel.VAL)`.
The enable status of HealthCheck can not be checked during registration of Metrics metadata which is by default registered and during response writer operation, health check status is checked.

### Testing Performed
````
asadmin start-domain
asadmin set-healthcheck-service-configuration --serviceName=stuck-thread --enabled=true
asadmin restart-domain
````
Open `http://localhost:8080/metrics/vendor` URL which so=houl contains the following result:
````
# TYPE vendor_system_cpu_load gauge
# HELP vendor_system_cpu_load Display the "recent cpu usage" for the whole system. This value is a double in the [0.0,1.0] interval. A value of 0.0 means that all CPUs were idle during the recent period of time observed, while a value of 1.0 means that all CPUs were actively running 100% of the time during the recent period being observed. All values between 0.0 and 1.0 are possible depending of the activities going on in the system. If the system recent cpu usage is not available, the method returns a negative value.
vendor_system_cpu_load 0.2421606522975489
# TYPE vendor_thread_stuck_count gauge
# HELP vendor_thread_stuck_count Displays the stuck thread count which is blocked, and can't return to the threadpool for a certain amount of time.
vendor_thread_stuck_count 1
````
Now disable health check stuck-thread:
````
asadmin set-healthcheck-service-configuration --serviceName=stuck-thread --enabled=false
asadmin restart-domain
````
Open `http://localhost:8080/metrics/vendor` URL which so=houl contains the following result:
````
# TYPE vendor_system_cpu_load gauge
# HELP vendor_system_cpu_load Display the "recent cpu usage" for the whole system. This value is a double in the [0.0,1.0] interval. A value of 0.0 means that all CPUs were idle during the recent period of time observed, while a value of 1.0 means that all CPUs were actively running 100% of the time during the recent period being observed. All values between 0.0 and 1.0 are possible depending of the activities going on in the system. If the system recent cpu usage is not available, the method returns a negative value.
vendor_system_cpu_load 0.2421606522975489
````

### Testing Environment
Windows 11, JDK 11.0.16

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MetricsMetadataHelper.java
+++ b/appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MetricsMetadataHelper.java
@@ -41,6 +41,7 @@ package fish.payara.microprofile.metrics.jmx;
 
 import fish.payara.microprofile.metrics.healthcheck.HealthCheckCounter;
 import fish.payara.microprofile.metrics.healthcheck.HealthCheckGauge;
+import fish.payara.microprofile.metrics.healthcheck.ServiceExpression;
 import fish.payara.nucleus.healthcheck.HealthCheckStatsProvider;
 import java.util.ArrayList;
 import java.util.List;
@@ -64,7 +65,6 @@ import org.eclipse.microprofile.metrics.MetricUnits;
 import org.eclipse.microprofile.metrics.Tag;
 import org.glassfish.api.admin.ServerEnvironment;
 import org.glassfish.hk2.api.ServiceLocator;
-import org.glassfish.internal.api.Globals;
 import org.jvnet.hk2.annotations.Service;
 
 @Service
@@ -130,14 +130,15 @@ public class MetricsMetadataHelper {
                     }
                     metricRegistry.register(beanMetadata, type, tags.toArray(new Tag[tags.size()]));
                 } else {
-                    HealthCheckStatsProvider healthCheck = habitat.getService(HealthCheckStatsProvider.class, beanMetadata.getService());
+                    ServiceExpression expression = new ServiceExpression(beanMetadata.getService());
+                    HealthCheckStatsProvider healthCheck = habitat.getService(HealthCheckStatsProvider.class, expression.getServiceId());
                     if (healthCheck != null) {
                         switch (beanMetadata.getTypeRaw()) {
                             case COUNTER:
-                                type = new HealthCheckCounter(healthCheck);
+                                type = new HealthCheckCounter(healthCheck, expression);
                                 break;
                             case GAUGE:
-                                type = new HealthCheckGauge(healthCheck);
+                                type = new HealthCheckGauge(healthCheck, expression);
                                 break;
                             default:
                                 throw new IllegalStateException("Unsupported type : " + beanMetadata);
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: appserver/payara-appserver-modules/microprofile/metrics/src/main/java/fish/payara/microprofile/metrics/jmx/MetricsMetadataHelper.java [hunk 1]
```diff
@@ -108,18 +112,33 @@ public List<MBeanMetadata> registerMetadata(MetricRegistry metricRegistry,
...
+                            break;
+                        case GAUGE:
+                            type = new MBeanGuageImpl(mBeanExpression);
+                            break;
+                        default:
+                            throw new IllegalStateException("Unsupported type : " + beanMetadata);
+                    }
+                    metricRegistry.register(beanMetadata, type, tags.toArray(new Tag[tags.size()]));
+                } else {
+                    HealthCheckStatsProvider healthCheck = Globals.getDefaultHabitat().getService(HealthCheckStatsProvider.class, beanMetadata.getService());
```
pdudits, line 129: I'd recommend `@Inject`ing a service locator instead, `Globals` lookup should be left for places when we are outside of HK2 control, but this one is a `@Service` already. We're overusing this pattern a lot and then we get random explosions on startup ;)

What is the behavior in error cases? What happens when invalid service name is given?

pdudits (approved):

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

Pandrex247 (commented):
Is there a reason it seems to always say there is a stuck thread?

jGauravGupta:
jenkins test please

jGauravGupta:
Changing the base branch to Payara6-tck as JavaEE8 Samples requires CDI 4.0, see https://github.com/payara/patched-src-javaee8-samples/commit/5956e8413fe4dbfc05e31b55fb954d53bea78958#diff-d26e10d72530f9acdf8b358d1aa078c602ed4aa38dea26ca966117cb1857e3a2

jGauravGupta:
jenkins test please

jGauravGupta:
Hi @Pandrex247,

At the start of the request to `http://localhost:8080/metrics/vendor`, the Stuck thread count is incremented by 1, and end of the request decremented as the metric writer processing is done in between it always shows at least 1.

` (org.glassfish.grizzly.threadpool.DefaultWorkerThread) Thread[http-thread-pool::http-listener-1(4),5,main]`

Pandrex247:
I don't understand.
If I turn on the log notifier there are no listed stuck threads - I don't see why a call to this endpoint has a discrepancy, it seems like a bug

Pandrex247 (changes requested):
Yeah I think this is a bug.
Setting the stuck thread threshold to 8 minutes and poking the endpoint still has it show 1 stuck thread.

Pandrex247:
We're closing the Payara6-tck branch, changing base.

jGauravGupta:
jenkins test please

jGauravGupta:
jenkins test please

Pandrex247 (approved):

PR #5966, opened 2022-09-28
FISH-6568 Jakarta Persistence Descriptor Not Parsed Correctly due to Oracle servers failure

nicolasduminil:
# FISH-6568: Jakarta Persistence Descriptor Not Parsed Correctly due to Oracle servers failure

## Description
This is a reproducer and a work-around the FISH-6568 issue concerning the indiscriminate unavailability of the web servers hosting XML schemas

### Blockers
In order to experience the reproducer and the work-around, a local Docker daemon is required

### New tests
Deploys a test WAR exposing a simple JAX-RS API on a Payara Micro 5 server. This simple API uses JPA to persist a test entity into an Oracle 11g XE database.

In order to guarantee predictability, the sample is using`testcontainers`to instantiate and run two Docker containers, as follows:

- a Docker container running an Oracle 11g XE database image;
- a Docker container running a Payara Micro 5.2022.3-JDK11 image.

An integration test is ran in the`verify`phase of Maven by the`failsafe`plugin. It allows to reproduce the issue, when present.

### Testing Performed
The integration test checks whether:

- the two Docker containers are up and running
- the deployment of teh WAR archive succeeds without raising SAX parser exceptions
- the JPA operation are performed as expected.

### Testing Environment
Ubuntu 20.04 LTS, Java 11, Maven 3.6.0

## Documentation
README.md file in the project

## Notes for Reviewers
Given the indiscriminate nature of the issue, its reproducibility isn't guarantee. During a two days testing period, I only succeeded to reproduce it once, for several minutes.

In order to run the sample, execute the command `mvn clean install`. This will execute the integration test which will instantiate the two containers and will deploy the test WAR on the Payara Micro instance. Then, the JAX-RS API will be invoked by the test class and will persist some dummy data into the Oracle database. A database schema is automatically created upon the Docker container startup. The test table is created automatically by the JPA `schema-generation`feature.

Once that you've succeeded to reproduce the issue, in order to work-around it, you need to run `mvn -Pfish6568 clean install`. Using this profile will allow you to take advantage of the `xml-maven-plugin` which leverages the OASIS XML catalog concept such that to solve the reference to the `persitence_3_0.xsd` resource in `persistence.xml` by the local file `src/main/resources/schema/persistence_3_0.xsd`.


Pandrex247 (changes requested):
This appears to break the build.

```
project reproducer: Fatal error compiling: invalid flag: --release
```

breakponchito:
Jenkins test please

PR #5971, opened 2022-10-03
FISH-6371 Adjust for Removal of 'web-naming' Module

Pandrex247:
## Description

Since what was previously the `web-naming` dependency has been rolled back into the Catalina dependency itself (pulled in as a `web-core`), we have to adjust the dependencies to account for this. The only module that uses `web-naming` was `war-util`, so this is the only module that needs reworking.

This also drops the l10n for the `web-naming` module, as it wasn't removed previously.

### Blockers
None

### New tests
None

### Testing Performed
N/A - Branch does not compile

### Testing Environment
N/A

## Documentation
N/A

## Notes for Reviewers
The `web-naming-l10n` module was removed since upon review of the unaltered _Payara6_ branch none of the string replacements appear to actually be used - it's dead code even without this rework.

The removal of `web-naming` actually only affects 2 classes within this `war-util` module: `WebappClassLoader` and `WarHandler`. None of the other classes in this module have any direct connection to Catalina - they are predominantly "util" classes (e.g. `LogFacade`, `ServletContainerInitializerUtil`).

I elected to only move these "problem" classes rather than doing a module restructure because upon review I'm not actually entirely sure why they're in that module to begin with and maintaining a "util" module isn't necessarily bad idea:
* `WarHandler` is a service - it's never directly referenced by anything. Given that the `war-util` module is a part of the same package as `web-core` and `web-glue` (as in `appserver/packager` package, not java package), there's nothing stopping me from therefore shifting it around
* `WebappClassLoader` is debatably a part of `web-core` or `web-glue` since it's directly related to being a classloader **for** it. 
  * The only classes which use it are as follows:
    * `JdbcLeakPrevention` (within `war-util`) - this class is actually directly linked to it and only used by it
      * When I say "directly" linked, I mean it's only used by `WebappClassLoader` and only in an extremely hacky way
    * `VirtualServer` (within `web-glue`).
      * So `WebappClassLoader` needs to exist in `web-core`, in `web-glue`, or for the classes to be reworked such that `WebappClassLoader` doesn't rely on Catalina APIs or `VirtualServer` doesn't rely on this classloader.
    * `WebContainer` (within `web-glue`)
      * So `WebappClassLoader` needs to exist in `web-core`, in `web-glue`, or for the classes to be reworked such that `WebappClassLoader` doesn't rely on Catalina APIs or `WebContainer` doesn't rely on this classloader.
    * `ACLSingletonProvider` (within `weld-integration`)
      * So `WebappClassLoader` needs to exist in `web-core`, in `web-glue`, in `weld-integration` or for the classes to be reworked such that `WebappClassLoader` doesn't rely on Catalina APIs or `ACLSingletonProvider` doesn't rely on this classloader.
  * I can see an argument for it belonging in `web-core` rather than `web-glue`, since it's actually a replacement classloader - it's a fork of the `org.apache.catalina.loader.WebappClassLoader`. 
    * Within Catalina there's actually now a `WebappClassLoaderBase` class, with two implementations (`WebappClassLoader` and `ParallelWebappClassLoader`) - **A terrifying/fun thought is that we may want to actually write an implementation based on this class rather than forking!**
    * Having said that though a number of our other "replacement" classes are in `web-glue`, so we may want to have a discussion of the bigger picture e.g. don't have a `web-core` module, or move a chunk of `web-glue` into `web-core` etc.
  * I've left the `WarHandler` package as-it-is, though I can also see an argument for changing it to align with the new module (I'm just not sure where!)
    * It could go in `com.sun.enterprise.web.deploy` - so dropping the `glassfish` bit
    * It could go in `org.glassfish.web.deployment` - so shifting to the "newer" namespace
    * Somewhere else - it's a fairly standalone class, it just needs to exist _somewhere_
  * I shifted `JdbcLeakPrevention` with the `WebappClassLoader` (see above on how it's used).
    * This is actually another class that's forked from Apache, and since the classe sare almost identical I've elected to use it albeit with a fallback until tested that the Catalina class can actually be found.
  * I brought back to life the `JarFileResourcesProvider` - I don't think this is actually an Apache class (despite the old `org.apache` package name)? It used GlassFish copyright without any mention of Payara
    * I've brought this back in the interim, but we probably do need to look at dropping/rewriting this anyway. It looks to be a part of the now removed "many `DirContext` implementations" way of working within Apache, which has been replaced with `WebResourceRoot`. So IMO bring it back to for now, and mark for removal (hence deprecated).


breakponchito:
Jenkins test please

breakponchito:
LGTM

breakponchito:
Jenkins test please

pdudits (approved):

pdudits:
As a rule of thumb, I'd for now keep `web-core` for basic web stack, i. e. anything we need to feed catalina's low level services, so things where we integrate Payara/Grizzly into catalina.
`web-glue` would be for things where we integrate catalina into Payara, like specific classloaders or integrating it in deployment process (what is the use case of `WarHandler`, I believe).

PR #5972, opened 2022-10-03
FISH-6500 hazelcast-configuration-file Domain Property Ignored

luiseufrasio:
## Description
Payara seems to completely ignore hazelcast config file

### New tests
None

### Testing Performed
1. Build the project: mvn clean install -DskipTests
2. Start Payara Server
3. Go to localhost:4848 and click on Data Grid
4. Save the file Hazelcast-test.xml available at: https://payara.atlassian.net/browse/FISH-6500
5. Fill input: "Override configuration file" with the path of "Hazelcast-test.xml"
6. Click on Save button
7. Check content of "DAS Start Port"... It should be changed to 8876

### Testing Environment
Open JDK 1.8.0_312 on Windows 10 with Maven 3.8.4


luiseufrasio:
jenkins test please

luiseufrasio:
jenkins test please

JamesHillyard:
jenkins test please

JamesHillyard (changes requested):
Functionally this seems to work well under ideal conditions! Under non-ideal conditions some things break: 
-  Putting an invalid file path just silently causes nothing to happen
- Putting a malformed file (Missing a closing tag) silently causes nothing to happen
- If you modify the file, then restart the server the new changes don't take effect. You have to click save again to trigger the 'Restart Required' event
- If you make a config-breaking change (In the example change the interface to `127.0.0.2`) apply and restart, Payara will fail to start. Reverting this change still causes Payara to fail to start as it doesn't reapply the new config

breakponchito:
Jenkins test please

luiseufrasio:
> Functionally this seems to work well under ideal conditions! Under non-ideal conditions some things break:
> 
> * Putting an invalid file path just silently causes nothing to happen
> * Putting a malformed file (Missing a closing tag) silently causes nothing to happen
> * If you modify the file, then restart the server the new changes don't take effect. You have to click save again to trigger the 'Restart Required' event
> * If you make a config-breaking change (In the example change the interface to `127.0.0.2`) apply and restart, Payara will fail to start. Reverting this change still causes Payara to fail to start as it doesn't reapply the new config

@JamesHillyard could you review it again?

JamesHillyard:
Jenkins test please

JamesHillyard (commented):
The properties in the admin console now update immediately, meaning the `restart required` warning doesn't show, looking at the instances these properties aren't shown until a restart happens 
image

Running `asadmin get-hazelcast-configuration` does show the updated properties, but running ` netstat -aon | findstr 4900` shows something running on port 4900. 

Restarting the server then causes the Data Grid instances page to update, and nothing running on port 4900, so I don't believe the properties to be updating dynamically. (This isn't a requirement, so forcing a restart is acceptable to solve this)

---

- Testing with an invalid file path works well
- Testing with an invalid file works well
- Testing with a server-breaking config causes the server to fail to start (As expected), then changing the file to be valid again causes the server start again successfully - works well
- When trying to remove the hazelcast config file by clearing the field or setting it to the default `hazelcast-config.xml` the value is accepted (Unlike an invalid path) but the field is changed back to what it was before and requests a restart - **Needs Fixing**

---

Another observation, when starting the server with a custom hazelcast file, between the two logs shared below the server appears to hang for 15-20 seconds with no indication what it's doing. This causes slower server startups (Which might be fine) but a log to say what it's doing (I suspect reading my hazelcast file) might be good so it's clear what's happening

```
[#|2022-10-12T10:55:25.585+0100|INFO|Payara 5.2022.4-SNAPSHOT|org.hibernate.validator.internal.util.Version|_ThreadID=30;_ThreadName=RunLevelControllerThread-1665568523911;_TimeMillis=1665568525585;_LevelValue=800;|
  HV000001: Hibernate Validator 6.1.5.Final|#]

[#|2022-10-12T10:55:41.952+0100|INFO|Payara 5.2022.4-SNAPSHOT|fish.payara.nucleus.hazelcast.HazelcastCore|_ThreadID=30;_ThreadName=RunLevelControllerThread-1665568523911;_TimeMillis=1665568541952;_LevelValue=800;|
  Hazelcast Instance Bound to JNDI at payara/Hazelcast|#]
```

luiseufrasio:
jenkins test please

JamesHillyard (commented):
When using a file to change the hazelcast port the properties appear to update dynamically, checking port 4900 with `netstat -aon | findstr 4900` shows nothing. Running the same command but with the new port ` netstat -aon | findstr 9955` also shows nothing, as if the properties were changed but haven't been applied. A restart then shows port 9955 as in use.

This seems to be stuck between working dynamically and still requiring a restart, I'm happy for it to work in either way but at the moment it appears to work dynamically but doesn't actually work until a restart happens which is a bit confusing!

luiseufrasio:
jenkins test please

luiseufrasio:
jenkins test please

JamesHillyard (commented):
Hazelcast properties only take affect after the required restart now which seems good. Only thing I noticed is setting the URL back to the default (hazelcast-config.xml) after applying changes from my file doesn't cause it to go back to the defaults, it leaves the previous properties with the new, default, file path.

Other than that all looks good

luiseufrasio:
jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

luiseufrasio:
jenkins test please

luiseufrasio:
jenkins test please

breakponchito:
Jenkins test please

nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/hazelcast/HazelcastRuntimeConfiguration.java
```diff
@@ -51,7 +51,11 @@
 @Configured
 public interface HazelcastRuntimeConfiguration 
     extends ConfigBeanProxy, DomainExtension {
-    
+
```
breakponchito, line 54: please add missed copyright to this file

breakponchito (approved):
LGTM

breakponchito:
please add missed copyrights where needed

JamesHillyard (commented):
There seems to be an issue when returning to the defaults and restarting it requests a second restart:

- Change the file to the hazelcast-test.xml file to change the properties
- Restart the server
- Change the file to the default hazelcast-config.xml
- Restart the server

The server still says it needs another restart?

Another nice to have would be the help text for the file property on the admin console should tell you how to restore the defaults

luiseufrasio:
jenkins test please

luiseufrasio:
jenkins test please

Pandrex247 (approved):

Pandrex247:
Jenkins test please

PR #5986, opened 2022-10-18
FISH-6566 Only create an action report for action report files

kalinchan:
## Description

On JDK 11.0.16 if a WAR contained a .json file, Payara would create an ActionReport when it is not needed, this change ensures that for only ActionReport files an ActionReport is created.


kalinchan:
jenkins test please

breakponchito (approved):
LGTM

JamesHillyard:
Fixes https://github.com/payara/Payara/issues/5987

PR #5991, opened 2022-10-20
FISH-6588 Add Extra Run Levels for Post-Boot and Deployment and Resolve Payara Micro Variable Expansion Error

Pandrex247:
## Description
This PR attempts to sanitise the boot sequence for services, post-boot commands, and application deployments.

In our previous attempts to allow variable substitution in post-boot command files and make post-boot commands consistent, we ended up frequently causing unintentional side effects typically to do with deployment or command timings.

This PR introduces two new run levels, which are used to hopefully make everything consistent and better space out the server boot lifecycle timings.

Before:
* Run level 10 (in no particular order - all run at the same run level):
  * Previously deployed applications
  * Post-boot commands
  * Startup services (e.g. MicroProfile Config)
* Run level 20:
  * Autodeploy applications (not previously deployed ones)
    * Applications in the autodeploy folder
    * Admin Console

After:
* Run level 10:
  * Startup services (e.g. MicroProfile Config)
* Run level 12:
  * Post-boot commands
* Run level 15: 
  * Previously deployed applications
* Run level 20:
  * Autodeploy applications (not previously deployed ones)
    * Applications in the autodeploy folder
    * Admin Console

Additionally, this PR fixes an issue where Payara Micro attempts to perform variable expansion as it parses the command files during the pre-boot phase - variable expansion is post-boot only.

### Blockers
None

### New tests
Pending, will look at automating the manual tests below...

### Testing Performed
Tested the following reproducers against Payara Server:
* [FISH-5976](https://payara.atlassian.net/browse/FISH-5976)
* [FISH-6484](https://payara.atlassian.net/browse/FISH-6484)
* [FISH-6492](https://payara.atlassian.net/browse/FISH-6492)
* [FISH-6501](https://payara.atlassian.net/browse/FISH-6501)

Tested the following reproducer against Payara Micro:
* [FISH-5976](https://payara.atlassian.net/browse/FISH-5976)
* [FISH-6492](https://payara.atlassian.net/browse/FISH-6492)

### Testing Environment
Windows 11, JDK 8.
WSL, JDK 8.

## Documentation
Public: N/A
Internal analysis and solution architecture: [Confluence Page](https://payara.atlassian.net/wiki/spaces/PAYAR/pages/3488448526/Start-up+Post-Boot+Deployment+and+Post-Deploy)

## Notes for Reviewers
I'm not sure if the `ApplicationLoaderService` should still have a non-validating run level. What this means is that if something with a lower run level asks for it (e.g. a startup service of run level 10), it will load _before_ it's nominal run level of 15. If we make it validating, what will happen is that the HK2 service will simply error out (and in all likelihood cause the server to fail to boot).

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
+++ b/appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/PayaraMicroImpl.java
@@ -1053,7 +1053,7 @@ public class PayaraMicroImpl implements PayaraMicroBoot {
             gfproperties.setConfigFileURI(runtimeDir.getDomainXML().toURI().toString());
 
             try {
-                configureCommandFiles();
+                configurePreBootCommandFiles();
             } catch (IOException ex) {
                 LOGGER.log(Level.SEVERE, "Unable to load command file", ex);
             }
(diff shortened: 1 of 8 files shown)


Pandrex247:
Jenkins test please

JamesHillyard (commented):
Tested this in a few different scenarios, including configuring instances, deployment groups, password aliases, jdbc connections, deploying apps and using set commands on http listeners, all worked exactly as expected

aubi (approved):
LGTM

PR #6009, opened 2022-10-28
FISH-6317: Integrate catalina stack into web connector

pdudits:
## Description

This aims to integrate core grizzly services of server with new catalina connector which is represented by `CatalinaWebStack`.

After having deciphered the startup sequence, the solution is to set `HttpHandler` that represents catalina as default handler of respective `Mapper`, that represents individual network listener. 

At the same time it is apparent, that we don't need to manage mappings on context root level. These were just optimization of grizzly mapping process, which is now useless, as the partial mapping cannot be reused when request is processed by Catalina.

Since facade for Tomcat is now represented by `CatalinaWebStack` any invocations to `Engine` or `Service` moved into this class. This made original `EmbeddedWebContainer` useless, and it was not an independent service anyway. Therefore any relevant methods moved into `WebContainer`.

More classes were marked for deletion, ultimately the entire hierarchy of `WebConnector` and its supporting classes should be no longer necessary.

## Testing

Once we get into runnable state we will need to verify following:

* behavior upon reconfiguration of VirtualServers and their assignment to network listeners
* HTTP Service property changes -- should get applied onto Grizzly Http filter chain

Other related issues: FISH-6141, FISH-6147, FISH-6162


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/pluggable/WebContainerFeatureFactory.java
```diff
@@ -54,7 +54,7 @@ public interface WebContainerFeatureFactory {
     
     public HealthChecker getHADBHealthChecker(WebContainer webContainer);
     
-    public ReplicationReceiver getReplicationReceiver(EmbeddedWebContainer embedded);
+    public ReplicationReceiver getReplicationReceiver(WebContainer webContainer);
```
Pandrex247, line 57: Is it necessary to keep this parameter? Its our own interface, and there's only one impl of it which ignores the parameter.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/pluggable/WebContainerFeatureFactory.java
```diff
@@ -54,7 +54,7 @@ public interface WebContainerFeatureFactory {
```
Pandrex247, line 54: Missing copyright update.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebContainerStartStopOperation.java
```diff
@@ -63,5 +63,5 @@ public interface WebContainerStartStopOperation {
```
Pandrex247, line 63: Missing copyright update.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEWebContainerStartStopOperation.java
```diff
@@ -55,20 +55,11 @@
  * @author lwhite
```
Pandrex247, line 55: Missing copyright update.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEWebContainerFeatureFactoryImpl.java
```diff
@@ -63,8 +63,8 @@ public HealthChecker getHADBHealthChecker(WebContainer webContainer) {
     }
```
Pandrex247, line 63: Missing copyright update.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEReplicationReceiver.java
```diff
@@ -56,7 +56,7 @@
 public class PEReplicationReceiver implements ReplicationReceiver {
```
Pandrex247, line 56: Missing copyright update.

Pandrex247 (approved):
I'm a little torn on the ripping out of `EmbeddedWebContainer`.
It was our "API" class for `Tomcat`, and now seems to have been awkwardly rolled into `CatalinaWebStack`.

I recognise one doesn't live without the other, but I just worry that `CatalinaWebStack` is going to end up as a bit of a monolith.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEWebContainerFeatureFactoryImpl.java
```diff
@@ -63,8 +63,8 @@ public HealthChecker getHADBHealthChecker(WebContainer webContainer) {
...
     
     @Override
-    public ReplicationReceiver getReplicationReceiver(EmbeddedWebContainer embedded) {
-        return new PEReplicationReceiver(embedded);
+    public ReplicationReceiver getReplicationReceiver(WebContainer webContainer) {
```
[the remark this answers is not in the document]
pdudits: > I recognise one doesn't live without the other, but I just worry that `CatalinaWebStack` is going to end up as a bit of a monolith

It should end up as a Facade - no tomcat data manipulation should happen outside it, and therefore we know exactly what we do with it. So yes, it might grow large, but no details of plugging into Catalina -- especially modifying catalina configuration -- would be present elsewhere.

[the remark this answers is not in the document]
pdudits: Regarding parameter, I didn't try to change the type parameters too much, because I would otherwise be tempted to delete the entire class and all its neighbors :D

appserver/web/web-core/src/main/java/fish/payara/appserver/web/core/CatalinaWebStack.java
```diff
@@ -97,8 +105,20 @@ public void addHost(Host host) {
...
+    public Host getHost(String name) {
+        return (Host) engine.findChild(name);
+    }
+
     public void start() throws LifecycleException {
-        server.start();
+        if (!hasConnectors) {
+            service.addConnector(new GrizzlyConnector());
+        }
+        tomcat.init();
```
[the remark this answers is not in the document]
pdudits: Yes, that's what appears to be correct startup sequence for Tomcat. We don't need explicit `init` command, because it is tomcat implementation detail not relevant for WebContainer.

PR #6053, opened 2022-11-16
FISH-6047: Fix Single Sing On handling for Jaspic applications, Single Sign Off for all SSO-enabled apps

### LINKED ISSUE #5551: SSO not executing single logout properly /FISH-6047
We have developed a number of applications (WARs) and have deployed them on Payara 5-2021-10 with “Enable SSO” checked. These apps use Soteria (org.glassfish.soteria:jakarta.security.enterprise:1.0.1). We use the following annotation for our login bean:

`@FacesConfig(
	// Activates CDI build-in beans
	version = JSF_2_3 
)
@CustomFormAuthenticationMechanismDefinition(
    loginToContinue = @LoginToContinue(
            loginPage="/login.xhtml",
            errorPage="" // DRAFT API - must be set to empty for now
    )
)`

For this example, we will refer to two apps, “A” and “B.” For user convenience, both apps have “logout” buttons. 

We log in to app “A” and do some work. We then use a browser Favorites shortcut to go to app “B.” This works as expected- we are not asked to log in again. We do some work in app “B” and click that app’s logout button. This redirects us to the home page of app "B" and we see the login screen as expected. We then use a shortcut on our browser’s Favorites bar to go back to app “A.” Rather than asking for a new login, as it should (because the SSO session has been invalidated), app “A” allows us to continue working as if we had never logged out. 

The URLs for both apps have identical prefixes, i.e., https://domain/etc. In other words, they are using the same port number, machine domain, and protocol. 

This behavior is observed in a number of browsers (Chrome, Edge/Chromium, and Firefox). 

It seems that Payara is creating two different sessions for the two apps, and is not coordinating them on logout. Obviously this is a serious security issue. Any idea as to what may be causing it?

The logout logic is as follows:

`
// Try to retrieve our current session
            HttpSession session = request.getSession(false);
                        
            // Logout current request
            request.logout();
            
            // Invalidate session so user becomes anonymous.
            if (session == null) {
                LOGGER.log(Level.INFO, "Identity: No session to invalidate for user: {0}", userId);
            } else {
                session.invalidate();
                LOGGER.log(Level.INFO, "Session invalidated for user: {0}/{1}", new Object[] { userId, sessionId } );
            }
`

pdudits:
## Description
Fixes #5551. The session that estabilishes SSO ID is now added into SSO tracking object.

Upon logout, the session is now explicitly invalidated before new SSO ID is generated. 

Empty realm is now handled the same as `null` realm. Therefore to enable SSO between Jaspic, or Jakarta EE Security, following must be present in `web.xml`:

```
<login-config>
  <realm-name>anything</realm-name>
</login-config>
```
This doesn't need to refer to any existing realm name, it's a mean to enable applications to opt-in to SSO feature. Otherwise any two applications would be treated as SSO ones, regardless of their identity store or even authentication mechanism.

### Testing Performed

Test project per our internal JIRA now behaves as expected (after making the change to `web.xml`).

### Testing Environment

```
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: C:\ProgramData\chocolatey\lib\maven\apache-maven-3.8.6
Java version: 11.0.12, vendor: Azul Systems, Inc., runtime: C:\Program Files\Zulu\zulu-11
```

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/AuthenticatorBase.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/authenticator/AuthenticatorBase.java
@@ -768,6 +768,9 @@ public abstract class AuthenticatorBase extends ValveBase
         assert (realm != null);
         sso.register(value, principal, authType, username, password, realm);
         // END S1AS8 PE 4856080,4918627
+        if (session != null) {
+            sso.associate(value, 0, session);
+        }
 
         request.setNote(Constants.REQ_SSOID_NOTE, value);
         if (sso.isVersioningSupported()) {
@@ -807,6 +810,9 @@ public abstract class AuthenticatorBase extends ValveBase
         if (session != null) {
             session.setPrincipal(null);
             session.setAuthType(null);
+            if (session.getSsoId() != null) {
+                session.expire();
+            }
         }
 
         // principal and authType set to null in the following
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/security/web/GlassFishSingleSignOn.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/security/web/GlassFishSingleSignOn.java
@@ -337,7 +337,7 @@ public class GlassFishSingleSignOn extends SingleSignOn
         }
 
         String realmName = realm.getRealmName();
-        if (realmName == null) {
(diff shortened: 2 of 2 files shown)


pdudits:
Jenkins test please

arieki (approved):

PR #6073, opened 2022-11-23
DEVOPS-565 Automate bundled certificates

kalinchan:
## Description
In order to avoid expired certificates when releasing Payara this feature has been implemented.

This feature only works on a Linux environment and is activated with the following maven profile `automate-certificates` 

This profile will execute a bash script which copies all of the certificates from the current JDK (if JAVA_HOME is set) and add them to the keystore which comes bundled with Payara, all of the expired certificates are then removed from the keystore.

You can see the results of this here https://www.diffchecker.com/oe1PfBgV (left is the original and right is with this feature)


cubastanley (approved):
Lgtm

Pandrex247 (approved):
Possible improvement would be to activate the profile if on Linux automatically - would we ever **not** want to remove the expired certificates?

kalinchan:
If the profile is activated by default on Linux environments, Developers building Payara and making pull requests will constantly be making changes to the keystores in each Pull Request (even when not required) unless they undo the changes manually, I don't believe this is ideal.

Pandrex247:
I think that point is moot: whenever you create a domain you're not getting the same certificates.
The create-domain command should be clearing out all expired certificates in the trust store and copying across those from the JDK.

Which raises the question of why this PR is necessary in the first place.

See:  https://github.com/payara/Payara/blob/master/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/KeystoreManager.java#L293

kalinchan:
Closing as feature is already implemented in Payara like @Pandrex247 mentioned.

PR #6080, opened 2022-11-28
FISH-6775 Authorization Constraints Ignored When Using Path Traversal Penetration Using Default Virtual Module (Payara6)

Pandrex247:
## Description
Security fix.

It was possible to circumvent JACC authentication using a `./` traversal, since the request normalisation which normally stops these attacks actually happens _after_ the JACC check has occurred. So while the path is normalised as expected, the actual JACC check has already occurred and passed when it would fail against the normalised path.

This PR reactivates the old normalisation in the CoyoteAdapater which was commented out many many moons ago without explanation when integrating Grizzly with the forked Tomcat in the days of GlassFish 2 & 3.

What happens now is that if a request which would fail its normalisation checks occur, a HTTP 400 request is returned. This is, from a quick scan of their code, in line with how Tomcat / Catalina functions today.
The criteria for failing normalisation are:

* Request not starting with a "/"
* Request with backslashes in (if a property is not enabled: `org.glassfish.grizzly.tcp.tomcat5.CoyoteAdapter.ALLOW_BACKSLASH`)
* Request with null bytes in
* Requests which try to escape the context

### Blockers
None

### New tests
None

### Testing Performed
* Ran the reproducer for FISH-6775
* Ran the reproducer for FISH-6603

### Testing Environment
WSL, JDK 11

## Documentation
None - no CVE

## Notes for Reviewers
None

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/CoyoteAdapter.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/CoyoteAdapter.java
@@ -428,6 +428,13 @@ public class CoyoteAdapter extends HttpHandler {
             response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid URI");
             return false;
         }
+
+        // Normalize Decoded URI
+        if (!normalize(decodedURI)) {
+            res.setStatus(400);
+            res.setDetailMessage(("Invalid URI"));
+            return false;
+        }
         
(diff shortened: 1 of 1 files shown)


Pandrex247:
Jenkins test please

pdudits:
Please make clone of the issue in Catalina epic, so we would retest this with new stack as well.

pdudits (approved):

PR #6098, opened 2022-12-02
FISH-6775 Update

Pandrex247:
## Description
Update to https://github.com/payara/Payara/pull/6079 based on feedback from GlassFish.

Reworks the solution so that exceptions aren't used.
Also fixes the detail message so that it actually gets set.

### Blockers
None

### New tests
None

### Testing Performed
_Pending_
* Ran the reproducer for FISH-6775

### Testing Environment
WSL, JDK 8

## Documentation
N/A - internal

## Notes for Reviewers
None


Pandrex247:
Jenkins test please

PR #6105, opened 2022-12-07
FISH-5971 Upgrade Catalina to 10.1.1

JamesHillyard:
Updates Catalina to 10.1.1, fixing new compilation errors in `web-core`, `web-glue` and `websecurity`. Unable to test this further as the branch doesn't compile.

Can't upgrade to Catalina 10.1.2 as this bug is present https://bz-he-de.apache.org/bugzilla/show_bug.cgi?id=66353 causing test failures. 10.1.3 will fix this when it's released


pdudits:
Which is interesting because it is a class from Jakarta migrator they state they don't have required dependency on (and we shouldn't have one as well)

JamesHillyard:
This is a bug in tomcat-catalina 10.1.2, it has been fixed in 10.1.3 which is unreleased, so reverting back to 10.1.1. https://bz-he-de.apache.org/bugzilla/show_bug.cgi?id=66353.

Pandrex247 (approved):

PR #6118, opened 2022-12-13
FISH-6578 Use Zip64 as-needed

JamesHillyard:
## Description
Payara Enterprise Embedded All exceeded the 65,535 file limit for a zip/jar so Zip64 was required. This however caused an invalid cen header / class not found exception for all other embedded distributions. This specifically only impacted Payara 5 on JDK 8. It appears to be related to this JDK bug which was fixed in JDK 10 https://bugs.openjdk.org/browse/JDK-8186464 hence only affecting JDK 8.

Use the 'as-needed' option instead so there is no issue with jars containing fewer than 65,535 files but will use Zip64 for distributions with more files, allowing both to compile and run as expected.

### Testing Performed
Tested Embedded Web & Embedded All manual test.

### Testing Environment
Ubuntu 20 (WSL), Zulu JDK 11, Maven 3.6.3

## Documentation
N/A

### DIFF
--- a/appserver/extras/embedded/build.xml
+++ b/appserver/extras/embedded/build.xml
@@ -126,7 +126,7 @@
         <defaultexcludes add="META-INF/**.inf"/>
         <defaultexcludes add="META-INF/**.SF"/>
 
-        <rejar destfile="${finaljar}" duplicate="preserve" zip64Mode="always">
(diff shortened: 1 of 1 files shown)


JamesHillyard:
Jenkins test please

Pandrex247 (approved):

JamesHillyard:
Jenkins test please

PR #6131, opened 2022-12-23
FISH-6596 Upgrade Mojarra to Patched Source 4.0.0-p2

aubi:
## Description
This change of Mojarra uses regex for jakarta/javax copy of POST headers.
Using startsWith failed in Faces TCK, when some option was in fact an event.

### Testing Performed
Faces TCK

### Testing Environment
OpenJDK, Linux


aubi:
Jenkins test please

breakponchito (approved):
LGTM

PR #6174, opened 2023-01-30
FISH-6152 Resolve Compilation Errors in ErrorPageDecorator

JamesHillyard:
## Description
Resolves compilation errors in ErrorPageDecorator. 

ErrorPage has moved to https://github.com/apache/tomcat/blob/10.1.x/java/org/apache/tomcat/util/descriptor/web/ErrorPage.java

RequestUtil.urlDecode has been replaced with UDecoder.URLDecode. This always requires a charset so to maintain previous behaviour I have set this to null as was previously happening https://github.com/apache/tomcat/blob/7.0.x/java/org/apache/catalina/util/RequestUtil.java#L178 
The same null charset behaviour was maintained https://github.com/apache/tomcat/blob/10.1.x/java/org/apache/tomcat/util/buf/UDecoder.java#L186

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/deploy/ErrorPageDecorator.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/deploy/ErrorPageDecorator.java
@@ -37,11 +37,12 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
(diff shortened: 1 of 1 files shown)


luiseufrasio (approved):
LGTM!

should we remove the unused "location" property?

Pandrex247 (approved):

PR #6182, opened 2023-02-01
FISH-6167 Resolve Compilation Errors in CacheModule

JamesHillyard:
Resolves the compilation errors in CacheModule. FilterDef and FilterMap were part of `web-core` which was removed, I believe these changes will maintain the original behaviour although they are different.

FilterDef `setFilterClassName` was replaced with `setFilterClass` however these both take strings and appear to be doing the same function. A null check is lost, however `CACHING_FILTER_CLASSNAME` is a static final string so could never be null hence I didn't feel the need to re-implement this.

Rather than a FilterMap having one servlet name and URL pattern, it instead has two lists which I have modified by adding rather than setting these properties. `setDispatcherTypes` with an EnumSet of DispatcherType is completely removed, being replaced with `setDispatcher` requiring a string. This new method then performs a similar action to the removed code, getting the DispatcherType in the `setDispatcher` method. As the comment indicates, these calls are cumulative and should still work within the loop as expected.

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/CacheModule.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/CacheModule.java
@@ -37,6 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2023] Payara Foundation and/or affiliates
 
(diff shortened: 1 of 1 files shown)


breakponchito (approved):
LGTM

Pandrex247 (approved):

PR #6185, opened 2023-02-03
FISH-6969: Use OpenTelemetry as basis for OpenTracing Support

pdudits:
## Description

As transitional step, we use OpenTracing shim to bridge all use of opentracing in the server with OpenTelemetry.

Whenever OpenTracing is used, OpenTelemetry SDK is auto-configured on per-application basis respecting system properties and environment variables.

### New tests
So far there is no application API beyond using OpenTracing to get to OTEL, so they will arrive later

### Testing Performed
Payara Samples and OpenTracing TCKs should still pass without problems, let's see what Jenkins thinks about it.


pdudits:
Jenkins test please

pdudits:
Jenkins test please

arieki (approved):
Some of copyright needs to be updated

pdudits:
Will update that in followup PR.

PR #6186, opened 2023-02-03
FISH-6769 OpenApi Support for 'extensions'

aubi:
## Description
Adding support

### Testing Performed
Going through MP OpenApi TCK -- test, which deploy (yes, the TCK is broken) pass!

There is a setup required -- disable https (Configurations, server-config, Network Config, edit http-listener-2, set disabled; this is registered as second server, tests expectat only one server, specifically `testStaticDocument`).

### Testing Environment
Linux, OpenJDK

## Note for reviewer
It was necessary to distinguish between "no extensions defined" and "empty extensions defined". The TCK requires, that annotation `@Extension` is propagated to `@APIResponses/@APIResponse`, but only if not defined. In the next code:

```
    @APIResponses(extensions = @Extension(name = "x-responses-ext", value = "test-responses-ext"), value = {
        @APIResponse(responseCode = "200", description = "successful operation",
                extensions = @Extension(name = "x-response-ext", value = "test-response-ext")),
        @APIResponse(responseCode = "500", description = "server error", extensions = {}),
        @APIResponse(responseCode = "503", description = "service not available",
                content = @Content(extensions = @Extension(name = "x-notavailable-ext", value = "true"))),})
    @Operation(summary = "Returns pet inventories by status",
            description = "Returns a map of status codes to quantities")
    @Extension(name = "x-operation-ext", value = "test-operation-ext")
    public java.util.Map<String, Integer> getInventory() {
```

The extension "x-operation-ext" is propagated to code 503 (no other extension is defined), but not to 200  (extension "x-response-ext" defined), nor to 500 (empty extensions!).

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) [2018-2020] Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) [2018-2023] Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
@@ -105,10 +105,6 @@ public class OperationImpl extends ExtensibleImpl<Operation> implements Operatio
         List<AnnotationModel> responses = annotation.getValue("responses", List.class);
         if (responses != null) {
             APIResponsesImpl apiResponsesImpl = APIResponsesImpl.createInstance(annotation, context);
-// FIXME: implement!
-//            for (AnnotationModel response : responses) {
-//                from.getResponses().addAPIResponse(response.getValue("responseCode", String.class), apiResponsesImpl);
-//            }
         }
         from.setExtensions(parseExtensions(annotation));
         extractAnnotations(annotation, context, "callbacks", "name", CallbackImpl::createInstance, from::addCallback);
--- a/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
+++ b/appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
@@ -57,6 +57,7 @@ import java.util.List;
 import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import fish.payara.microprofile.openapi.impl.model.responses.APIResponseImpl;
 
 import org.eclipse.microprofile.openapi.models.ExternalDocumentation;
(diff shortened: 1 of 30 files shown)


luiseufrasio (changes requested):
The following command: 
**curl http://localhost:8080/openapi?format=json | jq '.paths."/partne
r/inventory".get.responses."200"'**
returns 
**{
  "description": "successful operation"
}**
It should return 
**{
  "description": "successful operation",
  "x-responses-ext": "test-responses-ext"
}**

aubi:
> It should return
> **{ "description": "successful operation", "x-responses-ext": "test-responses-ext" }**

It should return 
```json
{
  "description": "successful operation",
  "x-response-ext": "test-response-ext"
}
```
Which is does now.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
```diff
@@ -102,7 +101,16 @@ public static Operation createInstance(AnnotationModel annotation, ApiContext co
...
         }
-        extractAnnotations(annotation, context, "responses", "responseCode", APIResponseImpl::createInstance, from.responses::addAPIResponse);
+        //extractAnnotations(annotation, context, "responses", "responseCode", APIResponseImpl::createInstance, from.responses::addAPIResponse);
+        List<AnnotationModel> responses = annotation.getValue("responses", List.class);
+        if (responses != null) {
+            APIResponsesImpl apiResponsesImpl = APIResponsesImpl.createInstance(annotation, context);
+// FIXME: implement!
+//            for (AnnotationModel response : responses) {
+//                from.getResponses().addAPIResponse(response.getValue("responseCode", String.class), apiResponsesImpl);
+//            }
```
Pandrex247, line 111: What is this? Is this required by the spec? Do we have an issue for this? Is this not required for the TCK?

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/ContentImpl.java
```diff
@@ -60,6 +62,8 @@ public class ContentImpl extends LinkedHashMap<String, MediaType> implements Con
 
     private static final long serialVersionUID = 1575356277308242221L;
 
+    private Map<String, Object> extensions = new HashMap<>(); // workaround as Content doesn't extend Extendable!
```
Pandrex247, line 65: Is this fine to not be null? You drew attention to the extension map needing to be null previously in another class

Pandrex247 (commented):
Couple of comments

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/media/ContentImpl.java
aubi: Actually it is ok, because the whole stuff around "null" is just to solve TCK requirement, which is discussed and will be probably removed (usage of `@Extension` on method without being part of `@APIResponse(s)`).
Anyway, I set to to null, there may be some other case I don't see now.

appserver/payara-appserver-modules/microprofile/openapi/src/main/java/fish/payara/microprofile/openapi/impl/model/OperationImpl.java
aubi: Solved now in other way, removing comment.

aubi:
Jenkins test please

Pandrex247 (commented):
That failing test appears to be valid.
You probably need to add null checks (or just go with your original gut instinct that it doesn't matter if it isn't null)

aubi:
Jenkins test please

aubi:
Jenkins test please

aubi:
Now, all the tests in TCK related to extensions pass correctly.

aubi:
Jenkins test please

Pandrex247 (approved):

luiseufrasio (approved):
LGTM!

PR #6189, opened 2023-02-06
FISH-7063: Use only single property for Jakarta API version

pdudits:
## Description

Payara BOM declared `jakarta-web-api` version of 9.1.0, as well as for Platform BOM. This was an oversight where we in fact had two properties for platform version.

It also address eclipse-ee4j/jakartaee-api#133, where we exclude the transitional dependencies by default.


pdudits:
Jenkins test please

pdudits:
Jenkins test please

arieki (approved):
LGTM

PR #6196, opened 2023-02-13
Log JDBC Realm Login Failures, User Not Found or Password Mismatch

aubi:
## Description
Setting JDBC Realm is a bit frustrating, because only successful login verifies the settings. In case of problems, setting  logger `javax.enterprise.system.core.security` to the `FINEST` level helps. But if the user is not logged in, only debugging server help to find why.

Therefor I added logging if user wasn't found of if password didn't match. Username is logged, password is not.
All these logs are done on the `FINEST` level.

### Testing Performed
Switch logger  `javax.enterprise.system.core.security` to the `FINEST` level
Setup JDBC Realm, take app using it, try to login with 
1) non-existing user, logs: `User 'abc' not found in the database!`
2) wrong password, logs: `User 'abc' password mismatch!`

It's necessary to search the log entry, running admin gui fills it quickly at this level.

### Testing Environment
Linux, OpenJDK


aubi:
Jenkins test please

Pandrex247 (approved):
Eyeballed - LGTM
WCGW?

PR #6207, opened 2023-02-17
FISH-6733 FISH-6169 FISH-6170 FISH-6171 FISH-6173 FISH-6193 FISH-6141 FISH-6138 FISH-6145 FISH-6161 FISH-7106: Remove WebConnector and PWC classes

pdudits:
## Description

This removes remaining classes with connector responsibilities that are now handled by `CatalinaStack` as well as any references to them. Basically all of `com.sun.enterprise.web` package is now without compilation errors.

It highlights another feature that will go away and that is programmable access to WebContainer and VirtualServers in embedded mode.

Also touches some of the features created to support High availability, where extra attributes are in session.

We also had special methods to reconfigure SSO on the fly, by having access to AuthenticatorBase. Catalina's authenticator base has all of that support for SSO, but only at restart. Updated FISH-6270 with this observation.


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEVirtualServer.java
```diff
@@ -50,11 +50,4 @@ public final class PEVirtualServer extends VirtualServer {
```
Pandrex247, line 50: Should we not just delete this class now?
It seems like an extension without changing anything from the parent?

Pandrex247 (approved):

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEVirtualServer.java
pdudits: Right, I think this was just a run of "Delete unused members", didn't notice it.

PR #6231, opened 2023-03-11
FISH-7203 Add Null Check for OpenTelemetry

Pandrex247:
## Description
Adds a null check for the OpenTelemtry service to avoid a NPE when running the embedded tests.

### Blockers
None

### New tests
None

### Testing Performed
Built and ran all unit tests: `mvn clean install -PBuildExtras`
Ran payara-samples.
Ran OpenTracing and OpenTelemetry TCKs.

### Testing Environment
Windows 11, Zulu 11.

## Documentation
N/A

## Notes for Reviewers
None


Pandrex247:
Jenkins test please

pdudits:
Doesn't that mean we forgot to package something into embedded? Or is it the usecase of using jaxrs client with embedded on classpath?

In latter case  I'd prefer having this check inside PayaraTracingServices, as that was reason the class exists. 

But we can have that in next iteration,since it's Saturday night 😁

pdudits (approved):

Pandrex247:
> Doesn't that mean we forgot to package something into embedded? Or is it the usecase of using jaxrs client with embedded on classpath?

Would need to debug that. Embedded has the `requesttracing` and `microprofile` packages in its dependencies, which _look_ like they should pull in the services such that they'd resolve.

PR #6233, opened 2023-03-14
FISH-6479 Metro and xmlsec Repackaging

aubi:
## Description
Metro and xmlsec use and also exported SLF4J. Payara therefor exported them as well, but deployed applications didn't run successfully anyway.
The fix is to stop exporting SLF4J. Both libraries are repackaged now, forbidding SLF4J export.

### Testing Performed
Deployed both reproducers, one is JEE 8, other JEE 9, one uses SLF4J v 2, the other version 1. Both now work correctly.
Reproducers are attached to the Jira ticket, use the *-fixed versions. The fix was in packaging slf4j with the apps, Payara now doesn't provide slf4j.

### Testing Environment
Linux, OpenJDK 11

### DIFF
--- a/appserver/packager/external/metro-xmlsec/pom.xml
+++ b/appserver/packager/external/metro-xmlsec/pom.xml
@@ -1,0 +1,136 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~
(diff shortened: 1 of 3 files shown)


aubi:
Jenkins test please

breakponchito (approved):
LGTM

PR #6246, opened 2023-03-24
FISH-6607: Introduce core-parent

pdudits:
## Description

Oooh, it's a big one. Build separation of core modules from the other ones, while not moving them away from their current locations in source tree.

* The modules that had `core` in `groupId` now have parent `core-parent`, and version `6.0.0-SNAPSHOT`
* all these modules are removed from their previous parents and aggregates and are aggregated with `core-aggregator` (which is  not their parent now, but one level higher)
* any dependencies core modules had are now managed via `core-bom`. (and `payara-bom` reexports `core-bom`)
* any properties required for those and parent setup moved to `core-aggregator`
* therefore `core-aggregator` is parent of `payara-aggregator` so that we don't have duplicated properties

All of that was plugged back to old aggregator, so that no immediate change to build process is required.

It wasn't the aim of remove duplication of all of the plugin configs and such, rather to keep `core-parent` minimal, as it will grow another responsibilities. Therefore non-core modules did not get `core-parent` as (transitive) parent.
 
I think this is best reviewed by commits:
* 406fefb introduces core-aggreator, moving properties into it
* d570565 adds core-parent
* next commits add modules that turned out to be dependencies of core
* except for metro-xmlsec-repackaged which shouldn't have been in core
* 969dcfc then replaces all places that referred to core modules with explicit `${project.version}`. There were lots of them.
* Then relevant modules to allow for making admingui plugins were added (`console-common` and its dependencies)
* And as a verification I transformed certificate-management to only depend on payara-core in b93b560d470ff8464f3a9c496ec4de3a19da539d

### Testing Performed

Builds with empty repo, starts, payara-samples pass.
`core/` builds with empty repo on its own.

### Testing Environment

Maven 3.9.2, Ubuntu 22.04, Zulu-11


pdudits:
Jenkins test please

Pandrex247 (changes requested):
Doesn't build with a clean repo using the `BuildExtras` profile

```
[ERROR] Failed to execute goal on project certificate-management-cli: Could not resolve dependencies for project fish.payara.extras.certificate-management:certificate-management-cli:glassfish-jar:6.2023.4-SNAPSHOT: Could not find artifact fish.payara.server.internal.cluster:cluster-common:jar:6.2023.4-SNAPSHOT
```

pdudits:
Fixed that one. But it opens a question of `certificate-management` in general. Do you feel it should be eligible to be out-of-tree component? Because it still has few internal dependencies that should move to core in such case - `server-mgmt`, `cluster-cli` and some admingui ones.

Pandrex247:
Yes certificate management should be an external component.
The only reason it still exists in the codebase is because it was created before we had such a process and we never went back to it.

Pandrex247:
Given that we're releasing Enterprise 6.0.0 _now_ - should the core version still be 6.0.0 or should it be 6.1.0 (or higher)?

api/payara-api/pom.xml
Pandrex247, line 1: Probably separate to this PR - but I think this probably needs a look at while we're rejigging all the APIs

core/core-parent/pom.xml
Pandrex247, line 1: Having the parent depend on the aggregator seems odd to me - unless the aggregator is itself individually versioned (and therefore essentially just a second parent) isn't this a little pointless?

core/pom.xml
```diff
@@ -0,0 +1,407 @@
...
+                <enabled>false</enabled>
+            </snapshots>
+        </pluginRepository>
+    </pluginRepositories>
+
+    <modules>
+        <module>core-bom</module>
+        <module>core-parent</module>
+        <!--        fish.payara.server.internal.common      glassfish-api-->
+        <module>../nucleus/common/glassfish-api</module>
```
Pandrex247, line 239: a) 🤢- why not just move them under core?
b) what's the distinction between core and nucleus now?

Pandrex247 (commented):
Yeah I think I need a sit down with you and a whiteboard.

pom.xml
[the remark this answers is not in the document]
pdudits: As long as these will remain the same as core, it doesn't. But it wasn't a decision I wanted to make right now.
Profiles should stay here and almost none of them should be in aggregator.

pom.xml
```diff
@@ -47,12 +47,19 @@
     <groupId>fish.payara.server</groupId>
     <artifactId>payara-aggregator</artifactId>
     <version>6.2023.4-SNAPSHOT</version>
+    <parent>
+        <groupId>fish.payara.server.core</groupId>
+        <artifactId>core-aggregator</artifactId>
+        <version>6.0.0-SNAPSHOT</version>
+        <relativePath>core</relativePath>
+    </parent>
```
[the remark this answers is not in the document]
pdudits: Given the difficulty of working with upstream repos I would keep it with the main tree for the time being, but with independent release cycle. There should be convenient way of working against snapshot, as well as convenient way of releasing core and switching to release version.

core/core-parent/pom.xml
pdudits: parent has build rules, aggregator just the metadata (i. e. component and plugin versions). Aggregator is also referred to from top level build exactly to fetch this metadata.

core/pom.xml
pdudits: a) Payara 5 backport simplification and minimizing merge disruption. But ultimately, we're not done with this project yet, we can still move things away later.
b) Core are the libraries lots of code (that we might want to externalize) ended up being dependent upon. There is no aim of building shippable distribution out of core.

pdudits:
Jenkins test please

pdudits:
> Yes certificate management should be an external component.
> The only reason it still exists in the codebase is because it was created before we had such a process and we never went back to it.

I added admingui core components now. As we discussed I said I would stop and reconsider if it is over 10 modules. It was exactly 10 modules.

However, I had to break depenendcy between `console-common` and `rest-service`, which structurally unsound anyway, see 278ec5d, otherwise it would end up being much more modules.

And as a verification I transformed `certificate-management` to only depend on `payara-core` in b93b560

pdudits:
> Given that we're releasing Enterprise 6.0.0 now - should the core version still be 6.0.0 or should it be 6.1.0 (or higher)?

Yes, it should but I leave the joy of doing that to whomever prepares release scripts. The core modules now seem simple enough to directly support maven release plugin if it doesn't complain too much about the scattered nature of the modules.

appserver/extras/certificate-management/pom.xml
```diff
@@ -44,9 +44,9 @@
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
-        <groupId>fish.payara.extras</groupId>
-        <artifactId>extras</artifactId>
-        <version>6.2023.4-SNAPSHOT</version>
+        <groupId>fish.payara.server.core</groupId>
+        <artifactId>core-parent</artifactId>
+        <version>6.0.0-SNAPSHOT</version>
```
Pandrex247, line 49: Should the relative path not be updated?

core/pom.xml
```diff
@@ -0,0 +1,421 @@
...
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <deploy.skip>true</deploy.skip>
+        <javadoc.skip>true</javadoc.skip>
+    </properties>
+
+    <scm>
+        <connection>scm:git:git@github.com:payara/payara.git</connection>
+        <url>scm:git:git@github.com:payara/payara.git</url>
+        <developerConnection>scm:git:git@github.com:payara/payara.git</developerConnection>
+        <tag>payara-server-6.2023.3-SNAPSHOT</tag>
```
Pandrex247, line 186: This is now wrong.
Also, I feel like the core should probably have its own tag?

Something along the lines of `payara-core-6.x.y`?

core/pom.xml
```diff
@@ -0,0 +1,421 @@
...
+    <artifactId>core-aggregator</artifactId>
+    <version>6.0.0-SNAPSHOT</version>
+
+    <packaging>pom</packaging>
+    <name>Payara Core Aggregator</name>
+
+    <description>All of Payara Core modules</description>
+
+    <properties>
+        <payara.core.version>6.0.0-SNAPSHOT</payara.core.version>
```
Pandrex247, line 57: I'm a little unsure on if this property should exist?

Within core, I don't foresee a case where this would never not just be project.version?

Outside of core, It only appears to be used within the api/payara-bom, and I'd argue that the property shouldn't belong within the core aggregator itself but instead within the server or server-api aggregators - that way it actually has meaning rather than just being something set by what will be the hard-coded parent version (and therefore slightly pointless?)

Pandrex247:
As another comment, with this rejig of modules, while it is possible to build core on its own, it doesn't seem to be possible to generate sources or javadoc specifically for the core modules - depending on how we're going to locate the core modules in the future this could be important.

Pandrex247:
Also is this intentionally not building the l10n modules as a part of core?

Pandrex247:
Also also, isn't the api/pom.xml a little pointless now?
It's now a parent pom for one child module (payara-bom).

pdudits:
> it doesn't seem to be possible to generate sources or javadoc specifically for the core modules

I agree, is this something we can put on backlog, or does it needs solving right now?

> Also is this intentionally not building the l10n modules as a part of core?

They're not needed for building components, only for distribution, so yes, it was intentional.

Pandrex247:
> > it doesn't seem to be possible to generate sources or javadoc specifically for the core modules
> 
> I agree, is this something we can put on backlog, or does it needs solving right now?

It depends where we're intending to deploy these.
If we're intending to deploy _all_ of these core artefacts to Maven Central, then I believe it's still a requirement that anything that isn't a pom/bom still has to be deployed with javadoc and sources. While these _can_ just be dummy files (e.g. containing a README), we'd still need to somehow script this for all of the artefacts.

If we're deploying to our own maven repository, then it would still be _nice_ to have this set up. Past this singular point in time where currently everything is `6.0.0-SNAPSHOT`, once it has been released the actual dependency version we're all going to be working from will be 6.0.0, whereas the version present in the checked out repo will presumably be 6.1.0 - our IDE's will throw a hissy fit. 

> > Also is this intentionally not building the l10n modules as a part of core?
> 
> They're not needed for building components, only for distribution, so yes, it was intentional.

So the l10n stuff will have a completely separate release process? So we release core to wherever we publish it with its own version _and it's own string substitutions_, but then have what actually gets subtituted in released separately and with a completely separate version?

Also, thinking ahead to where the core exists in the future where it's possibly in its own repo, we're going to have the l10n stuff where? In this repo awkwardly divorced from what it's actually the text for?

> > Also also, isn't the api/pom.xml a little pointless now?
> 
> It is, bom should be treated as distribution artifact, but we're not quite there yet.

Are we not? Could you not just remove it in this PR? As of _this_ PR it's a useless pom - you could leave the directory structure the same but simply remove it and adjust the payara-bom to be a child of the top-level aggregator

pdudits:
Jenkins test please

pdudits:
* I dropped `payara.core.version` from dependency management -- `mvn release:prepare` completed up to the point where it wasn't properly configured and tried to immediately push a tag to github
* Merged in current master, respecing version upgrades to Hazelcast and EclipseLink
* Reset versions back to `6.2023.5-SNAPSHOT` with `mvn versions:set`. Current state should not block releasing next version until we decide further details on how the core vs main release will be done, and fine tune the breadth of core modules.

appserver/extras/certificate-management/certificate-management-admin/pom.xml
```diff
@@ -46,7 +46,7 @@
     <parent>
         <groupId>fish.payara.extras.certificate-management</groupId>
         <artifactId>certificate-management-parent</artifactId>
-        <version>6.2023.5-SNAPSHOT</version>
+        <version>6.0.0-SNAPSHOT</version>
```
Pandrex247, line 49: Wrong version.

```suggestion
        <version>6.2023.5-SNAPSHOT</version>
```

appserver/extras/certificate-management/certificate-management-cli/pom.xml
```diff
@@ -46,7 +46,7 @@
     <parent>
         <groupId>fish.payara.extras.certificate-management</groupId>
         <artifactId>certificate-management-parent</artifactId>
-        <version>6.2023.5-SNAPSHOT</version>
+        <version>6.0.0-SNAPSHOT</version>
```
Pandrex247, line 49: Wrong version

```suggestion
        <version>6.2023.5-SNAPSHOT</version>
```

appserver/extras/certificate-management/certificate-management-common/pom.xml
```diff
@@ -46,7 +46,7 @@
     <parent>
         <groupId>fish.payara.extras.certificate-management</groupId>
         <artifactId>certificate-management-parent</artifactId>
-        <version>6.2023.5-SNAPSHOT</version>
+        <version>6.0.0-SNAPSHOT</version>
```
Pandrex247, line 49: Wrong version

```suggestion
        <version>6.2023.5-SNAPSHOT</version>
```

appserver/extras/certificate-management/certificate-management-console-plugin/pom.xml
```diff
@@ -46,34 +46,30 @@
     <parent>
         <groupId>fish.payara.extras.certificate-management</groupId>
         <artifactId>certificate-management-parent</artifactId>
-        <version>6.2023.5-SNAPSHOT</version>
+        <version>6.0.0-SNAPSHOT</version>
```
Pandrex247, line 49: Wrong version

```suggestion
        <version>6.2023.5-SNAPSHOT</version>
```

appserver/extras/certificate-management/pom.xml
Pandrex247: Wrong version.
Relative path still wrong.

```suggestion
        <version>6.2023.5-SNAPSHOT</version>
        <relativePath>../../../core/core-parent/pom.xml</relativePath>
```

Pandrex247 (changes requested):

pdudits:
Jenkins test please

appserver/extras/certificate-management/pom.xml
pdudits: Thanks, forgot about cert-management (again)

Pandrex247 (approved):

pdudits:
Jenkins test please

pdudits:
Jenkins test please

pdudits:
Jenkins test please

Issue #6274, opened 2023-05-01
Bug Report: DataSource.getConnection() blows up on transaction with wrong status/FISH-7451


thehpi:
### Brief Summary

When using two entity listeners where one is using @Transactional(Transactional.TxType.REQUIRES_NEW) this causes problems when doing DataSource.getConnection() after observing an event using @Observes(during = TransactionPhase.AFTER_SUCCESS).

See the reproducer: https://github.com/thehpi/payara-after-success

### Expected Outcome

I expect to always be able to get a connection from a datasource, regardless of what other processing has been done before using entity listeners or observers.

### Current Outcome

An exception is thrown:

    RAR5029:Unexpected exception while registering component
    java.lang.IllegalStateException: Transaction is not active in the current thread.
    at com.sun.enterprise.transaction.JavaEETransactionImpl.checkTransationActive(JavaEETransactionImpl.java:756)
    at com.sun.enterprise.transaction.JavaEETransactionImpl.enlistResource(JavaEETransactionImpl.java:692)
    at com.sun.enterprise.transaction.jts.JavaEETransactionManagerJTSDelegate.enlistDistributedNonXAResource(JavaEETransactionManagerJTSDelegate.java:298)
    at com.sun.enterprise.transaction.JavaEETransactionManagerSimplified.enlistResource(JavaEETransactionManagerSimplified.java:469)
    at com.sun.enterprise.resource.rm.ResourceManagerImpl.registerResource(ResourceManagerImpl.java:152)
    at com.sun.enterprise.resource.rm.ResourceManagerImpl.enlistResource(ResourceManagerImpl.java:112)
    at com.sun.enterprise.resource.pool.PoolManagerImpl.getResource(PoolManagerImpl.java:211)
    at com.sun.enterprise.connectors.ConnectionManagerImpl.getResource(ConnectionManagerImpl.java:360)
    at com.sun.enterprise.connectors.ConnectionManagerImpl.internalGetConnection(ConnectionManagerImpl.java:307)
    at com.sun.enterprise.connectors.ConnectionManagerImpl.allocateConnection(ConnectionManagerImpl.java:196)
    at com.sun.enterprise.connectors.ConnectionManagerImpl.allocateConnection(ConnectionManagerImpl.java:171)
    at com.sun.enterprise.connectors.ConnectionManagerImpl.allocateConnection(ConnectionManagerImpl.java:166)
[...]

PR #6275, opened 2023-05-02
FISH-7451 Fixes for Issue 6274

thehpi:
## Description
This PR fixes issue [6274](https://github.com/payara/Payara/issues/6274).

When using @Transactional(Transactional.TxType.REQUIRES_NEW) the interceptor suspends and resumes the transaction but also the transaction in the current invocation. But when resuming it sets the tx in the invocation also when there was none when the interceptor started. This causes issues when using an observer method with @Observes(during = TransactionPhase.AFTER_SUCCESS) because when this method is executed the tx is already committed and when calling datasource.getConnection() this causes an exception because it is checking the tx status in the current invocation.

### New tests
No new tests because the current code do not allow for setting or mocking the invocationmanager/current invocation.

### Testing Performed
Reran the reproducer, no more exception.

### Testing Environment
MacOS Ventura 13.3.1, openjdk 11.0.17 2022-10-18, mvn version 3.8.6

## Documentation
See reproducer: https://github.com/thehpi/payara-after-success

## Notes for Reviewers
This is a quick fix, it might be better to fix it in TransactionalInterceptorBase so other interceptor implementations are also fixed but I cannot oversee this.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/cdi/transaction/TransactionalInterceptorRequiresNew.java
@@ -54,6 +54,7 @@ import jakarta.transaction.TransactionManager;
 import jakarta.transaction.TransactionalException;
 
 import com.sun.enterprise.transaction.TransactionManagerHelper;
+import org.glassfish.api.invocation.ComponentInvocation;
 
 /**
  * Transactional annotation Interceptor class for RequiresNew transaction type, ie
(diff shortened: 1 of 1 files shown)


thehpi:
Dear payara, I found a bug and spent considerable time to a) create a reproducer and b) to create a fix. I would be grateful if you could acknowledge this is a bug and have a look at the fix.

I would be happy to improve on the fix or maybe apply it in other places (interceptors).

bramn:
+1 for this bug fix.

aubi:
Jenkins test please

aubi:
@thehpi 
Thank you for your PR, I was happy to review it and it looks well and works well.
The only problem was the reproducer, PostgreSQL complained during delete about the UUID:
```
db_1      | ERROR:  operator does not exist: character varying = uuid at character 28
db_1      | HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
db_1      | STATEMENT:  DELETE FROM HANS WHERE (ID = $1)
```

As the topic of this PR is somewhere else, I modified it to use String instead and I've got the expected results.
So I'm going to merge it.
Thank you!

aubi (approved):
LGTM, verified

PR #6291, opened 2023-06-01
Fish 7426 fix payara config cache timing comm

aubi:
## Description
Renaming variables according to their content and fix code accordingly. Also do synchronization only when setting data, otherwise use the fast path.

The problem was, that the config didn't cache at all and always did search for the DurationSeconds.

### Testing Performed
I use the reproducer from Jira ticket, using only the simple request (basically simple REST):

```
wrk -c 400 -t 400 -d 120 http://localhost:8080/jee-mongo/api/hello-world/
```

Comparison:
* Glassfish: `Requests/sec:   3543.07`
* Current Payara 6: `Requests/sec:   1400.88`
* This version: `Requests/sec:  35984.01`

Feel free to correct this result as it seems really to fast.

### Testing Environment
Linux, OpenJDK 11 (compile) 17 (run)


aubi:
Jenkins test please

aubi:
Jenkins test please

Pandrex247 (approved):

PR #6297, opened 2023-06-06
FISH-7339: Update span attributes for MP OpenTelemetry Tracing 1.1

pdudits:
## Description

This PR adds new span attributes that will be required by Microprofile OpenTelemetry Tracing TCK 1.1 (a388a3f).

Then it tries to minimize number of MP Config lookups and reflection done on every request by introducing REST application-scoped `ResourceCache`, where information is cached for every invoked resource (Jakarta Rest method).

And finally rearranges helper classes, makes them package private where possible  and removes dead code from them.

Also adds new alias for span strategy of Microprofile Telemetry 1.0 -- `opentelemetry-1.13`, as that is more fair designation of where that strategy comes from.

### Blockers
Depends on payara/MicroProfile-TCK-Runners#226, as in-server workaround for OpenTracing TCK has been removed.

### Testing Performed

OpenTracing TCK (with patch above), MP Telemetry Tracing 1.1-SNAPSHOT

### Testing Environment

```
Apache Maven 3.8.7 (b89d5959fcde851dcb1c8946a785a163f14e1e29)
Java version: 11.0.18, vendor: Azul Systems, Inc.
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.15.0-69-generic", arch: "amd64", family: "unix"
```

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/OpenTracingHelper.java
+++ b/appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/OpenTracingHelper.java
@@ -149,6 +149,10 @@ public class OpenTracingHelper {
         determineSpanStrategy().augmentSpan(spanBuilder);
     }
 
+    public String getHttpRoute(ContainerRequest request, ResourceInfo resourceInfo) {
+        return SpanStrategy.OTEL_MP_TCK.determineSpanName(request, resourceInfo);
+    }
+
     /**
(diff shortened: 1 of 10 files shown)


aubi:
Jenkins test please

aubi (approved):
I tried build and OpenTelemetry TCK locally - both successfully.

pdudits:
Turns out we need https://github.com/payara/MicroProfile-TCK-Runners/pull/227 to make Telemetry Tracing 1.0 happy now, since we removed the workarounds for the TCKs from the server code.

pdudits:
Jenkins test please

PR #6300, opened 2023-06-07
FISH-6607: Remove dependency on core-bom from payara-bom

pdudits:
## Description

Payara BOM should not depend on Payara Core BOM, as that one contains lots of internal information. The public dependencies from Core BOM has been copied to Payara BOM (most of them were already there).

Core BOM import moved into nucleus parent, which is next in inheritance hierarchy.

Identified few of the properties that core BOM used, but didn't actually define, they well still in root.

### DIFF
--- a/api/payara-bom/pom.xml
+++ b/api/payara-bom/pom.xml
@@ -108,15 +108,6 @@
             </dependency>
 
             <dependency>
-                <groupId>fish.payara.server.core</groupId>
-                <artifactId>core-bom</artifactId>
-                <version>${payara.core.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-
-
-            <dependency>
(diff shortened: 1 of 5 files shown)


pdudits:
Jenkins test please

Pandrex247:
Adjusting target branch to not alter in-progress release as it shouldn't break anything.
We can discuss next week what we think the BOMs should look like rather than making snap decisions in the heat of the moment.

Pandrex247 (changes requested):
Needs rebase after altering branch target to not pull in extraneous commit

pdudits:
Jenkins test please

Pandrex247 (approved):
LGTM.
Tried building core solo - still worked.
Eyeballing the payara-bom and payara-api boms - seem fine (alebit payara-api is missing a project `<url>` but we've got a separate issue for that).

PR #6313, opened 2023-06-22
FISH-7499 Fix Missing and Incorrect URLs Tags

Pandrex247:
## Description
Some flattened poms were missing the URL tag (e.g. payara-api) which meant they could not be used to upload to Maven Central. This fixes that, and also corrects the incorrect tags that the flatten plugin was creating. The flatten plugin was appending the artifact IDs to the "base" URL and leading to entries such as `https://github.com/payara/Payara/payara-nucleus-parent/payara-parent/ejb/ejb-http-remoting/ejb-http-client` which don't resolve.

Also updates the flatten plugin to 1.5.0 because I hoped they might have fixed the URL issue themselves (but haven't). When doing so I noticed a number of duplications of the flatten plugin configuration which I've removed.

### Blockers
None

### New tests
None

### Testing Performed
Built with BuildExtras - checked all POMs had correct URL tag in flattened POM.

### Testing Environment
Windows 11, Zulu 11

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/api/payara-api/pom.xml
+++ b/api/payara-api/pom.xml
@@ -53,6 +53,9 @@
     <name>Payara-API</name>
     <description>Artefact that exposes public API of Payara Application Server</description>
 
+    <!-- Override to stop flatten plugin resolving it to an incorrect URL using parent artifact IDs -->
+    <url>https://github.com/payara/Payara/</url>
+
     <properties>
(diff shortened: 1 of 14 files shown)


Pandrex247:
Jenkins test please

breakponchito (approved):
LGTM

PR #6319, opened 2023-06-26
FISH-7452 FISH-7453 Start/Create Instance When Using JKS Keystores

aubi:
## Description
We add support to run with jks keystores, not only p12.
When an instance is created, there is a list of files to copy. The jks variants were added to this list (they are ignored if they don't exist).
Instances synchronize files with DAS, therefor it was necessary to add jks there as well. They are required, so a simple list was changed to a list of alternatives -- the first existing file is copied.

### Testing Performed
* Check out and build Payara/Payara6
* Replace cacerts.p12 with cacerts.jks, keystore.p12 with keystore.jks (e.g. from Payara 5 build)
* Change settings in config.xml in domain1, replace `.p12` with `.jks` for both keystores
* Start the domain: `payara/bin/asadmin start-domain -v --debug`
* Load the admin console
* **Create a new instance** (FISH-7453) from admin gui - it is now normally created.
* **Start the instance** (FISH-7452) in `payara5/glassfish`: `lib/nadmin  start-local-instance --debug  --node localhost-domain1 --sync normal --timeout 120 i1`
  * Alternatively, start it from admin gui.
  * Instance normally starts.

### Testing Environment
Linux, OpenJDK 11

### DIFF
--- a/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/SecureAdminBootstrapHelper.java
+++ b/nucleus/cluster/admin/src/main/java/com/sun/enterprise/v3/admin/cluster/SecureAdminBootstrapHelper.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2018] Payara Foundation and/or affiliates
+// Portions Copyright 2018-2023 Payara Foundation and/or affiliates
 
 package com.sun.enterprise.v3.admin.cluster;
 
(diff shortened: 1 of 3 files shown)


aubi:
Jenkins test please

Pandrex247 (changes requested):
This doesn’t work in its current state - it only works when you try to start the instance twice. The first attempt fails without a reason.

Reproduced on Windows:

* Set JDK version to 11
* Unzip Payara Enterprise 5 (commit ID `fb511f360eeff6b5dcba47fbc5688d642837f09c`)
* Remove `-SNAPSHOT` from property in `glassfish-version.properties`
* Start domain: `.\payara5\bin\asadmin start-domain --verbose --debug`
* Load admin console in a private browsing window
* Create _Deployment Group_ named `Group1` via admin console
* Create an _Instance_ with an autogenerated name and add to `Group1` via admin console
* Deploy `clusterjsp.war` to `Group1` via admin console
* Start `Group1` via admin console
  * Should start successfully
* Stop `Group1` via admin console
* Stop domain: `ctrl+c` in terminal window
* Stage server upgrade to Payara 6 (commit ID `3c1c6fdec8f21bf390187b51b1ffef9a3ab26843`): `.\payara5\bin\asadmin upgrade-server --usedownloaded D:\Git\Payara2\appserver\distributions\payara\target\payara.zip --stage`
  * Should work without issue
* Apply staged upgrade: `.\payara5\glassfish\bin\applyStagedUpgrade.bat`
  * Should work without issue
* Start domain again: `.\payara5\bin\asadmin start-domain --verbose --debug`
* Start admin console in a new private browsing window
* Start Deployment Group via admin console: **kabloom!**
  * No error given, just "could not start instance"
  * A look at the log of the instance (`payara5\glassfish\nodes\localhost-domain1\Laconic-Medusafish\logs\server.log`) just shows the original successful boot and shutdown, followed by it attempting to start again where it only prints the JVM options and nothing else

Interestingly attempting to start the instance again in exactly the same manner (via the _Deployment Groups_ page) will have it start successfully.

This can also be reproduced by attempting to start the instance via the _Instances_ page: it will fail to start on the first attempt without reason, and succeed on the second attempt.

aubi:
@Pandrex247 as agreed, this will be fixed as part of FISH-7563

eliotmartin:
Overriding merge protection at Engineering request.

eliotmartin (approved):

PR #6320, opened 2023-06-26
FISH-7480 Reduce Duplication in POMs & BOMs

Pandrex247:
## Description
Reduces the duplication between the POMs and BOMs.

Should hopefully be clear from the commit descriptions, but a summary of changes:
* Removes a completely dead POM
* Removes duplicate POM sections (developer, issueManagement, licenses, scm, url, flatten)
* Removes the temporary Jakarta staging repo which should have been cleared away ages ago
* Fixed the SCM metadata - they were falling afoul of the same issue as the URL in FISH-7499, and were wrong to begin with
* Makes sure repositories are resolved in the correct order for both Core and Server (check Central first)
* Removed the properties from the flattened BOMs - they don't need to be there(?) due to the following point
* Set dependency management to expand during flatten so versions are defined rather than hidden behind a property
  * This has the side effect that BOMs imported into the BOM (e.g. hk2-bom) are exploded into their constituent parts, but I don't think there's anything inherently wrong with that
* Corrects the developer section for Server distributions
* Removes the `<tag>` field - it's not required as per https://central.sonatype.org/publish/requirements/#scm-information
* Removes the dependencies from the flattened pom of Embedded.
  * The POMs we previously published to Central didn't include them, so I assume they're not currently excluded by flatten to be an accident.
* Removes the `jdk8` profiles which were dotted around - pointless and flatten wasn't removing it from Embedded.

### Blockers
None

### New tests
None

### Testing Performed
* First off, make a clean Maven repo to use: `mkdir D:\Downloads\FISH-7480`
* Build & install Core: `mvn clean install "-Dmaven.repo.local=D:\Downloads\FISH-7480" -f core/pom.xml`
  * You should see it resolve from Central first before attempting our own repos.
* Remove Core from reactor by commenting out `<module>core</module>` from top-level aggregator pom.
* Build all Server distributions: `mvn clean package "-Dmaven.repo.local=D:\Downloads\FISH-7480" -PBuildExtras`
  * Note I use `package` instead of install here - for reasons explained later
* Check that the server starts and admin console loads

Opened up the flattened poms of the following artefacts to compare them against their original unaltered counterparts to check no dependencies were missing (currently verifying...):
* Core-BOM
* Payara-API
* Payara-BOM
* Payara Server
* EJB HTTP Client
* Payara Micro
* Payara Embedded All

The BOMs will be exploded so there will be "extra" dependencies for things like Jersey and HK2, but I couldn't see anything missing

I also then checked it was possible to use the Core-BOM to build an extension outside of the reactor, picking [Notifiers](https://github.com/payara/Notifiers). This is why I didn't install the "Server" portion in our temporary repo:

* Check out the `Payara6-Community` branch
* Adjust the `payara.version` property to `6.2023.7-SNAPSHOT`
* Remove all dependencies from the top-level POM's dependency management section
* Add `core-bom` to top-level POM's dependency Management section
  ```
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>fish.payara.server.core</groupId>
                <artifactId>core-bom</artifactId>
                <version>${payara.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
    ```
* Replace in all POMs `fish.payara.server.internal.admingui` with `fish.payara.server.core.admingui`
* Build using our temporary repo: `mvn clean package -Dmaven.repo.local=D:\Downloads\FISH-7480`
* Drop `teams-notifier-core` and `teams-notifier-console-plugin` into our built Payara Server's `payara6\glassfish\modules` directory
* Start server
* Load admin console
* Check that Teams notifier is present under server-config > Notification

### Testing Environment
Windows, Zulu 11

## Documentation
N/A

## Notes for Reviewers
See above


Pandrex247:
Jenkins test please

Pandrex247:
Jenkins test please

api/payara-bom/pom.xml
```diff
@@ -953,6 +495,9 @@
                 <configuration>
                     <flattenMode>bom</flattenMode>
                     <flattenDependencyMode>all</flattenDependencyMode>
+                    <pomElements>
+                        <dependencyManagement>expand</dependencyManagement>
+                    </pomElements>
```
pdudits, line 500: Oh yes, the brutalist explicitness! Will probably save a lot of trouble down the road.

As a both positive and negative of this approach I would note, that this also expands the `<scope>import</scope>` ones resulting in quite long flattened pom, but on the other hand it will be always clear where this comes from.

pdudits (approved):

PR #6339, opened 2023-07-17
FISH-7620 Fixed Test cases for MacOS build

rohan267:
FISH-7620 Fix timeout issue for Mac OS build

JIRA: https://payara.atlassian.net/browse/FISH-7620 - Fixing Mac OS build

## Description
At present, WatchService API takes around 10 seconds to update file system on Mac OS, so added delay of 10 seconds in case of assertion failure which fix Mac OS build. Have to ignore the assertion In a case where WatchService is failing to read the path pattern `watcher-files.example`

### New tests
Fixed existing tests with appropriate delays in case of Mac OS

### Testing Performed
Maven built succesfully

### Testing Environment
Maven: 3.9.2
Java version: 11.0.18, vendor: Oracle Corporation
Default locale: en_IN, platform encoding: UTF-8
OS name: "mac os x", version: "13.4", arch: "aarch64", family: "mac"

### DIFF
--- a/nucleus/payara-modules/nucleus-microprofile/config-service/src/test/java/fish/payara/nucleus/microprofile/config/source/DirConfigSourceTest.java
+++ b/nucleus/payara-modules/nucleus-microprofile/config-service/src/test/java/fish/payara/nucleus/microprofile/config/source/DirConfigSourceTest.java
@@ -77,7 +77,9 @@ public class DirConfigSourceTest {
     private static DirConfigSource source;
     private static ScheduledExecutorService exec = Executors.newScheduledThreadPool(3);
     private static ConfigProviderResolverImpl configService;
-    
+    private final int DELAY_MILLISECONDS = 10;
+    private final int TIMEOUT_MILLISECONDS = 10000;
+
     @BeforeClass
     public static void setUp() throws IOException {
         testDirectory = Files.createTempDirectory("microprofile-config-test-");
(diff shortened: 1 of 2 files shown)


rohan267:
Jenkins Test please

nucleus/payara-modules/nucleus-microprofile/config-service/src/test/java/fish/payara/nucleus/microprofile/config/source/DirConfigSourceTest.java
```diff
@@ -77,7 +77,9 @@ public class DirConfigSourceTest {
     private static DirConfigSource source;
     private static ScheduledExecutorService exec = Executors.newScheduledThreadPool(3);
     private static ConfigProviderResolverImpl configService;
-    
+    private final int DELAY_MILLISECONDS = 10000;
```
pdudits, line 80: Delay should not be greater than 10 milliseconds. We don't want to slow down other OS, just the ones with slow file watcher. Timeout on the other hand can safely be 10 or 15 seconds.

pdudits (changes requested):

nucleus/payara-modules/nucleus-microprofile/config-service/src/test/java/fish/payara/nucleus/microprofile/config/source/DirConfigSourceTest.java
rohan267: Delay is occurring only in case of assertion error (Line 482), which will happen mostly likely on Mac OS only

pdudits: We don't need the `while(true)` if the assertion fails once. It is safe method to try many times so the test case passes as soon as possible.

rohan267:
Jenkins tes please

nucleus/payara-modules/nucleus-microprofile/config-service/src/test/java/fish/payara/nucleus/microprofile/config/source/DirConfigSourceTest.java
```diff
@@ -470,20 +470,17 @@ public void testPropertyWatcher_RunFilesNewUpdate() throws Exception {
 
     static void await(long timeout, int delay, Runnable test) {
         long expireAt = System.currentTimeMillis() + timeout;
-        while(true) {
```
pdudits, line 472: Why did you remove the loop? It now only tests once.

rohan267:
Jenkins test please

pdudits (approved):

Issue #6344, opened 2023-07-25
Bug Report: Payara doesn't run on latest JDK patch levels / FISH-7665


lprimak:
### Brief Summary

With latest JDK patch levels (20.0.2, 17.0.8, 11.0.20) there is a breaking change that prevents Payara from running at all.
I put the following at the end of `glassfish/config/asenv.conf` to get it to run:
```
export JAVA_TOOL_OPTIONS
JAVA_TOOL_OPTIONS="-Djdk.util.zip.disableZip64ExtraFieldValidation=true --add-opens=java.base/java.io=ALL-UNNAMED"
```
Below is the exception:
```
java.util.zip.ZipException: Invalid CEN header (invalid zip64 extra data field size)
	at java.base/java.util.zip.ZipFile$Source.zerror(ZipFile.java:1750)
	at java.base/java.util.zip.ZipFile$Source.checkExtraFields(ZipFile.java:1276)
	at java.base/java.util.zip.ZipFile$Source.checkAndAddEntry(ZipFile.java:1219)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1689)
	at java.base/java.util.zip.ZipFile$Source.<init>(ZipFile.java:1464)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1427)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:723)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:250)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:179)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:193)
	at org.apache.felix.framework.util.SecureAction.openZipFile(SecureAction.java:802)
	at org.apache.felix.framework.util.WeakZipFileFactory$WeakZipFile.<init>(WeakZipFileFactory.java:169)
	at org.apache.felix.framework.util.WeakZipFileFactory$WeakZipFile.<init>(WeakZipFileFactory.java:151)
	at org.apache.felix.framework.util.WeakZipFileFactory.create(WeakZipFileFactory.java:78)
	at org.apache.felix.framework.cache.JarRevision.<init>(JarRevision.java:83)
	at org.apache.felix.framework.cache.BundleArchive.createRevisionFromLocation(BundleArchive.java:804)
	at org.apache.felix.framework.cache.BundleArchive.reviseInternal(BundleArchive.java:491)
[...]

PR #6262, opened 2023-04-15
Bump felix bundle plugin to 5.1.8

ctabin:
Updates the Apache Felix [maven-bundle-plugin](https://mvnrepository.com/artifact/org.apache.felix/maven-bundle-plugin) to latest version.

This fixes a ConcurrentModificationException while [building](https://docs.payara.fish/community/docs/General%20Info/Build%20Instructions.html) the source with Apache Maven 3.9.0 and JDK 17.

PR #6348, opened 2023-08-01
FISH-7665: fixing issue creating osgi bundle for libpam4j

breakponchito:
Fixing issue to start server caused by the libpam4j-repackaged.jar dependency

## Description

This is fix for a bug when creating the jar bundle for the libpam4j, older version of the maven-bundle-plugin 
is generating a jar file with error in zip headers. To fix, it is needed to use last version of
the plugin 5.1.9

### Testing Performed

Manual testing with most recent versions of JDK 11 and 17. Now the server can start with both JDK's

### Testing Environment

windows 11, Azul JDK 11.0.20 and Azul JDK  17.0.8

### DIFF
--- a/appserver/packager/external/libpam4j/pom.xml
+++ b/appserver/packager/external/libpam4j/pom.xml
@@ -77,6 +77,7 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
+                <version>5.1.9</version>
                 <configuration>
                     <instructions>
                         <Embed-Dependency>


breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

Pandrex247 (approved):

ctabin:
Hi, I wonder if this would also be related to #6262 ? :thinking:

Pandrex247:
It is related, but it's more focussed on fixing https://github.com/payara/Payara/issues/6344.
We can't currently bump the bundle plugin version globally at this point in time due to changes in the expected OSGi imports (see my comment [here](https://github.com/payara/Payara/pull/6262#issuecomment-1557154535)).

Pending us properly looking into how we intend to deal with OSGi (sprinkling `-noimportjava` everywhere vs. reconfiguring the imports/exports for JDK packages), we're just incrementing the version here to solve the immediate issue with the zip headers.

PR #6355, opened 2023-08-03
Fish 7356 automate cleanboot test

simonladen:
Automated test using Python 3 and Playwright.
The goal is to automate the Clean Boot test performed with each release candidate:
- start the domain
- create an instance, start the instance, check the logs for any warning/error, stop the instance, delete the instance
- create a deployment group with 2 instances, start the group, check the logs for any warning/error, stop the group, delete the group and the instance
- stop the domain

Instructions to run it are in the README.md

Tested with Payara 5 and 6

### DIFF
--- a/appserver/tests/functional/cleanboot/README.md
+++ b/appserver/tests/functional/cleanboot/README.md
@@ -1,0 +1,34 @@
+# Cleanboot File Tests
+
+Test a clean boot with Payara Platform.
+
+Creates an instance, starts it, checks the logs for any trace above INFO, stops the instance(s), deletes the instance(s)
+
+Creates a deployment group, create instances for that group, starts the group, checks the logs for any trce above INFO, stops the instances, delete the instances, delete the group
+
+Logs above INFO are collected in files, to be assessed manually (some warnings are known issues and can be ignored)
+
+
+## Setup
+
+Set PAYARA_HOME to the tested Payara. For example:
+
+    PAYARA_HOME=/path-to-my-payara-src/appserver/distributions/payara/target/stage/payara6
+
+Install the Pytest plugin:
+
+    pip install pytest-playwright
+
+Install the required browsers:
+
+    playwright install
+
+## run the tests:
+
+    pytest testcleanbootcommand.py
+
+use `--headed` to run with headless=False
+
+use `-s` to see the standard output of the tests
+
+to debug the test, set the environment variable PWDEBUG=1
--- a/appserver/tests/functional/cleanboot/README.md
+++ b/appserver/tests/functional/cleanboot/README.md
@@ -13,7 +13,7 @@ Logs above INFO are collected in files, to be assessed manually (some warnings a
 
 Set PAYARA_HOME to the tested Payara. For example:
 
-    PAYARA_HOME=/path-to-my-payara-src/appserver/distributions/payara/target/stage/payara6
+    export PAYARA_HOME=/path-to-my-payara-src/appserver/distributions/payara/target/stage/payara6
 
 Install the Pytest plugin:
 
@@ -22,6 +22,8 @@ Install the Pytest plugin:
 Install the required browsers:
 
     playwright install
+
+Note: you may also need to run the command `playwright install-deps`
 
 ## run the tests:
 
--- a/appserver/tests/functional/cleanboot/admingrouplib.py
+++ b/appserver/tests/functional/cleanboot/admingrouplib.py
@@ -1,0 +1,131 @@
+#!/bin/python3
+from playwright.sync_api import Page, expect
+
+
+def create_group(page: Page, nameGroup, nameInstances):
+	# Open the admin page 
+	page.goto('http://localhost:4848')
+	page.wait_for_selector('div[id="treeForm:tree_children"]')
+	
+	# Click on the option Deployment Groups in the menu
+	instance_button = page.get_by_role("link",name="Deployment Groups", exact=True)
+	instance_button.click()
+	page.wait_for_selector('div[id="propertyForm:dgTable"]')
+
+	# Expect the title to contain Instances
+	expect(page).to_have_title("Deployment Groups")
+
+	# Create new instance
+	page.get_by_role("Button", name="New...").click()
+	page.wait_for_selector('input[id="propertyForm:propertySheet:propertySectionTextField:NameTextProp:NameText"]')
+	page.get_by_role("textbox").fill(nameGroup)
+	page.get_by_role("Button", name="OK").click()
+	page.wait_for_selector('input[value="New..."]')
+
+	# Check for the presence of the new instance in the table
+	group_link = page.get_by_role("link", name=nameGroup, exact=True)
+	expect(group_link).to_be_visible()
+	group_link.click()
+	page.wait_for_selector('table.Tab1TblNew_sun4')
+
+	# Create instances in the group
+	for nameInstance in nameInstances:
+		group_tabs = page.locator('table.Tab1TblNew_sun4')
+		group_instance_tab = group_tabs.get_by_role("link", name="Instances", exact=True)
+		group_instance_tab.click()
+
+		# Create new instance
+		page.get_by_role("Button", name="New...").click()
+		page.wait_for_selector('input[id="propertyForm:propertySheet:propertSectionTextField:NameTextProp:NameText"]')
+		page.get_by_role("textbox").fill(nameInstance)
+		page.get_by_role("Button", name="OK").click()
+		page.wait_for_selector('input[value=" Save "]')
+
+		# Check for the presence of the new instance in the table
+		instance_link = page.get_by_role("link", name=nameInstance, exact=True)
+		expect(instance_link).to_be_visible()
+
+		group_general_tab = group_tabs.get_by_role("link", name="General", exact=True)
+		group_general_tab.click()
+
+
+	# return to homepage
+	page.goto('http://localhost:4848')
+
+def start_groups(page: Page):
+	# Open the admin page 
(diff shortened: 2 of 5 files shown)


simonladen:
Jenkins test please

lprimak:
Not sure about all this python stuff. Payara already had arquillian drone / grapheme way of testing UIs. Not sure that it’s worth introducing a totally different language and framework

phillipross:
playwrite does have a java api, yes?

graphene and drone were promising back in the day but are barely maintained now 😞  it seems people have moved on to raw selenium, selenide, testcontainers, et al.

aubi:
Yes, Playwright has Java API. But when it is about controlling U + running commands, it goes quickly beyond Selenium/Playwright. So it is easier to write Python tests for such usecases.
Java API for running commands is not that simple.

lprimak:
> it goes quickly beyond Selenium/Playwright

That's why Arquillian / Graphene / Drone are so good. Biggest bang for minimum code.
And it's begin maintained... lots of momentum and new updates coming this year.
Arquillian 1.7 has been now released and Drone / Graphene to be followed shortly.

phillipross:
> > it goes quickly beyond Selenium/Playwright
> 
> That's why Arquillian / Graphene / Drone are so good. Biggest bang for minimum code. And it's begin maintained... lots of momentum and new updates coming this year. Arquillian 1.7 has been now released and Drone / Graphene to be followed shortly.

This is exactly what Arquillian is for... orchestrating the app server (et al) for functional tests.  Plus it has the added benefit of running on the JVM and being served out of maven central.  no extra runtimes, no npm, no pypi.  Payara literally maintains arquillian containers for the appserver.  Doesn't jenkins use groovy for scripting?  🤷‍♂️

simonladen:
Now, maybe I was wrong, but the purpose of using Python was to automate the manual test case, as close as I could to the "real user's" experience. Meaning that it starts and stops the domain running the actual asadmin commands, not via an integrated solution handled by Arquillian. (and if Arquillian actually uses asadmin commands to start/stop the domain, then I'm wrong)
Also, this test will be performed using different JDK (usually jdk 8, 11 and 17 - jdk8 is only for Payara5), so it made sense in my mind to not rely on Java to execute the test. The use case would be to set a JDK, run the test, check the results, set a different JDK, run the test, check the results... If the test was run in Java, there would be a risk that it would behave differently with a different JDK, and potentially alter the results. What I want to check is that Payara Server behaves correctly with different JDKs, not that this test is compatible with these JDKs.

phillipross:
Each arquillian container is implemented separately depending on whether the use cases involves invoking the full appserver locally, a remote domain server, a payara microinstance, an embedded instance, etc.  I believe the payara managed arquillian container does use the java process builder api to build a commandline but @lprimak probably can speak more to those details as I think I've seen his username in the controller code commits.

Maven has "toolchains" configuration that allows multiple JDKs to be configured and various plugins (surefire/failsafe) are toolchains-aware.  I believe the payara arquillian containers project is already using the maven invoker plugin in its integration test module.  The invoker plugin invokes a separate maven process using separate toolchains (JDKs) that are part of the integration test configurations.

As @aubi may be implying, this maven way of doing things may be too inflexible, constrictive, too much work for simple use cases.  This may be true, and is usually what I hear from users of gradle, but gradle uses groovy and kotlin for scripting rather than over-optimizing and leveraging a completely different ecosystem like python or node.  Payara uses Jenkins for CI, right?  And Jenkins uses groovy?  and for scripting purposes they prescribe just using shell scripts? 

These things in the other ecosystems often come with their own runtime and dependency compatibility issues, so I usually recommend against that.  I'd prefer to stay on the JVM when my opinion counts 😉

lprimak:
> These things in the other ecosystems often come with their own runtime and dependency compatibility issues, so I usually recommend against that. I'd prefer to stay on the JVM when my opinion counts

This is exactly what I am trying to imply as well.
There is nothing that python can do that java can't do better, and introducing a whole new toolchain / language / framework inside Payara tests just seems to bring up questions and cognitive load that simply do not need to be brought up.

aubi:
Thanks everybody for opinions, it's good to discuss options we could miss.
The driving motivation is to make the tests readable and easy to write. We need a combination of running commands (mostly asadmin, but not only) and UI control.
I like the basic idea to stay at JVM/maven, because it's simpler for setup.
It would be also great to have it a part of JUnit, so it is a natural part testing.
BUT: Arquillian works with one deployment. Stopping server, changing parameters a running it again is a problem.
Using Java to run system commands is much more complex than in Python.
Groovy has IMHO crazy syntax.
Using Maven for more comolex tests lead to huge poms...
Therefor we see Python as the easiest way.

Can you provide some examples, how to solve the issues with your suggested ways? Thank you!

lprimak:
> The driving motivation is to make the tests readable and easy to write.

Agree 100%. However, with my Java / Payara background, I can't make heads or tails of the code in this PR. This just proves that this methodology doesn't make the cut for the above motivation :)

>  Stopping server, changing parameters a running it again is a problem.

I don't believe so.

> Can you provide some examples

Payara samples already has infrastructure to change parameters and restart the server. You can use those as an example.

More examples:

https://docs.flowlogix.com/#section-jee - See section about `Configuring ShrinkWrap (Arquillian) test artifacts`

Also, see https://github.com/flowlogix/flowlogix/tree/main/jakarta-ee/jee-examples for some very simple Drone / Graphene tests.

FYI: I have nothing against python / playwrite in general, however, in Payara's particular use case, I think it's a no-brainer to stick with the Java / Jakarta EE / Arquillian / Graphene - it's just simpler that way.

phillipross:
> Thanks everybody for opinions, it's good to discuss options we could miss. The driving motivation is to make the tests readable and easy to write. We need a combination of running commands (mostly asadmin, but not only) and UI control. I like the basic idea to stay at JVM/maven, because it's simpler for setup. It would be also great to have it a part of JUnit, so it is a natural part testing.

Maven's failsafe and surefire plugins support JUnit, TestNG, cucumber, spock, and probably others, so there's a lot of options and flexibility available for you there.  Test suites can be very simple and straightforward, or intricate and complex, but the factors that seem to be the most important over time is maintainability (which can be a function of ease of reading, ease of writing, and ease of understanding).

> BUT: Arquillian works with one deployment. Stopping server, changing parameters a running it again is a problem.

I'm not sure this is entirely true, but I suspect you're referring to specific use cases.  In general, it's
possible to use multiple deployments with arquillian.  I scanned the flowlogix code that @lprimak provided and
didn't actually see the examples, but I saw a dependency on the "arquillian-suite-extension" which is one of the
extensions that let's you structure deployments and tests across your suite definitions.

The maven invoker plugin, which I didn't see in the flowlogix project but I've seen in the payara arquillian
containers project, allows you to invoke a separate maven process from the one doing the primary build.  These
separate invocations can be with different JDKs and different profile specifications than the primary one.

> Using Java to run system commands is much more complex than in Python.

Is it?  If I look at the code in this PR, I see a bunch of string concatenation with some error checking, and then
passing the final result to subprocess.  The java equivalent would be the same but passing the final result to runtime
exec.  There are more complicated facilities that Java has, but you're not forced to use them if you don't care
about all the edge cases or sync/async, piping/streaming, threading, interprocess comms, etc.

> Groovy has IMHO crazy syntax.

I'm surpised to see this statement, as I'd say groovy looks closer to python, but I guess it depends on the person's
prior knowledge and experience.  In my experience, a lot of folks initially introduce groovy and/or kotlin into the codebases
via the test code, to begin learning via transition.  What I specifically like about kotlin is that it interoperates
with java code and libraries so well, and it allows you to decorate existing code with extension functions/operators and
build up your own domain-specific language.  Maintainers of test code especially seem to love this as it's something
that is not forced on them, but available to them to learn over time.

> Using Maven for more comolex tests lead to huge poms... Therefor we see Python as the easiest way.

The size of the poms will depend on how you implement it, but complex tests are... complex.  There are different ways to manage that complexity and maven provides a lot of flexibility there (with learning curve of course).

If the purpose is to build/maintain a test suite for internal use only, then it would make sense to tap other techs and ecosystems to optimize for the goals.  But if the goal is to have the test suite part of the build, and the build tool is maven, and community collaboration/contribution is expected to be leveraged, then it's probably better to stick to the conventions of the community and expect that contributors eager to collaborate will baulk at having to use npm, pip, pytest, etc when there are already existing alternatives within the same ecosystem.

The references provided by @lprimak are actually a really nice model representation of already existing and mature alternatives.  I suspect that a big problem faced by many technologists today are the lack of accessibility to the knowledge that these things already exist 😬

lprimak:
Just want to reiterate that payara samples asadmin tests already do the change configuration / payara restart tests that you are looking for: https://github.com/payara/Payara/tree/master/appserver/tests/payara-samples/asadmin

aubi:
@lprimak I tried to look at these examples and they prove it's not that easy:

Unstable, "Requires completely new domain. Side effects of other tests break this one.": https://github.com/payara/Payara/blob/master/appserver/tests/payara-samples/asadmin/src/test/java/fish/payara/samples/asadmin/SetHazelcastConfigurationTest.java#L53

Payara restart during test is a bit dirty hack, because it doesn't exit Java. What is NOT possible, is to stop the domain and start again AFAIK.

FYI: I already did one test in Python, which I didn't find a way, how to do it in maven/Arquillian: https://github.com/payara/Payara/tree/master/appserver/tests/functional/postboot

Anyway -- I'll start a new PR, trying to follow your suggestions and do the same test all in Java. And I'll ask you to review it as well.

aubi (approved):
LGTM, we will further investigate a way, how to do these tests in Java

PR #6384, opened 2023-08-29
FISH-7740 Disable Extra Zip Validation in Docker Images

Pandrex247:
## Description
Disables the extra zip validation added in JDK 11.0.20 & 17.0.8 which breaks the libpam4j package we bundle in Payara Server & Micro. While no longer an issue on the **most recent** releases due to fix introduced in https://github.com/payara/Payara-Enterprise/pull/899, if we ever decide to rebuild an older Community image utlising the newest JDKs we either need to add this option or backport the fix to all previous releases!

Since the `create-jvm-options` command requires the domain be running (and it cannot start due to the missing JVM option!) I've had to resort to inserting the option via sed (because I can't think of a better way). I have attempted to preserve the indentation because it pleases me.

### Blockers
None

### New tests
None

### Testing Performed
Built latest Payara Server & Micro and their respective Docker images and let the unit tests run - all worked: `mvn clean install -PBuildDockerImages`
Started Payara Micro image and deployed clusterjsp - worked.

Rebuilt Docker images utilising an older version of Payara without the fix: `mvn clean verify -Dpayara.version=6.2023.4 -PBuildDockerImages -pl :docker-images -amd`

### Testing Environment
Windows 11

## Documentation
https://github.com/payara/Payara-Documentation/pull/307

## Notes for Reviewers
Also adds missing copyright blocks so you may wish to filter the diff on only the first two commits.

The Payara Micro way of adding it as a default JVM_ARG is slightly fragile in that it can be overridden, but I feel that it's slightly better than hardcoding it into the entrypoint.sh file like -Xss is since this should only be required for older versions of Payara.

### DIFF
--- a/appserver/extras/docker-images/basic/pom.xml
+++ b/appserver/extras/docker-images/basic/pom.xml
@@ -1,4 +1,45 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+
+    Copyright (c) 2023 Payara Foundation and/or its affiliates. All rights reserved.
+
+    The contents of this file are subject to the terms of either the GNU
+    General Public License Version 2 only ("GPL") or the Common Development
+    and Distribution License("CDDL") (collectively, the "License").  You
+    may not use this file except in compliance with the License.  You can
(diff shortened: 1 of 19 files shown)


aubi (approved):
LGTM, tried both

PR #6395, opened 2023-09-04
Increment version numbers 6.2023.10

abdulrahim458:
Title - increments version for release and next dev iteration


abdulrahim458:
Jenkins test please

Pandrex247 (approved):

lprimak:
@abdulrahim458 @Pandrex247 Please reconsider changing the version # in git repo for every release. This is wasteful and introduces much pollution in the repo.

https://www.cloudbees.com/blog/apache-maven-continuous-deliverydeployment-devoptics-teams-approach

Especially this:
> Solution
Who says we have to ever push those [maven-release-plugin] prepare … commits, what would happen if we only pushed the tags?
The kind of Git History we want, linear commits to master with tags branching off for the Maven Release plugin's commits
The pom.xml versions stay the same because they are the same:
Merge conflicts resulting from the Maven Release Plugin commits are eliminated - because the master branch never sees those commits
Tags reflect releases
There are some potential negatives:
Depending on how you automate things, you may need to use throw-away checkout for releasing
It may be harder to determine what version a feature landed in
We have to determine the version number for each release somehow as it is no longer stored in the pom.xml

I have been using this for the last 2+ years and it works great!

abdulrahim458:
Hi @lprimak , thanks for solution. We will look in to this.

PR #6400, opened 2023-09-08
FISH-6759 change search for CDI context, try JNDI first

aubi:
## Description
Search for CDI doesn't work well in EAR packaging.
The current way searches through all modules, which can find the calling class. And because EJB module sees the WAR classes, it may be (and often is) chosen instead of the WAR, causing further CDI search failures.

This change tries to find "java:comp/BeanManager" in JNDI, which works more reliably.

The change is inspired by same behaviour in Jersey: https://github.com/eclipse-ee4j/jersey/blob/2.x/ext/cdi/jersey-cdi1x/src/main/java/org/glassfish/jersey/ext/cdi1x/internal/DefaultBeanManagerProvider.java#L45

### Testing Performed
Trying to run TCKs (TO BE UPDATED)

### Testing Environment
Linux, OpenJDK 8+11

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/GlassFishWeldProvider.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/GlassFishWeldProvider.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-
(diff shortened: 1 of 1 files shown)


aubi:
Jenkins test please

aubi:
Jenkins test please

aubi:
Not needed for Payara 6, closing

PR #6425, opened 2023-10-02
FISH-6519 Don't Ever Delegate JSF Classes if using Bundled JSF

Pandrex247:
## Description
Port from Enterprise 5 to keep things in sync. This fix was primarily written for the case of attempting to use a JSF 2.2 implementation bundled with PrimeFaces while the server was using JSF 2.3, but the behaviour change is probably still applicable to Payara 6.

The `useBundledJsf` property can be defined in an application's deployment descriptor to specify that the JSF version bundled with the application should be used over what is provided by the server. This allows you to use something like MyFaces, a different version of Mojarra, or some other JSF implementation.

PrimeFaces has a mechanism where it determines which version of JSF to activate (since it supports multiple versions) by doing a lookup on classes. In the case of PrimeFaces 11 it determines if an application is JSF 2.3 by attempting to look up the `javax.faces.component.UIImportConstants` class. Before this change, on Enterprise 5 if bundling an application with JSF 2.2 the PrimeFaces lookup on the JSF 2.3 class would still succeed because after the `WebappClassLoader` had attempted (and failed) to find the class by looking in the bundled JSF implementation, it would then delegate the loading to its parent which _would_ find it since JSF 2.3 _is_ on the class path.

This change makes it so that this delegation never happens - if it can't find it "locally", just give up and throw a CNFE. 

Before: override any JSF classes with those that can be found in the bundled JSF implementation, but still attempt to provide JSF classes from the server's implementation if they can't be found in the bundled JSF implementation.
After: Only attempt to load JSF classes using the bundled implementation.

Note, that this only applies to WARs. You cannot have a JSF implementation added as a library to an EAR and have a WAR also within the EAR use that implementation. You would have to have the bundled JSF within the WAR within the EAR.

Original skeleton for reproducer app taken from https://github.com/StevenHachel/glassfish-test-jsf22

### Blockers
None

### New tests
New sample under _payara-samples_.
New tests to demonstrate the application deploying successfully, and the use (or disuse) of a bundled JSF implementation.

### Testing Performed
Ran new arquillian tests.

### Testing Environment
Windows 11

## Documentation
https://github.com/payara/Payara-Documentation/pull/324

## Notes for Reviewers
Unsure if I should also be cancelling delegation lookup for JSF resources (e.g. `jakarta/faces/LogStrings.properties`)...


Pandrex247:
Jenkins test please

luiseufrasio (approved):
LGTM!

PR #6442, opened 2023-10-10
Increment version numbers core 6.9.0 snapshot

abdulrahim458:
Increments the version number for Core.


abdulrahim458:
Jenkins test please

Pandrex247 (changes requested):
This is wrong, should be 6.7.0

PR #6482, opened 2023-11-16
FISH-8068 - fix cleanboot test for linux

simonladen:
the cleanboot test fails when run in Linux. It appears that the first log entry opened in a browser has a blank log message, and it is surely caused by the lack of waiting for the page to load. (it is not reproduced when doing it manually)


simonladen:
jenkins test please

simonladen:
jenkins test please

simonladen:
jenkins test please

PR #6494, opened 2023-11-29
ITOPS-2602 Make Grizzly HTTP Test More Predictable

JamesHillyard:
## Description
In new Jenkins, the Grizzly tests failed ~ 20% of the time due to `Address already in use` errors. It's fairly well documented that after closing a socket, the port will enter a TIME WAIT state to prevent delayed packets arriving to a different listener. The OS will make the port available again after a period of time that is not deterministic.

This PR introduces a wait period of up to 1 minute to ensure the port is available before beginning the test.

### Blockers
None

### Testing Performed
Ran the tests locally where it was always observed to pass and they continue to pass 100% of the time with no added delay in running the test suite. 
Ran 15 Jenkins builds and none failed where it's expected ~ 3-5 would fail. Verified some tests took 20+ seconds showing the await functionality was activated to make this test pass.

### Testing Environment
New Engineering Jenkins:
- Maven 3.6.3
- Zulu 11.0.19
- Ubuntu 22.04 LTS

## Documentation
N/A

## Notes for Reviewers
This does **not** force all tests to wait for 1 minute before starting. Awaitility will continue to check the condition and start the test immediately after the condition returns true. In most cases there is no observed time gain. I have observed in some cases the tests take ~ 30-50 seconds to execute up from ~ 2 seconds. This is infrequent and significantly better than the build failing.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -190,6 +190,8 @@
 
         <payara.deployment.transformer.version>1.3</payara.deployment.transformer.version>
         <payara.transformer.version>0.2.12</payara.transformer.version>
+
+        <awaitility.version>4.2.0</awaitility.version>
     </properties>
 
     <issueManagement>
--- a/nucleus/grizzly/config/pom.xml
+++ b/nucleus/grizzly/config/pom.xml
@@ -144,5 +144,11 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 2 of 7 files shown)


eliotmartin (approved):
I still think someone should sort the duplicate shutdownNetwork/shutdown methods in GrizzlyConfig but this is shown to fix the issue with the OS not closing ports in a timely manner and should be merged.

aubi (approved):

PR #6522, opened 2024-01-09
FISH-8072 - cleanboot test with arquillian

simonladen:
Cleanboot test converted in Java. 
It uses Arquillian to start/stop the server.
Playwright opens a browser (in headless mode by default), loads the admin console, then runs 2 tests
- create an instance, start the instance, check the logs, stop the instance and delete the instance. It fails if the logs contain errors.
- create a deployment group with 2 instances, start the group, check the logs for both instances, stop the group, delete the group and the instances. It fails if the logs contain errors.
Any warning in the logs is printed in the output, for manual verifications when necessary.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,7 +80,7 @@ pipeline {
         stage('Run Payara Samples Tests') {
             steps {
                 echo '*#*#*#*#*#*#*#*#*#*#*#*#  Running test  *#*#*#*#*#*#*#*#*#*#*#*#*#*#*#'
-                sh """mvn -V -B -ff clean install --strict-checksums -Ppayara-server-remote \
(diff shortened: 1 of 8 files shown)


aubi:
It is necessary to install dependencies beforehand or have working sudo:
```
[INFO] --- exec-maven-plugin:3.1.1:exec (install-playwright-dependencies) @ cleanboot ---
Installing dependencies...
Switching to root user to install dependencies...
[sudo] password for aubi: 
```

aubi:
Because this test requires some system libraries and it is not able to install them alone (requires sudo), I suggest to create a profile for tests using Playwright.

aubi (approved):
Great, let's start writing such tests!

PR #6523, opened 2024-01-10
FISH-7821 FISH-8190 Multiplatform Docker Images for Server and Micro

Pandrex247:
## Description
Follow up / replacement for https://github.com/payara/Payara/pull/6521

Allows creation and deployment of multiplatform Docker images.

Azul provide AMD64 and ARM64 versions of their azul/zulu-openjdk and azul/zulu-openjdk-alpine images, which we can leverage to create [multiplatform images](https://docs.docker.com/build/building/multi-platform/).

Buildx has a quirk that it cannot utilise locally cached images - it can **only** pull from _published_ images. This comes up when trying to build the Payara Server images, which pull from the payara/basic image. Since the payara/basic image utilised by all Server images is versioned in the same manner as them currently (as in, the `payara/server:6.2024.1-SNAPSHOT` image expects to pull from a `payara/basic:6.2024.1-SNAPSHOT` image), this breaks the build unless we _always_ publish (no local builds allowed).  The logic behind the original split was to reduce redundancy and optimise the build, but this splitting off wasn't done with buildx in mind. The move to buildx makes maintaining this image awkward, and it will quite possibly frequently catch us out. As an example, if you update the JDK version you **won't** necessarily see that it breaks the server images until you fully deploy the image and _then retest_ because the server images will still pull down the previously published image. So your local maven build will pass, but upon deployment and retesting it may suddenly start failing. Removing the base image makes the build of the images slower, and there's duplicate steps, but it makes it much clearer if something is breaking or not.

Unfortunately (as far as I can tell anyway), a combination of the way we have the build configured with various executions for the different JDK versions rather than a single "default" execution, and the way buildx works where when deploying it actually builds it again first, we have to essentially have a duplicate build config.

There is currently a bug with the maven plugin where the _filter_ settings are ignored during the _deploy_ phase - combined with the above point of buildx requiring us to rebuild the image, this has led me to come up with a workaround where we filter using ant as we create the individual docker files for the various JDK versions (rather than during the build of the Docker images).

I have also had to alter the basic (now server-base) image a little as it was directly downloading Tini. This led to an issue on ARM machine because it was an AMD64 version of Tini. Tini is now installed via the package manager to get around this.

### Blockers
None

### New tests
None

#### On my local machine
* Built Server and Micro so that they're present in maven cache: `mvn clean install -DskipTests -T 1C`
* Built all Docker images: `mvn clean install -pl .\appserver\extras\docker-images -amd "-PBuildDockerImages`
  * Images built successfully
  * Checking the build logs, you should see that the images were built using buildx: `[INFO] DOCKER> docker --config ... **-------> buildx <-------** build --progress=plain --builder maven --platform linux/amd64 --tag payara/micro:latest ...`
* Build all Docker images utilising an older version of Payara: `mvn clean install -pl .\appserver\extras\docker-images -amd -PBuildDockerImages "-Dpayara.version=6.9.0"`
  * Images built and tests ran successfully
  * Should see in the logs that it's downloading and tagging for the older payara version

#### Deployment test from WSL
* Built and deployed all Docker images with new _generate-multiplatform-docker-images_ profile active: `mvn clean deploy -pl ./appserver/extras/docker-images -amd -PBuildDockerImages -DskipTests`
  * Images built and deployed successfully
  * End result of micro visible [here](https://hub.docker.com/repository/docker/payara/micro/tags?page=1&ordering=last_updated) where you can see that there are _amd_ and _arm_ digests.

#### On an ARM machine I set up in AWS
* Built Server and Micro so that they're present in maven cache: `mvn clean install -DskipTests -T 1C`
* Built all Docker images: `mvn clean install -pl ./appserver/extras/docker-images -amd -PBuildDockerImages`
  * Images built and tests ran successfully.
  * Checking the build logs, saw that the images were built using buildx
* Clear docker cache: `docker system prune -a`
* Run the Docker tests (so that it downloads and tests the images published to Docker hub): `mvn clean install -pl ./appserver/extras/docker-images/tests -PBuildDockerImages`
  * Tests ran successfully

### Testing Environment
Windows 11.
WSL OpenSUSE
Amazon Linux (ARM).

## Documentation
N/A

## Notes for Reviewers
Related PR to update the release jobs: https://github.com/payara/EngineeringJenkinsjobs/pull/56


aubi:
Trying on Odroid machine (ARM64 architecture):
`docker run --publish 8080:8080 --publish 4848:4848 payara/server-full:6.2024.1-SNAPSHOT`
* downloads, starts
Text from JVM report:
```
Operating System Information:
Name of the Operating System: Linux
Binary Architecture name of the Operating System: aarch64, Version: 4.19.219-odroid-arm64
```

aubi (approved):
Great, let's start writing such tests!

PR #6535, opened 2024-01-23
FISH-8215 : port solution from Glassfish

luiseufrasio:
## Description
package java.security.acl does not exist

### Blockers
None

### New tests
None

### Testing Performed
Samples, quicklook etc

### Testing Environment
Zulu JDK 11.0.11 on Windows 11 with Maven 3.8.6

## Documentation
None

## Notes for Reviewers
None

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java
@@ -55,11 +55,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2019] Payara Foundation and/or affiliates
+// Portions Copyright [2019-2024] Payara Foundation and/or affiliates
 
 package org.apache.catalina.realm;
 
 
+import com.sun.enterprise.security.GroupPrincipal;
 import org.apache.catalina.Container;
 import org.apache.catalina.LifecycleException;
 import org.apache.catalina.LogFacade;
@@ -67,7 +68,6 @@ import org.apache.catalina.LogFacade;
 import javax.security.auth.Subject;
 import javax.security.auth.login.*;
 import java.security.Principal;
-import java.security.acl.Group;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Enumeration;
@@ -438,15 +438,13 @@ public class JAASRealm
                 roles.add(principal.getName());
             }
             // Same as Jboss - that's a pretty clean solution
-            if( (principal instanceof Group) &&
-                    "Roles".equals( principal.getName())) {
-                Group grp=(Group)principal;
-                Enumeration en=grp.members();
-                while( en.hasMoreElements() ) {
-                    Principal roleP=(Principal)en.nextElement();
-                    roles.add( roleP.getName());
+            if ((principal instanceof GroupPrincipal) && "Roles".equals(principal.getName())) {
+                GroupPrincipal grp = (GroupPrincipal) principal;
+                Enumeration<? extends Principal> membersEnum = grp.members();
+                while (membersEnum.hasMoreElements()) {
+                    Principal roleP = membersEnum.nextElement();
+                    roles.add(roleP.getName());
                 }
-
             }
         }
 
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java
@@ -56,6 +56,7 @@
  * limitations under the License.
  */
 // Portions Copyright [2019-2024] Payara Foundation and/or affiliates
+// Payara Foundation and/or its affiliates elects to include this software in this distribution under the GPL Version 2 license
 
 package org.apache.catalina.realm;
 
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/GroupPrincipal.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/GroupPrincipal.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2021 Contributors to Eclipse Foundation. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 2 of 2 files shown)


luiseufrasio:
jenkins test please

OndroMih:
Hello, @luiseufrasio ,

This PR as it is now is breaking the license of the code ported from Eclipse GlassFish. 

The file in `JAASRealm.java` in this PR contains code both from Payara master branch, which is licensed under CDDL + GPL2 with classpath exception licenses, and it also contains code from Eclipse GlassFish master branch ([JAASRealm.java#L379](https://github.com/eclipse-ee4j/glassfish/blob/b0fe34c7f7eec2a6ae01937475bbfa5150545175/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java#L379)), which is under the EPL2 + GPL2 with classpath exception licenses. Copying code from Eclipse GlassFish requires that the license is not changed and is clearly stated in the code, which is not the case here.

I recommend that you either copy the whole file `JAASRealm.java` from Eclipse GlassFish, so that it has the correct license. Or you extract the code copied from GlassFish into a separate file, which is then distributed under the EPL2 + GPL2 with classpath exception licenses.

The [LICENSE.txt](https://github.com/payara/Payara/blob/master/LICENSE.txt) file in the Payara source code should also mention EPL2 license to clarfiy that some parts of the Payara source code are distributed under this license and not under the CDDL license. This is required by the EPL2 license. For more information, read here: https://www.eclipse.org/legal/epl-2.0/ (especially the 3. Requirements section).

nucleus/security/core/src/main/java/com/sun/enterprise/security/GroupPrincipal.java
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the
+ * Eclipse Public License v. 2.0 are satisfied: GNU General Public License,
+ * version 2 with the GNU Classpath Exception, which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ */
+// Portions Copyright 2024 Payara Foundation and/or affiliates
```
Pandrex247, line 16: Portions copyright is only a thing with our usual license.
I believe for EPL you'd just add an entry for us below the "Contributors to Eclipse Foundation" line.

Pandrex247:
@OndroMih 

> This PR as it is now is breaking the license of the code ported from Eclipse GlassFish.
> 
> The file in `JAASRealm.java` in this PR contains code both from Payara master branch, which is licensed under CDDL + GPL2 with classpath exception licenses, and it also contains code from Eclipse GlassFish master branch ([JAASRealm.java#L379](https://github.com/eclipse-ee4j/glassfish/blob/b0fe34c7f7eec2a6ae01937475bbfa5150545175/appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java#L379)), which is under the EPL2 + GPL2 with classpath exception licenses. Copying code from Eclipse GlassFish requires that the license is not changed and is clearly stated in the code, which is not the case here.
> 
> I recommend that you either copy the whole file `JAASRealm.java` from Eclipse GlassFish, so that it has the correct license. Or you extract the code copied from GlassFish into a separate file, which is then distributed under the EPL2 + GPL2 with classpath exception licenses.

I'm not a copyright expert, but I believe if we expressly license the file under GPL we're OK? 
[As an example](https://github.com/payara/Payara/blob/master/appserver/security/ejb.security/src/main/java/com/sun/enterprise/iiop/security/GSSUPToken.java#L41)
I'd have to go check the history (as it was literally years ago now), but I believe we ported some GlassFish changes to this class previously and I asked what the correct procedure for doing so would be with regards to licensing.

Since this file exists from the original Oracle GlassFish days, I'm also fairly certain we can't just strip the original copyright from it.
I'll ask around internally for advice.

> The [LICENSE.txt](https://github.com/payara/Payara/blob/master/LICENSE.txt) file in the Payara source code should also mention EPL2 license to clarfiy that some parts of the Payara source code are distributed under this license and not under the CDDL license. This is required by the EPL2 license. For more information, read here: https://www.eclipse.org/legal/epl-2.0/ (especially the 3. Requirements section).

The EPL license will get mentioned in the third party license file we generate and include in the release but I'll ask for advice on whether it also needs to be reflected in that file. My understanding is that is for _our_ license, not necessarily the license of everything covered within the codebase.

Again though, not a copyright expert!
Thanks for the heads up!

appserver/web/web-core/src/main/java/org/apache/catalina/realm/JAASRealm.java
```diff
@@ -55,19 +55,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2019] Payara Foundation and/or affiliates
+// Portions Copyright [2019-2024] Payara Foundation and/or affiliates
```
Pandrex247, line 58: License specifically under GPLv2 for GlassFish port.
```suggestion
// Portions Copyright [2019-2024] Payara Foundation and/or affiliates
// Payara Foundation and/or its affiliates elects to include this software in this distribution under the GPL Version 2 license
```

nucleus/security/core/src/main/java/com/sun/enterprise/security/GroupPrincipal.java
Pandrex247: As this is a direct port from GlassFish without changes, we don't need our own copyright.
```suggestion
```

Pandrex247 (changes requested):

OndroMih:
Hi, @Pandrex247. I think if you only use the GPL license, then it's OK. But the PR as it is now re-licenses the code from Eclipse GlassFish under both CDDL and GPL, which I think you can't do without approval of the original copyright holders within the Eclipse Foundation. 

If you strip the CDDL license from the file, then the file is distributed under the GPL+CP exception only, which should be fine. As a result, you cannot distribute the whole code base under the CDDL license, only under the GPL+CP exception. I think this is already the case, as some files in your existing code base have CDDL and GPL+CP license, and some other files have only EPL and GPL+CP license, so it's not possible to distribute all the code or final binaries under CDDL or EPL, only under GPL+CP exception.

I'm not a lawyer either :) But I suggest getting an advice from a real lawyer.

Pandrex247 (approved):

PR #6546, opened 2024-01-30
FISH-7048 FISH-7253 FISH-7254 FISH-7255 FISH-8219 FISH-8220 Upgrade OSGi

Pandrex247:
## Description
Upgrades OSGi and its related Felix components.

Does a small amount of cleanup with the osgi.properties files. I converted those that I needed to touch into Maven config, but I elected not to nuke all of them as it would be a) time-consuming and b) make this PR huge.

I've elected to leave the stuff for Knopflerfish and Equinox in for now, though this is a similar thing which we may wish to clean up at some point.

The properties I've removed are resolved and provided by Felix.

The `org.osgi.framework.system.packages.extra` property is now resolved programmatically by reading the `simple-glassfish-api.jar` manifest - this module contains the bootstrap APIs.

Also makes a change to the `OpenAPISupplier` class to deal with the duplicate type conflicts that come about from this change: now that Felix is exporting JDK classes, it detects them while doing parsing (see [here](https://github.com/payara/patched-src-hk2/blob/3.0.1-payara-maintenance/class-model/src/main/java/org/glassfish/hk2/classmodel/reflect/util/CommonModelRegistry.java#L63)). Without this change classes such as `Enum` are not detected while parsing - they are now. Payara 5 apparently always had some conflict resolution where it would simply use the first found which I've ported across and added some logging to - it Shouldn't™ come up as I filter out classes which aren't a part of the application itself, but it's there as a safety net. 
This _possibly_ slows down parsing as it's now able to resolve these classes and therefore potentially parses more "transitively" - I'd need to do some performance testing to confirm. To be clear, the `Enum` class will always be detected while parsing, but what may have changed _now_ is that anything defined within the `Enum` class **itself** may also now be parsed.

I have a follow-up PR which updates Felix & OSGi in HK2 [here](https://github.com/payara/Payara/pull/6565).

### Blockers
None

### New tests
None

### Testing Performed
Started Payara Server and loaded the admin console.
Started Payara Micro and deployed an application.
Started Payara Embedded and deployed an application.

Additionally started Payara Server and loaded the admin console on Java 17, and 21.

Ran the MicroProfile OpenAPI TCK on JDK 11 and 21.

### Testing Environment
Windows 11, Java 11.0.22, 17.0.10, 21.0.2

## Documentation
N/A

## Notes for Reviewers
* I would recommend filtering out the last formatting commit to aid readability of the `MainHelper` class.
* The change to `UberJarOSGiGlassFishRuntimeBuilder` is actually to dead code - this isn't used anywhere. It's for a distribution of embedded that we've never published.


aubi (approved):
LGTM

PR #6580, opened 2024-02-29
Fish 7358 automate test payara micro

simonladen:
Adds a functional test with Payara Micro.
It starts 2 instances, deploys ClusterJsp in both and expect them to cluster. Using Playwright, in a browser, the test adds an attribute in the first instance. Then opens the second instance, checks for attribute 1, and adds attribute 2. Then navigates to Instance1, and checks for attribute1 and attribute2.

The test depends on some utilities from payara-samples, and requires to build payara-samples before running the test. Because of that, it requires to specify the profile `payara-micro-managed` when executing the test with maven.

The test uses Playwright. By default, it will use Playwright CLI command to install the dependencies. In Linux, this will use sudo to install the dependencies. To disable that installation, disable the profile `install-deps`
(see the readme for the relevant commands)

### DIFF
--- a/appserver/tests/functional/payara-micro/README.md
+++ b/appserver/tests/functional/payara-micro/README.md
@@ -1,0 +1,47 @@
+# Payara Micro Tests
+
+Starts 2 instances of Payara Micro
+
+On each instance is deployed ClusterJsp. This application should cluster the different instances and share the httpsession.
+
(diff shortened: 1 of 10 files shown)


aubi (approved):
LGTM

luiseufrasio (commented):
LGTM!

PR #6586, opened 2024-03-15
FISH-8174: adding fix to guarantee each thread updates lastaccesstime…

breakponchito:
Adding fix to guarantee each thread update lastaccesstime and thisAccessedTime after validating if session hasExpired

## Description

This is a fix for a reported bug regarding http session management when using cluster configuration with lite and normal Payara micro instances
### Testing Performed

Manual testing with reproducers included on the ticket:
https://payara.atlassian.net/browse/FISH-8174
not only can be tested with docker configuration but also with local environment with stand alone Payara Micro. Here the steps:
Use following war file, and configuration files on the following zip:
[Reproducer.zip](https://github.com/payara/Payara/files/14609980/Reproducer.zip)

unzip on the place where the payara micro jar is installed:
image

the application on the war file set the session timeout limit to 1 minute, if you need to configure more time use the following zip file with the code and customize the time as you need: 
[Shared-Session-Code.zip](https://github.com/payara/Payara/files/14610013/Shared-Session-Code.zip)

then create two instances of Payara Micro with the following commands:
instance1:
java -jar payara-micro.jar --clustermode multicast --clusterName myService-cluster --deploy Shared-Session-1.0.war --nohostaware --logProperties logging.properties --hzConfigFile payara-hazelcast.xml

instance2:
java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=9002 -jar payara-micro.jar --lite --clustermode multicast --clusterName myService-cluster --deploy Shared-Session-1.0.war --nohostaware --logProperties logging.properties --hzConfigFile payara-hazelcast.xml --port 8081

image

open browser with two tabs for the following URLs: 
open first and with a chronometer, from the mobile calculate 1 minute 
http://localhost:8080/Shared-Session-1.0/safe
this is going to create the session

then open second with the other url:
before to achieve the timeout (60 seconds), for example second 57 or 58 
http://localhost:8081/Shared-Session-1.0/safe

then you will see that the session is not recreated, just the values for the lastaccesstime and the thisAccessedTime are updated this will fix the problem of the gap for the session. From the reported issue the session is expired and created a new one 2 or 3 seconds before the timeout

image

image

I tested the following configuration of timeouts:

- 1 minute
- 2 minutes
- 30 minutes
- 60 minutes
-  120 minutes
from all of them now the session is not expired prematurely

### Testing Environment

Windows 11,  Azul JDK 11, Maven 3.9.5

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/connector/Request.java
@@ -2910,7 +2910,7 @@ public class Request implements HttpRequest, HttpServletRequest {
 
     // ------------------------------------------------------ Protected Methods
 
-    protected synchronized Session doGetSession(boolean create) {
+    protected Session doGetSession(boolean create) {
         // There cannot be a session if no context has been assigned yet
         if (context == null) {
             return null;
@@ -2953,8 +2953,11 @@ public class Request implements HttpRequest, HttpServletRequest {
             if (session != null && !session.isValid()) {
                 session = null;
             }
+            
             if (session != null) {
-                session.access();
+                synchronized (this) {
+                    session.access();
+                }
                 return session;
             }
         }
@@ -3036,7 +3039,9 @@ public class Request implements HttpRequest, HttpServletRequest {
         }
 
         if (session != null) {
-            session.access();
+            synchronized (this) {
+                session.access();
+            }
             return session;
         } else {
             return null;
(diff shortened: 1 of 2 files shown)


breakponchito:
Jenkins test please

jGauravGupta:
Instead of synchronizing the entire method, Can we target synchronization to critical sections where concurrency issues might arise to prevent unnecessary contention? I suggest further diagnosis and identify the critical section where concurrent access might lead to issues. And locking the critical section to ensure thread safety within that specific scope.

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

PR #6606, opened 2024-03-25
FISH-8453 Specify Previous Version Directly, No Automation

aubi:
## Description
We relied on finding latest version of previously published API, but this fails for a deployment of branch 6.x.y, while there is already published branch 7.

It will be necessary to update value `payara.core.compare-version` every release. Therefor it is next to the version number.

### Testing Performed
Local build.

### Testing Environment
Linux, OpenJDK 11

### DIFF
--- a/core/core-parent/pom.xml
+++ b/core/core-parent/pom.xml
@@ -616,19 +616,6 @@
             </activation>
             <build>
                 <plugins>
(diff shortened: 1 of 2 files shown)


breakponchito (approved):
LGTM

PR #6611, opened 2024-03-29
FISH-8248 Increment Version Numbers 7.2024.1.Alpha2-SNAPSHOT

Pandrex247:
## Description
Increments the version numbers for the next alpha release.

### Blockers
Cargotracker will not like the fact we're changing the version number, because it uses a maven shrinkwrap resolver which can only read the pom as-is (ignoring our usual strategy of overriding the payara dependency version using a property)
https://github.com/payara/cargotracker/pull/35

### New tests
Narp

### Testing Performed
Niet

### Testing Environment
Jenkins

## Documentation
N/A

## Notes for Reviewers
Narp


kalinchan:
Cargotracker test doesn't seem happy

Pandrex247:
> Cargotracker test doesn't seem happy

Cargotracker will not like the fact we're changing the version number, because it uses a maven shrinkwrap resolver which can only read the pom as-is (ignoring our usual strategy of overriding the payara dependency version using a property)
https://github.com/payara/cargotracker/pull/35

😉

luiseufrasio (approved):
LGTM!

luiseufrasio:
jenkins test please

PR #6665, opened 2024-04-29
FISH-8497 add dependabot conditions

abdulrahim458:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,3 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
-# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
-
 version: 2
 updates:
   # Configuration for Maven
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,3 +7,30 @@ updates:
...
+      - dependency-name: "jakarta.*"
+        update-types: ["version-update:semver-minor"]
+      - dependency-name: org.eclipse.microprofile*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "org.eclipse.microprofile*"
+        update-types: ["version-update:semver-minor"]
+  - package-ecosystem: "gitsubmodule"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
Pandrex247, line 33: Do we need to define a schedule? It seems a bit odd to scan daily to ignore everything

.github/dependabot.yml
```diff
@@ -12,3 +7,30 @@ updates:
...
+      jakarta:
+        patterns:
+        - "jakarta.*"
+        update-types:
+        - "patch"
+      microprofile:
+        patterns:
+        - "org.eclipse.microprofile*"
+        update-types:
+        - "patch"
```
Pandrex247, line 20: I don't think we want these grouped. I'd continue letting them being updated one at a time
```suggestion
```

.github/dependabot.yml
```diff
@@ -12,3 +7,30 @@ updates:
...
+    ignore:
+      - dependency-name: "jakarta.*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "jakarta.*"
+        update-types: ["version-update:semver-minor"]
+      - dependency-name: org.eclipse.microprofile*"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "org.eclipse.microprofile*"
+        update-types: ["version-update:semver-minor"]
```
Pandrex247, line 29: Do we need separate entries for ignoring major and minor?
It's an array isn't it?
Can we not have:
```
    ignore:
      - dependency-name: "jakarta.*"
        update-types: ["version-update:semver-major", "version-update:semver-minor"]
      - dependency-name: org.eclipse.microprofile*"
        update-types: ["version-update:semver-major", "version-update:semver-major"]
```

Pandrex247 (changes requested):

.github/dependabot.yml
abdulrahim458: updated in single entries

.github/dependabot.yml
abdulrahim458: removed groupng

Pandrex247: I don't think you have?

.github/dependabot.yml
abdulrahim458: I created a new PR. Iam not able to push to this branch as its not from my Forked repo.

PR #6630, opened 2024-04-04
Master Sync 10/05/2024

Pandrex247:
## Description
Syncs applicable changes from the master branch.
Hopefully didn't miss anything.

### Blockers
None

### New tests
None

### Testing Performed
Built the server and loaded the admin console.

### Testing Environment
Windows 11, Zulu 21.0.2

## Documentation
N/A

## Notes for Reviewers
None

PR #6693, opened 2024-05-08
FISH-8635 FISH-8653 Upgrade HK2 and Introduce EE11 Shim

Pandrex247:
## Description
Upgrades HK2 to the version aligned with EE11 APIs (4.0.0) with our patches reapplied as we need them for compilation.

Also introduces the EE11 shim to fudge the OSGi bundle version validation, allowing us to avoid having to do a "big bang" update of all the dependencies all at once. 

The shim as introduced here has an upgrade shim for Annotations 2.1 to 3.0, as HK2 now requires Annotations 3.0 but the server is currently providing 2.1 (upgrading annotations would kick off a large dependency chain and will be tackled separately).
The shim also contains a downgrade shim for HK2 itself, as we're not upgrading Jersey at this moment (Jersey 3.1 still uses HK2 3.0) as it also kicks off a chain of dependency updates e.g. Grizzly. This shim also covers the HK2 requirement in InSights, which we will upgrade to a Payara 7 aligned version in a separate piece of work.

### Blockers
Requires release of HK2 p2: https://github.com/payara/patched-src-hk2/pull/34
Requires the Javassist upgrade from https://github.com/payara/Payara/pull/6630

### New tests
None

### Testing Performed
Built the server, started the admin console.

### Testing Environment
Windows 11, Zulu Java 21.0.3

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/appserver/packager/appserver-core/pom.xml
+++ b/appserver/packager/appserver-core/pom.xml
@@ -2,7 +2,7 @@
 <!--
  DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
 
- Copyright (c) 2019-2022 Payara Foundation and/or its affiliates. All rights reserved.
+ Copyright (c) 2019-2024 Payara Foundation and/or its affiliates. All rights reserved.
 
  The contents of this file are subject to the terms of either the GNU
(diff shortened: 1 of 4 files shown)


kalinchan (approved):
lgtm

PR #6706, opened 2024-05-10
FISH-8497 update dependabot conditions

abdulrahim458:
•	Jakarta EE APIs - Ignore major and minor version increases (patch updates only) on "jakarta*"

•	Jakarta EE Implementations - ignore major version increases (minor and patch updates only)

•	MicroProfile APIs - Ignore major and minor version increases (patch updates only). Can probably wildcard on "org.eclipse.microprofile.*"

•	Ignore all submodules

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,147 @@ updates:
       interval: "daily"
     open-pull-requests-limit: 5
     target-branch: "master"
+    ignore:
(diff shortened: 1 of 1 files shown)


Pandrex247 (changes requested):
Should also have 
```
      - dependency-name: "fish.payara.*"
        update-types: ["version-update:semver-major"]
```

Pandrex247 (approved):

PR #6713, opened 2024-05-10
FISH-8688 improvement: optimize MR-jar class loading

### LINKED ISSUE #6554: Bug Report: Application deployment and AS startup takes more RAM (ever growing) and too long (more than twice the time it took on previous releases)
### Brief Summary

Since Payara release 6.2023.9 we observe the application deployment time is at least twice as much as it was in 6.2023.8 and earlier versions. This includes the latest 6.2024.1 release.

We have tried to trace this performance regression and there seems to be no obvious reason for it:

 - We see no extra errors or warnings on the logs before or while deploying the application ear.
 - Increasing the available resources does not help (It reduces the application deployment time, but it's still nowhere near previous releases).
 - We see [a commit](https://github.com/payara/Payara/commit/34153c8741382e051df38717b92b07905837bf4b) added after 2023.8 that could be causing this issue.
 - A simple reproducer does not reproduce the issue.

We need a mitigation or fix to prevent the upgrade to the latest Payara release from severely impacting our instance deployment times. 

We want to know:

 - Is anyone else experiencing this issue?
 - Is there a documented way to optimise deployment times we may have missed to help us get 2023.8 and earlier deployment times in the latest Payara release for the affected application?

### Expected Outcome

Application deployment time remains approximately the same after upgrading from Payara 6.2023.8 to latest.

### Current Outcome

Application deployment time more than doubles after upgrading from Payara 6.2023.8 to latest. No reason given in the logs. We do not know which component is causing the delay.

### Reproducer

I wrote a reproducer deploying pretty much every Jakarta component we use (JAX-RS, JPA, @Startup EJBs) packed on a single ear, but it only helped to discard those components, at least on a basic level: https://github.com/elcodedocle/payara-sscce/tree/latest 

We'd appreciate any pointers on how to extend it to reproduce (Perhaps running a profiler during the startup of the affected application would be more helpful, for example?)

### Operating System

Debian bookworm

### JDK Version

Debian openjdk 17 JRE (17.0.10+7-1~deb12u1) / application built with target 11 on Debian openJDK 11.0.22+7-1~deb11u1

### Payara Distribution

Payara Server Full Profile

lprimak:
## Description
Optimize multi-release JAR processing, remove resource leak from MR-JarFile remaining open
Opening MR-Jars correctly by using `RUNTIME_VERSION`
fixes #6554 

### Testing Performed
Deploy original reproducer for this issue

### Testing Environment
Any

### DIFF
--- a/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
+++ b/nucleus/common/common-util/src/main/java/com/sun/enterprise/loader/ASURLClassLoader.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
- // Portions Copyright [2016-2023] [Payara Foundation and/or its affiliates]
+ // Portions Copyright [2016-2024] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.loader;
 
@@ -95,6 +95,7 @@ public class ASURLClassLoader extends CurrentBeforeParentClassLoader
 
     /** logger for this class */
     private static final Logger _logger = CULoggerInfo.getLogger();
+    private static final Runtime.Version RUNTIME_VERSION = Runtime.Version.parse(System.getProperty("java.version"));
 
     /*
        list of url entries of this class loader. Using LinkedHashSet instead of original ArrayList
@@ -613,12 +614,9 @@ public class ASURLClassLoader extends CurrentBeforeParentClassLoader
         return AccessController.doPrivileged((PrivilegedAction<byte[]>) () -> {
             try {
                 if (res.isJar) { // It is a jarfile..
-                    JarFile zip = res.zip;
-                    JarFile jar = new JarFile(res.file, false, JarFile.OPEN_READ,
-                            Runtime.Version.parse(System.getProperty("java.version")));
-                    JarEntry entry = jar.getJarEntry(entryName);
+                    JarEntry entry = res.zip.getJarEntry(entryName);
                     if (entry != null) {
-                        byte[] classData = getClassData(zip.getInputStream(entry));
+                        byte[] classData = getClassData(res.zip.getInputStream(entry));
                         res.setProtectionDomain(ASURLClassLoader.this, entry.getCertificates());
                         return classData;
                     }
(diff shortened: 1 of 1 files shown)


lprimak:
jenkins test

lprimak:
jenkins test please

RInverid:
I get a deployment failure with this PR - an `@DependsOn` annotation fails to resolve another singleton in the same ejb package. (the singleton that is depended on implements an interface, and that interface is injected into the singleton with the `@DependsOn` annotation)

lprimak:
@RInverid do you have a reproducer or at least a stack trace? There isn’t much to go on in your comment

RInverid:
Not yet, but I have more practical feedback: this MR does not load Bouncycastle libs if running on JDK 17. BC libs are multirelease, but they only have class files for jdk9. Your PR skips loading BC classes becuase it cannot find classes for jdk17

lprimak:
Hmmmm. I specifically tested BC libs in 11 (but not 17) so I am not sure how 17 would be any different.
Looking forward to a reproducer.

Pandrex247:
Had a quick look myself, I think you can use the reproducer provided against the original change PR to see the error (albeit against embedded): https://github.com/ctabin/gf-test-lucene

Without this PR, it works as it's able to load the `Analyzer` class.
With this PR, issues start arising again with it unable to find the `Analyzer` class

Not yet determined the failing difference _Yet™_

lprimak:
Ok, I didn't realize that [commit](https://github.com/payara/Payara/commit/34153c8741382e051df38717b92b07905837bf4b) was not only optimization but a bug fix. I re-incorporated the bug fix into this PR and tested it with https://github.com/ctabin/gf-test-lucene and it works

RInverid:
Thanks for the changes, I re-tested this MR. The deployment time and memory Issues are seeing are less of an issue, but still present. Probably because the problematic change (`JarFile jar = new JarFile` etc) is still called, but significantly less often.

lprimak:
@RInverid how about now?

lprimak:
Jenkins test

lprimak:
Jenkins test please

RInverid:
Yes this resolves my issues, as I've also found with my testing. Not sure what the change in `ProtectedJarFile` does though, its at least not necessary to resolve my issues according to my testing.

Pandrex247 (approved):

Pandrex247:
Thanks both! 👌

PR #6732, opened 2024-05-22
FISH-8588 Cleanup POMs

kalinchan:
## Description

Altered some properties to align with conventions (too many to do)

Removed unused profiles
Removed findbugs
### Testing Performed

Build Payara with BuildExtras profile enabled


pom.xml
```diff
@@ -145,35 +145,35 @@
...
 
         <javase.version>${jdk.version}</javase.version>
         <!-- BOM-referenced versions -->
 
         <osgi-resource-locator.version>1.0.3</osgi-resource-locator.version>
         <snakeyaml.version>2.2</snakeyaml.version>
-        <jstl-impl.version>3.0.1</jstl-impl.version>
-        <mojarra.version>4.0.0.payara-p2</mojarra.version>
+        <jakarta.servlet.jsp.jstl.version>3.0.1</jakarta.servlet.jsp.jstl.version>
+        <jakarta.faces.version>4.0.0.payara-p2</jakarta.faces.version>
```
Pandrex247, line 156: I'm not sure about this change. I think our convention should be that we use project names where they exist.
So Mojarra should stay as `mojarra.version`, EclipseLink should stay as `eclipselink.version`.

We can then extend this for any sub-project artefacts e.g. eclipselink.asm.version

This helps to differentiate the properties from their specification API versions e.g. Jakarta Faces API version does not necessarily align with the Jakarta Faces implementation (Mojarra) version.

Pandrex247 (changes requested):

api/payara-bom/pom.xml
```diff
@@ -483,7 +483,7 @@
             <dependency>
                 <groupId>org.eclipse.angus</groupId>
                 <artifactId>angus-activation</artifactId>
-                <version>${angus.activation.version}</version>
+                <version>${angus-activation.version}</version>
```
Pandrex247, line 486: By my suggested "use project names" convention this should probably just be `angus.version`

api/payara-bom/pom.xml
```diff
@@ -472,7 +472,7 @@
             <dependency>
                 <groupId>org.glassfish</groupId>
                 <artifactId>jakarta.enterprise.concurrent</artifactId>
-                <version>${concurrent.version}</version>
+                <version>${jakarta.enterprise.concurrent}</version>
```
Pandrex247, line 475: By my suggested "use project names" convention this should probably just be `concurro.version`

api/payara-bom/pom.xml
```diff
@@ -441,12 +441,12 @@
             <dependency>
                 <groupId>org.glassfish.mq</groupId>
                 <artifactId>mq-distribution</artifactId>
-                <version>${mq.version}</version>
+                <version>${mq-distribution.version}</version>
```
Pandrex247, line 444: By my suggested "use project names" convention this should probably be `openmq.version`

api/payara-bom/pom.xml
```diff
@@ -419,7 +419,7 @@
             <dependency>
                 <groupId>org.apache.santuario</groupId>
                 <artifactId>xmlsec</artifactId>
-                <version>${org.apache.santuario.version}</version>
+                <version>${xmlsec.version}</version>
```
Pandrex247, line 422: By my suggested "use project names" convention this should probably be `santuario.version`

api/payara-bom/pom.xml
```diff
@@ -341,7 +341,7 @@
             <dependency>
                 <groupId>com.fasterxml.woodstox</groupId>
                 <artifactId>woodstox-core</artifactId>
-                <version>${woodstox.version}</version>
+                <version>${woodstox-core.version}</version>
```
Pandrex247, line 344: By my suggested "use project names" convention this should probably remain as `woodstox.version`

api/payara-bom/pom.xml
Pandrex247: You're correct!
We should probably leave this at it is then

Pandrex247 (approved):
LGTM
If you really wanted to make me happy you could also get rid of the doclint-java8-disable profile which is never not active 😏

PR #6784, opened 2024-06-27
FISH-8855-Added Auto name for deployment group

NotedSalmon:
FISH-8855 added auto name for deployment group

## Description

I have made changes to the strings.properties file so that instead of saying autoname instance, it says autoname instance/deployment group so it saves creating another autoname.help text.

Changed dgNew.jsf to include the checkbox button and changed the commands block for the new button so that the Deployment group label does not show as required (as autoname is an option) and if autoname is selected, it will use the same code as the instance autoname.

### Testing Performed

Built and ran the server, made sure it was working. Tested auto naming multiple deployment groups and instances to make sure both worked correctly. Tested leaving everything blank to display error messages and tested setting own name + ticking the box to make sure that if the label is populated, it will default the name to that even if the box is checked.

## Notes for Reviewers

I could of made the same exact functions that the instance auto name uses and changed to deployment group but I felt it would be unnecessary code.

### DIFF
--- a/appserver/admingui/cluster/src/main/resources/org/glassfish/cluster/admingui/Strings.properties
+++ b/appserver/admingui/cluster/src/main/resources/org/glassfish/cluster/admingui/Strings.properties
@@ -448,4 +448,4 @@ clusters.ejbTimers.TimerCountCol=Timer Count
 general.serverNameCol=Server Name
 
 autoName=Auto Name
-autoName.help=When enabled, will generate a name for your instance if none is provided, or will otherwise resolve any name conflicts. 
+autoName.help=When enabled, will generate a name for your instance/deployment group if none is provided, or will otherwise resolve any name conflicts. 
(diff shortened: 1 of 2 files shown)


appserver/admingui/cluster/src/main/resources/org/glassfish/cluster/admingui/Strings.properties
```diff
@@ -448,4 +448,4 @@ clusters.ejbTimers.TimerCountCol=Timer Count
 general.serverNameCol=Server Name
 
 autoName=Auto Name
-autoName.help=When enabled, will generate a name for your instance if none is provided, or will otherwise resolve any name conflicts. 
+autoName.help=When enabled, will generate a name for your instance/deployment group if none is provided, or will otherwise resolve any name conflicts.
```
kalinchan, line 451: Missing copyright update

aubi (approved):
LGTM

PR #6792, opened 2024-06-30
FISH-8879 : jakarta authentication upgraded to 3.1.0

luiseufrasio:
## Description
For Payara 7:
Upgrade to the final version if available, otherwise the “best” milestone release.

### Blockers
None

### New tests
None

### Testing Performed
build and run

### Testing Environment
Zulu JDK 11.0.11 on Windows 11 with Maven 3.8.6


Pandrex247 (changes requested):
Do we not need to also update the shim?
The Exousia explodes when I try to start the admin console.
It doesn't explicitly give an OSGi error, but Felix can't find the `GlassfishRoleMapper` class

PR #6816, opened 2024-07-08
Increment version numbers core 6.18.0 snapshot

abdulrahim458:
Increments the version number for Core.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -46,7 +46,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>fish.payara.server.core</groupId>
     <artifactId>core-aggregator</artifactId>
-    <version>6.17.0</version>
+    <version>6.18.0-SNAPSHOT</version>
 
     <packaging>pom</packaging>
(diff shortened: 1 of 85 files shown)


simonladen (approved):
LGTM

core/pom.xml
```diff
@@ -46,15 +46,15 @@
...
 
     <packaging>pom</packaging>
     <name>Payara Core Aggregator</name>
 
     <description>All of Payara Core modules</description>
 
     <properties>
-        <payara.core.version>6.17.0-SNAPSHOT</payara.core.version>
+        <payara.core.version>6.18.0-SNAPSHOT</payara.core.version>
         <payara.core.compare-version>6.16.0</payara.core.compare-version>
```
Pandrex247, line 58: At some point it would be nice to have automatic incrementing of this property.
Probably can't be done as a part of this PR as 6.17.0 is obviously what we're trying to release - we can't compare against it until it's at least staged somewhere.

Pandrex247 (approved):

PR #6837, opened 2024-07-22
FISH-8874 Replace occurrences of master with main

kalinchan:


Pandrex247 (approved):
.github/dependabot.yaml also needs updating

PR #6873, opened 2024-08-08
FISH-9175 spotbugs concatenated strings in loop

NotedSalmon:
FISH-9175 spotbugs concatenated strings in loop

## Description

Previous strings used += to add strings in a for loop which SpotBugs complained. Changed to use StringBuilder

### Testing Performed

Ran SpotBugs against changes

### DIFF
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2022] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2024] [Payara Foundation and/or its affiliates]
 
 package org.glassfish.enterprise.iiop.impl;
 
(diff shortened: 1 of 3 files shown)


appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
```diff
@@ -952,19 +952,18 @@ private String[] getORBInitRef(String endpoints) {
...
     private String getCorbalocURL(Object[] list) {
 
-        String corbalocURL = "";
+        StringBuilder corbalocURL = new StringBuilder();
         //convert list into corbaloc url
         for (int i = 0; i < list.length; i++) {
             logger.log(Level.INFO, "list[i] ==> {0}", list[i]);
-            if (corbalocURL.equals("")) {
-                corbalocURL = IIOP_URL + ((String) list[i]).trim();
+            if (corbalocURL.toString().equals("")) {
```
svendiedrichsen, line 959: toString might be too expensive. Maybe use length() == 0?

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
```diff
@@ -952,19 +952,18 @@ private String[] getORBInitRef(String endpoints) {
...
+                corbalocURL.append(IIOP_URL).append(((String) list[i]).trim());
             } else {
-                corbalocURL = corbalocURL + "," +
-                        IIOP_URL + ((String) list[i]).trim();
+                corbalocURL.append(",").append(IIOP_URL).append(((String) list[i]).trim());
             }
         }
-        logger.log(Level.INFO, "corbaloc url ==> {0}", corbalocURL);
-        return corbalocURL;
+        logger.log(Level.INFO, "corbaloc url ==> {0}", corbalocURL.toString());
```
svendiedrichsen, line 965: Same here. Maybe use Logger.log with supplier.

appserver/payara-appserver-modules/payara-micro-service/src/main/java/fish/payara/appserver/micro/services/asadmin/SendAsadminCommand.java
```diff
@@ -202,21 +202,21 @@ public void execute(AdminCommandContext context)
...
                             // Only print out the messages if verbose was chosen
                             if (verbose) {
-                                actionReport.setMessage(output);
+                                actionReport.setMessage(output.toString());
                             }
 
                             // Only log the messages if logOutput was chosen
                             if (logOutput) {
-                                Logger.getLogger(SendAsadminCommand.class.getName()).log(Level.INFO, output);
+                                Logger.getLogger(SendAsadminCommand.class.getName()).log(Level.INFO, output.toString());
```
svendiedrichsen, line 219: Maybe use the log Method with Supplier here as well.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/GlassFishORBManager.java
aubi: It will be converted to String anyway -- just assign it to a local variable and use both here and in return.
When you touch it -- supplier is really better!
```
String corbalocURLString = corbalocURL.toString();
log.info(() -> "corbaloc url ==> " + corbalocURLString);
```

aubi (changes requested):
Tiny improvements, looks good otherwise.
The changes with suppliers in logs are good, please do them.

aubi (approved):
LGTM

PR #6882, opened 2024-08-13
FISH-8880 Upgrade Exousia to 3.0.0-M3

kalinchan:
## Description

Changed from the SimplePolicyProvider to the default

Uses the exousia authorization service


appserver/security/core-ee/src/main/java/com/sun/enterprise/security/web/integration/GlassFishToExousiaConverter.java
```diff
@@ -0,0 +1,122 @@
...
+import com.sun.enterprise.deployment.WebBundleDescriptor;
+import com.sun.enterprise.deployment.WebComponentDescriptor;
+import com.sun.enterprise.deployment.web.SecurityRoleReference;
+
+import jakarta.servlet.annotation.ServletSecurity.TransportGuarantee;
+
+/**
+ * This class converts from GlassFish security types to Exousia security types.
+ *
+ * @author arjan
```
breakponchito, line 45: Should we need to maintain the author name?

breakponchito (approved):
LGTM

appserver/security/core-ee/src/main/java/com/sun/enterprise/security/web/integration/GlassFishToExousiaConverter.java
Pandrex247: If this is a port from GlassFish then yes. Would be nice to have the full name though 🤣

nucleus/deployment/common/src/main/java/org/glassfish/deployment/common/SecurityRoleMapper.java
```diff
@@ -37,16 +37,18 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
+// Portions Copyright [2024] Payara Foundation and/or affiliates
```
OndroMih, line 40: Code in this file is copied from Eclipse GlassFish. Copyright should be attributed to "Contributors to the Eclipse Foundation" or OmniFish. The copied code is licensed under EPL or GPL, it cannot be distributed under the CDDL license.

OndroMih (changes requested):

OndroMih:
Hi @kalinchan, how are you? I left a few comments where I think this PR violates license or copyright. I hope it's easy to fix.

kalinchan:
Hi @OndroMih,

I have made the changes, let me know if you believe more are required.

OndroMih:
Hi, @kalinchan,

Even though you added the correct copyright, both files are still distributed under both CDDL and GPL licenses. Even if you choose to use the Eclipse Foundation code under the GPL, you cannot redistribute it under a different license. So the whole file should be now only distributed under the GPL license, without CDDL.

Pandrex247:
`// Payara Foundation and/or its affiliates elects to include this software in this distribution under the GPL Version 2 license`
With this addition it _is_ only distributed under GPL. 

```
 * If you wish your version of this file to be governed by only the CDDL or
 * only the GPL Version 2, indicate your decision by adding "[Contributor]
 * elects to include this software in this distribution under the [CDDL or GPL
 * Version 2] license."  If you don't indicate a single choice of license, a
 * recipient has the option to distribute your version of this file under
 * either the CDDL, the GPL Version 2 or to extend the choice of license to
 * its licensees as provided above.  However, if you add GPL Version 2 code
 * and therefore, elected the GPL Version 2 license, then the option applies
 * holder.
```

Pandrex247 (approved):
Server starts and admin console loads. I think this is enough for us to merge this initial PR, and make any additional changes incrementally.

I've created an issue, [FISH-9478](https://payara.atlassian.net/browse/FISH-9478), to review whether we should retain our `PolicyConfigurationFactoryImpl` and `PolicyProviderImpl` classes or not, as they were introduced to allow us to provide [JACC per application functionality](https://docs.payara.fish/community/docs/Technical%20Documentation/Payara%20Server%20Documentation/Security%20Guide/Administering%20System%20Security.html#how-to-install-a-custom-jacc-provider-per-application), but this seems to now be a [part of the specification itself](https://github.com/jakartaee/authorization/issues/98) and may be provided by Exousia natively.

PR #6914, opened 2024-08-28
WIP - FISH-9332 resync Cargotracker with the new EE repo

astoica1:
Only for Payara 6 - I have introduced the new Cargotracker repo which is actively maintained.


kalinchan (commented):
Do we want to use payara-micro-managed and shouldn't we specify the Payara version?

You also don't need the setupDomain() if we are using the managed profile.

kalinchan:
You should be able to specify the Payara version with `-Dpayara.version=6.2024.x`

Jenkinsfile
```diff
@@ -201,6 +201,7 @@ pipeline {
                         -Djavax.net.ssl.trustStore=${env.JAVA_HOME}/lib/security/cacerts \
                         -Djavax.xml.accessExternalSchema=all -Dpayara.version=${pom.version} \
                         -Ppayara-server-remote -DtrimStackTrace=false"""
```
kalinchan, line 203: The payara-server-remote profile does not exist?

PR #6915, opened 2024-08-29
FISH-9146 Upgrade Jakarta Pages API

kalinchan:
## Description

Upgrades Jakarta Pages Api which requires a downgrade shim.

Also updated the wrong JDK message.


kalinchan:
Will close this PR as Quicklook fails - this works with the WASP upgrade so [6916](https://github.com/payara/Payara/pull/6916) will resolve this

PR #6918, opened 2024-09-02
FISH-9571 FISH-9572 Upgrade JSTL API

kalinchan:
## Description

WASP has replaced the impl which is why references are replaced

Shim is not needed anymore due to this upgrade.

### DIFF
--- a/api/payara-bom/pom.xml
+++ b/api/payara-bom/pom.xml
@@ -444,9 +444,9 @@
                 <version>${openmq.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.glassfish.web</groupId>
-                <artifactId>jakarta.servlet.jsp.jstl</artifactId>
-                <version>${jakarta.servlet.jsp.jstl.version}</version>
+                <groupId>org.glassfish.wasp</groupId>
+                <artifactId>wasp</artifactId>
+                <version>${jsp-impl.version}</version>
             </dependency>
             <dependency>
                 <groupId>org.eclipse</groupId>
--- a/api/payara-bom/pom.xml
+++ b/api/payara-bom/pom.xml
@@ -444,11 +444,6 @@
                 <version>${openmq.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.glassfish.wasp</groupId>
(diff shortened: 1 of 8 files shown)


api/payara-bom/pom.xml
```diff
@@ -444,9 +444,9 @@
...
             </dependency>
             <dependency>
-                <groupId>org.glassfish.web</groupId>
-                <artifactId>jakarta.servlet.jsp.jstl</artifactId>
-                <version>${jakarta.servlet.jsp.jstl.version}</version>
+                <groupId>org.glassfish.wasp</groupId>
+                <artifactId>wasp</artifactId>
+                <version>${jsp-impl.version}</version>
             </dependency>
             <dependency>
```
Pandrex247, line 451: Don't think this is required at all now - it would be duplicated from the core bom.
```suggestion
```

appserver/packager/glassfish-web/pom.xml
```diff
@@ -162,8 +162,8 @@
             <artifactId>jakarta.servlet.jsp.jstl-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.glassfish.web</groupId>
-            <artifactId>jakarta.servlet.jsp.jstl</artifactId>
+            <groupId>org.glassfish.wasp</groupId>
+            <artifactId>wasp</artifactId>
         </dependency>
```
Pandrex247, line 167: Duplicate - it's already defined a few lines above.

pom.xml
```diff
@@ -163,7 +163,6 @@
 
         <osgi-resource-locator.version>1.0.3</osgi-resource-locator.version>
         <snakeyaml.version>2.2</snakeyaml.version>
-        <jakarta.servlet.jsp.jstl.version>3.0.1</jakarta.servlet.jsp.jstl.version>
```
Pandrex247, line 166: You also need to remove/replace this here: https://github.com/payara/Payara/blob/5f3c6cbf93e1fc490b6de6c9519b287748a3f755/appserver/pom.xml#L279

Pandrex247 (changes requested):

PR #6937, opened 2024-09-09
FISH-9553 Add system property validation

kalinchan:
## Description

Uses Regex to validate the system property creation ensuring that no html tags are present.

Validated after TranslatedConfigView on purpose.
### Testing Performed

Reproducer in JIRA

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CreateSystemProperties.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CreateSystemProperties.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2024] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.v3.admin;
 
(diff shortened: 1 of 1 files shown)


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CreateSystemProperties.java
```diff
@@ -154,6 +157,17 @@ public void execute(AdminCommandContext context) {
                 } finally {
                     TranslatedConfigView.doSubstitution.set(Boolean.TRUE);
                 }
+
+                String HTML_JS_PATTERN = "<[^>]+>";
+                Pattern pattern = Pattern.compile(HTML_JS_PATTERN);
```
svendiedrichsen, line 162: The Pattern and its String can be made static to avoid compilation on every execution.

nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/admin/CreateSystemProperties.java
```diff
@@ -154,6 +159,15 @@ public void execute(AdminCommandContext context) {
                 } finally {
                     TranslatedConfigView.doSubstitution.set(Boolean.TRUE);
                 }
+
+                Matcher matcher = HTML_JS_PATTERN.matcher(properties.getProperty(propName));
+                if (matcher.find()) {
+                    report.setMessage(localStrings.getLocalString("create.system.properties.failed",
+                            "System property {0} creation failed", sysPropName));
+                    report.setActionExitCode(ActionReport.ExitCode.FAILURE);
+                    return;
```
svendiedrichsen, line 168: Wouldn't it be clearer in terms of code structure to throw an IllegalArgumentException instead of returning directly out of the loop? This way you could have a more specific error message and leave the proper error handling to the catch below.

PR #6943, opened 2024-09-10
FISH-9553 Sanitise properties in admin console

kalinchan:
## Description

Escapes properties in admin console
### Testing Performed

Reproducer in JIRA


Pandrex247 (changes requested):
You also need to sanitise the following page: Configs > $config > System Properties > Instance Values

Pandrex247 (approved):
I've clicked around and can't find any more places

PR #6955, opened 2024-09-17
FISH-6030: Remove noisy and all references to it.

Viii3:
The noisy parameter has been removed from all Admin Console pages and has been marked as deprecated within the API.

## Description

- Resolves FISH-6030

### Testing Performed

Built the server, checked the output of various `asadmin get-XYZ-configuration` commands, checked the configuration pages in the admin console to ensure the "noisy" option has been successfully removed.

### Testing Environment

Maven home: 3.9.6


Pandrex247 (commented):
I think we need a feature branch, either that or this PR needs to be reworked with backwards compatibility in mind.

The `SetNotificationConfiguration` command is shared with Enterprise, where we can't just drop support for the noisy option. Even marking it as `obsolete` in the `Param` annotation means the option gets ignored and would thus change the behaviour.

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/admin/BaseGetNotifierConfigurationCommand.java
```diff
@@ -151,13 +150,8 @@ protected String listConfiguration(C configuration) {
...
 
-        if (configuration != null) {
-            map.put("Enabled", configuration.getEnabled());
-            map.put("Noisy", configuration.getNoisy());
-        } else {
-            map.put("Enabled", FALSE.toString());
-            map.put("Noisy", TRUE.toString());
-        }
+        if (configuration != null) map.put("Enabled", configuration.getEnabled());
+        else map.put("Enabled", FALSE.toString());
```
Pandrex247, line 154: As an side, this is cursed formatting - please put the curly braces back in 😝

nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/admin/BaseSetNotifierConfigurationCommand.java
```diff
@@ -92,6 +92,7 @@ public abstract class BaseSetNotifierConfigurationCommand<C extends PayaraNotifi
     @Param(name = "enabled")
     protected Boolean enabled;
 
+    @Deprecated
     @Param(name = "noisy", optional = true, defaultValue = "true")
```
Pandrex247, line 96: Deprecated is only a note for the IDE - you probably want to add the `obsolete` parameter to the `Param` annotation too

appserver/admingui/eventbus-notifier-console-plugin/src/main/resources/eventbus/eventbusNotifierConfiguration.jsf
```diff
@@ -105,11 +101,6 @@ holder.
                           helpText="$resource{i18nebn.notifier.eventbus.configuration.enabledLabelHelpText}">
                 <sun:checkbox id="enabledProp" selected="#{pageSession.enabledSelected}" selectedValue="true" />
             </sun:property>
-            <sun:property id="noisy" labelAlign="left" noWrap="#{true}" overlapLabel="#{false}"
```
Pandrex247, line 108: I would also remove the string property from the `Strings.properties` file and the l10n equivalents.

Pandrex247 (changes requested):

Pandrex247 (approved):

PR #6637, opened 2024-04-04
Fish 8174 improve thread expire validation to read lastaccesstime

breakponchito:

PR #6956, opened 2024-09-18
FISH-9748 bugfix: fix HA session race condition leading to session attribute loss

lprimak:
## Description

Fixed regression introduced in #6637 

## Important Info
- compare update times to stored session in HA scenarios only
- Use `store.load()` instead of `swapIn()` to load sessions during update time comparison, as `swapIn()` has side effects, such as replacing session object in the session map, and calling users' activation handlers every time a session is retrieved, leading to session loss and performance issues

To further explain this, when the session is replaced in the map by `swapIn()`, other copies of the session hang around in other threads.
Sessions get stored in HA store at the end of each request, and only the session currently in the map gets updated.
Those other copies do not get stored in the HA store when request ends, since they are no longer in the session map.
This leads to the data loss.

Performance problem is caused by the session map thrashing, and session creation callbacks being called every time the `swapIn()` method is called.

## Testing
Manual testing done

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/Store.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/Store.java
@@ -55,6 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+// Portions Copyright 2016-2024 Payara Foundation and/or its affiliates
 
 package org.apache.catalina;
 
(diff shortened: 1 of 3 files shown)


lprimak:
Don't forget about me :) :)

lprimak:
@Pandrex247 dont forget me :) thank you

breakponchito:
@lprimak @Pandrex247 I already tested with my reproducer and the resulting behavior is what we expect.  On the following url you will fins the steps I tested: https://github.com/payara/Payara/pull/6637

breakponchito (approved):
LGTM

PR #6967, opened 2024-09-23
FISH-9771 FISH-9688 spot bugs random only used once

NotedSalmon:
Used `SecurityUtils` to generate a random and use it within the respective classes. Spot bugs issue. (check description)
## Description
Spotbugs warning:
```
Random object created and used only once in com.sun.enterprise.admin.cli.cluster.ImportSyncBundleCommand.backupInstanceDir()

This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.

If it is important that the generated Random numbers not be guessable, you must not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).random objects.
```
### Testing Performed

built payara and ran spotbugs against the changes made.

### DIFF
--- a/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
+++ b/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
@@ -77,6 +77,7 @@ import java.io.File;
 import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Random;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -612,9 +613,9 @@ public interface Cluster extends ConfigBeanProxy, PropertyBag, Named, SystemProp
          */
         @Override
         public void decorate(AdminCommandContext context, final Cluster instance) throws TransactionFailure, PropertyVetoException {
-            SecureRandom random = new SecureRandom();
             Logger logger = ConfigApiLoggerInfo.getLogger();
             LocalStringManagerImpl localStrings = new LocalStringManagerImpl(Cluster.class);
+            Random random = new SecureRandom();
             Transaction t = Transaction.getTransaction(instance);
             //check if cluster software is installed else fail , see issue 12023
             final CopyConfig command = (CopyConfig) runner
--- a/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
+++ b/nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
@@ -615,7 +615,6 @@ public interface Cluster extends ConfigBeanProxy, PropertyBag, Named, SystemProp
         public void decorate(AdminCommandContext context, final Cluster instance) throws TransactionFailure, PropertyVetoException {
             Logger logger = ConfigApiLoggerInfo.getLogger();
             LocalStringManagerImpl localStrings = new LocalStringManagerImpl(Cluster.class);
-            Random random = new SecureRandom();
             Transaction t = Transaction.getTransaction(instance);
             //check if cluster software is installed else fail , see issue 12023
             final CopyConfig command = (CopyConfig) runner
@@ -725,7 +724,7 @@ public interface Cluster extends ConfigBeanProxy, PropertyBag, Named, SystemProp
 
                         // generate a random port since user did not provide one.
                         // better fix in future would be to walk existing clusters and pick an unused port.
-                        TCPPORT = Integer.toString(random.nextInt(9200 - 9090) + 9090);
+                        TCPPORT = Integer.toString(SecurityUtils.nextInt(9200 - 9090) + 9090);
 
                         // hardcode all instances to use same default port.
                         // generate mode does not support multiple instances on one machine.
@@ -746,7 +745,7 @@ public interface Cluster extends ConfigBeanProxy, PropertyBag, Named, SystemProp
                             gmsListenerPortSysProp.setName(propName);
                             if (TCPPORT == null || TCPPORT.trim().charAt(0) == '$') {
                                 String generateGmsListenerPort = Integer.toString(
-                                        random.nextInt(9200 - 9090) + 9090);
+                                        SecurityUtils.nextInt(9200 - 9090) + 9090);
                                 gmsListenerPortSysProp.setValue(generateGmsListenerPort);
                             } else {
                                 gmsListenerPortSysProp.setValue(TCPPORT);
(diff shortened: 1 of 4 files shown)


nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
```diff
@@ -93,6 +93,8 @@
 @ReferenceConstraint(skipDuringCreation=true, payload=Cluster.class)
 public interface Cluster extends ConfigBeanProxy, PropertyBag, Named, SystemPropertyBag, ReferenceContainer, RefContainer, Payload {
 
+    SecureRandom R = new SecureRandom();
```
svendiedrichsen, line 96: Pretty thrifty name for it.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ImportSyncBundleCommand.java
```diff
@@ -132,6 +132,7 @@ public class ImportSyncBundleCommand extends LocalInstanceCommand {
     private File backupDir;
 
     private static final String RENDEZVOUS_PROPERTY_NAME = "rendezvousOccurred";
+    private static final SecureRandom r = new SecureRandom();
```
svendiedrichsen, line 135: Thrifty as well.

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
pzygielo: It is static already.

aubi (approved):
LGTM

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
```diff
@@ -614,6 +615,7 @@ class Decorator implements CreationDecorator<Cluster> {
         public void decorate(AdminCommandContext context, final Cluster instance) throws TransactionFailure, PropertyVetoException {
             Logger logger = ConfigApiLoggerInfo.getLogger();
             LocalStringManagerImpl localStrings = new LocalStringManagerImpl(Cluster.class);
+            Random random = new SecureRandom();
```
svendiedrichsen, line 618: * Is it nessessary to use a SecureRandom for this? SecureRandom might block if the underlying system runs out of random byte input.
* wouldn't it be sufficient to have a static Random member

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ImportSyncBundleCommand.java
```diff
@@ -132,6 +132,7 @@ public class ImportSyncBundleCommand extends LocalInstanceCommand {
     private File backupDir;
 
     private static final String RENDEZVOUS_PROPERTY_NAME = "rendezvousOccurred";
+    private static final Random random = new SecureRandom();
```
svendiedrichsen, line 135: Is it nessessary to use a SecureRandom for this?

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
pzygielo, line 618: 1. Now `random` gets created even if it's not going to be used.
2. I can not see `random` to be used more than once. Both references seem to be in exclusive branches.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ImportSyncBundleCommand.java
aubi: "usually static properties are written in capital letters"

https://www.oracle.com/technetwork/java/codeconventions-150003.pdf
I suppose you mean "Constant", e.g. chapter 9. Random is not a constant, although it is static and final.

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
aubi: You are right, just `Random` is really much faster (I tested via JMH).
This place is executed during cluster-related commands, e.g. once only. So for this usage, I suggest to leave the original SecureRandom.

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ImportSyncBundleCommand.java
aubi: ad SecureRandom -- see my comment up

nucleus/admin/config-api/src/main/java/com/sun/enterprise/config/serverbeans/Cluster.java
aubi: The new solution solves this issue -- the static field in a different class is created only if it is used.
Ad used more than once: it's what spotbugs reported, in fact it is called only once, of course.

aubi (approved):
LGTM

nucleus/common/common-util/src/main/java/org/glassfish/common/util/RandomUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.security.SecureRandom;
+
+/**
+ * Centralize usage of random number generator.
+ *
+ * @author: Petr Aubrecht
+ */
+public class RandomUtils {
+    private static SecureRandom random = new SecureRandom();
```
svendiedrichsen, line 50: Can and should be final, shouldn't it?

nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/security/SecurityUtils.java
```diff
@@ -48,9 +48,9 @@
  * @author Byron Nevins
  */
 public final class SecurityUtils {
+    private static SecureRandom random = new SecureRandom();
 
     public static String getSecureRandomHexString(int numBytes) {
-        SecureRandom random = new SecureRandom();
         byte[] bb = new byte[numBytes];
         random.nextBytes(bb);
```
jGauravGupta, line 55: Use `RandomUtils` here to consolidate the `SecureRandom` usage into a single utility class to maintain consistency and manage randomness more effectively.

jGauravGupta (changes requested):

nucleus/common/common-util/src/main/java/com/sun/enterprise/universal/security/SecurityUtils.java
jGauravGupta: Should be fixed separately as may cause OSGI issue, due to package naming.

jGauravGupta (approved):

PR #6977, opened 2024-09-27
FISH-9803 - NPE prevented by update of index in array to prevent null value as input.

adLuk:
…ameter.

### DIFF
--- a/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/monitoring/WebServiceTesterServlet.java
+++ b/appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/monitoring/WebServiceTesterServlet.java
@@ -538,7 +538,7 @@ public class WebServiceTesterServlet extends HttpServlet {
             logger.log(Level.SEVERE, LogUtils.CREATE_DIR_FAILED, classesDir);
         }
 
-        String[] wsimportArgs = new String[14];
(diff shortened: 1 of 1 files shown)


simonladen (approved):

appserver/webservices/jsr109-impl/src/main/java/org/glassfish/webservices/monitoring/WebServiceTesterServlet.java
```diff
@@ -538,7 +538,7 @@ private String wsImport(URL wsdlLocation) throws IOException {
             logger.log(Level.SEVERE, LogUtils.CREATE_DIR_FAILED, classesDir);
         }
 
-        String[] wsimportArgs = new String[14];
+        String[] wsimportArgs = new String[13];
```
svendiedrichsen, line 541: Why not use a List<String> here? Why bother with indexes?

adLuk: To be honest I was trying to keep changes as small as possible. Just to fix problem not really improve code. But I will do it, also I can open one more issue related to this class as technical depth and try to at least remove some of these potecially problematic parts of code.

aubi:
The new version is fine, waiting for CLA.

smillidge:
CLA received last week

PR #6988, opened 2024-10-01
FISH-9198 Add Microprofile Config TOML Support

### LINKED ISSUE #6822: Enhancement: TOML Config Source for MicroProfile Config /FISH-9198
### Brief Summary

I would like to be able to use a TOML file as a source for configuration values. This is a standard feature for SmallRye and should be included in Payara, too.

### Expected Outcome

Just configure a source and read away that TOML file!

### Current Outcome

I'm stuck with adding such a config source to the application, which does not allow me to provide values for early init stuff like `${MPCONFIG=...:...}` variable replacements.

### Alternatives

Implement a fake JDBC driver to read TOML files. 

[Been there, done that: gdcc/jdbc-conffile-driver](https://github.com/gdcc/jdbc-conffile-driver) Works, but is hacky. Also: a config source that reads a local file can be updated on change and doesn't need 60 secs default caching! But this will require more control over what's going on, which is just not possible with the JDBC workaround.

### Context

As with the DirConfigSource (#5006 / #5007), I would be willing and able to create a pull request and donate the necessary code. If accepted, we should talk about the location for this config source (`fish.payara.microprofile.config.extensions` vs `fish.payara.nucleus.microprofile.config.source`) and potentially about using a different caching strategy for it.

kalinchan:
## Description

Resolves #6822

This feature introduces changes to the Admin Console and new AsAdmin Commands.

By default the ordinal for TOML is `95`.

The properties will only be updated when first initialised or if the last modified time has been changed.
### Admin Console
(Configurations -> server-config -> Microprofile -> Config -> TOML)
image

### AsAdmin Commands
* `get-toml-configuration`: Returns the path to the TOML file and the maximum recursion depth (to prevent potential stackoverflow)
* `set-toml-configuration`: `path` parameter to specify location of toml file, can be relative to current domain. `depth` parameter to specify the depth of recursion.

### Blockers

n/a
### New tests

none
### Testing Performed

Specified a TOML file with various different tables and arrays and deployed a simple application that prints out all properties, the ordinal and specific values.
### Testing Environment

Java version: 11.0.23, vendor: Azul Systems, Inc., runtime: /Users/kalinchan/.sdkman/candidates/java/11.0.23-zulu/zulu-11.jdk/Contents/Home
OS name: "mac os x", version: "14.6.1", arch: "aarch64", family: "mac"
## Documentation

to come... 
## Notes for Reviewers

`tomlj` had to be repackaged due to not being an OSGi bundle and it also has a depedency on Antlr4


luiseufrasio (changes requested):
When the file does not exists on Windows it returns:

```
**asadmin set-toml-configuration --path C:\config.tom --depth 10**
Picked up JAVA_TOOL_OPTIONS: -Duser.language=en -Duser.country=US -Dfile.encoding=UTF-8
remote failure: java.nio.file.InvalidPathException: Illegal char <:> at index 105: C:\Users\luise\git\Payara\appserver\distributions\payara\target\stage\payara6\glassfish\domains\domain1\C:\config.tom
Illegal char <:> at index 105: C:\Users\luise\git\Payara\appserver\distributions\payara\target\stage\payara6\glassfish\domains\domain1\C:\config.tom
Command set-toml-configuration failed.
```

core/pom.xml
```diff
@@ -138,6 +138,7 @@
         <monitoring-console-api.version>2.0.2</monitoring-console-api.version>
         <metainf-services.version>1.11</metainf-services.version>
         <ldapbp.version>1.0</ldapbp.version>
+        <tomlj.version>1.1.1</tomlj.version>
```
Pandrex247, line 141: Does this need to be here? I can't see any core modules depending on this.

Pandrex247 (commented):
I would also question whether the TOML config source should exist under the nucleus package - would it not be better homed [alongside the other extensions](https://github.com/payara/Payara/tree/main/appserver/payara-appserver-modules/microprofile/config-extensions/src/main/java/fish/payara/microprofile/config/extensions)?

core/pom.xml
kalinchan: removed

luiseufrasio (approved):
LGTM!

appserver/packager/external/tomlj-repackaged/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.tomlj</groupId>
+            <artifactId>tomlj</artifactId>
+            <version>${tomlj.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
+            <version>4.11.1</version>
```
Pandrex247, line 117: Why not also have these configured via property?

appserver/packager/external/tomlj-repackaged/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+                        </Embed-Dependency>
+                        <Export-Package>
+                            !org.antlr.*,!org.checkerframework.checker.nullness.qual,!com.ibm.icu.*,!org.abego.treelayout.*,!org.stringtemplate.*,*;version=${tomlj.version}
+                        </Export-Package>
+                        <Import-Package>
+                            !org.antlr.*,!org.checkerframework.checker.nullness.qual,!com.ibm.icu.*,!org.abego.treelayout.*,!org.stringtemplate.*,*
+                        </Import-Package>
+                        <!-- Exclude unnecessary imports -->
+                        <Private-Package>!*</Private-Package>
+                        <_exportcontents>*</_exportcontents>
```
Pandrex247, line 91: I don't think these are actually doing anything

appserver/packager/external/tomlj-repackaged/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+                <configuration>
+                    <instructions>
+                        <Embed-Dependency>
+                            *;scope=compile;inline=true
+                        </Embed-Dependency>
+                        <Export-Package>
+                            !org.antlr.*,!org.checkerframework.checker.nullness.qual,!com.ibm.icu.*,!org.abego.treelayout.*,!org.stringtemplate.*,*;version=${tomlj.version}
+                        </Export-Package>
+                        <Import-Package>
+                            !org.antlr.*,!org.checkerframework.checker.nullness.qual,!com.ibm.icu.*,!org.abego.treelayout.*,!org.stringtemplate.*,*
```
Pandrex247, line 87: ICU4J is licensed under unicode. This is not currently under our approved license list.
This will need reverifying if it's a compatible license.

poikilotherm:
Potentially stupid question, late to the party. Instead of using TOMLJ and adding another dependency, would it help to use Jackson Dataformat TOML, which is already around as a lib?

Pandrex247:
> Potentially stupid question, late to the party. Instead of using TOMLJ and adding another dependency, would it help to use [Jackson Dataformat TOML](https://github.com/FasterXML/jackson-dataformats-text/tree/2.19/toml), which is already around as a lib?

My memory may be failing me, but I don't recall us discussing this library in particular, and it doesn't seem to be indexed highly in Google - it's entirely possible we may have simply overlooked the existence of this library.

Thanks for the suggestion @poikilotherm: we will investigate if it's suitable. We already use Jackson for XML and YAML parsing in the server so yes, all things being equal, it would be the logical library to use over Tomlj.

appserver/packager/external/tomlj-repackaged/pom.xml
kalinchan: No longer used

PR #7023, opened 2024-10-11
FISH-9695 Upgrade Service

Viii3:
## Description

- Changes the internal default string value of the notification filtering level.
- Part of FISH-9695

### Blockers

- Relates to [the upgrade service PR](https://github.com/payara/Payara-Enterprise/pull/1655).
  - Enables the upgrade service to disable itself for notifiers that have had their values manually set.

### Testing Performed

- Tested as part of the related PR.

### Testing Environment

Maven version: 3.9.6

### DIFF
--- a/nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotifierConfiguration.java
+++ b/nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotifierConfiguration.java
@@ -56,7 +56,7 @@ public interface PayaraNotifierConfiguration extends ConfigBeanProxy {
 
     String DEFAULT_ENABLED_VALUE = "false";
     String DEFAULT_NOISY_VALUE = "true";
-    String DEFAULT_EVENT_FILTER = "WARNING";
+    String DEFAULT_EVENT_FILTER = "WARNING_(DEFAULT)"; // A safe dummy value used for the upgrade service.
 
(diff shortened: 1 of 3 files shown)


nucleus/common/internal-api/src/main/java/fish/payara/internal/notification/PayaraNotifierConfiguration.java
```diff
@@ -56,7 +56,7 @@ public interface PayaraNotifierConfiguration extends ConfigBeanProxy {
 
     String DEFAULT_ENABLED_VALUE = "false";
     String DEFAULT_NOISY_VALUE = "true";
-    String DEFAULT_EVENT_FILTER = "WARNING";
+    String DEFAULT_EVENT_FILTER = "WARNING_(DEFAULT)"; // A safe dummy value used for the upgrade service.
```
Pandrex247, line 59: I'm not sure how I feel about this name. It's a little misleading, as it's not really the default?
The get command also now _expects_ to encounter an exception with this default.

I don't think I'm a fan of this strategy. I shall mull on it over the weekend.

Pandrex247: See https://github.com/payara/Payara-Enterprise/pull/1655#discussion_r1810438628

Pandrex247 (changes requested):

PR #7099, opened 2024-11-28
FISH-10146 FISH-9875 Deployment Group Nullpointer

Viii3:
## Description

- Resolves FISH-10146
  - Instances no longer strip data from the `<deployment-groups>` block of the domain xml.
  - Instances have improved checking for deciding whether or not to discard a config from their own domain xml.
    - Previously instances would discard any config for instances not in their _first_ deployment group. They now check all of their deployment groups.
 - Resolves FISH-9875
   - Deployment groups that share instances no longer exhibit inconsistent behaviour.

### Testing Performed

Tested using a modified reproducer for FISH-10146:

The following batch file was used for testing.
```
@ECHO OFF
CALL asadmin start-domain
echo.
CALL asadmin create-local-instance instance1
echo.
CALL asadmin start-local-instance instance1
echo.
CALL asadmin create-deployment-group group1
echo.
CALL asadmin add-instance-to-deployment-group --instance instance1 --deploymentgroup group1
echo.
CALL asadmin create-local-instance instance2
echo.
CALL asadmin start-local-instance instance2
echo.
CALL asadmin add-instance-to-deployment-group --instance instance2 --deploymentgroup group1
echo.
CALL asadmin stop-local-instance instance1
echo.
CALL asadmin stop-local-instance instance2
echo.
CALL asadmin stop-domain
```

### Testing Environment

Maven version: 3.9.6

### DIFF
--- a/nucleus/admin/config-api/src/main/java/org/glassfish/config/support/DomainXmlPreParser.java
+++ b/nucleus/admin/config-api/src/main/java/org/glassfish/config/support/DomainXmlPreParser.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2018-2023] Payara Foundation and/or affiliates
+// Portions Copyright [2018-2024] Payara Foundation and/or affiliates
 
 package org.glassfish.config.support;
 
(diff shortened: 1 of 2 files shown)


breakponchito:
Jenkins test please

Viii3:
There appears to be a Java versioning error in some of the tests, one of the test classes has been compiled with Java 17.
```
Caused by: org.apache.maven.plugin.PluginContainerException: An API incompatibility was encountered while executing org.apache.felix:maven-bundle-plugin:6.0.0:bundle: java.lang.UnsupportedClassVersionError: aQute/bnd/osgi/Analyzer has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

breakponchito:
I already tested and the issue now is not presented

breakponchito (approved):
LGTM

PR #7132, opened 2024-12-16
FISH-10274 Revert Parsson Upgrade

abdulrahim458:


Pandrex247 (changes requested):
Shouldn't be required?

PR #7133, opened 2024-12-17
FISH-10318 Add create-system-property command

Viii3:
## Description

- Resolves FISH-10318
  - Adds a `create-system-property` asadmin command.
  - Adds an associated manpage.
  - Adjusts existing manpages for relevant commands to refer to the new one.

### Testing Performed

- Built the server and used the command to create a new system property.  
- Verified the property exists with the `list-system-properties` command.

### Testing Environment

Maven version: 3.9.6

### DIFF
--- a/nucleus/core/kernel/src/main/java/fish/payara/enterprise/admin/CreateSystemProperty.java
+++ b/nucleus/core/kernel/src/main/java/fish/payara/enterprise/admin/CreateSystemProperty.java
@@ -1,0 +1,147 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2024 Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ * The contents of this file are subject to the terms of either the GNU
+ * General Public License Version 2 only ("GPL") or the Common Development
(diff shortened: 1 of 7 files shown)


nucleus/core/kernel/src/main/java/fish/payara/enterprise/admin/CreateSystemProperty.java
```diff
@@ -0,0 +1,147 @@
...
+    final private static LocalStringManagerImpl localStrings = new LocalStringManagerImpl(CreateSystemProperties.class);
+
+    @Param(optional=true, defaultValue=SystemPropertyConstants.DAS_SERVER_NAME)
+    String target;
+
+    @Param(name="name")
+    String propertyName;
+
+    @Param(name="value")
+    String propertyValue;
```
Pandrex247, line 86: I think it would make more sense from a user POV to just have a single parameter to provide, splitting on '`=`'.
So you could replace these with one parameter, and mark it as `primary=true` so that on the command line it's a nice simple `asadmin create-system-property --target=server-config wibbles=wobbles`.

So something like:

```suggestion
    @Param(name = "name_value", primary = "true")
    String propertyNameValue;
```

Pandrex247 (dismissed):

PR #7163, opened 2025-01-20
FISH-8646 Allow Mojarra to Initialise Weld / Cannot Deploy Sprint Boot 3

Pandrex247:
## Description
Attempts to deploy Spring Boot 3 applications following our recommendations with class loading delegation disabled and CDI bean discovery set to _none_, even very basic "hello world" ones, will lead to an error being thrown by Mojarra that CDI is not available.

This is caused by the fact that the `spring-web` dependency (which you can't really exclude) has a Jakarta Faces class [DelegatingPhaseListenerMulticaster](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/jsf/DelegatingPhaseListenerMulticaster.java) which the Servlet context initialiser in Mojarra [FacesInitializer](https://github.com/payara/patched-src-mojarra/blob/mojarra-4.0.9.payara-p1/impl/src/main/java/com/sun/faces/config/FacesInitializer.java) has a `HandlesTypes` for, inducing Mojarra to initialise its Servlet context listener.

This is arguably a valid initialisation attempt, since even though the simple "hello world" application may not contain Jakarta Faces content, it bundles a dependency which does. 

[Mojarra currently has functionality which will attempt to Weld if it detects that CDI has not yet been enabled](https://github.com/payara/patched-src-mojarra/blob/mojarra-4.0.9.payara-p1/impl/src/main/java/com/sun/faces/config/FacesInitializer.java#L227), though this is specifically written for use with the [EnhancedListener](https://github.com/weld/core/blob/5.0.1.Final/environments/servlet/core/src/main/java/org/jboss/weld/environment/servlet/EnhancedListener.java) class contained within the `weld-servlet-core` dependency, which we don't bundle in the server and looks to be intended for use in Servlet containers rather than "full" Jakarta EE application servers; if nothing else it isn't OSGi-ified and bundling it as a dependency in an application leads to class cast exceptions with loading the `WeldModule` interface implementations provided by the Weld modules bundled in Payara (it says they don't implement the interface when they do - presumably an issue with different class loaders being used to load the respective classes).

This PR somewhat replicates the functionality in Mojarra by adding a deployment sniffer which will look for Faces content in the application and force the Weld container to initialise.

I've added a means of skipping this change in behaviour as the class scanning conducted by HK2 is pretty extensive, and may start causing Weld to initialise in cases where it arguably shouldn't. This can be done via system property, `fish.payara.faces.integration.allowFacesCdiInitialisation`, or via deploy command property `allowFacesCdiInitialisation`.

### Blockers
None

### New tests
None

### Testing Performed
* Deployed the simple reproducer application attached to the issue to Payara Server - sniffer is skipped and application fails to deploy.
* Deployed the simple reproducer application attached to the issue to Payara Server with `allowFacesCdiInitialisation` deployment property set to true (`asadmin deploy --properties "allowFacesCdiInitialisation=true" demo-0.0.1-SNAPSHOT.war`) - application deploys and endpoint is accessible.
* Deployed the simple reproducer application attached to the issue to Payara Server with `fish.payara.faces.integration.allowFacesCdiInitialisation` system property set to true in `server-config` - application deploys and endpoint is accessible.
* Deployed the simple reproducer application attached to the issue to Payara Server with `fish.payara.faces.integration.allowFacesCdiInitialisation` system property set to true in `server-config` and with `allowFacesCdiInitialisation` deployment property set to false (`asadmin deploy --properties "allowFacesCdiInitialisation=false" demo-0.0.1-SNAPSHOT.war`) - sniffer is skipped and application fails to deploy.
* Deployed the simple reproducer application attached to the issue to Payara Micro - sniffer is skipped and application fails to deploy.
* Deployed the simple reproducer application attached to the issue to Payara Micro with `fish.payara.faces.integration.allowFacesCdiInitialisation` system property set to true (`java "-Dfish.payara.faces.integration.allowFacesCdiInitialisation=true" -jar --nocluster payara-micro.jar demo-0.0.1-SNAPSHOT.war`) - application deploys and endpoint is accessible.
* Kicked off a full TCK run: passed

### Testing Environment
Windows 11, Zulu Java 21, Maven 3.9.9

## Documentation
https://github.com/payara/Payara-Documentation/pull/543

## Notes for Reviewers
None


lprimak:
I would hold off on this somewhat. HK2 Types are going away replaced by Jandex. That will simplify things overall as well.

luiseufrasio (commented):
I tested on my local and the application is deployed, but the endpoint `/hello-world` returns **HTTP 500** in all scenarios, including the payara micro one.

Pandrex247:
HTTP 500? What's the error?

luiseufrasio:
> HTTP 500? 🤔 What's the error?

@Pandrex247 I tested agin in WSL and I got a 404-Not Found

luiseufrasio (approved):
LGTM!

PR #7165, opened 2025-01-21
[WIP] Replace HK2/Types/Parser/Scanner with Jandex and cache in the domain

### LINKED ISSUE #7079: Enhancement: Payara should use Jandex to speed up CDI deployments
### Brief Summary

CDI deployment speed could use improvement

Jandex is much faster than using class scanning, because it's done at build time and can be cached.
Using Jandex could speed to deployment by orders of magnitude.
Jandex indexes should be used and / or cached for all internal Payara CDI extensions

### Expected Outcome

Jandex is used for all CDI modules, including internal ones.

### Current Outcome

Jandex is not used

### Alternatives

None

### Context

_No response_

lprimak:
Deploy performance improvements 

## This is an experiment (for now)
This PR achieves at least 2X improvement when dependencies are not cached, which is a worst-case scenario.
I have seen as much as 5X performance improvement on a warmed-up JVM with dependencies cached,
with many dependencies.

## Feedback is appreciated
 Thank you.

Fixes #7079

- [x] Save jandex to transient metadata
- [x] Refactor HK2-Types stuff in WeldUtils into separate file
- [x] Use Jandex in addition to Types in WeldUtils (get jandex-based WAR to deploy) - ignoring dependencies for now
- [x] How does Scanner / Parser parse war-libraries? - what's the calling flow? - in ApplicationLifecycle (kernel)
- [x] Save / cache Types or use Jandex for dep libs? - Jandex
- [x] Index (via Jandex) 
- [x] Index via Jandex - WAR dependencies
- [x] Restore shared WARlib functionality
- [x] Cache Jandex indexes in the domain - file cache on disk
- [x] Load Jandex from cache, if exists
- [x] Does old scanner get invoked for WARlibs? - No, scanner doesn't get invoked for warlibs
- [x] DOL Factory using Jandex
- [x] WarScanner using Jandex
- [x] EAR support (DOL module, etc)
- [x] Do not cache -SNAPSHOT dependencies
- [x] DOL: remove handler stack, make it no-op
- [x] Sync w/main branch
- [x] Profile the code, see where the bottlenecks actually are
- [x] Check default context push / pop
- [x] Re-introduce parallel scanning. Hopefully causes much more performance improvement
- [x] Delay writing jandex cache to disk until after deployment
- [x] Check for file leaks
- [x] Check for ClassLoader leaks
- [x] Check new SpringBoot-related code and remove types from there / cc: @Pandrex247 
- [x] Emailmanager doesn't deploy on remote container
- [x] Remove AnnotationClassModel and use Jandex-based model directly
- [x] Test / fix `issue-reproducers/cdi-unknownbda` reproducer
- [x] Do not cache exploded WARs / JARs
- [x] Get FlowLogix tests to work
- [x] WeldUtils - cache CDI results in context, per-index
- [x] Check old lifecycle to see if anything is missing that needs to be restored
- [x] Remove APF/hk2 folders (old lifecycle)
- [x] Reintroduce `StructuredDeploymentTracing tracing = StructuredDeploymentTracing.load(context);`
- [x] Reintroduce `tracing.startSpan(DeploymentTracing.AppStage.CLASS_SCANNING, source.getName());` for each source
- [x] Reintroduce getExternalLibraries() into Lifecycle
- [x] Reintroduce `DeploymentProperties.SKIP_SCAN_EXTERNAL_LIB` support
- [ ] Remove TODOs
- [ ] Look at `GenericAnnotationDetector` to see if it's getting used anywhere / use Jandex for it
- [ ] Get current test to deploy (duplicate classes in root EAR vs. EJB-JAR
- [ ] Code cleanup
- [ ] Get tests to pass again


nucleus/common/internal-api/src/main/java/org/glassfish/internal/deployment/JandexIndexer.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+@Contract
+public interface JandexIndexer {
+    void index(DeploymentContext deploymentContext) throws IOException;
+    void reindex(DeploymentContext deploymentContext) throws IOException;
+    boolean isJakartaEEApplication(DeploymentContext deploymentContext) throws IOException;
+    Index getRootIndex(DeploymentContext deploymentContext);
```
svendiedrichsen, line 57: You might want to avoid spreading the JBoss classes like i.e. `Index` around in the Payara codebase.

lprimak: ha, that's kinda the point of this PR.
The issue is that current annotation scanning framework is not made for CDI, is slow, and can't be cached.
Jandex is a small, separate project, has no dependencies, and is tailor made for CDI.
While it did come from JBoss, it's no longer embedded in that ecosystem and can be used freely.

Any effort to isolate `Index` class will result in an unnecessary wrapper that would not add any value.

svendiedrichsen: It would result in a wrapper class. I agree. But I would call it a useful one. It allows to control and limit the interface towards payara code and allows to switch technology if needed/wanted.

svendiedrichsen: And you could have convenience methods useful for the payaras interaction with the index.

lprimak: Let's see where this goes... I think it's only used in one place anyway

lprimak:
Currently, performance is same or better, so primary goal looking good.

nucleus/common/internal-api/src/main/java/org/glassfish/internal/deployment/JandexIndexer.java
lprimak: Had to create a wrapper class anyway, so this has been resolved.

lprimak:
Update on the progress:

Dependencies of this issue ballooned from a 3-day project to a 8-month behemoth. What this means is that I won't have any time to work on this til probably 2026 or later.
I am going to put this work on hold to more concentrate on some strategic things that I need to do.
If anyone is interested to accelerate this work, please contact me to discuss.

Just to reiterate, I am not going abandon this (I never abandon anything) but currently the timing is not in this task's favor.

PR #7193, opened 2025-02-18
FISH-10527 added nofollow to index page

NotedSalmon:
## Description

Added nofollow and tracking for links. Updated copyright year

### DIFF
--- a/nucleus/admin/template/pom.xml
+++ b/nucleus/admin/template/pom.xml
@@ -97,16 +97,7 @@
                         <configuration>
                             <artifactItems>
                                 <!-- The index.html for the "docroot" application,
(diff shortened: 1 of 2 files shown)


Pandrex247 (changes requested):
Confirmed, this isn't enough as-is.
We need to look at removing or overriding https://github.com/NotedSalmon/Payara/blob/FISH-10527-add-nofollow-to-index-page/nucleus/admin/template/pom.xml#L101-L108 

This is an extremely old package which overrides essentially this entire module.
It appears to have always been here, something we inherited from GlassFish 3/4 ¯\\_(ツ)\_/¯

Pandrex247 (approved):

PR #7196, opened 2025-02-18
FISH-10578 FISH-10597 update Payara 7 samples

astoica1:
This PR just bumps the Jakarta EE11 Api version to 11.0.0-M5


Pandrex247 (approved):
Kicked off a run against all of our currently passing TCKs, still waiting on Faces but looks like all should be green 👍

PR #7199, opened 2025-02-19
FISH-8545 SameSite Cookie

Viii3:
## Description

- Resolves FISH-8545
  - Makes Payara parse the payara-web.xml file correctly.
  - Adds SameSite to various cookie configurations.

### Testing Performed

Tested against the reproducer.

### Testing Environment

Maven version: 3.9.6

### DIFF
--- a/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
+++ b/appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2023] [Payara Foundation and/or its affiliates]
+// Portions Copyright 2016-2025 Payara Foundation and/or its affiliates
 
 package com.sun.enterprise.glassfish.web;
 
@@ -636,7 +636,7 @@ public class WarHandler extends AbstractArchiveHandler {
                         versionIdentifier = parser.getElementText();
                     } else if (RuntimeTagNames.PAYARA_WHITELIST_PACKAGE.equals(name)) {
                         application.addWhitelistPackage(parser.getElementText());
-                    } else if ("cookie-properties".equals(name)) {
+                    } else if ("session-config".equals(name)) {
                         readCookieConfig();
                     } else {
                         skipSubTree(name);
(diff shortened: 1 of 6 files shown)


appserver/web/war-util/src/main/java/com/sun/enterprise/glassfish/web/WarHandler.java
```diff
@@ -636,7 +636,7 @@ protected void read(InputStream input) throws XMLStreamException {
                         versionIdentifier = parser.getElementText();
                     } else if (RuntimeTagNames.PAYARA_WHITELIST_PACKAGE.equals(name)) {
                         application.addWhitelistPackage(parser.getElementText());
-                    } else if ("cookie-properties".equals(name)) {
+                    } else if ("session-config".equals(name)) {
```
breakponchito, line 639: Could this be a constant value instead to compare literal value directly?

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
```diff
@@ -2072,6 +2072,8 @@ private void configureCookieProperties(CookieProperties bean) {
                         cookieConfig.setSecure(value);
                     } else if("cookieHttpOnly".equalsIgnoreCase(name)) {
                         cookieConfig.setHttpOnly(Boolean.valueOf(value));
+                    } else if("cookieSameSite".equalsIgnoreCase(name)) {
```
breakponchito, line 2075: probably same here move the literal value to a constant and then compare

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/session/SessionCookieConfig.java
```diff
@@ -282,6 +293,10 @@ public String toString() {
             sb.append(", domain=");
             sb.append(_domain);
         }
+        if (sameSite != null) {
+            sb.append(", sameSite=");
```
breakponchito, line 297: here also split more the append calls and use sameSite literal value on a constant

breakponchito:
@Viii3 I tested the cookie is there but when selecting any of the values from the console on the SameSite value it is always set strict either if you select none or lax from the admin console

image

breakponchito (approved):

breakponchito:
little comments but in general this is working as expected

PR #7216, opened 2025-02-25
FISH-10521 Escape HTTP Characters in REST Interface

Pandrex247:
## Description
Escapes HTTP characters in the REST interface, to help lock down any XSS attempts someone may try if they have access to the application server's filesystem.

Most config which would be displayed here is protected by validation of the domain.xml, as that prevents you from having the "<" or ">" characters as values. The key files however don't have this validation, and so we must escape them.

### Blockers
None

### New tests
None

### Testing Performed
On a clean domain:
* Add the following to the `admin-keyfile` file under domain config: `user3;{SSHA256}qXV9vLHbkJRoMaJkGsNUrz5jSHDzMTR0CyFmZjUt8HJQVlIi2X8XaA==;asadmin`
* Add the following to the `keyfile` file under domain config: `wibbles;{SSHA256}6v8s3F0zGwF9bi9s1tygCvTz+az/Im55nS6Z3Z1sCIS9ng9aLAykCQ==;`
* Ping the REST interface http://localhost:4848/management/domain/configs/config/server-config/security-service/auth-realm/file/list-users - no message box should appear
* Ping the REST interface http://localhost:4848/management/domain/configs/config/server-config/security-service/auth-realm/admin-realm/list-users
* Create a standalone instance called _Elated-Whalefish_
* From the admin console, edit the value of its `ASADMIN_LISTENER_PORT` property to `24848` - Instances > Elated-Whalefish > Properties
  * This was previously fixed in https://github.com/payara/Payara/pull/6943
* Load this page in the REST interface, no message box should appear - http://localhost:4848/management/domain/servers/server/Elated-Whalefish/system-property/ASADMIN_LISTENER_PORT

### Testing Environment
Windows 11, Maven 3.9.9, Zulu JDK 11.0.26

## Documentation
https://github.com/payara/Payara-Documentation/pull/552

## Notes for Reviewers
None

### DIFF
--- a/appserver/admingui/common/src/main/resources/appServer/serverInstAdminPassword.jsf
+++ b/appserver/admingui/common/src/main/resources/appServer/serverInstAdminPassword.jsf
@@ -39,6 +39,7 @@
     holder.
 
 -->
+<!-- Portions Copyright 2025 Payara Foundation and/or its affiliates -->
 
 <!-- appServer/serverInstAdminPassword.jsf -->
 
@@ -80,7 +81,7 @@
     <!-- Text Field section -->
     <sun:propertySheetSection id="propertSectionTextField">
         <sun:property id="userIdProp"  labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18nc.domain.Userid}">
-            <sun:staticText id="UserId" text="#{sessionScope.userName}"/>
+            <sun:staticText id="UserId" text="#{sessionScope.sanitisedUserName}"/>
         </sun:property>
         <sun:property id="groupListProp"  labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18nc.domain.GroupList}">
             <sun:staticText id="GroupList" text="#{pageSession.group}"/>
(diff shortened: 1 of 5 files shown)


breakponchito:
@Pandrex247 same is happening with Payara 6,  I tried to test and it seems something is missing because after editing the admin-keyfile and including an script I'm still seeing the message when loading the server:

Content of file:
`admin;{SSHA256}Q3p+X8y0tC4olXfIFMAmKXq+19eFXaDm+C8tBucgdZGfIIM3gl5mNA==;asadmin`

image

on the other cases it seems that is working as expected, just the case with the file admin-keyfile

I tested on windows 11, zulu jdk 11 and maven 3.9.5

Pandrex247:
I just tried again, it worked for me.
Are you sure you built my branch and were testing against that build?
I recommend using private browsing to ensure nothing is cached

Your URL doesn't look correct either? http://localhost:4848/common/index.jsf is the admin console, not the REST management interface

If you've found another vulnerability in the admin console that would be a separate issue/CVE

Pandrex247:
>  after editing the admin-keyfile and including an script
What script are you referring to? There's no reference to editing a script in the reproducer

luiseufrasio (approved):
LGTM!

breakponchito:
@Pandrex247 with latest's changes now I can't see the dialog trying to inject java script code

breakponchito (approved):
LGTM

PR #7222, opened 2025-02-27
Fish 9891 adding parsing module for repository annotation

breakponchito:
## Description

Adding structure for module to parse Repository annotation from Jakarta Data

### New tests

Manual testing deploying an application using Repository annotation from Jakarta and doing debugging session to review if the server identify the annotation

here the app:

[JakartaDataReproducer.zip](https://github.com/user-attachments/files/19019997/JakartaDataReproducer.zip)

example of debug session to identify the Repository annotation:

image

### Testing Performed

Manual testing using debugger
### Testing Environment

windows 11, zulu JDK 21, maven 3.9.5


appserver/persistence/pom.xml
```diff
@@ -62,5 +62,6 @@
         <module>entitybean-container</module>
         <module>jpa-container-l10n</module>
         <module>cmp-l10n</module>
+        <module>data-core</module>
```
Pandrex247, line 65: I wouldn't necessarily have this under the `persistence` module.
I think this should simply sit above under appserver.

I'm imagining a structure somewhat akin to:

* appserver
  * data
    * data-core
    * data-persistence
    * data-nosql

WDYT?

appserver/packager/glassfish-jpa/pom.xml
```diff
@@ -95,6 +95,11 @@
            <artifactId>persistence-common</artifactId>
            <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>fish.payara.server.internal.persistence</groupId>
+            <artifactId>data-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
Pandrex247, line 102: See my other comment, I wouldn't necessarily have this as a part of persistence as it's not a part of JPA (just interacts with it).

I would have it as a standalone `data` package

Pandrex247 (commented):
Ran through all of our currently passing TCKs - works!
I have some thoughts about the module placement however

PR #7230, opened 2025-03-03
FISH-10526: update of corba to use new concurrency implementation instead of synchronized blocks

breakponchito:
Updating corba version to use new concurrency implementation
## Description

This is a feature to improve the concurrency functionality of the corba implementation when processing big messages. The process to read big messages cause to split in multiple fragments internally processed by multithread functionality. The reported issue from customer mentioned that in some moment most of the threads are in the wait state causing an startvation because any of those threads can get time to continue processing. This will impact the system using most of the resources available and in some moment restart the system to clear resources
### Testing Performed

You need to deploy the following EJB on the server:

[remoteview.zip](https://github.com/user-attachments/files/19061160/remoteview.zip)

And use the same as a lib for the following stand alone application:

[EJBRemoteClient6.zip](https://github.com/user-attachments/files/19061167/EJBRemoteClient6.zip)

This is a remote client implementation that will send 300 requests to the server sending big message each one. This will cause to split the big message in fragments using the internal fragment processing from corba.

On the Main class you will see the remote call configuration to execute the tests with multiple threads creating 300 requests to the server. something important for the standalone application is to indicate the correct path for the modules folder and resolve all dependencies for the EJB Client call configuration:

image

to enable the new implementation with locks you need to set the following property on the system:

`com.sun.corba.ee.protocol.enablingNewFragmentProcess=true`

by default this is set as false to use the synchonized mechanism

also you can set the timeout time of the internal management of the process to wait for information on the queue with the following property:

`com.sun.corba.ee.protocol.newFragmentEmptyConditionTimeout=10000`

by default this is set with 10000 milliseconds, the unit for the value is milliseconds and this value is only used when the previous property is set as true

when you enable the new implementation you can send multiple calls of the client sending 300 messages and check with a tool like visualvm the behavior of the system. Here the example of results:

image

### Testing Environment

windows 11, azul JDK 11 and maven 3.8.6

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -127,7 +127,6 @@
         
         <!-- implementation and build dependencies -->
         <glassfish-management-api.version>3.2.3.payara-p1</glassfish-management-api.version>
-        <!--glassfish-corba.version>4.2.4.payara-p2</glassfish-corba.version-->
         <glassfish-corba.version>4.2.4.payara-p3</glassfish-corba.version>
         <junit.version>4.13.2</junit.version>
         <mockito.version>2.28.2</mockito.version>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -127,7 +127,7 @@
         
         <!-- implementation and build dependencies -->
         <glassfish-management-api.version>3.2.3.payara-p1</glassfish-management-api.version>
-        <glassfish-corba.version>4.2.4.payara-p3</glassfish-corba.version>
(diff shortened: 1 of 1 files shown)


breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

Pandrex247 (approved):

PR #7238, opened 2025-03-10
Fish 8858 docker images do not log to file

luiseufrasio:
## Description
Payara Docker Images do not log to file server.log

### Blockers
None

### New tests
None

### Testing Performed
1. Use in WSL
2. mvn clean install -DskipTests at root directory
3. cd appserver/extras/docker-images/server-full
4. mvn clean package
5. cd target/docker
6. docker build -t payara/server-full:local -f ./payara/server-full/build/Dockerfile.jdk11 ./payara/server-full/tmp/docker-build/
7. docker run -it -e ENABLE_FILE_LOGS=true -p 8080:8080 -p 4848:4848 --name payara_docker payara/server-full:local
8. Verify that server.log file now is filled with logs and the system property com.sun.enterprise.server.logging.GFFileHandler.logtoFile is enabled

### Testing Environment
Zulu JDK 11 on Ubuntu 22.04 with Maven 3.9.0

## Documentation
PR will be created

## Notes for Reviewers
None

### DIFF
--- a/appserver/extras/docker-images/server-full/src/main/docker/bin/entrypoint.sh
+++ b/appserver/extras/docker-images/server-full/src/main/docker/bin/entrypoint.sh
@@ -41,10 +41,8 @@
 
 if [ "${ENABLE_FILE_LOGS:-false}" == "true" ]; then
       echo "[Entrypoint] Enabling Logs into server.log"
-      ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} start-domain ${DOMAIN_NAME}
-      ${PAYARA_DIR}/bin/asadmin --user=${ADMIN_USER} --passwordfile=${PASSWORD_FILE} \
-            set-log-attributes com.sun.enterprise.server.logging.GFFileHandler.logtoFile=true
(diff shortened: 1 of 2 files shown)


kalinchan:
Why not just change the value to true in the Dockerfile? 

If a user wants to disable this, they can do so via preboot/postboot commands?

luiseufrasio:
> Why not just change the value to true in the Dockerfile?
> 
> If a user wants to disable this, they can do so via preboot/postboot commands?

I suppose there is a reason for setting the value to false by default. Only one customer asked for the option to change it to true, so I did not want to change the default behavior for everyone.

kalinchan:
> I suppose there is a reason for setting the value to false by default. Only one customer asked for the option to change it to true, so I did not want to change the default behavior for everyone.

Looking at the commit history there doesn't seem to be a specific reason, I also don't see any potential errors that could arise from this. However, I suppose this is probably a product matter.

Pandrex247 (changes requested):
You don't _have_ to use domain1 in our docker images, it's just the default.

I'm not a fan of this method, and upon review I'm not even sure what the point of this is - you can just re-enable the log settings in a postboot script.

Pandrex247:
We will simply update the documentation

PR #7271, opened 2025-03-27
FISH-9749 Add option to skip jdbc client info validation

kalinchan:
## Description

Add option to admin console to skip client info validation.
Also throw a restart required when changed.

### Blockers

n/a
### New tests

none
### Testing Performed

Deployed an simple jpa example from Payara-Examples to verify if client info was skipped when set.
Also verified behaviour isn't changed dynamically
### Testing Environment

Java version: 21.0.5, vendor: Azul Systems, Inc., runtime: /Users/kalinchan/.sdkman/candidates/java/21.0.5-zulu/zulu-21.jdk/Contents/Home
OS name: "mac os x", version: "15.1.1", arch: "aarch64", family: "mac"

## Documentation

https://github.com/payara/Payara-Documentation/pull/578

### DIFF
--- a/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
+++ b/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
@@ -102,7 +102,7 @@ public class ConnectionHolder40 extends ConnectionHolder {
      */
     protected void init() {
         try {
-            if (!skipClientInfoValidation && isSupportClientInfo()) {
+            if (isSupportClientInfo()) {
                 defaultClientInfo = getClientInfo();
             }
         } catch (Exception e) {
@@ -400,6 +400,9 @@ public class ConnectionHolder40 extends ConnectionHolder {
      * @since 1.6
      */
     private boolean isSupportClientInfo() {
+        if (skipClientInfoValidation) {
+            return false;
+        }
         Boolean isSupportClientInfo = getManagedConnection().isClientInfoSupported();
         if (isSupportClientInfo != null) {
             return isSupportClientInfo;
--- a/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
+++ b/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
@@ -102,7 +102,7 @@ public class ConnectionHolder40 extends ConnectionHolder {
      */
     protected void init() {
         try {
-            if (isSupportClientInfo()) {
+            if (!skipClientInfoValidation && isSupportClientInfo()) {
                 defaultClientInfo = getClientInfo();
             }
         } catch (Exception e) {
--- a/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
+++ b/appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  *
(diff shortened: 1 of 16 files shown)


appserver/resources/resources-connector/src/main/java/org/glassfish/resources/admin/cli/ResourceConstants.java
```diff
@@ -259,6 +259,7 @@ public final class ResourceConstants {
     public static final String THREAD_LIFETIME_SECONDS = "thread-lifetime-seconds";
     public static final String TASK_QUEUE_CAPACITY = "task-queue-capacity";
     public static final String MAX_ASYNC = "max-async";
+    public static final String SKIP_CLIENT_INFO_VALIDATION = "skip-client-info-validation";
```
Pandrex247, line 262: You should move this up to under ` // JDBC Connection pool` - where it is you've put it under `// concurrent resource objects`

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/service/JdbcServiceConfigListener.java
```diff
@@ -0,0 +1,87 @@
...
+import java.beans.PropertyChangeEvent;
+import java.util.ArrayList;
+import java.util.List;
+
+@Service
+@RunLevel(StartupRunLevel.VAL)
+public class JdbcServiceConfigListener implements ConfigListener {
+
+    @Inject
+    private JdbcConnectionPool jdbcConnectionPool;
```
Pandrex247, line 62: Unused?
```suggestion
```

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/service/JdbcServiceConfigListener.java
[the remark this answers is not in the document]
kalinchan: Didn't work for me when testing - will try again

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/service/JdbcServiceConfigListener.java
kalinchan: According to the ConfigListener interface, it only registers for Injected @Configured classes hence the unused injection.

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/service/JdbcServiceConfigListener.java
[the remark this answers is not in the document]
Pandrex247: Actually there is a difference - having the ConfigListener under `JdbcAdminServiceImpl` has the "Restart Required" only get triggered on the DAS, not on all applicable instances.

So it's valid to have this in its own class

appserver/jdbc/jdbc-runtime/src/main/java/org/glassfish/jdbcruntime/service/JdbcServiceConfigListener.java
kalinchan: Seems like the Transaction registering the listener is enough. Removed.

appserver/resources/resources-connector/src/main/java/org/glassfish/resources/admin/cli/ResourceConstants.java
kalinchan: Resolved

appserver/jdbc/jdbc-ra/jdbc40/src/main/java/com/sun/gjc/spi/jdbc40/ConnectionHolder40.java
Pandrex247, line 1: Should we also be skipping in the close method? That also runs the `isSupportClientInfo` method

It might be worth moving `skipClientInfoValidation` into the method itself

Pandrex247 (changes requested):

Pandrex247 (approved):

luiseufrasio (approved):
LGTM

PR #7275, opened 2025-03-31
FISH-9527 Implement Movable Master Password

Viii3:
## Description

- Implements FISH-9527
  - Instead of saving the master password to the default location, the user has the option to save the master password to a different location.
  - A `masterpasswordlocation` parameter has been added to `create-domain`, `change-master-password`, `create-local-instance`.

### Testing Performed

- Created a new domain without a master password.
- Created a new domain with a master password (default location).
- Created a new domain with a master password (custom location).
- Changed master password to a new location.
- Created a new local instance with a master password (default location).
- Created a new local instance with a master password (custom location).

### Testing Environment

Maven version: 3.9.6

### DIFF
--- a/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
+++ b/nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
@@ -37,7 +37,7 @@
  * only if the new code is made subject to such option by the copyright
  * holder.
  */
-// Portions Copyright [2016-2024] [Payara Foundation and/or affiliates]
+// Portions Copyright [2016-2025] [Payara Foundation and/or affiliates]
 
 package com.sun.enterprise.admin.servermgmt.cli;
 
@@ -61,6 +61,7 @@ import com.sun.enterprise.util.io.FileUtils;
 import com.sun.enterprise.util.io.ServerDirs;
 import org.glassfish.api.ActionReport;
 import org.glassfish.api.admin.CommandException;
+import org.glassfish.grizzly.utils.Charsets;
 import org.w3c.dom.Document;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
@@ -80,14 +81,17 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.nio.file.Files;
 import java.security.KeyStore;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Stream;
 
 import static com.sun.enterprise.admin.servermgmt.domain.DomainConstants.MASTERPASSWORD_FILE;
+import static com.sun.enterprise.admin.servermgmt.domain.DomainConstants.MASTERPASSWORD_LOCATION_FILE;
 
 /**
  * A class that's supposed to capture all the behavior common to operation on a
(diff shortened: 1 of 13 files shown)


nucleus/admin/server-mgmt/src/main/java/com/sun/enterprise/admin/servermgmt/cli/LocalServerCommand.java
```diff
@@ -61,6 +61,7 @@
 import com.sun.enterprise.util.io.ServerDirs;
 import org.glassfish.api.ActionReport;
 import org.glassfish.api.admin.CommandException;
+import org.glassfish.grizzly.utils.Charsets;
```
svendiedrichsen, line 64: Why not the JDK StandardCharsets?

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/ChangeNodeMasterPasswordCommand.java
```diff
@@ -65,6 +66,7 @@
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.CommandException;
 import org.glassfish.api.admin.CommandValidationException;
+import org.glassfish.grizzly.utils.Charsets;
```
svendiedrichsen, line 69: Why not the JDK StandardCharsets?

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/CreateLocalInstanceCommand.java
```diff
@@ -56,21 +56,28 @@
...
 import java.util.ArrayList;
 import java.util.logging.Level;
+import java.util.logging.Logger;
+
 import org.glassfish.api.ActionReport;
 import org.glassfish.api.I18n;
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.CommandException;
 import org.glassfish.api.admin.CommandValidationException;
+import org.glassfish.grizzly.utils.Charsets;
```
svendiedrichsen, line 74: Why not the JDK StandardCharsets?

nucleus/cluster/cli/src/main/java/com/sun/enterprise/admin/cli/cluster/LocalInstanceCommand.java
```diff
@@ -67,6 +70,7 @@
 import org.glassfish.api.Param;
 import org.glassfish.api.admin.CommandException;
 import org.glassfish.api.admin.CommandValidationException;
+import org.glassfish.grizzly.utils.Charsets;
```
svendiedrichsen, line 73: Why not the JDK StandardCharsets?

Pandrex247 (changes requested):
You're missing an update to the `change-master-password.1` help page

I'm not sure this is correctly working with nodes either - a master-password file is still being created / left behind in the default location, and it doesn't seem to be creating a new one

`asadmin change-master-password --savemasterpassword --masterpasswordlocation D:\Downloads\test-password localhost-domain1`

Pandrex247 (changes requested):
There is some odd behaviour if saving the password to a directory.
If saving a password to a directory with nothing present, this wipes out the directory and replaces it with the master password file.
I would expect that when given a directory it doesn't delete the directory, and instead stores the password in that directory with the default name. That, or just errors out with a message along the lines of "Cannot save master password - file already exists at location"

If saving the password to a directory with stuff in it, it doesn't nuke the directory (good), but it fails without telling you why. You just get a generic error:

```
CLI130: Could not create domain, test-domain-2
Command create-domain failed.
```

Also, you're missing help text in `create-domain.1` and `create-local-instance.1` for `create-domain --help` and `create-local-instance --help` respectively

Pandrex247:
Also please address Sven's comments - you don't need to use Grizzly charsets

Pandrex247 (approved):
Feature branch created to allow us to work on this iteratively.
Outstanding issues will be addressed in follow-up PRs.

PR #7280, opened 2025-04-01
FISH-10857: adding changes to fix class issues to resolve dynamic proxy class creation

breakponchito:
Adding fixes to resolve dynamic proxy creation for Jakarta Data Implementation

## Description

This is a feature to fix issues related to resolve the dynamic proxy creation for Jakarta Data implementation. This solves an additional ticket here:  [FISH-10733](https://payara.atlassian.net/browse/FISH-10733)

### New tests

Deploy the example application with an interface annotated with Repository annotation from Jakarta Data. Here the reproducer:

[JakartaDataReproducer.zip](https://github.com/user-attachments/files/19545037/JakartaDataReproducer.zip)

after deploying and calling the following endpoint we will see the following message on the logs:

`http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world`

image

this means that the endpoint is calling a method to execute save operation using a proxy component that represents the Repository interfaces from Jakarta Data.

Something that helps to make this is the usage of the following dependency:

`
       <dependency>
            <groupId>io.github.classgraph</groupId>
            <artifactId>classgraph</artifactId>
            <version>${classgraph.version}</version>
        </dependency>
`

using this api will help us to resolve full information of the class that implements the Repository interfaces, including the Generic Information. This is needed to resolve the injection point during the lifecycle of the CDI:

image

This was the only way I found to get the full information. I tried to use the BeforeBeanDiscovery event from CDI with the method addAnnotatedType to register the desired annotation, but this doesn't work for me. Also I used another strategy to find the annotation with the annotation @AnnotationHandlerFor within a service class to locate and get the Interface type but this has partial information of the class that didn't resolve the injection point.  For now I don't found any other strategy to load the classes with complete information and just with the usage of the classgraph dependency this issue can be fixed.

### Testing Environment

OS windows 11, JDK zulu 21, maven 3.9.5
## Notes for Reviewers

[FISH-10733]: https://payara.atlassian.net/browse/FISH-10733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
```diff
@@ -0,0 +1,132 @@
...
+        repositoryTypes.forEach(t -> {
+            afterBeanDiscovery.addBean(new DynamicInterfaceDataProducer<>(t, manager, this));
+        });
+    }
+
+    public Set<Class<?>> getRepositoryClasses() {
+        Set<Class<?>> repositories = new HashSet<>();
+        //here is the place to start to use the classgraph api to get information from the classes that are 
+        //annotated by the jakarta data annotation 
+        try (ScanResult result = new ClassGraph().enableAllInfo().scan()) {
```
kalinchan, line 82: Would ClassGraph().enableClassInfo() be better here?

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/DynamicInterfaceDataProducer.java
```diff
@@ -0,0 +1,131 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) [2025] Payara Foundation and/or its affiliates. All rights reserved.
```
Pandrex247, line 4:
```suggestion
 *    Copyright (c) 2025 Payara Foundation and/or its affiliates. All rights reserved.
```

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
```diff
@@ -0,0 +1,132 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) [2025] Payara Foundation and/or its affiliates. All rights reserved.
```
Pandrex247, line 4:
```suggestion
 *    Copyright (c) 2025 Payara Foundation and/or its affiliates. All rights reserved.
```

core/pom.xml
```diff
@@ -203,6 +203,8 @@
         <payara.transformer.version>0.2.13</payara.transformer.version>
 
         <awaitility.version>4.3.0</awaitility.version>
+        <!-- classgraph dependency used to solve issue when creating proxy implementations for jakarta data-->
+        <classgraph.version>4.8.179</classgraph.version>
```
Pandrex247, line 207: This dependency doesn't seem to be used in core and so shouldn't be here, it should be moved to the top-level aggregator or appserver pom

appserver/packager/data/pom.xml
```diff
@@ -89,6 +89,16 @@
...
         </dependency>
+        <dependency>
+            <groupId>fish.payara.server.internal.data</groupId>
+            <artifactId>data-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.github.classgraph</groupId>
+            <artifactId>classgraph</artifactId>
+            <version>${classgraph.version}</version>
```
Pandrex247, line 100: This should be defined in dependency-management in the top-level or appserver parent pom

Pandrex247:
Ran through our currently passing TCKs, nothing breaks 👍

PR #7296, opened 2025-04-11
FISH-10499: Adding simple crud operation using jakarta data interface methods

breakponchito:
Adding implementation of simple CRUD operations using jakarta data interfaces

## Description

This is a feature to include the execution of CRUD operations using the Jakarta Data interfaces

### Testing Performed

Manual testing using the following r
[JakartaDataReproducer.zip](https://github.com/user-attachments/files/19769500/JakartaDataReproducer.zip)
eproducer:

The current implementation on the server can resolve the following operation for a simple entity on the default database on Payara Server:

We need a class  on an application that implements the CrudRepository interface and also this interface should need to be annotated by the annotation @Repository, here the example:

image

where the Author is a valid entity defined on the application that follows Jakarta Persistence rules for its definition

then on the application we can define endpoints to use our interface as follows:

image

we need to inject the interface to use the methods provided by the Jakarta Data interfaces. From the reproducer we have the following endpoints to interact with the database using Jakarta Data CRUD operation:

This will insert a new entity on the database
- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/insert
This will save a new entity on the database
- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/save
This will update the indicated id for an entity
- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/update/1
this will delete the indicated id from the database
- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/delete/1
this will get all available entities from the database
- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/find

For now the implementation is using default embedded H2 database from the server. You can create a connection to the database to verify the previous behavior described here:

create a connection to the default  location of the embedded database

`
C:/java/projects/Payara/appserver/distributions/payara/target/stage/payara7/glassfish/domains/domain1/lib/databases/embedded_default.mv.db
`

image

then execute querys to verify results from each operation

image

### Testing Environment

Windows 11,  zulu 21 jdk, maven 3.9.5

### DIFF
--- a/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
+++ b/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
@@ -66,6 +66,15 @@ import java.util.stream.Collectors;
 public class JakartaDataExtension implements Extension {
 
     private static final Logger logger = Logger.getLogger(JakartaDataExtension.class.getName());
+    
+    private String applicationName;
+
+    public JakartaDataExtension() {
+    }
+    
+    public JakartaDataExtension(String appName) {
+        this.applicationName = appName;
+    }
 
     public void afterBeanDiscovery(@Observes AfterBeanDiscovery afterBeanDiscovery, BeanManager manager) {
         logger.info("Finishing scanning process");
@@ -79,7 +88,7 @@ public class JakartaDataExtension implements Extension {
         Set<Class<?>> repositories = new HashSet<>();
         //here is the place to start to use the classgraph api to get information from the classes that are 
         //annotated by the jakarta data annotation 
-        try (ScanResult result = new ClassGraph().enableAllInfo().scan()) {
+        try (ScanResult result = new ClassGraph().enableClassInfo().enableAnnotationInfo().scan()) {
             repositories.addAll(locateAndGetRepositories(result));
         }
         return repositories.stream().filter(cl -> {
@@ -128,5 +137,9 @@ public class JakartaDataExtension implements Extension {
             return entity;
         }
         return null;
+    }
+
+    public String getApplicationName() {
+        return applicationName;
     }
 }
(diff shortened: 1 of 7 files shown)


appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
```diff
@@ -79,7 +88,7 @@ public Set<Class<?>> getRepositoryClasses() {
         Set<Class<?>> repositories = new HashSet<>();
         //here is the place to start to use the classgraph api to get information from the classes that are 
         //annotated by the jakarta data annotation 
-        try (ScanResult result = new ClassGraph().enableAllInfo().scan()) {
+        try (ScanResult result = new ClassGraph().enableClassInfo().enableAnnotationInfo().scan()) {
```
kalinchan, line 91: Do we need to include `.enableClassInfo()`?

If the goal is to just get specific annotated classes just `.enableAnnotationInfo()` should be enough?

kalinchan:
Functionality wise, looks good and the reproducer works for me, running through the Jakarta EE 11 TCKs (https://jenkins.payara.fish/view/TCKs/job/JakartaEE-11-TCK/170)

kalinchan (approved):
LGTM - tcks pass and reproducer works, other than general comments

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/JakartaDataExtension.java
breakponchito: yes you are right after removing the enableClassInfo the implementation still worked well, so we don't nee the ClassInfo just the AnnotationInfo, good catch

breakponchito: Already done!!

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/QueryData.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class QueryData {
+
+    private Class<?> repositoryInterface;
+    private Method method;
+    private Class<?> entityParamType;
+    private Class<?> declaredEntityClass;
+    private QueryType queryType;
+
+    public enum QueryType {
```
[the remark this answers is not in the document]
breakponchito: already separate the enum class

PR #7297, opened 2025-04-16
FISH-10876 Fix error page property parsing

kalinchan:
## Description

When specifying a custom error page within a preboot file, this wouldn't work due to the property being parsed incorrectly.

This PR will strip the property value of any additional quotation marks.

i.e.
property = "code=500 reason=example path=/wibbles/wobbles.html"
will result in = code=500 reason=example path=/wibbles/wobbles.html

Note the stripped quotation marks.

### Blockers

n/a
### New tests

Created payara micro programmatic test 

go to payara-samples directory and run `mvn clean verify -Ppayara-micro-managed -Dpayara.version=6.2025.5-SNAPSHOT -Dtest=CustomErrorPageTest` (might have to comment out other modules)
### Testing Performed

1. Created a 404 html page somewhere on my local device.
2. Created preboot command to specify the property:
`set configs.config.server-config.http-service.virtual-server.server.property.send-error_1="code=404 path=/Users/kalinchan/Downloads/custom-404.html reason=not_found"`
3. Ran payara-micro
`java -jar payara-micro.jar --prebootcommandfile preboot.txt --deploy clusterjsp.war`
## Notes for Reviewers

Payara-micro-progamattic tests wouldn't compile so made some changes to fix that 

There are also some issues with payara-micro-programmatic, but that is another issue

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
@@ -38,7 +38,7 @@
  * holder.
  */
 
-// Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2025] [Payara Foundation and/or its affiliates]
 
 package com.sun.enterprise.web;
 
@@ -1166,6 +1166,7 @@ public class VirtualServer extends StandardHost implements org.glassfish.embedda
                 _logger.log(Level.WARNING, LogFacade.NULL_VIRTUAL_SERVER_PROPERTY, getID());
                 continue;
             }
+            propValue = propValue.replace("\"", "");
 
             if (!propName.startsWith("send-error_")) {
                 continue;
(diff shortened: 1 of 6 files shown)


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
```diff
@@ -1166,6 +1166,7 @@ void configureErrorPage() {
                 _logger.log(Level.WARNING, LogFacade.NULL_VIRTUAL_SERVER_PROPERTY, getID());
                 continue;
             }
+            propValue = propValue.replace("\"", "");
```
aubi, line 1169: I think the purpose was to remove the quotes from beginning and end, not all of them.
I would expect some test, if there is a quote char at the first and last place and then remote them.

aubi (changes requested):

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
breakponchito: also agree with this comment will suggest to add some king of integration tests for this

breakponchito:
@kalinchan Could you include your 404.html page to reproduce the issue? I know that is a simple page but to reproduce please include the content needed

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/VirtualServer.java
kalinchan: Added a test

breakponchito:
@kalinchan Already tested, looks great

breakponchito (approved):
LGTM

aubi (approved):
LGTM

PR #7306, opened 2025-04-23
FISH-10805 Remove Semantic Versioning Check from Payara Core

Pandrex247:
## Description
The experiment is over, we have decided that getting the codebase into a less monolithic state with individually versioning APIs and modules is too much effort, and even when complete adds a lot of mental overhead for all of the different versions of Enterprise we support and adds complexity to any attempt to stagger changes between Community and Enterprise.

This is the first PR of a few which will look to remove the concept of Core vs. Server.
Step 1 (FISH-10805) - removing the semantic versioning enforcement we have.

### Blockers
None

### New tests
None

### Testing Performed
* Built the server - no errors.
* Made a breaking API change and compiled the module - no errors

### Testing Environment
Windows 11, Maven 3.9.9, Zulu Java 11.0.26

## Documentation
_Pending..._

## Notes for Reviewers
None

PR #7308, opened 2025-04-25
FISH-10806 FISH-10807 FISH-10809 Remove Core and Realign Modules 

Pandrex247:
## Description
Second PR, follow up to https://github.com/payara/Payara/pull/7306

Removes the `core-aggregator`, `core-bom`, and `core-parent` modules, and reintegrates and aligns the Core modules these managed back into the respective `payara-aggregator`, `payara-nucleus-parent`, and `payara-parent` poms.

Also adds applicable dependencies back into `payara-bom`, and adds the required `${project.version}` dependency version back for those internal dependencies not covered by the `payara-bom`.

I have taken efforts to keep the property and plugin config aligned with where they were before the Core vs. Server split - even when it would be potentially beneficial to move them. The intent is that these will be moved in follow up commits.

Additional weirdness and tasks I spotted which I will also look to address separately (if time permits):
* Fix `org.glassfish.hk2:config-generator` - this is a "dead" module, but shouldn't be. Where it currently exists is a cyclic dependency, which we "get around" by not actually using the version in our code and instead pull in an extremely old version published to Maven Central (2.5.0-b53)
* Flashlight module parents - these don't seem to use their own parent module and instead use `nucleus/common`? I assume this is a copy-paste error that occurred at some point, as it's not immediately obvious to me why this is the case
* Remove `production` domain - this is unused in Community
* Misaligned HK2 dependency in `appserver/payara-appserver-modules/jmx-monitoring` - it's using `org.glassfish.hk2:core:2.3.0` instead of `org.glassfish.hk2:hk2-core:3.0.1.payara-p4`
* Remove unused `appserver/admin/runtime` module?
* Remove unused `appserver/registration` module?
* Remove unused `nucleus/cluster/uc`, `nucleus/cluster/windows`, `nucleus/cluster/uc-l10n` modules
  * At the very least, we should exclude `uc-l10n` from the build, as the module it's meant to be localising isn't included!
* ~Remove or reorder MicroProfile repository from `appserver/pom` - its placement has it get pinged first instead of Maven Central, plus I'm not sure we need it anyway~ - now also included
* ~Remove duplicate `payara-nexus-artifacts` repository entry from `nucleus/pom`~ - now also included
* ~Remove duplicate `maven-deploy-config` `<dependencyManagement>` entry in `nucleus/pom`~ - now also included
* Fix Javadoc generation for the distribution - it's broken and just generates an empty jar (it was broken before outside of the Core modules as well)
* Change `deploy.skip` to true except for distributions, which should have `deploy.skip` set to false - tackled [here](https://github.com/payara/Payara/pull/7309)

### Blockers
None

### New tests
None

### Testing Performed
Wiped maven repo and built the server: Server, Micro, Embedded, and Docker all resolve.
Started Payara Server and loaded admin console - no explosions.

### Testing Environment
Windows 11, Maven 3.9.9, Zulu JDK 11.0.26

## Documentation
Pending...

## Notes for Reviewers
Trust Me™


kalinchan (approved):

breakponchito (approved):
LGTM

PR #7323, opened 2025-04-30
FISH-10506: adding By annotation processing for methods deleteById and findById

breakponchito:
Add support for @By annotation on methods deleteById and findById from Repository Interfaces

## Description

This is a feature to include more functionality for Jakarta Data implementation. In this case we are including the support of the @By annotation for two methods from the Repository interfaces, the list of methods are the following:

From BasicRepository

`
@Find
Optional<T> findById(@By("id(this)") K id);
`

`
@Delete
void deleteById(@By("id(this)") K id);
`

### Testing Performed

Manual testing using the following reproducer:

[JakartaDataReproducer.zip](https://github.com/user-attachments/files/19971058/JakartaDataReproducer.zip)

From the reproducer use the following endpoint to create 3 new entities:

- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/insertList

then you can use the new endpoint to find specific id

image

- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/findById/1

image

image

the other new method is to delete specific entity by id with the following endpoint:

image

- http://localhost:8080/JakartaDataReproducer-1.0-SNAPSHOT/api/hello-world/deleteById/1

image

you can delete all the entities available and then execute the find operation to see empty result:

image

### Testing Environment

Windows 11, zulu 21 jdk, maven 3.9.5

### DIFF
--- a/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/RepositoryImpl.java
+++ b/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/cdi/extension/RepositoryImpl.java
@@ -39,30 +39,35 @@
  */
 package fish.payara.jakarta.data.core.cdi.extension;
 
+import fish.payara.jakarta.data.core.util.FindOperationUtility;
+import jakarta.data.repository.By;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.EntityManagerFactory;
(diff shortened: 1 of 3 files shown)


breakponchito:
Jenkins test please

kalinchan (approved):

PR #7379, opened 2025-05-19
FISH-10885 DON'T MERGE - Dummy change go test with JDK 11 also

stian-sigvartsen:
## Description
Just for the purpose of testing tests with JDK 11


stian-sigvartsen:
Jenkins run revealed that https://payara.atlassian.net/browse/FISH-11145 does not occur with JDK 11. The PR has served its purpose.

PR #7404, opened 2025-05-27
(Empty) Bundle for c.s.l.e.s.c.ejb

pzygielo:
## Description

To avoid log:
```
[#|2025-05-27T17:21:40.854+0200|INFO|Payara 6.2025.5||_ThreadID=33;_ThreadName=RunLevelControllerThread-1748359277816;_TimeMillis=1748359300854;_LevelValue=800;|
  Cannot find the resource bundle for the name com.sun.logging.enterprise.system.container.ejb for class org.glassfish.ejb.persistent.timer.TimerState using fish.payara.server.internal.ejb.ejb-full-container [70]|#]
```

### Testing Performed

Deployed app using TimerService. Before this change, the entry as above is present in server.log. After this change it's absent.

### Testing Environment

JDK 21


Pandrex247 (approved):
Thanks

PR #7437, opened 2025-06-13
FISH-11244 Parse Query Method

aubi:
## Description
Adding parser for the Jakarta Data query method name. So far, it just parses the name and provides data from the name.
Next task: integrate with the annotations and generate the final JPQL.

### New tests
There are 24 new JUnit tests. All queries mentioned in the specification are passing. I suppose there will be some work with the TCKs.

### Testing Environment
OpenJDK 21, Linux

## Documentation
https://jakarta.ee/specifications/data/1.0/jakarta-data-addendum-1.0

### DIFF
--- a/appserver/data/data-core/src/test/java/fish/payara/jakarta/data/core/querymethod/QueryMethodParserTest.java
+++ b/appserver/data/data-core/src/test/java/fish/payara/jakarta/data/core/querymethod/QueryMethodParserTest.java
@@ -390,20 +390,24 @@ public class QueryMethodParserTest {
     }
 
     // ----------------- Aditional Tests ---------------------
-
+    /**
+     * Validate Or.
+     *
+     * @throws QueryMethodSyntaxException
+     */
     @Test
     public void testFindByTitleOrAuthor() throws QueryMethodSyntaxException {
         QueryMethodParser parser = new QueryMethodParser("findByTitleOrAuthor").parse();
         assertEquals(
-            List.of(
-                    new QueryMethodParser.Condition(LogicalOperator.NONE, "Title", false, false, null),
-                    new QueryMethodParser.Condition(LogicalOperator.OR, "Author", false, false, null)
-            ),
-            parser.getConditions());
+                List.of(
+                        new QueryMethodParser.Condition(QueryMethodParser.LogicalOperator.NONE, "Title", false, false, null),
+                        new QueryMethodParser.Condition(QueryMethodParser.LogicalOperator.OR, "Author", false, false, null)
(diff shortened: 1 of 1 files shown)


appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/querymethod/OptionalPartsParser.java
```diff
@@ -0,0 +1,244 @@
...
+        updatedQueryString.append(queryString.substring(position));
+    }
+
+    /**
+     * Add ORDER BY with sorts. The original query shouldn't have ORDER BY if there are sorts,
+     * otherwise we'll get a corrupted query, with 2 ORDER BY clauses
+     */
+    private void appendSorts() {
+        if (sorts != null && !sorts.isEmpty()) {
+            updatedQueryString.append("ORDER BY ");
```
luiseufrasio, line 136: Is it possible to the updateQueryString already has an ORDER BY clause? Should you verify it before appending?

luiseufrasio (commented):
Added some tests and question about appending ORDER BY

aubi:
@luiseufrasio Thanks for the tests!

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/querymethod/OptionalPartsParser.java
aubi:
```
    /**
     * Add ORDER BY with sorts. The original query shouldn't have ORDER BY if there are sorts,
     * otherwise we'll get a corrupted query, with 2 ORDER BY clauses
     */
    private void appendSorts() {
````
The idea is, that the orders come from the annotation and it's forbidden to use both, annotation and ORDER BY in the query. So it's completely OK to fail. It's user's mistake.

This piece of code doesn't parse the whole JPQL! It parses only the necessary beginning to be able to fill missing SELECT and FROM. Everything after FROM is just copied. And optionally adding GROUP BY from the annotation.
A space should be added, of course.

aubi: Also, it is only TEMPORARY solution! In the next version of JPQL, the missing parts will be allowed by the spec and we will just pass the query to JPA. Just now, the Jakarta Data for some strange reason uses future extension of JPA.

luiseufrasio:
LGTM!

luiseufrasio (approved):
LGTM!

PR #7452, opened 2025-06-19
FISH-11377 add dependabot payara7

abdulrahim458:


Pandrex247 (approved):

Pandrex247:
In hindsight, I think this is a bad idea

PR #7485, opened 2025-06-27
FISH-11198 Fix Missing Message for Empty Deployment Group Name

aubi:
## Description
There was same message for missing name of Instance and Deployment Group. This fix makes them separate.
Also i18n was used instead of a hardcoded message.

### Testing Performed
Steps to Reproduce:
* Go to deployment groups
* Click “New…”
* Do NOT enter Deployment Group Name nor click Auto Name
* Click OK
* A message “No deployment group name provided and Auto Name is not enabled.”

Similarly with Instances, the message will be "No instance name provided and Auto Name is not enabled.".

### Testing Environment
Linux, OpenJDK 21


jtarry-payara (approved):
LGTM

PR #7494, opened 2025-06-30
FISH-11379 FISH-11381 FISH-11132 integrate count/delete/exists with parse by name

luiseufrasio:


stian-sigvartsen:
@breakponchito the issue seems to only happen when you request a different entity (Company) which has a relationship to the deleted entity (Employee). I am still debugging...

stian-sigvartsen:
I believe we can remove `QueryDate#setParamIndex` and only call `Query#setDeclaredEntityClass` during deployment time to make `QueryData `thread safe. I can push some commits.

With that we can move the use of `MethodNameParser `to deployment time also, and achieve an abstraction.

breakponchito:
@luiseufrasio I tested and now the behavior looks correct

PR #7526, opened 2025-07-09
FISH-11618 bugfix(cdi-provider): getBundleDescriptor() added a check for non-exi…

### LINKED ISSUE #7492: Bug Report: database transaction with EntityListener using CDI in @Asynchronous process results in NPE
### Brief Summary

We have the following situation:

- An `@Asynchronous` process, which manipulates an entity in a database using JPA
- An EntityListener on said entity which performs some processing on post-persist
- The EntityListener requires some dependencies retrieved using `CDI.current().select(...)`

This results in a NullPointerException when the EntityListener is created and tries to fetch dependencies.

### Expected Outcome

Database transaction succeeds without NullPointerExceptions.

### Current Outcome

Database transaction fails because the EntityListener cannot be created with the following exception:

```
java.lang.NullPointerException: Cannot invoke "org.glassfish.api.invocation.ComponentInvocation.getJNDIEnvironment()" because the return value of "org.glassfish.api.invocation.InvocationManager.getCurrentInvocation()" is null
	at org.glassfish.weld.GlassFishWeldProvider.getBundleDescriptor(GlassFishWeldProvider.java:166)
	at org.glassfish.weld.GlassFishWeldProvider.getCDI(GlassFishWeldProvider.java:149)
	at jakarta.enterprise.inject.spi.CDI.getCDIProvider(CDI.java:78)
	at jakarta.enterprise.inject.spi.CDI.current(CDI.java:65)
	at com.github.bvschaik.reproducer.cdi.SampleEntityListener.<init>(SampleEntityListener.java:15)
    ... more
```

Digging deeper into `EjbAsyncTask`, `BaseContainer` and `InvocationManager`, it appears that the current invocation is cleared in postInvoke() _before_ the transaction is committed. So when the EntityListener is created, there no longer is a current invocation, causing the NPE.

### Reproducer

https://github.com/bvschaik/payara-cdi-error

### Operating System

MacOS, Linux

### JDK Version

openjdk 21.0.6

### Payara Distribution

Payara Micro

lprimak:
…stent invocation

## Description
This is a bug fix. `getBundleDescriptor()` was causing a NPE during asynchronous operations
because invocation stack was null.
The fix lets CDI use an existing fallback method of finding out the correct WAR via current ClassLoader.

fixes #7492

## Testing
Tested with both reproducers for the attached issue.

### DIFF
--- a/appserver/web/weld-integration/src/main/java/org/glassfish/weld/GlassFishWeldProvider.java
+++ b/appserver/web/weld-integration/src/main/java/org/glassfish/weld/GlassFishWeldProvider.java
@@ -163,7 +163,10 @@ public class GlassFishWeldProvider implements CDIProvider {
 
     private static BundleDescriptor getBundleDescriptor() {
         BundleDescriptor bundle = null;
-        Object componentEnv = invocationManager.getCurrentInvocation().getJNDIEnvironment();
+        Object componentEnv = null;
+        if (!invocationManager.isInvocationStackEmpty()) {
+            componentEnv = invocationManager.getCurrentInvocation().getJNDIEnvironment();
+        }
         if( componentEnv instanceof EjbDescriptor) {
             bundle = (BundleDescriptor)
                     ((EjbDescriptor) componentEnv).getEjbBundleDescriptor().


lprimak:
@Pandrex247 a simple one for you :) Thank goodness!

lprimak:
jenkins test

Pandrex247 (approved):

Pandrex247:
Thanks!

PR #7527, opened 2025-07-09
FISH-9460 FISH-10142 Woodstox and Metro

Viii3:
## Description

- Resolves FISH-9460
- Resolves FISH-10142

### Blockers

- Depends on [Metro 4.0.4](https://github.com/payara/patched-src-metro-wsit/pull/24) patched project.


breakponchito:
@Viii3 This is still on draft

PR #7529, opened 2025-07-11
FISH-11465 Existing Realms Don’t Display Any Properties, Just Name - Fix Missing Properties

NotedSalmon:
FISH-11465 changed realmAttrs so that the extra properties are set to visible if the condition is met (if the realm class name = the specific section)

## Description

in `realmAttrs` the extra properties visibility was set to false which resulted in them not showing. I have changed it so that the extra properties are visible if the class matches. It makes sure it displays the correct section relating to the class name.

### Testing Performed

Run Payara 6, Security > Realms > fileRealm > make sure extra properties is visible > change something & save > reload realm to make sure change is made


jGauravGupta (approved):
LGTM

PR #7561, opened 2025-07-28
FISH-11624 Add surefire and failsafe rerun options

astoica1:
## Description
This PR just adds surefire and failsafe rerun options for the test stages


Pandrex247 (changes requested):
Also needs adding for each MicroProfile TCK

Pandrex247 (approved):

PR #7562, opened 2025-07-28
FISH-11591 Jakarta Data - Add resolution for multiple persistence units

kalinchan:
## Description

Introduces a new property "fish.payara.data.usePU" to specify the persistence unit to use when multiple are present.

Only uses the one which is true, if there are multiple, an exception is thrown.
### Blockers

n/a
### New tests

n/a
### Testing Performed

Added another persistence unit:
```
<persistence-unit name="pu2" transaction-type="JTA">
        <provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
        <jta-data-source>jdbc/__default</jta-data-source>

        <properties>
            <property name="fish.payara.jakarta.data.usePU" value="true"/>
        </properties>
    </persistence-unit>
```

### DIFF
--- a/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/DataCommonOperationUtility.java
+++ b/appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/DataCommonOperationUtility.java
@@ -64,13 +64,13 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.WildcardType;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
+
 import org.glassfish.internal.api.Globals;
 import org.glassfish.internal.data.ApplicationInfo;
 import org.glassfish.internal.data.ApplicationRegistry;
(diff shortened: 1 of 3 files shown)


appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/UnambiguousPersistenceUnitException.java
```diff
@@ -0,0 +1,46 @@
...
+ *     recipient has the option to distribute your version of this file under
+ *     either the CDDL, the GPL Version 2 or to extend the choice of license to
+ *     its licensees as provided above.  However, if you add GPL Version 2 code
+ *     and therefore, elected the GPL Version 2 license, then the option applies
+ *     only if the new code is made subject to such option by the copyright
+ *     holder.
+ */
+package fish.payara.jakarta.data.core.util;
+
+public class UnambiguousPersistenceUnitException extends RuntimeException {
```
Pandrex247: Is this unambiguous? 
I thought the issue is that it **is** ambiguous?

This being an error would mean the correct state is that there **should** be ambiguity?

Pandrex247 (dismissed):

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/DataCommonOperationUtility.java
```diff
@@ -124,10 +124,26 @@ public static EntityManager getEntityManager(String applicationName) {
...
             return factory.createEntityManager();
         }
-        return null;
+
+        List<EntityManagerFactory> result = factoryList.stream().filter(factory -> {
+            Map<String, Object> properties = factory.getProperties();
+            if (properties == null || properties.isEmpty()) {
+                return false;
+            }
+            if (!properties.containsKey("fish.payara.jakarta.data.usePU")) {
```
Pandrex247: Not a blocker, but I'm not a fan of this name.
I'd drop the jakarta.

Also, since this is being used in 3 separate locations here - an enum or static final string for this property name seems appropriate.

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/DataCommonOperationUtility.java
```diff
@@ -124,10 +126,26 @@ public static EntityManager getEntityManager(String applicationName) {
...
+                return false;
+            }
+            return Boolean.parseBoolean((String) properties.get(PERSISTENCE_UNIT_ENABLED_PROPERTY));
+        }).toList();
+
+        if (result.size() == 1) {
+            EntityManagerFactory factory = result.getFirst();
+            return factory.createEntityManager();
+        }
+        throw new AmbiguousPersistenceUnitException(String.format("Multiple persistence units found for application '%s', try setting the '%s' property to true", applicationName, PERSISTENCE_UNIT_ENABLED_PROPERTY));
```
Pandrex247: This text is potentially misleading.
Would you not also throw this exception if you had multiple persistence units with the property enabled?

Pandrex247 (commented):
I wonder if there is a way to use qualifiers instead of this flag... 🤔 
It's potentially a more elegant and flexible solution (assuming it makes sense)

Would need to look into it a bit more.

This is probably fine as an iterative step towards that though

appserver/data/data-core/src/main/java/fish/payara/jakarta/data/core/util/DataCommonOperationUtility.java
Pandrex247: Also it doesn't tell you _where_ to set the property to true

PR #7603, opened 2025-08-18
Fish 11805 fix failing assertion errors entity tests

luiseufrasio:

### DIFF
--- a/appserver/data/data-core/src/main/java/fish/payara/data/core/util/FindOperationUtility.java
+++ b/appserver/data/data-core/src/main/java/fish/payara/data/core/util/FindOperationUtility.java
@@ -201,11 +201,9 @@ public class FindOperationUtility {
 
             Query query = em.createQuery(dataForQuery.getQueryString());
 
-            // Set parameters
-            int paramIndex = 1;
             for (int i = 0; i < args.length; i++) {
                 if (!excludeParameter(args[i])) {
-                    query.setParameter(paramIndex++, args[i]);
+                    query.setParameter(i + 1, args[i]);
                 }
             }
 
--- a/appserver/data/data-core/src/main/java/fish/payara/data/core/util/QueryByNameOperationUtility.java
+++ b/appserver/data/data-core/src/main/java/fish/payara/data/core/util/QueryByNameOperationUtility.java
@@ -202,11 +202,6 @@ public class QueryByNameOperationUtility {
 
         DataParameter parameter = extractDataParameter(args);
         Limit limitFromArgs = parameter.limit();
-                Arrays.stream(args)
-                        .filter(arg -> arg instanceof Limit)
-                        .map(arg -> (Limit) arg)
-                        .findFirst()
-                        .orElse(null) : null;
 
         try {
             QueryMethodParser parser = new QueryMethodParser(methodName).parse();
@@ -247,6 +242,7 @@ public class QueryByNameOperationUtility {
 
         buildQueryClause(jpql, executionAction, dataForQuery, rootAlias);
 
+        DataParameter parameter = extractDataParameter(args);
         List<Sort<?>> dynamicSorts = parameter.sortList();
         buildJoins(joinClause, parser, rootEntityType, rootAlias, joinAliases);
         if (executionAction == QueryMethodParser.Action.FIND) {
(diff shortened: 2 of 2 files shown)


appserver/data/data-core/src/main/java/fish/payara/data/core/util/FindOperationUtility.java
```diff
@@ -191,6 +314,13 @@ public static void createQueriesForPagination(PageRequest pageRequest, Method me
                                                   StringBuilder builder,
                                                   Object[] args, List<Sort<?>> sortList, String rootAlias) {
         boolean forward = pageRequest == null || pageRequest.mode() != PageRequest.Mode.CURSOR_PREVIOUS;
+
+        System.out.println("=== DEBUG createQueriesForPagination ===");
```
breakponchito: Could you add logger instead of using System.out.println?

breakponchito:
> It looks as if you are building JPQL parser from scratch. Eclipselink contains such a parser. So if this is of interest: https://eclipse.dev/eclipselink/api/4.0/org.eclipse.persistence.core/org/eclipse/persistence/internal/jpa/jpql/HermesParser.html

not exactly JPQL, this is a new query format named Query By Method Name, here the spec:https://jakarta.ee/specifications/data/1.0/jakarta-data-addendum-1.0 also we need to provide the other new format named JDQL, here the spec: https://jakarta.ee/specifications/data/1.0/jakarta-data-1.0#_jakarta_data_query_language

appserver/data/data-core/src/main/java/fish/payara/data/core/util/FindOperationUtility.java
```diff
@@ -144,21 +188,24 @@ public static Object processFindByOperation(Object[] args, EntityManager em, Que
...
-            //check order conditions to improve select queries
+
             Query query = em.createQuery(dataForQuery.getQueryString());
+
+            // Set parameters
+            int paramIndex = 1;
             for (int i = 0; i < args.length; i++) {
                 if (!excludeParameter(args[i])) {
-                    query.setParameter(i + 1, args[i]);
+                    query.setParameter(paramIndex++, args[i]);
```
breakponchito: Here you are expanding the scope of the variable, if we maintain only the internal i+1, we are creating and deleting the variable until the for cycle finish. just in case you need the variable out of the scope of the for cycle this will make sense

appserver/data/data-core/src/main/java/fish/payara/data/core/util/QueryByNameOperationUtility.java
```diff
@@ -193,9 +198,15 @@ public static Object processExistsByNameOperation(Object[] args, QueryData dataF
     private static Object buildAndExecuteQuery(Object[] args, QueryData dataForQuery, EntityManager entityManager, QueryMethodParser.Action expectedAction) {
         Method method = dataForQuery.getMethod();
         String methodName = method.getName();
-
         boolean evaluatePages = paginationPredicate.test(dataForQuery.getMethod());
 
+        Limit limitFromArgs = args != null && args.length > 0 ?
```
breakponchito: Instead of calling this block of code you can reuse the method created to review sort criteria that return a reference of the DataParameter record with the available Limit, and sortlist if available, please check:
https://github.com/payara/Payara/blob/Payara7/appserver/data/data-core/src/main/java/fish/payara/data/core/util/DataCommonOperationUtility.java#L412
```suggestion
        DataParameter parameter = extractDataParameter(args);
        Limit limitFromArgs = parameter.limit();
```

appserver/data/data-core/src/main/java/fish/payara/data/core/util/QueryByNameOperationUtility.java
```diff
@@ -228,59 +245,77 @@ private static jakarta.persistence.Query buildQueryFromParser(QueryMethodParser
         Map<String, String> joinAliases = new HashMap<>();
 
         buildQueryClause(jpql, executionAction, dataForQuery, rootAlias);
+
+        List<Sort<?>> dynamicSorts = collectSortsFromArgs(dataForQuery.getMethod(), args);
```
breakponchito: if you define the reference of the DataParameter, you also will have the sortlist available from the args, you don't need to call the collectSortsFromArgs, we need to see if this is needed because we have duplicated functionality
```suggestion
        List<Sort<?>> dynamicSorts = parameter.sortList();
```

appserver/data/data-core/src/main/java/fish/payara/data/core/util/QueryByNameOperationUtility.java
```diff
@@ -228,59 +245,77 @@ private static jakarta.persistence.Query buildQueryFromParser(QueryMethodParser
...
         return q;
     }
 
+    /**
+      * Collect all dynamic sorts received by argument:
+      * - jakarta.data.Order
+      * - jakarta.data.Sort
+      * - varargs Sort...
+      */
+    private static List<Sort<?>> collectSortsFromArgs(Method method, Object[] args) {
```
breakponchito: not needed if we reuse the functionality added here:  https://github.com/payara/Payara/blob/Payara7/appserver/data/data-core/src/main/java/fish/payara/data/core/util/DataCommonOperationUtility.java#L412

PR #7607, opened 2025-08-20
Reduced logging level for "Clustered CDI Event bus initialized"

mkarg:
Reducing from INFO to FINE as no administrator actually wants to see literally hundreds to thousands of time this message in real-word production scenarios. Developers can easily configure their log to see that message, if they really want to. The defaults should be for production use, not for developers.


lprimak (approved):
Nice one. This log is useless

Pandrex247 (approved):

PR #7610, opened 2025-08-21
FISH-11138 add support for warlibs to admin console

NotedSalmon:
FISH-11138 add support for warlibs to admin console

## Description

I have created a checkbox in the deployment page that uses the `strings.properties` for labels and also sets the parameter in `DeployCommandParameters` accordingly.

In `DeployCommand` I have added a check so that if the boolean value of the parameter is true, it will call `DeployCommandUtils.useWarLibraries` using the context and enable warlibs.

### Testing Performed

I added logs to the condition check to make sure it did get activated. I also tested with the clusterJSP and the ObservabilityTool (the only ones I had around) and there was no issues I could find.


nucleus/common/glassfish-api/src/main/java/org/glassfish/api/deployment/DeployCommandParameters.java
```diff
@@ -179,6 +179,10 @@ public Boolean isLoadOnly() {
       return loadOnly;
     }
 
+    @Param(name=ParameterNames.WARLIBS, optional=true, defaultValue="false")
+    public Boolean useWarLibs = false;
+    public Boolean isUseWarLibs() {return useWarLibs;}
```
Pandrex247: We already have a parameter for this?
It's defined via `--properties`

aubi: Following the example of KEEP_STATE, it's done the same way...
Anyway, after the second commit it seems working, at least DeploymentUtils.useWarLibraries returns true.

Pandrex247: From offline discussion: 

We should investigate configuring it in a similar way to Implicit CDI.

__Alternatively/in addition__: if we wish to retain the `useWarLibs` parameter, we will need to document the new parameter, including which parameter option takes precedence (`--properties` vs. `useWarLibs`) since there will be two ways of configuring it.

Pandrex247 (commented):
https://github.com/payara/Payara/pull/7610#discussion_r2297700503

nucleus/deployment/admin/src/main/manpages/org/glassfish/deployment/admin/deploy.1
```diff
@@ -281,6 +282,11 @@ OPTIONS
            the newly redeployed application deserializes the data that was
            previously saved.
 
+       --warlibs
```
Pandrex247, line 285: Which takes precedence? `--warlibs=true` or `--properties="warlibs=false"`?

Pandrex247: Also last I heard is that we weren't going to add the new parameter?
https://github.com/payara/Payara/pull/7610#discussion_r2297700503

Pandrex247: If we're adding the new parameter we may as well remove the existing --properties configuration mechanism - I don't see why we should have two ways of configuring it

aubi: `--warlibs=true` takes precedence

aubi: "if we wish to retain the useWarLibs parameter, " - which we do. We didn't find another way within a reasonable time.

Pandrex247: Please also make a PR to update the documentation then.
https://github.com/payara/Payara-Documentation/blob/main-6/docs/modules/ROOT/pages/Technical Documentation/Payara Server Documentation/Command Reference/deploy.adoc

Pandrex247 (changes requested):
This breaks the existing property.

`--properties warlibs=true` will for all intent and purposes be ignored, even if you don't specify `--useWarLibs`. It gets set to true, and then gets overridden by the default value of `useWarLibs` (just because you haven't defined it on the command line doesn't mean it doesn't have a default value).

Please also raise a PR to update the documentation.
The new parameter needs documenting [here](https://github.com/payara/Payara-Documentation/blob/main-6/docs/modules/ROOT/pages/Technical%20Documentation/Payara%20Server%20Documentation/Command%20Reference/deploy.adoc)

NotedSalmon:
> This breaks the existing property.
> 
> `--properties warlibs=true` will for all intent and purposes be ignored, even if you don't specify `--useWarLibs`. It gets set to true, and then gets overridden by the default value of `useWarLibs` (just because you haven't defined it on the command line doesn't mean it doesn't have a default value).
> 
> Please also raise a PR to update the documentation. The new parameter needs documenting [here](https://github.com/payara/Payara-Documentation/blob/main-6/docs/modules/ROOT/pages/Technical%20Documentation/Payara%20Server%20Documentation/Command%20Reference/deploy.adoc)
> 
> Tested using one of the original reproducers: https://gist.github.com/lprimak/167ff6d0e82f51f42abde57f17758026

The only issue I have with this, which I am not sure how true it is, is that if I remove warlibs from properties, I am not sure how the `--warlibs` will work since whatever `--warlibs` is, it sets `--properties warlibs` in accordance to the value which will then turn some magical button on for warlibs. Just cautious that if I remove properties and then set the param warlibs to true, it wont activate the magical button.

My other concern is if existing apps use warlibs, would they stop functioning if they tried to be redeployed? I have then changed the Documentation: https://github.com/payara/Payara-Documentation/pull/643 which will be subject to changes depending on what answer I get.

I have tried outputting an error message of sorts to warn the user it is using a property value which will get overwritten but I am unsure of how practical it is.

Pandrex247 (approved):

PR #7634, opened 2025-09-08
FISH-11174 Get instance info in parallel & display timeouts as Unknown

jtarry-payara:
## Description

- Modified `ClusterCommandHelper.java` to use `ExecutorServiceFactory` for thread pool management when executing commands on cluster instances, factoring out thread pool creation
- Introduced `ExecutorServiceFactory.java` to encapsulate the creation of `ExecutorService` instances
- Modified `ListInstancesCommand.java` to retrieve instance information in parallel using an `ExecutorService`
- Modified `InstanceInfo.java` to handle `TimeoutException`, setting the state to `UNKNOWN` 
- Updated `Strings.properties` and `LocalStrings.properties` to add support for new `Unknown` state with corresponding image and text.

### New tests

See https://payara.atlassian.net/browse/FISH-12070

### Testing Performed

- Created deployment group with remote instances
- Slowdown network on remote instances so they don't respond within timeout period (default 10 secs)
- Observe timeout is only ever 10 secs and never 10 secs per slow/unreachable instance
- Observe instances that do timeout become Unknown state

### Testing Environment

Ubuntu 24.04 openjdk 21.0.8 2025-07-15

### DIFF
--- a/appserver/admingui/cluster/src/main/resources/dg/dgGeneral.jsf
+++ b/appserver/admingui/cluster/src/main/resources/dg/dgGeneral.jsf
@@ -38,7 +38,7 @@
     only if the new code is made subject to such option by the copyright
     holder.
 
-    Portions Copyright [2018] [Payara Foundation and/or its affiliates]
+    Portions Copyright [2018-2025] [Payara Foundation and/or its affiliates]
 
 -->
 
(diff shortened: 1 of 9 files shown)


Pandrex247 (commented):
Would need to properly sit down and read through this.
This is potentially affecting how _all_ clustered commands run, not just the instance status admin console page.

Also: if we're refactoring its usage of an executor service, it may be worth making it use the shared internal `PayaraExecutorService` since this was originally introduced to stop every internal service creating and managing their own.

aubi:
I tested it using firewall dropping the packets to some of the instances to simulate unresponsive instance.
I found 3 scenarios to display all instances -- none of them is working properly.
Details and video with testing is in Jira ticket.

jtarry-payara:
Have address all issues raised by @aubi thanks for sharing detailed video explaining how to reproduce the issues. The deployment groups adminui pages were all coded to check each instance's status within a foreach loop. I have fixed this so it gets the instance statuses for the deployment group, which makes more sense especially as you are on the deployment group page

PR #7662, opened 2025-09-25
FISH-12205 Add Cache TTL to Application Deployment Properties

### LINKED ISSUE #7663: Enhancement: No way to turn off file caching for redeployment
### Brief Summary

When developing, apps need to be reloaded quickly, however, this is not possible due to 5s hard-coded cache value in the web framework

### Expected Outcome

Add a `cacheTTL` to app deployment properties to be able to set it to zero for dev mode

### Current Outcome

Even though cache is a variable, there is no way to set it.

### Alternatives

None

### Context

Dev mode

lprimak:
## Description

Currently, there is no way to set cache TTL for serving files. This prevents dev mode to function properly

Fixes #7663

## Documentation
add `--properties cacheTTL=0` to deployment

### DIFF
--- a/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
+++ b/appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
@@ -200,6 +200,7 @@ public class WebModule extends PwcWebModule implements Context {
 
     private static final String WS_SERVLET_CONTEXT_LISTENER =
         "com.sun.xml.ws.transport.http.servlet.WSServletContextListener";
+    public static final String CACHE_TTL_APP_PROPERTY = "cacheTTL";
 
     // ----------------------------------------------------- Instance Variables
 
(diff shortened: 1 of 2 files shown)


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/WebModule.java
```diff
@@ -1552,6 +1561,8 @@ private void configureProperty(
             vs.setDefaultWebXmlLocation(value);
         } else if(name.startsWith("alternatedocroot_")) {
             parseAlternateDocBase(name, value);
+        } else if(name.startsWith(CACHE_TTL_APP_PROPERTY)) {
```
Pandrex247: One small adjustment to bring it more in line with the other options.

```suggestion
        } else if(CACHE_TTL_APP_PROPERTY.equalsIgnoreCase(name)) {
```

Pandrex247 (commented):
It would be great if you could also document the new property in the `create-virtual-server.1` man page. I don't have permissions to push to your branch so can't do it myself.

Something like the following, added after `errorReportValve`

```
           cacheTTL
               Specifies the time to live (TTL) in milliseconds of the 
               virtual server cache for serving files.
               
               Defaults to 5000
```

lprimak:
> new property in the create-virtual-server.1 man page

Done, thanks!

PR #7686, opened 2025-10-05
FISH-8542 prime faces autocomplete has issues with the select event

luiseufrasio:


Pandrex247 (changes requested):
Why p2? What happened to p1?

Also I'm still not a fan of the fix you've done - I have doubts it'll be accepted by upstream Expressly, and it seems like the kind of change which should be handled by the WebappClassLoader

Pandrex247 (changes requested):
You're also missing the GitHub tag and release commits over in our patched project - the maintenance branch still lists the version as p2-snapshot.

You should be releasing 5.0.0.payara-p1

Pandrex247:
I've deleted the erroneous p2

PR #7698, opened 2025-10-10
FISH-10721 Add ** role for authenticated users

kalinchan:
## Description

Resolves java ee 7 sample test 

https://github.com/payara/patched-src-javaee7-samples/pull/105
### Testing Performed

https://jenkins.payara.fish/view/TCKs/job/JakartaEE-11-TCK/336/


Pandrex247 (commented):
Trying to run this through some TCKs but hitting issues with Jenkins...

Also trying to walk through P6 to compare against this to see how the code flow has changed (this class is new) and why we now need to explicitly add this here.

In other words: I'm being extra wary about granting permissions.

PR #7751, opened 2025-11-04
FISH-12541 Create SECURITY.md for security policy

Dedal0:
Added a security policy document outlining supported versions and vulnerability reporting, so vulnerabilities can be reported.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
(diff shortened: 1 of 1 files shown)


MeroRai:
@Dedal0, thank you for your report. You can find our current policies on how Payara handles security issues here: https://docs.payara.fish/community/docs/Security/Overview.html. As outlined there, the correct process is to email any identified flaws or vulnerabilities to security@payara.fish. You’ve already done that, and our team is currently reviewing your submission.

PR #7780, opened 2025-11-19
FISH-12431: upgrading version of trilead ssh2

breakponchito:
Changing eddsa dependency to use latest changes
## Description

This is a fix to upgrade a use latest changes for eddsa dependency
### Testing Performed

Manual testing with the following scope:

- Creating remote ssh node:
Start server and create remote ssh node:
image
-Ping the remote node
image
- Create remote instance
image
image
- Start remote instance 
image
- Stop remote instance
image
- Delete instance
image
- Delete ssh node
image

### Testing Environment

Window 11 Home edition, azul jdk 11, maven 3.9.11 for DAS host, Payara 6
OpenSuse Leap 15.6 on VM instance using Virtual Box 7.2.4, open jdk 11, Payara 6

### DIFF
--- a/nucleus/packager/external/trilead-ssh2/pom.xml
+++ b/nucleus/packager/external/trilead-ssh2/pom.xml
@@ -126,19 +126,18 @@
       		</exclusions>
         </dependency>
         <!-- Import seperately to avoid http block in Maven 3.8.1 -->
+        <dependency>
+            <groupId>com.bloxbean.cardano</groupId>
+            <artifactId>net-i2p-crypto-eddsa</artifactId>
+            <scope>compile</scope>
+            <optional>true</optional>
+            <version>0.3.1</version>
+        </dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


nucleus/packager/external/trilead-ssh2/pom.xml
```diff
@@ -126,19 +126,18 @@
...
 		    <version>1.0.2</version>
 		</dependency>
-
-        <dependency>
-            <groupId>net.i2p.crypto</groupId>
-            <artifactId>eddsa</artifactId>
-            <version>0.3.0</version>
-            <scope>compile</scope>
-            <optional>true</optional>
-        </dependency>
```
Pandrex247, line 141: The reordering of this makes it look like you're importing the new eddsa dependency to avoid the Maven issue rather than jbcrypt - you need to move the comment or leave the ordering as if

PR #7795, opened 2025-11-27
FISH-7667 PartItem memory leak

Viii3:
## Description

- Fixes FISH-7667
  - Rewrites the `PartItem` garbage collection runnable to prevent it from maintaining a reference to the object.

### Testing Performed

Tested using the reproducer provided in the ticket. Allowed the test method to iterate 200 times and observed no notable increases in the number of `PartItem` objects held in memory, the value stayed at a consistent 4.

### Testing Environment

Maven version: 3.9.6
Java version: 21.0.3, vendor: Eclipse Adoptium

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
@@ -812,6 +812,7 @@ class PartItem
     }
     
     private static class Cleaner implements Runnable {
+        private final static Logger logger = LogFacade.getLogger();
         private DeferredFileOutputStream fileStream;
 
         private void setFileStream(DeferredFileOutputStream fileStream) {
@@ -826,10 +827,11 @@ class PartItem
             
             File file = fileStream.getFile();
             if (file != null && file.exists()) {
-                if (!file.delete() && log.isLoggable(Level.FINE)) {
-                    log.log(Level.FINE, "Cannot delete file: " + fileStream);
+                if (!file.delete() && logger.isLoggable(Level.FINE)) {
+                    logger.log(Level.FINE, "Cannot delete file: " + fileStream);
                 }
             }
+            fileStream = null;
         }
     }
 }


appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
```diff
@@ -813,4 +810,26 @@ private void deleteFile(File file) {
             log.log(Level.FINE, "Cannot delete file: " + file);
         }
     }
+    
+    private static class Cleaner implements Runnable {
+        private DeferredFileOutputStream fileStream;
+
+        private void setFileStream(DeferredFileOutputStream fileStream) {
```
svendiedrichsen, line 818: Why using the stream and not the file directly? Seems only the file is of interest here while the stream may buffer some considerable amount of memory.

appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
```diff
@@ -813,4 +810,26 @@ private void deleteFile(File file) {
...
+                return;
+            }
+            
+            File file = fileStream.getFile();
+            if (file != null && file.exists()) {
+                if (!file.delete() && log.isLoggable(Level.FINE)) {
+                    log.log(Level.FINE, "Cannot delete file: " + fileStream);
+                }
+            }
+        }
```
svendiedrichsen, line 835: Additionally you might want to NULL the reference after processing thus only the empty Runnable instance might be referenced for longer.

appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
```diff
@@ -813,4 +810,26 @@ private void deleteFile(File file) {
...
+        @Override
+        public void run() {
+            if (fileStream == null) {
+                return;
+            }
+            
+            File file = fileStream.getFile();
+            if (file != null && file.exists()) {
+                if (!file.delete() && log.isLoggable(Level.FINE)) {
+                    log.log(Level.FINE, "Cannot delete file: " + fileStream);
```
svendiedrichsen: Referencing the outer classes static Logger in the inner class might prevent outer class unloading?

appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
svendiedrichsen: At least if you keep using the stream please consider closing it. It will be created by the PartItem and won't probably be closed by anyone else.

Pandrex247: An issue has been created for us to follow-up on this: FISH-12609
We'll look at working on it for the January release

PR #7808, opened 2025-12-08
bugfix: PostBoot commands now run after server ready event, not durin…

### LINKED ISSUE #7806: Bug Report: Warlibs are not honored from --postbootcommandfile script
### Brief Summary

Warlibs are not taken into account when using --postbootcommandfile script is used to deploy applications

### Expected Outcome

Self explanatory

### Current Outcome

Self explanatory

### Context
Auto-deploy and Auto-reload are not considered secure. In docker containers,
a more secure and performant way is to deploy applications via PostBoot command file.
Currently, Warlibs are loaded after PostBoot command processing, which is incorrect.

### Reproducer

TBD

### Operating System

Any

### JDK Version

7.2025.1

### Payara Distribution

Payara Server Full Profile

lprimak:
…g service construction.
fixes #7806

This is due to the fact that previously, deploy would fail due to server not being ready for deployment fixes #7806

### Testing Performed
Tested CDI app in postboot

### DIFF
--- a/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/bootstrap/BootCommandService.java
+++ b/nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/bootstrap/BootCommandService.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) [2016-2022] Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) [2016-2025] Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development


nucleus/core/kernel/src/main/java/com/sun/enterprise/v3/bootstrap/BootCommandService.java
```diff
@@ -86,8 +92,15 @@ public void doBootCommands(String file, boolean expandValues) {
...
+        if (event.is(EventTypes.SERVER_READY)) {
+            doBootCommands(startupContext.getArguments().getProperty("-postbootcommandfile"), true);
+        }
+    }
+
     @Override
     public void postConstruct() {
-        doBootCommands(startupContext.getArguments().getProperty("-postbootcommandfile"), true);
+        events.register(this);
     }
```
Pandrex247, line 105: See my other comment regarding sequencing.
This will likely reintroduce race conditions and sequencing issues: the postboot config file would be executed after all but new applications in the deploy dir have been deployed. Tying this service to an event rather than a specific run level will also mean that it races against all other services reacting to that event.

PR #7816, opened 2025-12-09
FISH-12613 Encode

luiseufrasio:
## Description

For details, see: https://payara.atlassian.net/browse/FISH-12613

## Testing Performed

Tested encoding works correctly as described in the Jira ticket.


Pandrex247 (approved):
Appears to work on my machine

PR #7830, opened 2025-12-21
FISH-12719 fixed another case of leaks when fails to deploy EjbApplication

lprimak:
## Description
Follow-up to #7763
Any type of deployment (war, jar, ear) is susceptible to more leaks when EJBs are not able to be deployed.
This PR fixes newly-discovered instances of this.


Pandrex247 (commented):
Do you have a reproducer for this? The changes to EjbApplication and EjbDeployer look a bit suspect 🤔

Pandrex247 (approved):

PR #7836, opened 2026-01-02
FISH-12609 Close file stream

kalinchan:
## Description

Title


appserver/web/web-core/src/main/java/org/apache/catalina/fileupload/PartItem.java
```diff
@@ -824,14 +824,25 @@ public void run() {
...
-            File file = fileStream.getFile();
-            if (file != null && file.exists()) {
-                if (!file.delete() && logger.isLoggable(Level.FINE)) {
-                    logger.log(Level.FINE, "Cannot delete file: " + fileStream);
+
+            try {
+                File file = fileStream.getFile();
+                if (file != null && file.exists()) {
+                    if (!file.delete() && logger.isLoggable(Level.FINE)) {
+                        logger.log(Level.FINE, "Cannot delete file: " + fileStream);
```
svendiedrichsen: Shouldn't this log the file instead?

PR #7881, opened 2026-01-14
FISH-10915 Remove ClassGraph

Pandrex247:
## Description
ClassGraph was introduced to aid with our Jakarta Data implementation, but given that we already do class scanning via HK2 it always felt somewhat redundant. To not halt development to conduct an investigation we pulled it in.

This PR iterates on this original implementation and removes the ClassGraph dependency to tidy things up, as after some experimentation: no we don't need it.
This is achieved by using the HK2 class canning which will already have been conducted and stored as deployment metadata to store the names of the annotated classes in the CDI extension, in a similar way as to how we store the application name.

### Blockers
None

### New tests
None

### Testing Performed
Ran against Data sample in Payara Samples
Ran against Data TCK

Would like to do some leak testing before I make this PR "final" - local poking seemed fine

### Testing Environment
Windows 11, Maven 3.9.11, Zulu JDK 21.0.9

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/appserver/data/data-core/pom.xml
+++ b/appserver/data/data-core/pom.xml
@@ -2,7 +2,7 @@
 <!--
    DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
 
-   Copyright (c) 2025 Payara Foundation and/or its affiliates. All rights reserved.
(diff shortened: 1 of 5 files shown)


breakponchito (approved):
LGTM

PR #7890, opened 2026-01-20
FISH-11726 reactivate and fix client-certificate-validator test payara7

astoica1:
## Description
This PR reactivates and fixes the client-certificate-validator module that was previously disabled.

### New tests
New test: ClientValidationTest

### Testing Performed
JDK21 , maven 3.9.9, macOS, against the latest Payara7 head.

mvn -V -ff clean install --strict-checksums -Ppayara-server-remote,jakarta-staging -Djavax.net.ssl.trustStore=<JAVA_PATH>/zulu-21.jdk/Contents/Home/lib/security/cacerts -Djavax.xml.accessExternalSchema=all -Dpayara.home="appserver/distributions/payara/target/stage/payara7" -f appserver/tests/payara-samples

### DIFF
--- a/appserver/tests/payara-samples/microprofile-endpoints/insecure/pom.xml
+++ b/appserver/tests/payara-samples/microprofile-endpoints/insecure/pom.xml
@@ -194,7 +194,7 @@
 
                             <!-- Restart server after undoing config -->
                             <execution>
-                                <id>stop-or-restart-payara-after-config</id>
+                                <id>stop-or-restart-payara-after-unapplying-config</id>
                                 <phase>post-integration-test</phase>
                                 <goals>
                                     <goal>exec</goal>
--- a/appserver/tests/payara-samples/microprofile-endpoints/secure/pom.xml
+++ b/appserver/tests/payara-samples/microprofile-endpoints/secure/pom.xml
@@ -232,7 +232,7 @@
 
                             <!-- Restart server after undoing config -->
                             <execution>
-                                <id>stop-or-restart-payara-after-config</id>
+                                <id>stop-or-restart-payara-after-unapplying-config</id>
                                 <phase>post-integration-test</phase>
                                 <goals>
                                     <goal>exec</goal>


appserver/tests/payara-samples/samples/client-certificate-validator/src/test/java/fish/payara/samples/security/validation/ClientValidationTest.java
```diff
@@ -48,39 +48,49 @@
...
+import static org.junit.Assert.fail;
 
-@Ignore
 @RunWith(PayaraArquillianTestRunner.class)
 @SincePayara("5.2021.8")
 public class ClientValidationTest {
 
     private static String certPath;
-
     private static final String CERTIFICATE_ALIAS = "omnikey";
```
Pandrex247: This shares an alias name with the multiple key stores test - might be worth making it different

appserver/tests/payara-samples/samples/client-certificate-validator/src/test/java/fish/payara/samples/security/validation/ClientValidationTest.java
```diff
@@ -98,41 +108,227 @@ public static WebArchive deploy() {
 
     @Test
     @InSequence(1)
-    public void generateCertsInTrustStore() throws IOException {
-        if (ServerOperations.isServer()) {
-            certPath = ServerOperations.generateClientKeyStore(true, true, CERTIFICATE_ALIAS);
+    public void generateCertsInTrustStore() throws Exception {
+        logger.info("Entering generateCertsInTrustStore");
+        // Generate the keystore to ensure certPath is set
+        certPath = ServerOperations.generateClientKeyStore(true, true, CERTIFICATE_ALIAS);
```
Pandrex247: I'm still a hater of this method.
I feel we should be using key tool for all certificate generation, additions, and removals.

Otherwise further down the road this is likely going to need to be rewritten from scratch when we start making the samples work with payara micro

Pandrex247 (commented):
The test only seems to work on its own: when run as a part of the full suite it fails.

appserver/tests/payara-samples/samples/client-certificate-validator/src/test/java/fish/payara/samples/security/validation/ClientValidationTest.java
```diff
@@ -98,41 +108,227 @@ public static WebArchive deploy() {
...
+            HttpsURLConnection.setDefaultSSLSocketFactory(sslContext.getSocketFactory());
+
+            // For testing, accept all hostnames
+            HttpsURLConnection.setDefaultHostnameVerifier((hostname, session) -> true);
+
+        } catch (Exception e) {
+            logger.log(Level.SEVERE, "Failed to configure SSL context: " + e.getMessage(), e);
+            throw e;
         }
     }
```
Pandrex247: Am I mad or this entire test not actually testing anything? Isn't this just a glorified `@Before`?

The "test" is to generate a certificate and set up a TLS context, but it never actually _calls_ the endpoint?

PR #7912, opened 2026-01-27
FISH-12906 Fixing com.sun.jdo.spi.persistence.support.sqlstore.connection.ConnectionManager memory leak

svendiedrichsen:
## Description
using internal static class to wrap ConnectionManager state to allow shutdown if ConnectionManager instance is GC'd

Briefly mentioned in and aiming to fix for #7849

### Testing Performed
PR pipeline

## Notes for Reviewers
see description

### DIFF
--- a/appserver/persistence/cmp/support-sqlstore/src/main/java/com/sun/jdo/spi/persistence/support/sqlstore/connection/ConnectionManager.java
+++ b/appserver/persistence/cmp/support-sqlstore/src/main/java/com/sun/jdo/spi/persistence/support/sqlstore/connection/ConnectionManager.java
@@ -56,7 +56,6 @@ import com.sun.jdo.spi.persistence.support.sqlstore.LogHelperSQLStore;
 import org.glassfish.hk2.utilities.CleanerFactory;
 
 
-import java.lang.ref.Cleaner;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


appserver/persistence/cmp/support-sqlstore/src/main/java/com/sun/jdo/spi/persistence/support/sqlstore/connection/ConnectionManager.java
```diff
@@ -1477,49 +1552,7 @@ public void startUp() throws ClassNotFoundException, SQLException {
...
-     */
-    public final void registerCloseEvent() {
-        CleanerFactory.create().register(this, () -> {
-            try {
-                shutDown();
-            } catch (SQLException se) {
-                // Ignore it.
-            }
-        });
+        this.cleanable.clean();
```
luiseufrasio: The original `shutDown()` threw `SQLException` from `conn.close()` to the caller. The new implementation delegates to `cleanable.clean()` which calls `CleanableConnectionManagerState.run()`, but `run()` catches and silently ignores the `SQLException`. This changes the contract of `shutDown()` — callers that previously handled `SQLException` will no longer receive it. `The run()` method (Cleaner path) should silently ignore exceptions, but `shutDown()` (explicit path) should propagate them. Consider splitting the logic: shutDown() calls the cleanup directly and throws, while run() calls the same logic but catches exceptions.

svendiedrichsen: Done as requested.

PR #7913, opened 2026-01-29
FISH-13732 Fix Issue 7197: Invalid No X509Certificate found WARNING spam from CertificateRealm

### LINKED ISSUE #7197: Bug Report: Invalid No X509Certificate found WARNING spam from CertificateRealm
### Brief Summary

Pull request https://github.com/payara/Payara/pull/5398 added a warning "No X509Certificate found" to the CertificateRealm which is spamming our logs every time a request is made.

The warning comes from this line:
https://github.com/payara/Payara/blame/97cfdaff4278107579041b228351f48bafba8b61/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/certificate/CertificateRealm.java#L194

It appears the code is trying to find the subject for the certificate with this getCertificateFromSubject method https://github.com/payara/Payara/blob/7e73ca597af46ed99827371d1c9def64aea826c6/nucleus/security/core/src/main/java/com/sun/enterprise/security/auth/realm/certificate/CertificateRealm.java#L234C4-L251C6

```java
private X509Certificate getCertificateFromSubject(Subject subject, X500Principal principal) {
        X509Certificate result = null;
        Set<Object> publicCredentials = subject.getPublicCredentials();
        for (Object publicCredential : publicCredentials) {
            if (publicCredential instanceof List) {
                List<Object> data = (List<Object>) publicCredential;
                for (Object item : data) {
                    if (item instanceof X509Certificate) {
                        X509Certificate certificate = (X509Certificate) item;
                        if (principal.equals(certificate.getIssuerX500Principal())) {
                            result = (X509Certificate) item;
                        }
                    }
                }
            }
        }
        return result;
    }
```

but the logic appears to be looking for a certificate where the principal equals the issuer (`if (principal.equals(certificate.getIssuerX500Principal())) {`) which I believe will only occur for self signed certificates.

Also even if the code was correct, the warning gives no context for the problem, where in the code the warning is coming from (none of filename, class, function, or line number is logged with the warning with the default payara configuration), nor how to fix it so it is not a useful warning.

We are using payara 6.2025.1

### Expected Outcome

No warning in the server.log about "No X509Certificate found"

### Current Outcome

Warning "No X509Certificate found" in the server.log 

### Reproducer

Setup a CertificateRealm as the login module (login.conf) and receive a request from a certificate that is not self signed.

### Operating System

RHEL 8.10

### JDK Version

openjdk JDK 21.04.

### Payara Distribution

Payara Server Full Profile

dx-coding:
## Description
Fixed a bug in `CertificateRealm.getCertificateFromSubject()` where the method was incorrectly matching certificates against the `IssuerX500Principal` instead of the `SubjectX500Principal`.

According to the method name and its purpose, it should match against the certificate's subject (the entity the certificate is issued to), not the issuer (the CA that signed the certificate).

### Blockers
N/A

### New tests

Created new test `nucleus/security/core/src/test/java/com/sun/enterprise/security/auth/realm/certificate/CertificateRealmTest.java`

### Testing Performed

Ran `mvn test -Dtest=CertificateRealmTest -pl nucleus/security/core` successfully
Verified the fix correctly identifies certificates by their subject rather than issuer

### Testing Environment
Ubuntu 22.04, Maven 3.6.3, JDK 21.0.7

## Documentation
Closes #7197

## Notes for Reviewers
N/A


MeroRai:
@Pandrex247 CLA has been recieved.

PR #7927, opened 2026-02-04
FISH-12815: adding module to register server runtime telemetry

breakponchito:
Adding module to configure MicroProfile OpenTelemetry for server runtime config
## Description

This is a feature to register OpenTelemetry at server runtime level reading properties from domain.xml and from environment variables.

### Testing Performed

Manual testing starting and running the server in debug mode to see if the new service starts before the LogManagerService and reads the properties to start OpenTelemetry SDK instance. We can set the property in the following way to start the server telemetry instance for runtime:

- Set the following JVM property for the server on the domain.xml:
```
<jvm-options>-Dotel.sdk.disabled=false</jvm-options>
```

- set environment variable: for windows I used powershell
```
$env:OTEL_SDK_DISABLED = "false"
```

also we can set other properties related to opentelemetry like the following:

```
<jvm-options>-Dotel.java.global-autoconfigure.enabled=true</jvm-options>
```

on debug mode we can see that the opentelemetry runtime instance is created. Also we can see from logs that the exporter starts to send information:

image

if we provide an instance of a reader like Jaege then we will receive some information:

image

image

image

### Testing Environment

Windows 11, Azul jdk 21, maven 3.9.11

### DIFF
--- a/nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
+++ b/nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
@@ -67,9 +67,6 @@ public class PayaraTelemetryBootstrapFactoryServiceImpl implements PayaraTelemet
     @PostConstruct
     public void init() {
         runtimeSdk = createTelemetryRuntimeInstance();
-        if (GlobalOpenTelemetry.get() == null) {
-            GlobalOpenTelemetry.set(runtimeSdk);
-        }
     }
 
     @Override


nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        Map<String, String> props = new HashMap<>();
+        Map<String, String> environmentOtelPropsAvailable = System.getenv().entrySet().stream()
+                .filter(e -> e.getKey().startsWith(OTEL_PROPERTIES_PREFIX))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        Map<String, String> systemOtelPropsAvailable = System.getProperties().entrySet().stream()
+                .filter(e -> ((String) e.getKey()).startsWith(OTEL_PROPERTIES_PREFIX))
+                .map(e -> Map.entry((String) e.getKey(), (String) e.getValue()))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        props.putAll(environmentOtelPropsAvailable);
+        props.putAll(systemOtelPropsAvailable);
```
pdudits: I believe the precedence should be opposite. And of course the naming amongst environment and system properties are different, so they need to be mapped. You'll need to replicate quite lots of behavior from MP config ConfigSource implementation for this, which is why I suggested to create server-level MP Config provider as well.

nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    
+    private OpenTelemetry runtimeSdk = null;
+    
+    private static final String OTEL_PROPERTIES_PREFIX = "otel";
+    private static final String OTEL_SYSTEM_PROPERTY_NAME = "otel.sdk.disabled";
+    private static final String OTEL_ENVIRONMENT_PROPERTY_NAME = "OTEL_SDK_DISABLED";
+
+    @PostConstruct
+    public void init() {
+        runtimeSdk = createTelemetryRuntimeInstance();
```
pdudits, line 69: You will need to register this as global telemetry instance as well. But because GlobalTelemetry is a singleton that's mutable only once, to support application-level SDKs you need to have an api for current `OpenTelemetryService` to register itself as a router to application-configured instances and then delegate to it once it registers.

(This will be fun to do in efficient manner, I think it is safe to assume that application-level opentelemetry will be used less often than server-level opentelemetry).

breakponchito: for the API now when you call the method getOpenTelemetrySdk automatically your are setting the sdk as global, we don't need to explicitly call the **GlobalOpenTelemetry.set()** method

nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
breakponchito: this will be attended on a continuation ticket

nucleus/core/telemetry-bootstrap/src/main/java/fish/payara/telemetry/service/PayaraTelemetryBootstrapFactoryServiceImpl.java
breakponchito: the implementation of the router mechanism will be implemented on a continuation ticket

breakponchito:
For now the pipeline is failing because of affectation on other parts of the telemetry implementation. Those issues will be fixed on following tickets

PR #7928, opened 2026-02-04
FISH-7515 Concurrency issues causing HTTP 403

stian-sigvartsen:
## Description
The issue is caused by some design problems:
1) A static "default" SecurityContext is shared between threads
2) ServiceContext is not immutable as appears was the original intention before 2011
3) The state is mutated and nullified at cleanup (affecting all threads)

### Testing Performed
Used the jaspic-bug.war load (attached to ticket) to make repeat authenticated requests and report on any HTTP 403 responses, whilst at the same time producing additional load with jmeter and unauthenticated requests. All requests processed successfully.

For detailed steps on setting this up, see the README.md file in the war's source code .zip .

### Testing Environment
OSX 15.1, Zulu JDK 21.0.8+9-LTS, Maven 3.9.9

### DIFF
--- a/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
+++ b/nucleus/security/core/src/main/java/com/sun/enterprise/security/SecurityContext.java
@@ -84,8 +84,6 @@ public class SecurityContext extends AbstractSecurityContext {
     // Did the client log in as or did the server generate the context
     private boolean serverGeneratedCredentials;
 
-    private Principal sessionPrincipal;
-    
     // ### Static public methods
     
     
@@ -306,14 +304,6 @@ public class SecurityContext extends AbstractSecurityContext {
     @Override
     public Subject getSubject() {
         return subject;
-    }
-
-    public Principal getSessionPrincipal() {
-        return sessionPrincipal;
-    }
-
-    public void setSessionPrincipal(Principal sessionPrincipal) {
-        this.sessionPrincipal = sessionPrincipal;
     }
 
     public Set<Principal> getPrincipalSet() {


stian-sigvartsen:
Passed TCKs:
- authentication 
- authorization 
- security 
- servlet 
- rest-platform

appserver/security/webintegration/src/main/java/com/sun/web/security/RealmAdapter.java
```diff
@@ -813,13 +810,19 @@ private Caller getCaller(Subject subject) {
...
+    private Subject reuseSessionSubject(Caller caller, RequestFacade requestFacade) {
+        // Get the session's WebPrincipal for enriching the authenticated principal upon Soteria callback
+        Principal callerPrincipal = caller.getCallerPrincipal();
+        Principal requestPrincipal = requestFacade.getRequestPrincipal();
 
-        if (returnedPrincipal instanceof WebPrincipal) {
-            return reuseWebPrincipal((WebPrincipal) returnedPrincipal);
-        }
+        // Check if the top level session principal is indeed wrapping our current principal
+        if (callerPrincipal != null
```
breakponchito, line 819: We don't need the null comparation since instanceof operator is doing that for us, if the case the reference is null the instanceof will return false
```suggestion
        if (!(callerPrincipal instanceof WebPrincipal)
                && requestPrincipal instanceof WebPrincipal webPrincipalFromSession
                && webPrincipalFromSession.getCustomPrincipal() == callerPrincipal) {
```

stian-sigvartsen: I think it is needed because the `instanceof` is negated?

stian-sigvartsen:
Thanks @breakponchito I completely forgot that I upgraded the app to EE 11 when I worked on the main branch! Please find it attached to the ticket.
For steps how to test it, see the README.md in the app .zip.

appserver/security/webintegration/src/main/java/com/sun/web/security/RealmAdapter.java
breakponchito: ok yes if we don't want to evaluate the other conditions in this case make sense to left the null comparation first, in case we don't at most two condition will need to be evaluated

stian-sigvartsen: The same conditional logic existed before my refactor, but I believe it is there to protect against the other condition returning true when callerPrincipal is `null`:

`webPrincipalFromSession.getCustomPrincipal() == callerPrincipal`

i.e. protecting against the case when `webPrincipalFromSession.getCustomPrincipal()` returns null which means the `WebPrincipal` is not previously associated with a caller principal during the previous username + password authentication.

stian-sigvartsen:
Hey @breakponchito . I have just now clean built the PR and I've been running the JaspicBug requests + jmeter for over 30 mins without any 403 response. Maybe try a clean build?

PR #7929, opened 2026-02-04
FISH-12784 update faces dependabot rule

abdulrahim458:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -47,10 +47,10 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: "org.glassfish.mq:mq-distribution"
         update-types: ["version-update:semver-major"]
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
Pandrex247: Won't this fully exclude this dependency?
We don't want to exclude dependabot from upgrading the "[real](https://github.com/payara/Payara/blob/main/pom.xml#L267)" dependency - we only want it to exclude this [specific test dependency](https://github.com/payara/Payara/blob/main/appserver/tests/payara-samples/samples/use-bundled-jsf-primefaces/pom.xml#L74-L80) which is intentionally configured to be an older version.

Pandrex247, line 51: We don't want to remove this rule

Pandrex247 (approved):

PR #7949, opened 2026-02-13
FISH-12926 FISH-13013 Disable Jenkins Testing for Pending MicroProfile TCKs

Viii3:
## Description

- Resolves FISH-12926
  - Comments out various MicroProfile TCKs to disable the Jenkins testing.
- Resolves FISH-13013
  - Comments out failing Samples test modules to disable Jenkins testing.


breakponchito:
@Viii3 It seems that the pipeline is still failing but we can remove some samples tests that use traces on a different ticket

PR #7952, opened 2026-02-13
FISH-12816: adding properties and setting Runtime SDK instance as global

breakponchito:
Seetting default properties for Telemetry Runtime SDK instance and setting as global
## Description

This is a feature to customize the Telemetry SDK instance in case the otel property to enable is available at startup time
### Testing Performed

Manual testing: starting server, checking in debug mode the creation of Telemetry SDK instance and reviewing how the configuration as global instance work on the OpenTelemtryService when asking for the Global instance
### Testing Environment

Windows 11, azul jdk 21, maven 3.9.11

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -152,43 +152,43 @@ pipeline {
                         }
                     }
                 }
-                stage('MicroProfile Fault Tolerance TCK') {
-                    agent {
-                        label 'general-purpose'
-                    }
-                    options {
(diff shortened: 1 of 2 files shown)


breakponchito:
For now it is expected to fail because some of the changes are affecting Open Trancing and Open Telemetry implementation

luiseufrasio (approved):
LGTM!

PR #7962, opened 2026-02-20
FISH-12951 Remove leading slash from MP JWT properties

kalinchan:
## Description

Removing the leading slash allows for the resolution of the properties file in external libraries - previously a zip entry with the exact name was required if a user wanted to load these properties.
### Blockers

n/a
### New tests

n/a
### Testing Performed

Manual testing of external libraries (domain/lib)
## Documentation

n/a
## Notes for Reviewers

n/a


appserver/payara-appserver-modules/microprofile/jwt-auth/src/main/java/fish/payara/microprofile/jwtauth/eesecurity/SignedJWTIdentityStore.java
svendiedrichsen: How about using try-with-resources for the opened stream to have it closed in the end?

PR #7969, opened 2026-02-24
FISH-12253: use Build job instead of building Payara locally (Payara7)

astoica1:


Jenkinsfile
```diff
@@ -664,13 +632,89 @@ void stopDomain() {
...
+     selector: specific("${buildId}"),
+     filter: artifactFilter,
+     target: 'artifacts/')
+
+    // If Maven repository is included, restore it
+    if (fileExists('artifacts/maven-repository.zip')) {
+        echo "Restoring Maven repository"
+        sh 'unzip -o artifacts/maven-repository.zip -d ~/.m2/repository'
+        echo "Maven repository restored to ~/.m2/repository"
+    }
```
Pandrex247, line 665: I don't believe we should be doing this for any test stage other than Payara Samples and Payara Functional

This is intentionally omitted to help ensure that we aren't "leaking" dependencies - our test repos shouldn't have dependencies on internal build artefacts.

Jenkinsfile
```diff
@@ -3,14 +3,20 @@
...
 def payaraBuildNumber
+def buildId
+
 pipeline {
     agent {
         label 'general-purpose'
     }
+    options {
+        buildDiscarder(logRotator(numToKeepStr: '30', daysToKeepStr: '14'))
+    }
```
Pandrex247, line 14: I'm not sure this is actually doing what we think?

We wanted to add this to the Build job, not here to this one.
I don't see why we would need to keep 30 builds of a single PR - I would change this down to only keeping 5 builds (the 14 days can stay).

Jenkinsfile
```diff
@@ -664,13 +632,92 @@ void stopDomain() {
...
+    sh installArtifactsScript
+
+    echo "Updating pom.xml payara.version property for Shrinkwrap resolver"
+    def updatePayaraVersionScript = """
+        sed -i "s/payara\\.version>.*<\\/payara\\.version>/payara\\.version>${payaraVersion}<\\/payara\\.version>/g" pom.xml
+    """
+    sh updatePayaraVersionScript
+
+    archiveArtifacts artifacts: 'artifacts/payara.zip', fingerprint: true
+    archiveArtifacts artifacts: 'artifacts/payara-micro.jar', fingerprint: true
```
Pandrex247: I don't think we need this?
It's just a duplicate of the Build job

```suggestion
```

Jenkinsfile
```diff
@@ -622,32 +600,22 @@ pipeline {
...
 void setupDomain() {
-    echo '*#*#*#*#*#*#*#*#*#*#*#*#  Unstash distributions  *#*#*#*#*#*#*#*#*#*#*#*#*#*#*#'
-    unstash name: 'payara-target'
-    echo '*#*#*#*#*#*#*#*#*#*#*#*#  Unstash maven repository  *#*#*#*#*#*#*#*#*#*#*#*#*#*#*#'
-    dir('/home/ubuntu/.m2/repository/'){
-        unstash name: 'payara-m2-repository'
-    }
+    echo '*#*#*#*#*#*#*#*#*#*#*#*#  Setting up domain  *#*#*#*#*#*#*#*#*#*#*#*#*#*#*#'
+    sh 'ls -la'
+    sh 'ls -la payara7'
```
Pandrex247: Is this necessary?
Seems like a debug statement?

Jenkinsfile
Pandrex247: PR raised to add limits the Build job here: https://github.com/payara/EngineeringJenkinsjobs/pull/433

PR #7972, opened 2026-02-25
FISH-12986 Eclipselink 5.0.0-B13

Viii3:
## Description

- Resolves FISH-12986
  - Updates the version in the pom to the latest patch.


Pandrex247:
By-and-large looks good: just kicking off a run against the Data TCK.
I noticed we have a b11 patched branch which we're not using for some reason: the reason turned out to be the Data TCK gets broken by it: https://payara.atlassian.net/browse/FISH-12468

PR #7978, opened 2026-02-26
FISH-12774: Validate property dataStore from @Repository to resolve persistence unit name

luiseufrasio:
## Description
Implements support for the `@Repository(dataStore)` attribute from Jakarta Data 1.0 to resolve which persistence unit to use when multiple are defined in `persistence.xml`.

Previously, Payara used a custom property `fish.payara.data.usePU=true` in `persistence.xml` to flag which persistence unit Jakarta Data should use. This was limited to marking only one PU globally. This PR replaces that mechanism with the standard Jakarta Data approach: `@Repository(dataStore = "myPU")`, where the `dataStore` value is matched against the persistence unit name stored in a name map populated by `JPADeployer` during deployment.

**Changes:**
- `JPADeployer`: Stores a `Map<String, EntityManagerFactory>` (PU name to EMF) alongside the existing EMF list in transient app metadata, enabling reliable PU name lookup
- `DataCommonOperationUtility`: Removed `fish.payara.data.usePU` property logic; added `getEntityManager(applicationName, dataStore)` that looks up the EMF by PU name from the name map
- `RepositoryImpl`: Propagates `dataStore` to all `getEntityManager()` calls
- `DynamicInterfaceDataProducer`: Extracts `dataStore` from `@Repository` annotation and passes it through

**Backward compatibility:** When only one persistence unit exists, it is used directly regardless of the `dataStore` value (existing behavior preserved).

### Blockers
None

### New tests
- `MultipleDataStoreTest.java`: Arquillian test that deploys a WAR with two persistence units (`firstPU` and `secondPU`) and verifies:
  - Both repositories (`@Repository(dataStore = "firstPU")` and `@Repository(dataStore = "secondPU")`) are injected correctly
  - CRUD operations (save, find, delete) work through each PU-specific repository
  - Query-by-method-name operations work on both repositories simultaneously
  - Count and existence checks work through specific PU repositories

### Testing Performed
- Verified compilation of `data-core` and `jpa-container` modules
- All new and existing test classes pass successfully on Jenkins

### Testing Environment
Windows 11, JDK 21, Maven 3.9.9

## Documentation
N/A

## Notes for Reviewers
- `JPADeployer.iterateInitializedPUsAtApplicationPrepare()` populates the EMF name map using `pud.getName()`
- `DataCommonOperationUtility.getEntityManager(String, String)` contains the core resolution logic using the name map
- The initial approach of reading `jakarta.persistence.unitName` from `EntityManagerFactory.getProperties()` did not work because EclipseLink does not automatically expose the PU name in that map — the PU name is an XML descriptor attribute, not a `<property>` entry

### DIFF
--- a/appserver/tests/payara-samples/samples/data/src/main/java/fish/payara/samples/data/DataSourceConfig.java
+++ b/appserver/tests/payara-samples/samples/data/src/main/java/fish/payara/samples/data/DataSourceConfig.java
@@ -1,0 +1,52 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ *    Copyright (c) [2026] Payara Foundation and/or its affiliates. All rights reserved.
+ *
+ *     The contents of this file are subject to the terms of either the GNU
+ *     General Public License Version 2 only ("GPL") or the Common Development
+ *     and Distribution License("CDDL") (collectively, the "License").  You
+ *     may not use this file except in compliance with the License.  You can
+ *     obtain a copy of the License at
+ *     https://github.com/payara/Payara/blob/main/LICENSE.txt
+ *     See the License for the specific
+ *     language governing permissions and limitations under the License.
+ *
+ *     When distributing the software, include this License Header Notice in each
+ *     file and include the License file at legal/OPEN-SOURCE-LICENSE.txt.
+ *
+ *     GPL Classpath Exception:
+ *     The Payara Foundation designates this particular file as subject to the "Classpath"
+ *     exception as provided by the Payara Foundation in the GPL Version 2 section of the License
+ *     file that accompanied this code.
+ *
+ *     Modifications:
+ *     If applicable, add the following below the License Header, with the fields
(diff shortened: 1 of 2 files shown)


stian-sigvartsen (commented):
@luiseufrasio as we discussed the runtime code looks correct, but I am not convinced it is actually working. I haven't yet done a manual test (can you verify?) because I wanted to share what I've got so far. 

With the changes the tests does now fail, which I think it should do because it is finding unexpected data in the `secondPU`. You'll notice that I created a 2nd H2 DataSource (& connection pool) for it, which appears to be working correctly (checked with debugger).

I made some small refactors to move some of the logic to deployment time and reduce the number of `EntityManagers` that are constructed. Those changes will conflict with payara/Payara/pull/7961 so we'll probably need to rebase one or the other.

cc @breakponchito

breakponchito:
@luiseufrasio in general looks great and we need to fix the test that is failing, also we need to consider to create a different ticket to implement search within the DataSourceDefinition annotations

stian-sigvartsen (approved):

PR #7984, opened 2026-03-01
FISH-13058 Session bugs in availability mode: Missing `JSESSIONIDVERSION` and parallel requests Lead To Session Not Being Correctly Retrieved In Relaxed Mode

lprimak:
…y retrieved in relaxed mode

This only happens on a first request that happens in parallel with session creation, so very sporadic.

Another issue in the same realm. During parallel requests, older version of the session can overwrite a newer version. This fixes that by preventing older version from ever overriding a newer one.

## Description
This is a bug fix. When a new session is being accessed in parallel with another thread creating it,
no version exists and thus session is not being found, or newer sessions being overwritten with older versions

### DIFF
--- a/appserver/web/web-core/src/main/java/org/apache/catalina/session/ManagerBase.java
+++ b/appserver/web/web-core/src/main/java/org/apache/catalina/session/ManagerBase.java
@@ -55,7 +55,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Portions Copyright [2016-2021] [Payara Foundation and/or its affiliates]
+// Portions Copyright [2016-2026] [Payara Foundation and/or its affiliates]
 
 package org.apache.catalina.session;
 
@@ -708,7 +708,14 @@ public abstract class ManagerBase implements Manager {
      */
     @Override
     public void add(Session session) {
-        sessions.put(session.getIdInternal(), session);
+        sessions.compute(session.getIdInternal(), (id, existingSession) -> {
+            if (isSessionVersioningSupported()
+                    && existingSession != null
+                    && existingSession.getVersion() > session.getVersion()) {
+                    return existingSession;
+                }
+            return session;
+        });
         int size = sessions.size();
         if (size > maxActive) {
             synchronized(maxActiveUpdateLock) {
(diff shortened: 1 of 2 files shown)


stian-sigvartsen:
HI @lprimak . Understandably it is very tricky to provide a reproducing steps for the bug, but can you clarify how the mentioned condition(s) is possible?
> This only happens on a first request that happens in parallel with session creation

> When a new session is being accessed in parallel with another thread creating 

I mean, until the session creating request has responded, the JSESSIONID is unknown to the client so there is no chance of another request being made referring to that. 

Presumably there is some possibility of multiple threads used by that session creating request causing the issue, but I cannot see it right now.

stian-sigvartsen:
Also the only way I have found to trigger your new if block in `ReplicationManagerBase#findSession` is following these steps (using the `clusterjsp` app, but any should do):

1) Clear all cookies
2) Request node 2's clusterjsp endpoint. This will cause new JSESSIONID and JSESSIONVERSION cookies to be set
3) Manually delete the JSESSIONVERSION cookie from the browser
4) Request node 1's clusterjsp endpoint. It is important that node 1 has no previous awareness of the session ID

But in real world usage, the JSESSIONVERSION will be set alongside the JSESSIONID and will be used in following requests, so it doesn't seem like a problem?

lprimak:
Hi, Stian, Thanks for looking into this.

The use case is triggered by parallel browser requests, along with JavaScript / Ajax requests on the same page. Browser will parallelize those requests as much as possible, resulting in different versions of the session being processed at the same time, possibly out of order.

Grizzly processes requests in an asynchronous manner, meaning that a multiple requests with the same `JSESSIONID` can be processed at the same time. Furthermore, since Grizzly processes requests in chunks, it can process the request headers from multiple requests at the same time, with the same `JSESSIONID`, and return different `JSESSIONVIERSION` in their responses, which can come back in a different order.
ClusterJSP sample's pages are not long enough or complicated enough to trigger this bug.

Session version is actually stored in an attribute (See `Request.getSessionVersionsRequestAttribute`)
and `Request.requestedSessionVersion` field is also used, so there are many instances where the version is not directly taken / parsed from the `JSESSIONIDVERSION` cookie.

stian-sigvartsen (approved):

PR #7986, opened 2026-03-02
Fish 13018 fix trace tck issues and reestructure telemetry adapter module

breakponchito:
FISH-13018 add changes to pass telemetry tracing TCK and reestructure telemetry adapter module
## Description

This is a feature to delete old functionality for old traces an fix issues related to latest version of telemetry-tracing to pass TCK
### Testing Performed

Execute the MicroProfile-TCK-Runner with configuration to execute tracing TCKs, check the following PR:  https://github.com/payara/MicroProfile-TCK-Runners/pull/292

After execution we will see the following output using changes for the PR from MicroProfile-TCK-Runners:

image

### Testing Environment

Ubuntu 20.04, Azul JDK 21, maven 3.9.11

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/opentracing-jaxws/src/main/java/fish/payara/opentracing/jaxws/JaxwsContainerRequestTracingFilter.java
+++ b/appserver/payara-appserver-modules/microprofile/opentracing-jaxws/src/main/java/fish/payara/opentracing/jaxws/JaxwsContainerRequestTracingFilter.java
@@ -436,6 +436,11 @@ public class JaxwsContainerRequestTracingFilter implements MonitorFilter {
 
         private final MultivaluedMap<String, String> map;
 
+        /**
+         * Initialises this object with the MultivaluedMap to wrap.
+         *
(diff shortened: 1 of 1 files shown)


appserver/payara-appserver-modules/microprofile/opentracing-jaxws/src/main/java/fish/payara/opentracing/jaxws/JaxwsContainerRequestTracingFilter.java
Pandrex247, line 1: If you're deleting this class - it's probably worth deleting `OpenTracingJaxwsCdiUtils`
At which point - do we even need the module?

breakponchito: We need this class to provide traces when the tracing mechanism is enabled, that is why I reverted this change and I fixed issues to import tracing classes

PR #7993, opened 2026-03-05
FISH-12994 Readd Corba FOLB

Pandrex247:
## Description
Readds the Corba Fail Over / Loading Balancing removed in https://github.com/payara/Payara/pull/7935

After reading into it some more, the implementation does make sense with some additional fixes.
It _specifically_ only makes sense if the instances being clustered are: in a single deployment group; the application is deployed to all instances in the deployment group; all instances within the deployment group share the same instance group; the instance group is not shared with any other instance outside of the deployment group: they cannot use the default _MicroShoal_ **unless** the intent is to have all instances within the domain (including the DAS) within the same deployment group. Any deviation from this will lead to errors: there is little validation here, but that's how it has always been and would be feature creep to start adding it in now.

This PR also includes a couple of fixes for race conditions on setting the instance group and name (plus the HazelcastCore#setAttribute method not actually setting the attribute you ask it to). The race condition is _still there_ in a sense that the instance name is only set after Hazelcast has booted and joined the cluster, but it has a fallback now with the initial instance group and name also set in the `MemberAttributeConfig` pre-bootstrap of Hazelcast.

### Blockers
N/A

### New tests
None

### Testing Performed

* Start the DAS
* Create a new config based off of _default-config_ called _Gruppy-config_
* Enable `FINEST` logging on the `javax.enterprise.resource.corba` logger in _Gruppy-config_
* Set the Hazelcast instance group in _Gruppy-config_ to _GruppyShoal_
* Create a deployment group named _Gruppy_
* Deploy the attached `remoteview.jar` to _Gruppy_
  * This comes from quicklook if you want to rebuild this JAR yourself
  * [remoteview.zip](https://github.com/user-attachments/files/25790465/remoteview.zip)
* Create a new instance named _Insty1_, referencing _Gruppy-config_ (not copying), and add it to _Gruppy_
* Create a new instance named _Insty2_, referencing _Gruppy-config_ (not copying), and add it to _Gruppy_
* Start _Insty1_ and let it sort itself out
  * You should see in the logs that Corba FOLB has started with only _Insty1_ registered: `org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient;MethodName=fineLog;|
  internalClusterInstanceInfo: currentMembers {Insty1=ClusterInstanceInfo[name=Insty1 weight=100 endpoints=[SocketInfo[type=orb-listener-1 host=Pandrex-PC port=23700], SocketInfo[type=SSL host=Pandrex-PC port=23820], SocketInfo[type=SSL_MUTUALAUTH host=Pandrex-PC port=23920]]]}|#]`
  * You should also see that the JAR was deployed and printed some messages: `HellohelloRemote.hello() says hello, world`
* Start _Insty2_ and let it sort itself out
  * You should see in the logs that Corba FOLB has started with **both** instances registered: `org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient;MethodName=fineLog;|
  internalClusterInstanceInfo: currentMembers {Insty2=ClusterInstanceInfo[name=Insty2 weight=100 endpoints=[SocketInfo[type=orb-listener-1 host=Pandrex-PC port=23701], SocketInfo[type=SSL host=Pandrex-PC port=23821], SocketInfo[type=SSL_MUTUALAUTH host=Pandrex-PC port=23921]]], Insty1=ClusterInstanceInfo[name=Insty1 weight=100 endpoints=[SocketInfo[type=orb-listener-1 host=Pandrex-PC port=23700], SocketInfo[type=SSL host=Pandrex-PC port=23820], SocketInfo[type=SSL_MUTUALAUTH host=Pandrex-PC port=23920]]]}|#]`
  * You should also see that the JAR was deployed and printed some messages: `HellohelloRemote.hello() says hello, world`
* You should also see in the logs of _Insty1_ that it has added _Insty2_ to its cluster: `IiopFolbGmsClient.addMember<-: Insty2|#]`

### Testing Environment
Windows 11, Maven 3.9.12, Zulu JDK 21.0.10

## Documentation
The docs needs updating to specifically mention the instance group, as it currently only makes mention of the deployment group part that's required here.

## Notes for Reviewers
None

### DIFF
--- a/nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/cluster/MemberEvent.java
+++ b/nucleus/payara-modules/hazelcast-bootstrap/src/main/java/fish/payara/nucleus/cluster/MemberEvent.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
- * Copyright (c) 2017-2020 Payara Foundation and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017-2026 Payara Foundation and/or its affiliates. All rights reserved.
  *
  * The contents of this file are subject to the terms of either the GNU
  * General Public License Version 2 only ("GPL") or the Common Development
@@ -59,11 +59,19 @@ public class MemberEvent {
     }
     
     public String getServer() {
-        return hzCore.getAttribute(member.getUuid(), HazelcastCore.INSTANCE_ATTRIBUTE);
+        String serverName = hzCore.getAttribute(member.getUuid(), HazelcastCore.INSTANCE_ATTRIBUTE);
+        if (serverName == null) {
+            serverName = member.getAttribute(HazelcastCore.INSTANCE_ATTRIBUTE);
+        }
+        return serverName;
     }
     
     public String getServerGroup() {
-        return member.getAttribute(HazelcastCore.INSTANCE_GROUP_ATTRIBUTE);
+        String serverGroup = hzCore.getAttribute(member.getUuid(), HazelcastCore.INSTANCE_GROUP_ATTRIBUTE);
+        if (serverGroup == null) {
+            serverGroup = member.getAttribute(HazelcastCore.INSTANCE_GROUP_ATTRIBUTE);
+        }
+        return serverGroup;
     }
     
     private final HazelcastCore hzCore;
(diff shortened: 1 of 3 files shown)


appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
```diff
@@ -0,0 +1,464 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import static org.glassfish.api.naming.NamingClusterInfo.IIOP_CLUSTER_UPDATE_PROPERTY;
+
+// REVISIT impl
+//import com.sun.corba.ee.impl.folb.ServerGroupManager;
```
breakponchito, line 82: Do we need this commented import?

PR #7998, opened 2026-03-06
FISH-12544 FISH-13103 Document session-token-timeout property precedence.

kalinchan:
## Description

No available documentation exists

### DIFF
--- a/appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
+++ b/appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
@@ -712,7 +712,7 @@ domain.ReloadHelp=Enables dynamic reloading of applications.
 domain.ReloadPollInterval=Reload Poll Interval:
 domain.ReloadPollIntervalHelp=Frequency for checking reload requests.
 domain.AdminTimeout=Admin Session Timeout:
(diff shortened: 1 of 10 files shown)


appserver/admingui/common/src/main/resources/org/glassfish/common/admingui/Strings.properties
```diff
@@ -712,7 +712,7 @@ domain.ReloadHelp=Enables dynamic reloading of applications.
 domain.ReloadPollInterval=Reload Poll Interval:
 domain.ReloadPollIntervalHelp=Frequency for checking reload requests.
 domain.AdminTimeout=Admin Session Timeout:
-domain.AdminTimeoutHelp=A value of <tt>0</tt> means the session never times out.
+domain.AdminTimeoutHelp=A value of <tt>0</tt> means the session will never time out. <br/>If this property is set to a value higher than <tt>session-token-timeout</tt>, you must align these values using <tt>_set-rest-admin-config --sessionTokenTimeout {value}</tt>, as the lower value takes precedence.
```
Pandrex247: I'd reword that slightly.

Something along the lines of: "If <tt>session-token-timeout</tt> is set to a lower value, it will take precedence over this timeout.

Also I'm not keen on us directly referring to a hidden asadmin command (prefixed with '_'); it might be better to just provide the equivalent `set` command.

PR #8034, opened 2026-03-23
FISH-13115 update nexus registry creds

abdulrahim458:


Pandrex247 (changes requested):
Community shouldn't require access to the private Enterprise repo

Pandrex247 (dismissed):

.github/dependabot.yml
```diff
@@ -1,8 +1,18 @@
...
+    url: https://nexus.dev.payara.fish/repository/payara-enterprise-artifacts-private
+    username: ${{secrets.PAYARA_NEXUS_USERNAME}}
+    password: ${{secrets.PAYARA_NEXUS_PASSWORD}}
+
 updates:
   # Configuration for Maven
   - package-ecosystem: "maven"
     directory: "/" # Location of package manifests
+    registries:
+      - payara-private-nexus
```
Pandrex247: It's not trying to check this repo, it's trying to check the private staging repo

Pandrex247 (changes requested):

.github/dependabot.yml
Pandrex247: We should either exclude this repo (as that repo should be disabled by default anyway), and we don't want dependabot prematurely creating PRs for things we're trying to patch - it shouldn't be creating PRs for *.payara-px anyway, and that's the only thing which should really be in there

.github/dependabot.yml
```diff
@@ -20,6 +20,12 @@ updates:
       - dependency-name: "fish.payara.*"
         update-types: ["version-update:semver-major"]
 
+      - dependency-name: "*payara-px*"
```
Pandrex247: I'm not sure this would match anything?
We don't have any dependencies following that naming convention unless I'm overlooking something

Pandrex247 (changes requested):

.github/dependabot.yml
abdulrahim458: else i can just ignore the failing dependencies to check

- dependency-name: "javax.xml.stream:stax-api"
- dependency-name: "javax.jbi:jbi"

this prevents from failures or else add the staging registry details to dependabot file

Pandrex247: Those dependencies will just be the first ones being polled - there's nothing special about them

abdulrahim458:
**Before**: Dependabot was returning 401 Unauthorized on every request to payara-staging, causing private_source_authentication_failure for javax.jbi:jbi and javax.xml.stream:stax-api. Because of these auth failures, the job was exiting with a non-zero exit code meaning not all dependencies were being evaluated  some updates were being silently skipped and no PRs were created for those dependencies.

**After**: With the registry credentials added, payara-staging now correctly returns 404 where artifacts aren't present and falls back to payara-artifacts for Payara-patched dependencies (e.g. *.payara-p1) and Maven Central for everything else. All dependencies are now fully evaluated. 

The job completes successfully and update PRs are created correctly.

Test run: https://github.com/abdulrahim458/Payara/actions/runs/24636034657/job/72031532827

CC: @Pandrex247

Pandrex247 (approved):
I still don't think this is required but I will approve for now to see what happens

PR #8106, opened 2026-04-23
FISH-13193 : throwing UnsupportedOperationException when the method is not supported by the Jakarta Data provider

luiseufrasio:
## Description
This PR addresses the issue where invoking an unsupported Jakarta Data repository method at runtime resulted in a `NullPointerException` instead of the expected `UnsupportedOperationException`.

Previously, if a repository method could not be mapped to a valid query during deployment (e.g., due to an unresolvable entity type or an invalid method signature), the method's metadata was not generated. This led to a null `QueryMetadata` being retrieved at runtime, causing an NPE when attempting to access its properties.

**Changes included:**
* **Exception Handling:** The `RepositoryImpl.invoke` method now checks if `queryMetadata` is null and throws `UnsupportedOperationException`.
* **Object Methods handling:** Included explicit handling for common `Object` methods (`toString`, `hashCode`, `equals`) to ensure the proxy behaves correctly.
* **Spec Compliance:** This ensures compliance with the Jakarta Data specification regarding methods that cannot be implemented at deployment time.

### Blockers
* N/A

### New tests
* `appserver/data/data-core/src/test/java/fish/payara/data/core/cdi/extension/RepositoryImplTest.java`
* `appserver/tests/payara-samples/samples/data/src/main/java/fish/payara/samples/data/repo/UnsupportedRepo.java`
* `appserver/tests/payara-samples/samples/data/src/test/java/fish/payara/samples/data/UnsupportedMethodTest.java`

### Testing Performed
* **Unit Testing:** Created `RepositoryImplTest.java` to verify the core logic of `RepositoryImpl` throwing `UnsupportedOperationException` when metadata is missing.
* **Integration Testing:** Added an Arquillian-based test (`UnsupportedMethodTest.java`) to confirm that invoking an unsupported method on an injected repository correctly results in the expected exception.

### Testing Environment
- JDK: Zulu JDK 21
- OS: Windows 11
- Maven: 3.9.9

## Documentation
* N/A

### DIFF
--- a/appserver/data/data-core/pom.xml
+++ b/appserver/data/data-core/pom.xml
@@ -70,11 +70,6 @@
             <artifactId>weld-integration</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
     
     <build>
--- a/appserver/data/data-core/pom.xml
+++ b/appserver/data/data-core/pom.xml
@@ -70,6 +70,11 @@
             <artifactId>weld-integration</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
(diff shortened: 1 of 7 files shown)


breakponchito (approved):
LGTM

PR #8174, opened 2026-05-22
FISH-13392: enabling unit tests after fixing issue for NullPointerException

breakponchito:
Enabling unit test for Fault Tolerance previously disabled
## Description

This is a fix to enable the unit test that previously were disabled when upgrading the version of Fault Tolerance for MicroProfile 7.1
### Testing Performed

Build server from pipeline without any issue after passing all unit tests and the TCK from Fault Tolerance:

image

Fault Tolerance TCK:
image

### Testing Environment

Ubuntu 20.04, azul jdk 21,  maven 3.9.11

### DIFF
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
@@ -309,9 +309,9 @@ public interface FaultToleranceMetrics {
                     {"result", "valueReturned", "exceptionThrown"}, fallbackTag});
             //place to register telemetry ft.invocations.total
             Meter currentMeter = getCurrentMeter();
+            FaultToleranceMethodContextImpl faultToleranceMethodContext = (FaultToleranceMethodContextImpl) context;
+            setClassAndMethodName(faultToleranceMethodContext.getClassName() + "." + faultToleranceMethodContext.getMethodName());
             if (currentMeter != null) {
-                FaultToleranceMethodContextImpl faultToleranceMethodContext = (FaultToleranceMethodContextImpl) context;
-                setClassAndMethodName(faultToleranceMethodContext.getClassName() + "." + faultToleranceMethodContext.getMethodName());
                 addFTInvocationTotalMeter(createFTInvocationTotalMeter(getClassAndMethodName(), currentMeter, policy.isFallbackPresent()));
             }
 
@@ -391,10 +391,11 @@ public interface FaultToleranceMetrics {
     }
 
     default Meter getCurrentMeter() {
-        if (getDefaultHabitat() == null) {
+        var habitat = getDefaultHabitat();
+        if (habitat == null) {
             return null;
         }
-        OpenTelemetryService service = getDefaultHabitat().getService(OpenTelemetryService.class);
+        var service = habitat.getService(OpenTelemetryService.class);
         return service != null ? service.getCurrentMeter() : null;
     }
     
--- a/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
+++ b/appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
@@ -309,8 +309,10 @@ public interface FaultToleranceMetrics {
                     {"result", "valueReturned", "exceptionThrown"}, fallbackTag});
             //place to register telemetry ft.invocations.total
             Meter currentMeter = getCurrentMeter();
-            FaultToleranceMethodContextImpl faultToleranceMethodContext = (FaultToleranceMethodContextImpl) context;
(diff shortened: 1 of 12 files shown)


appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
```diff
@@ -344,38 +349,54 @@ default FaultToleranceMetrics boundTo(FaultToleranceMethodContext context, Fault
...
         return this;
     }
+
+    default Meter getCurrentMeter() {
+        if (getDefaultHabitat() == null) {
+            return null;
+        }
+        OpenTelemetryService service = getDefaultHabitat().getService(OpenTelemetryService.class);
+        return service != null ? service.getCurrentMeter() : null;
+    }
```
luiseufrasio, line 402: Would it be better reusing the call to `getDefaultHabitat`:
```
default Meter getCurrentMeter() {
      var habitat = getDefaultHabitat();
      if (habitat == null) return null;
      var service = habitat.getService(OpenTelemetryService.class);
      return service != null ? service.getCurrentMeter() : null;
  }
```

WDYT?

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
```diff
@@ -302,17 +302,18 @@ public String getClassAndMethodName() {
...
                     {"result", "valueReturned", "exceptionThrown"}, fallbackTag});
             //place to register telemetry ft.invocations.total
-            Meter currentMeter = openTelemetryService.getCurrentMeter();
-            FaultToleranceMethodContextImpl faultToleranceMethodContext = (FaultToleranceMethodContextImpl) context;
-            setClassAndMethodName(faultToleranceMethodContext.getClassName() + "." + faultToleranceMethodContext.getMethodName());
-            addFTInvocationTotalMeter(createFTInvocationTotalMeter(getClassAndMethodName(), currentMeter, policy.isFallbackPresent()));
+            Meter currentMeter = getCurrentMeter();
+            if (currentMeter != null) {
+                FaultToleranceMethodContextImpl faultToleranceMethodContext = (FaultToleranceMethodContextImpl) context;
+                setClassAndMethodName(faultToleranceMethodContext.getClassName() + "." + faultToleranceMethodContext.getMethodName());
```
luiseufrasio: Should `setClassAndMethodName` be done before `if (currentMeter != null)` instead of inside it?

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
```diff
@@ -344,38 +349,54 @@ default FaultToleranceMetrics boundTo(FaultToleranceMethodContext context, Fault
...
+                        addFTBulkheadWaitingDuration(createFTBulkheadWaitingDuration(currentMeter));
+                    }
                 } else {
                     AtomicInteger running = context.getQueuingOrRunningPopulation();
                     register("ft.bulkhead.executionsRunning", null, running::get);
                     this.setExecutionBulkheadRunningSupplier(running::get);
-                    addFTBulkheadExecutionRunning(createFTBulkheadExecutionsRunning(currentMeter, this));
+                    if (currentMeter != null) {
+                        addFTBulkheadExecutionRunning(createFTBulkheadExecutionsRunning(currentMeter, this));
+                    }
```
luiseufrasio, line 387: Is there a way to avoid the repetition of `if (currentMeter != null)` ?

luiseufrasio (changes requested):

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
breakponchito: done

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
breakponchito: We have different conditions from the policy to validate if we can apply specific configuration, in this case it is needed to validate if the currentMeter is null on each condition and also to pass the junit test execution skipping null reference that can be provided by the lookup of the service

appserver/payara-appserver-modules/microprofile/fault-tolerance/src/main/java/fish/payara/microprofile/faulttolerance/FaultToleranceMetrics.java
breakponchito: done

breakponchito:
Jenkins test plese

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

luiseufrasio (approved):
LGTM

PR #8177, opened 2026-05-25
FISH-10485 FISH-13138 Move Micro Boot Properties Timing Earlier and Fix Launcher Not Reading Boot Properties

Pandrex247:
## Description
Moves the setting of the boot properties for Payara Micro earlier.

Changes to when Java first performs the lookup for the `java.net.preferIPv4Stack` property meant that when using Java 21 or higher and not defining a directory for Payara Micro to unpack into (default behaviour is to unpack into a temp directory) then the payara-boot.properties file had not been read and applied yet (which includes setting this property).

This property is only looked up once, meaning that it was forever initialising as null.

This moves the reading and setting of properties to before a temp dir is created (since this is what causes the lookup).

However, the "stronger" way of configuring this property remains to be defining it on the command line itself. I don't think there is a way around this, and figuring out if there is becomes troublesome as you start heading into native code.

This PR also fixes the Payara Micro Launcher not reading from the boot properties

### Blockers
None

### New tests
None

### Testing Performed
We never got a reproducer for the inciting issue to work on anything other than the issue raiser's environment. So all testing is essentially done via breakpoints.

Added break points to `InetAddress#initializePlatformLookupPolicy` (from the JDK itself) to see which class `impl` gets initialised as, and the values of `PREFER_IPV4_STACK_VALUE` and `PREFER_IPV6_ADDRESSES_VALUE`.

An interesting quirk which I can't yet fathom out (as it goes into `native` code) is why specifying `java.net.preferIPv4Stack` on the command line seems to be a "stronger" way of setting it: doing so makes `impl` be `Inet4AddressImpl` class rather than `Inet6AddressImpl` (default). Setting the property programmatically via `System.setProperty` does not replicate this behaviour, even though the static initialisation of the system properties has already happened by the time the class is initialised. 

Scenarios tested:
* No root dir, no property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar payara-micro.jar`
  * IPv4 Preferred = true
  * impl = Inet6AddressImpl
  * LookupPolicy = IPV4
* No root dir, property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar payara-micro.jar`
  * IPv4 Preferred = true
  * impl = Inet4AddressImpl
  * LookupPolicy = IPV4
* Clean root dir, no property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar payara-micro.jar --rootdir C:\Users\pandr\Downloads\FISH-10485\explode`
  * IPv4 Preferred = true
  * impl = Inet6AddressImpl
  * LookupPolicy = IPV4
* Clean root dir, property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar payara-micro.jar --rootdir C:\Users\pandr\Downloads\FISH-10485\explode`
  * IPv4 Preferred = true
  * impl = Inet4AddressImpl
  * LookupPolicy = IPV4
* Uberjar, no property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar payara-micro.jar --outputuberjar C:\Users\andrew.pielage\Downloads\FISH-10485\uberjar.jar && java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar C:\Users\andrew.pielage\Downloads\FISH-10485\uberjar.jar`
  * IPv4 Preferred = true
  * impl = Inet6AddressImpl
  * LookupPolicy = IPV4 
* Uberjar, property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar payara-micro.jar --outputuberjar C:\Users\andrew.pielage\Downloads\FISH-10485\uberjar.jar` && java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar C:\Users\andrew.pielage\Downloads\FISH-10485\uberjar.jar`
  * IPv4 Preferred = true
  * impl = Inet4AddressImpl
  * LookupPolicy = IPV4
* Launcher, no property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar payara-micro.jar --rootdir C:\Users\andrew.pielage\Downloads\FISH-10485\explode --outputlauncher && java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar C:\Users\andrew.pielage\Downloads\FISH-10485\explode\launch-micro.jar`
  * IPv4 Preferred = true
  * impl = Inet6AddressImpl
  * LookupPolicy = IPV4 
* Launcher, property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar payara-micro.jar --rootdir C:\Users\andrew.pielage\Downloads\FISH-10485\explode --outputlauncher && java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar C:\Users\andrew.pielage\Downloads\FISH-10485\explode\launch-micro.jar`
  * IPv4 Preferred = true
  * impl = Inet4AddressImpl
  * LookupPolicy = IPV4
* Nested: no property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" -jar payara-micro.jar --nested`
  * IPv4 Preferred = true
  * impl = Inet6AddressImpl
  * LookupPolicy = IPV4 
* Nested, property set: `java -Xdebug "-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=y" "-Djava.net.preferIPv4Stack=true" -jar payara-micro.jar --nested`
  * IPv4 Preferred = true
  * impl = Inet4AddressImpl
  * LookupPolicy = IPV4

### Testing Environment
Windows 11, Maven 3.9.15, Zulu JDK 21.0.11

## Documentation
Pending...
It's probably worth mentioning in our docs that this property is configured by default (if it isn't already).

It's probably also worth mentioning the _very similar_ Hazelcast property: `hazelcast.prefer.ipv4.stack` which is true by default - disabling Hazelcast from using IPv6.

From their own docs:

> VM has two system properties for setting the preferred protocol stack (IPv4 or IPv6) as well as the preferred address family types (inet4 or inet6). On a dual stack machine, IPv6 stack is preferred by default, you can change this through the java.net.preferIPv4Stack=<true|false> system property. When querying name services, JVM prefers IPv4 addresses over IPv6 addresses and returns an IPv4 address if possible. You can change this through java.net.preferIPv6Addresses=<true|false> system property. 
By default, IPv6 support has been switched off. Some platforms have issues using the IPv6 stack. Other platforms have no IPv6 support at all. To enable IPv6 support, just set the configuration property hazelcast.prefer.ipv4.stack to false.

## Notes for Reviewers
None

### DIFF
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/Launcher.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/Launcher.java
@@ -13,16 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Modified 2017 Payara Foundation
+// Modified 2017-2026 Payara Foundation
 package fish.payara.micro.boot.loader;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
 import java.net.URL;
 import java.security.CodeSource;
 import java.security.ProtectionDomain;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import fish.payara.micro.boot.loader.archive.Archive;
 import fish.payara.micro.boot.loader.archive.JarFileArchive;
@@ -35,6 +40,8 @@ import fish.payara.micro.boot.loader.archive.JarFileArchive;
  * @author Dave Syer
  */
 public abstract class Launcher {
+
+    private static final String BOOT_PROPS_FILE = "/MICRO-INF/payara-boot.properties";
 
     /**
      * Launch the application. This method is the initial entry point that
@@ -57,6 +64,8 @@ public abstract class Launcher {
             }
         }
 
+        setPayaraBootProperties();
+
         ClassLoader classLoader;
         if (!explode) {
             classLoader = createClassLoader(getClassPathArchives());
@@ -71,6 +80,26 @@ public abstract class Launcher {
             }
         }
         return launch(method, args, getMainClass(), classLoader);
+    }
+
+    public static Properties setPayaraBootProperties() {
+        Properties bootProperties = new Properties();
+
+        try (InputStream is = Launcher.class.getResourceAsStream(BOOT_PROPS_FILE)) {
+            if (is != null) {
+                bootProperties.load(is);
+                for (String key : bootProperties.stringPropertyNames()) {
+                    // Do not override an existing system property
+                    if (System.getProperty(key) == null) {
+                        System.setProperty(key, bootProperties.getProperty(key));
+                    }
+                }
+            }
+        } catch (IOException ioe) {
+            Logger.getLogger(Launcher.class.getName()).log(Level.WARNING, "Could not load the boot system properties from " + BOOT_PROPS_FILE, ioe);
+        }
+
+        return bootProperties;
     }
 
     /**
--- a/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/Launcher.java
+++ b/appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/Launcher.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-// Modified 2017-2026 Payara Foundation
+// Modified 2017-2026 Payara Foundation and/or its affiliates
 package fish.payara.micro.boot.loader;
 
 import java.io.File;
@@ -41,7 +41,7 @@ import fish.payara.micro.boot.loader.archive.JarFileArchive;
  */
 public abstract class Launcher {
 
-    private static final String BOOT_PROPS_FILE = "/MICRO-INF/payara-boot.properties";
+    public static final String BOOT_PROPS_FILE = "/MICRO-INF/payara-boot.properties";
 
     /**
      * Launch the application. This method is the initial entry point that
@@ -64,7 +64,9 @@ public abstract class Launcher {
             }
         }
 
-        setPayaraBootProperties();
+        try (InputStream inputStream = Launcher.class.getResourceAsStream(BOOT_PROPS_FILE)) {
+            setPayaraBootProperties(inputStream);
+        }
 
         ClassLoader classLoader;
         if (!explode) {
@@ -82,12 +84,20 @@ public abstract class Launcher {
         return launch(method, args, getMainClass(), classLoader);
     }
 
-    private static Properties setPayaraBootProperties() {
+    /**
+     * Loads the properties from the given {@link InputStream} and sets them as system properties if not already set.
+     *
+     * This method will not close the provided {@link InputStream}.
+     *
+     * @param inputStream the input stream to read the properties from.
+     * @return The properties read from the {@code inputStream} parameter.
+     */
+    public static Properties setPayaraBootProperties(InputStream inputStream) {
         Properties bootProperties = new Properties();
 
-        try (InputStream is = Launcher.class.getResourceAsStream(BOOT_PROPS_FILE)) {
-            if (is != null) {
-                bootProperties.load(is);
+        try {
+            if (inputStream != null) {
+                bootProperties.load(inputStream);
                 for (String key : bootProperties.stringPropertyNames()) {
                     // Do not override an existing system property
                     if (System.getProperty(key) == null) {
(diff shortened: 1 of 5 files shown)


pdudits:
You haven't tested outputlauncher as outlined in FISH-13138 and haven't touched `LauncherCreator`, so boot properties are nowhere on the classpath of Payara Micro started via `java -jar rootidr/launch-micro.jar`.

Pandrex247:
I have tested output launcher, check the _Testing Performed_ again - bullet points 7 and 8

pdudits:
Oh, sorry, let me re-verify that.

appserver/extras/payara-micro/payara-micro-boot/src/main/java/fish/payara/micro/boot/loader/Launcher.java
```diff
@@ -57,6 +66,12 @@ protected Object launch(String method, String[] args) throws Exception {
             }
         }
 
+        try (InputStream inputStream = Launcher.class.getResourceAsStream(BOOT_PROPS_FILE)) {
+            setPayaraBootProperties(inputStream);
+        } catch (FileNotFoundException fnfe) {
```
luiseufrasio: `Class.getResourceAsStream()` returns `null` when the resource is not found — it does not throw `FileNotFoundException`

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/RootDirLauncher.java
```diff
@@ -61,6 +67,14 @@ public static void main(String[] args) throws Exception {
         String rootDir = bootJar.getParentFile().getAbsolutePath();
         System.setProperty(BOOT_JAR_URL, bootJar.toURI().toString());
         System.setProperty(ROOT_DIR_PATH, rootDir);
+
+        String bootPropsFile = rootDir + File.separator + CONFIG_DIR_NAME + File.separator + BOOT_PROPS_FILE_NAME;
+        try (FileInputStream fis = new FileInputStream(bootPropsFile)) {
+            Launcher.setPayaraBootProperties(fis);
+        } catch (FileNotFoundException fnfe) {
```
luiseufrasio: `FileInputStream.close()` can throw `IOException`, and try-with-resources propagates suppressed exceptions. Only `FileNotFoundException` (a subclass) is caught here; a broader `IOException` from closing the stream would propagate uncaught to `main()`, which would terminate with a stack trace. Catch IOException instead. WDYT?

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/RootDirLauncher.java
```diff
@@ -61,6 +67,14 @@ public static void main(String[] args) throws Exception {
         String rootDir = bootJar.getParentFile().getAbsolutePath();
         System.setProperty(BOOT_JAR_URL, bootJar.toURI().toString());
         System.setProperty(ROOT_DIR_PATH, rootDir);
+
+        String bootPropsFile = rootDir + File.separator + CONFIG_DIR_NAME + File.separator + BOOT_PROPS_FILE_NAME;
```
luiseufrasio: Sugestion: `Path.of(rootDir, CONFIG_DIR_NAME, BOOT_PROPS_FILE_NAME).toString()` is more idiomatic and avoids separator edge cases.

luiseufrasio (changes requested):

appserver/extras/payara-micro/payara-micro-core/src/main/java/fish/payara/micro/impl/RootDirLauncher.java
Pandrex247: Hmm, on the one-hand your reasoning makes sense. On the other hand - if a full IOException gets thrown, perhaps we **want** to just abort...? 🤔

luiseufrasio (approved):
LGTM!

PR #8214, opened 2026-06-05
FISH-11606 Avoid parsing method names on every Jakarta Data HTTP request

luiseufrasio:
## Description

Jakarta Data "query by method name" repositories (`findBy...`, `deleteBy...`, `countBy...`, `existsBy...`) were re-parsing the method name on **every** HTTP request. The parsing result is fully determined by the method signature, so it is invariant for the lifetime of the deployment — recomputing it per request is wasted work on the hot path.

This PR moves the parsing to **deploy time** and caches the result:

- `QueryMethodParser.parse()` now returns an immutable `ParseResult` record (action, limit, orderBy, conditions) instead of mutating/returning the parser itself.
- The `ParseResult` is computed once in `DynamicInterfaceDataProducer#addQueries` (deploy time) and stored on the immutable `QueryMetadata`.
- `QueryByNameOperationUtility` now reads the cached `ParseResult` from `QueryMetadata` instead of constructing a new parser on each request.
- The `QueryType` for by-name methods is now derived from the parsed `Action` at deploy time (via `queryTypeForByNameAction`) rather than from ad-hoc `methodName.startsWith(...)` prefix checks.
- Dead `Action.FIND_FOR_DELETE` (unreachable — could only be produced by the non-identifier method name `find_for_delete*`) and its two references were removed.

**Behaviour change (intentional, fail-fast):** a malformed/unsupported by-name method previously fell through to `FIND_BY_NAME`, was registered, and only failed at request time with a `MappingException`. It now fails to parse at deploy time, is logged as a warning, and is never registered; invoking it throws `UnsupportedOperationException`. Valid queries are unaffected — the same parser is used, just earlier.

### Blockers

None. Builds on the preparatory commit `cd3361d4` (extraction of the `ParseResult` record), which is included in this branch.

### New tests

- `DynamicInterfaceDataProducerTest` — covers the deploy-time `Action -> QueryType` classification, both via the direct mapping and via the realistic method-name -> parser -> query-type path.
- `QueryMethodParserTest` — updated to the new `ParseResult` return type.

### Testing Performed

- `mvn -o -pl appserver/data/data-core test -Dtest=DynamicInterfaceDataProducerTest,QueryMethodParserTest` — **38 tests, 0 failures**.
- `mvn -o -pl appserver/data/data-core test-compile` — module compiles cleanly.

### Testing Environment

Zulu JDK 21.0.1 on Windows 11 with Apache Maven 3.9.9.

## Documentation

N/A — internal performance change, no public API or user-facing behaviour change for valid queries.

## Notes for Reviewers

- Good starting point: `DynamicInterfaceDataProducer#addQueries` (where parsing now happens once, inside the existing `try`) and `QueryByNameOperationUtility#buildAndExecuteQuery` / `findEntitiesForModification` (which now read `QueryMetadata#getParseResult()` instead of re-parsing).
- The first commit (`cd3361d4`, by @stian-sigvartsen) is preparatory groundwork: it introduces the `ParseResult` record and updates the parser API. The remaining commits move parsing to deploy time, cache the result, remove the dead action, and add the classification test.
- `getParseResult()` is nullable by design — it is only populated for `*_BY_NAME` query types; annotation-driven types (`@Find`, `@Query`, `@Save`, etc.) leave it null.


stian-sigvartsen (approved):
Looks great @luiseufrasio !

PR #8243, opened 2026-06-19
FISH-13207 Fix `restart-deployment-group` Not Displaying Progress.

Pandrex247:
## Description
Makes it so the `restart-deployment-group` command displays the progress of the command like the other group start/stop/restart commands.

Also clears away what seems like redundant code: the cluster command helper should be handling all of this.
_This isn't just cleanup, the progress mechanism only works when using `ClusterCommandHelper`_

### Blockers
None

### New tests
None

### Testing Performed
Create a deployment group with multiple instances in and start them all.

Run the `restart-deployment-group` command, you should see the instances restart in a rolling manner and the progress should be displayed (and not just jump from 0-100, it should work its way up based on the number of instances)

### Testing Environment
Windows 11, Maven 3.9.16, Zulu JDK 21.0.11

## Documentation
N/A

## Notes for Reviewers
None

### DIFF
--- a/nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/RestartDeploymentGroupCommand.java
+++ b/nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/RestartDeploymentGroupCommand.java
@@ -1,7 +1,7 @@
 /*
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
  *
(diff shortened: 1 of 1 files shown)


breakponchito:
@Pandrex247 now showing the increment of the percentage depending of the number of instances from the deployment group: 

image

looks ok

breakponchito (approved):
LGTM

PR #8244, opened 2026-06-19
Fish 13751 reimplementing ejb tracing with opentelemetry

breakponchito:
Re-implementing EJB trace mechanism to pass sample tests
## Description

This is a fix to use opentelemetry implementation to create traces and pass sample tests

### Testing Performed

Passing Jenkins pipeline and remote-ejb-tracing from samples:
image

passing from the Jenkins Pipeline:
image

### Testing Environment

Ubuntu 20.04, azul jdk 21, maven 3.9.11

### DIFF
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -338,9 +338,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
     private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
         Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(
                 Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)));
-
         SpanBuilder builder = tracer.spanBuilder("executeConcurrentContext");
-
         Context parentContext = handle.getParentTraceContext();
         if (parentContext != null) {
             builder.setParent(parentContext);
@@ -363,7 +361,6 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         }
 
         builder.setAttribute("Thread Name", Thread.currentThread().getName());
-
         Span span = builder.startSpan();
         currentConcurrentSpan.set(span);
         currentConcurrentScope.set(span.makeCurrent());
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -49,6 +49,12 @@ import com.sun.enterprise.deployment.util.DOLUtils;
 import com.sun.enterprise.security.SecurityContext;
 import com.sun.enterprise.transaction.api.JavaEETransactionManager;
 import com.sun.enterprise.util.Utility;
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
 import org.glassfish.api.invocation.ComponentInvocation;
 import org.glassfish.api.invocation.InvocationManager;
 import org.glassfish.concurrent.LogFacade;
@@ -106,6 +112,9 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
     public static final String CONTEXT_TYPE_SECURITY = "SECURITY"; // Concurrency 3.0: SECURITY
     public static final String CONTEXT_TYPE_NAMING = "NAMING"; // Concurrency 3.0: APPLICATION
     public static final String CONTEXT_TYPE_WORKAREA = "WORKAREA"; // Concurrency 3.0: TRANSACTION
+
+    private static final ThreadLocal<Span> currentConcurrentSpan = new ThreadLocal<>();
+    private static final ThreadLocal<Scope> currentConcurrentScope = new ThreadLocal<>();
 
     // TODO: do we need these booleans if we have sets?
     private boolean classloading, security, naming, workArea;
@@ -306,6 +315,10 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             transactionManager.clearThreadTx();
         }
 
+        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+            startConcurrentContextSpan(invocation, handle);
+        }
+        
         if (stuckThreads != null) {
             stuckThreads.registerThread(Thread.currentThread().threadId());
         }
@@ -320,6 +333,53 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
 
         return new InvocationContext(invocation, resetClassLoader, resetSecurityContext, handle.isUseTransactionOfExecutionThread(),
                 Collections.EMPTY_LIST, restorers);
+    }
+
+    private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
+        Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(
+                Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)));
+
+        SpanBuilder builder = tracer.spanBuilder("executeConcurrentContext");
+
+        Context parentContext = handle.getParentTraceContext();
+        if (parentContext != null) {
+            builder.setParent(parentContext);
+
+            String parentOperationName = Baggage.fromContext(parentContext).getEntryValue("operation.name");
+            if (parentOperationName != null) {
+                builder.setAttribute("Parent Operation Name", parentOperationName);
+            }
+        }
+
+        if (invocation != null) {
+            builder.setAttribute("App Name", invocation.getAppName())
+                    .setAttribute("Component ID", invocation.getComponentId())
+                    .setAttribute("Module Name", invocation.getModuleName());
+
+            Object instance = invocation.getInstance();
+            if (instance != null) {
+                builder.setAttribute("Class Name", instance.getClass().getName());
+            }
+        }
+
+        builder.setAttribute("Thread Name", Thread.currentThread().getName());
+
+        Span span = builder.startSpan();
+        currentConcurrentSpan.set(span);
+        currentConcurrentScope.set(span.makeCurrent());
+    }
+
+    private void stopConcurrentContextSpan() {
+        Scope scope = currentConcurrentScope.get();
+        Span span = currentConcurrentSpan.get();
+        if (scope != null) {
+            scope.close();
+            currentConcurrentScope.remove();
+        }
+        if (span != null) {
+            span.end();
+            currentConcurrentSpan.remove();
+        }
     }
 
     @Override
@@ -362,6 +422,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
         }
 
         if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+            stopConcurrentContextSpan();
             requestTracing.endTrace();
         }
         if (stuckThreads != null) {
--- a/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
+++ b/appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
@@ -315,7 +315,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
             transactionManager.clearThreadTx();
         }
 
-        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
+        if (requestTracing != null && openTracing != null && requestTracing.isRequestTracingEnabled()) {
             startConcurrentContextSpan(invocation, handle);
         }
         
@@ -336,8 +336,7 @@ public class ContextSetupProviderImpl implements ContextSetupProvider {
     }
 
     private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
-        Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(
-                Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)));
+        Tracer tracer = openTracing.getTracer(openTracing.getApplicationName(invocationManager));
         SpanBuilder builder = tracer.spanBuilder("executeConcurrentContext");
         Context parentContext = handle.getParentTraceContext();
         if (parentContext != null) {
--- a/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
+++ b/appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
@@ -92,6 +92,8 @@ import fish.payara.nucleus.requesttracing.RequestTracingService;
 import fish.payara.opentracing.OpenTracingService;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
 import jakarta.ejb.AccessLocalException;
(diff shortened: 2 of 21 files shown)


appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
```diff
@@ -109,17 +109,36 @@ public Object traceCdiCall(final InvocationContext invocationContext) throws Exc
...
+                try {
+                    return invocationContext.proceed();
+                } catch (Exception e) {
+                    LOG.log(Level.FINEST, "Setting the error to the active span ...", e);
+                    span.recordException(e);
+                    throw e;
+                }
+            } finally {
+                span.end();
+                LOG.log(Level.FINEST, "Don't close the span for now");
```
luiseufrasio: says the opposite of what it does?
span is closed in previous line?

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
```diff
@@ -109,17 +109,36 @@ public Object traceCdiCall(final InvocationContext invocationContext) throws Exc
...
+            try (Scope scope = span.makeCurrent()) {
+                try {
+                    return invocationContext.proceed();
+                } catch (Exception e) {
+                    LOG.log(Level.FINEST, "Setting the error to the active span ...", e);
+                    span.recordException(e);
+                    throw e;
+                }
+            } finally {
+                span.end();
```
luiseufrasio, line 125: Is this correct? It seems to be ending the caller's span

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
```diff
@@ -109,17 +109,36 @@ public Object traceCdiCall(final InvocationContext invocationContext) throws Exc
...
-            try {
-                return invocationContext.proceed();
-            } catch (Exception e) {
-                LOG.log(Level.FINEST, "Setting the error to the active span ...", e);
-                span.recordException(e);
-                throw e;
+
+        Span span = Span.current();
+        if (span.isRecording()) {
+            span = span.updateName(operationName).setAttribute("otel.service.name", applicationName);
```
luiseufrasio, line 115: Is renaming needed? The prior version named the span at build time via `spanBuilder(operationName)`

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
```diff
@@ -1719,6 +1746,44 @@ public JavaEETransaction createImportedTransaction(TransactionInternal jtsTx)
...
+                }
+            }
+        } else {
+            final PayaraTracingServices payaraTracingServices = new PayaraTracingServices();
+            final Tracer tracer = payaraTracingServices.getActiveTracer();
+            AttributesBuilder attrsBuilder = Attributes.builder();
+            spanLog.getLogEntries().forEach(attrsBuilder::put);
+            if (tx != null && tracer != null) {
+                span = tracer.spanBuilder("addJtaEventTraceLog").startSpan();
+                span.makeCurrent();
```
luiseufrasio, line 1768: use `try-with-resources` for the `Scope` and a `finally` with end():

```
span = tracer.spanBuilder("jtaTransactionEvent").startSpan();
try (Scope s = span.makeCurrent()) { 
  span.addEvent(...); 
  ... 
} finally { 
  span.end(); 
}
```

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
```diff
@@ -0,0 +1,189 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2020-2026 Payara Foundation and/or its affiliates. All rights reserved.
```
luiseufrasio, line 4: only 2026. It is a new file, right?

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+        } catch (BAD_PARAM e) {
+            // No OTel context was propagated by the client
+            return;
+        }
+
+        HashMap<String, String> contextMap;
+        try (ByteArrayInputStream bis = new ByteArrayInputStream(serviceContext.context_data);
+             ObjectInputStream in = new ObjectInputStream(bis)) {
+            contextMap = (HashMap<String, String>) in.readObject();
```
luiseufrasio: **Deserialization Regression**
I was looking through the history and noticed a critical security regression introduced in this PR (verified against commit 22eade5cd64^).

Previous Behavior: The legacy module used an `OpenTracingIiopObjectInputStream` that safely overrode `resolveClass()`. It operated with a strict allowlist, permitting only `OpenTracingIiopTextMap` and `HashMap` (throwing an `InvalidClassException` for anything else).

Current Behavior: The new implementation instantiates a raw `ObjectInputStream` over bytes arriving directly from a remote IIOP client. This reintroduces the classic Java deserialization gadget-chain attack surface.

**Proposed Fix**:
To close this vulnerability, we need to apply a restrictive filter before `readObject()` is invoked. I recommend one of the following approaches:

Option 1: Apply a restrictive `ObjectInputFilter` (Recommended)
```
// Configure the filter to allow only specific safe classes and reject everything else
in.setObjectInputFilter(ObjectInputFilter.Config.createFilter(
    "java.util.HashMap;java.lang.String;java.lang.Number;!*"
));
```
Option 2: Reintroduce the secure subclass

Reimplement a custom `ObjectInputStream` subclass that overrides `resolveClass()` with a strict allowlist validation pattern, mirroring the original pre-regression architecture.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -306,6 +315,10 @@ public ContextHandle setup(ContextHandle contextHandle) throws IllegalStateExcep
             transactionManager.clearThreadTx();
         }
 
+        if (requestTracing != null && requestTracing.isRequestTracingEnabled()) {
```
luiseufrasio: add `&& openTracing != null` to avoid NPE in line 339

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
```diff
@@ -4094,6 +4098,29 @@ final void onEjbMethodEnd(String method_sig, Throwable th) {
...
+                                .put("EJBMethod", info.getMethod().getName())
+                                .put("CallerPrincipal", String.valueOf(info.getCallerPrincipal()))
+                                .put("TX-ID", String.valueOf(info.getTransactionId()))
+                                .build());
+                    } else {
+                        _logger.log(Level.FINE, "Tracing is not enabled, skipping event {0}", eventName);
+                    }
+                } finally {
+                   span.end();
+                }
```
luiseufrasio: Unlike the JTA and `@Traced` paths (which add an event to `Span.current()`), this creates a new child span per method entry/exit just to carry a single event, then ends it.

The else branch is also effectively unreachable: a span freshly started via `startSpan()` is always `isRecording` (barring immediate sampler drop), so **"Tracing is not enabled"** never fires usefully.

Note: the `constructEjbMethodSpanLog()` / `addSpanLog()` lines just below are not new — they are the pre-existing legacy request-tracing pipeline. 
Running both pipelines may be intentional during migration; the real problem is the ephemeral span, not addSpanLog.

**Possible Fix**: add the `event` to `Span.current()` (consistent with `JTA/@Traced`) and remove the dead `else`; only create a new `span` if none is active.

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+        } catch (IOException | ClassNotFoundException e) {
+            throw new ForwardRequest(e.getMessage(), serverRequestInfo);
```
luiseufrasio, line 127: On deserialization failure, log at FINE and return (proceed without trace context) — instead of throw ForwardRequest.

WDYT?

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopInterceptorFactory.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS HEADER.
+ *
+ * Copyright (c) 2020-2026 Payara Foundation and/or its affiliates. All rights reserved.
```
luiseufrasio, line 4: only 2026

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopInterceptorFactory.java
```diff
@@ -0,0 +1,119 @@
...
+                try {
+                    clientRequestInterceptor = new OpenTelemetryIiopClientInterceptor();
+                } catch (NullPointerException nullPointerException) {
+                    logger.log(Level.WARNING, "Could not create OpenTracing IIOP Client Interceptor - Remote EJBs will not be traced");
+                    return null;
+                }
+            }
+        }
+
+        return clientRequestInterceptor;
```
luiseufrasio: Check-then-act with no synchronization on instance fields of a `@Singleton`. ORB initialization is usually serial, so practical impact is low, but under concurrent init two instances could be created and one discarded — and the ThreadLocals (currentSpan/currentScope) of the actually-registered interceptor could diverge, leaking spans. NullPointerException is also caught as flow control (anti-pattern).

Note: the catch (`NullPointerException`) blocks are dead code — the client constructor is no-arg (extends LocalObject) and the server constructor only assigns a field, so neither can throw NPE, and attemptCreation() already guarantees openTracingService != null before use. They can be removed entirely.

Fix: synchronize both create*RequestInterceptor() methods (they run only during ORB init — a cold path — so method-level synchronized is cheap) and drop the dead NPE catch. attemptCreation() is called only from these methods, so it is covered by the same lock.

```
@Override
public synchronized ClientRequestInterceptor createClientRequestInterceptor(ORBInitInfo info, Codec codec) {
    if (clientRequestInterceptor == null && attemptCreation()) {
        clientRequestInterceptor = new OpenTelemetryIiopClientInterceptor();
    }
    return clientRequestInterceptor;
}

@Override
public synchronized ServerRequestInterceptor createServerRequestInterceptor(ORBInitInfo info, Codec codec) {
    if (serverRequestInterceptor == null && attemptCreation()) {
        serverRequestInterceptor = new OpenTelemetryIiopServerInterceptor(openTracingService);
    }
    return serverRequestInterceptor;
}

private boolean attemptCreation() {
    if (serviceLocator == null) {
        serviceLocator = Globals.getStaticBaseServiceLocator();
        if (serviceLocator == null) {
            return false;
        }
    }
    if (openTracingService == null) {
        openTracingService = serviceLocator.getService(OpenTracingService.class);
    }
    return openTracingService != null;
}
```

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
```diff
@@ -1719,6 +1746,44 @@ public JavaEETransaction createImportedTransaction(TransactionInternal jtsTx)
...
 
+    private void addJtaEventTraceLog(RequestTraceSpanLog spanLog, JavaEETransaction tx) {
+        Span span = Span.current();
+        if (span.isRecording()) {
+            AttributesBuilder attrsBuilder = Attributes.builder();
+            spanLog.getLogEntries().forEach(attrsBuilder::put);
+            String eventName = spanLog.getLogEntries().getOrDefault("logEvent", "jtaTransactionEvent");
+            span.addEvent(eventName, attrsBuilder.build(), spanLog.getTimeMillis(), TimeUnit.MILLISECONDS);
+            // Add transaction ID as baggage item
+            if (tx != null) {
```
luiseufrasio, line 1758: `tx != null` is checked three times (line 1757 in the if; lines 1769 and 1775 in the else — the one at 1775 is dead). The TX-ID attribution block is duplicated across both branches, and the comment "Add transaction ID as baggage item" is wrong: the code sets a span attribute, not baggage. 

Extract a helper `setTransactionAttribute(Span, JavaEETransaction);` drop the dead check and the misleading comment.

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+            @Override
+            public Iterable<String> keys(HashMap<String, String> carrier) {
+                return carrier.keySet();
+            }
+
+            @Override
+            public String get(HashMap<String, String> carrier, String key) {
+                return carrier.get(key);
+            }
+        };
```
luiseufrasio: The stateless anonymous `TextMapGetter` is allocated on every receive_request (RMI hot path). 

Promote it to a `private static final TextMapGetter<Map<String, String>>` constant to remove GC pressure.

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
```diff
@@ -322,6 +335,50 @@ public ContextHandle setup(ContextHandle contextHandle) throws IllegalStateExcep
                 Collections.EMPTY_LIST, restorers);
     }
 
+    private void startConcurrentContextSpan(ComponentInvocation invocation, InvocationContext handle) {
```
luiseufrasio, line 338: `startConcurrentContextSpan` calls `Globals.getDefaultBaseServiceLocator().getService(InvocationManager.class)` on every dispatch, ignoring the already-injected `this.invocationManager` field (`line 97 / 143`). 

Use the cached field.

luiseufrasio (changes requested):

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopInterceptorFactory.java
breakponchito: nop this was a component previously used, we renamed

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
breakponchito: yes this is correct, ,the span will close until all the invocations are processed

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
breakponchito: yes because it is possible that when from the ejb scope this span will be related to a transaction and the span is created previously. That us why we renamed here, the other case is when no transaction is related and then the span is named with the operationName

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/TracedInterceptor.java
breakponchito: done

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
breakponchito: If we close the scope automatically the span is not promoted, that is why we need to only mark the span as current without close the scope, the span will be close more latter when evaluating the @Trace annotation within the interceptor TracedInterceptor

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
breakponchito: I fixed by returning old object with a rename

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
breakponchito: already done

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/BaseContainer.java
breakponchito: Already sone, I changed the logic to identify if we have a current span add the event, in other case we will create a new span to report the event and then close

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
breakponchito: the old implementation did in this way I prefer to maintain previous logic

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopInterceptorFactory.java
breakponchito: done

appserver/transaction/jta/src/main/java/com/sun/enterprise/transaction/JavaEETransactionManagerSimplified.java
breakponchito: done

appserver/payara-appserver-modules/microprofile/telemetry/src/main/java/fish/payara/microprofile/telemetry/tracing/ejb/iiop/OpenTelemetryIiopServerInterceptor.java
breakponchito: done

appserver/concurrent/concurrent-impl/src/main/java/org/glassfish/concurrent/runtime/ContextSetupProviderImpl.java
breakponchito: Trying to use this reference, done

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

luiseufrasio (approved):
LGTM!

PR #8266, opened 2026-07-01
FISH-9740 Implement MicroProfile 7.1 Support

luiseufrasio:
## Description
This PR brings MicroProfile 7.1 support to Payara Platform. It upgrades all MicroProfile spec dependencies to their 7.1-compatible versions and implements the necessary changes to pass the MicroProfile 7.1 TCK suite.

Key changes included:

- **FISH-13752** – Update MicroProfile Release version to 7.1
- **FISH-9744** – Implement MicroProfile REST Client 4.0
- **FISH-13069** – Implement JAX-WS telemetry tracing with OpenTelemetry
- **FISH-13662** – Enable MicroProfile OpenTelemetry JVM metrics and tracing/logging TCK tests
- **FISH-13751** – Reimplement EJB tracing with OpenTelemetry (IIOP client/server traces, TX traces)
- **FISH-12815 / FISH-12816** – Add OpenTelemetry SDK runtime module for server and admin GUI
- **FISH-13392** – Fix FaultTolerance NullPointerException when casting context
- **FISH-13519** – Enable Fault Tolerance tests for pipeline
- **FISH-12029** – Reactivate OpenAPI TCK in Jenkins
- **FISH-13739** – Remove snakeyaml dependency from opentelemetry-repackaged; fix telemetry-bootstrap version

### Blockers
None.

### New tests
New and updated TCK tests for MicroProfile OpenTelemetry (JVM metrics, tracing, logging), REST Client 4.0, EJB tracing, and JAX-WS tracing.

### Testing Performed
The following MicroProfile TCK suites were executed and passed via the Jenkins pipeline — results available at https://jenkins.payara.fish/job/TCKs/job/MP-TCKs/12363/parameters/:

- Config
- Health
- JWT-Auth
- Rest-Client
- Metrics
- Fault-Tolerance
- OpenAPI
- OpenTelemetry-JVM-Metrics-Runtime
- OpenTelemetry-JVM-Metrics-Application
- OpenTelemetry-Tracing-Logging

### Testing Environment
Ubuntu 20.04, Azul JDK 21, Maven 3.9.11

## Documentation
No additional documentation PR at this time.

## Notes for Reviewers
This is a long-running feature branch (~161 commits) representing the full MicroProfile 7.1 upgrade effort. It was kept in sync with `main` (PR #8253) before this final merge PR. Reviewers may want to focus on the OpenTelemetry and EJB tracing changes as the most significant new functionality.

PR #8262, opened 2026-06-30
FISH-13838 FISH-13839 FISH-13916 Improved runtime and application telemetry bootstrap

pdudits:
## Description

* Create Init run level configuration that is not dependent on current application and only depends on services available at that runlevel.
* Make all injections in MP Config provider explicit instead of relying on ServiceLocator lookup. 
* This resulted into move of ConfigProviderResolverImpl to Startup run level and we enforce early start with a Rank.
* Removed 2017-era migration guards in config injections
* Added explicit modification interface and service that was used by the CLI commands
* Password alias source now auto-refreshes on application deployment

### Extension to OpenTelemetry bootstrap
* Remove more references to OpenTracingService and deprecate it
* Rename the service to `OpenTelemetryBootstrap`
* Pass all `otel` prefixed microprofile properties to SDK autoconfigure
* Enable in-server customization via HK2 services, such as bridge from OTEL Tracing to Request Tracing Service
* centralizes SDK setup for both runtime and app-level SDK instances
* adds new resource attributes for payara-specific attributes

### ASadmin extensions
* monitoring service has new attribute `openTelemetryEnabled`
* added the options to `set-monitoring-service-configuration`, `get-monitoring-service-configuration`, `enable-monitoring`
* changing the option triggers unprocessed change (requests a restart)

## BREAKING CHANGES

Payara specific OpenTracing features, especially those reliant on OpenTracing `@Traced` annotations are not supported - JAXWS traces and Remote EJB traces and Transaction ID in propagated with the trace. These features will be implemented later with design more native to OpenTelemetry.

EclipseLink Moxy has been removed, as the shift in classloaders made it the default JAXB provider despite not being widely used for this purpose. In order to use Moxy in your application update the dependencies:

```
<dependency>
    <groupId>org.glassfish.jersey.media</groupId>
    <artifactId>jersey-media-moxy</artifactId>
    <scope>compile</scope>
</dependency>
<dependency>
    <groupId>org.eclipse.persistence</groupId>
    <artifactId>org.eclipse.persistence.core</artifactId>
    <scope>provided</scope>
</dependency>
```

### New Tests
* `payara-samples/opentelemetry` reenabled
* updated opentelemetry sample to new semantic conventions api
* new tests for new asadmin commands
* added tests for MP Config related asadmin commands

### Testing Performed

MP Config TCK
MP Telemetry TCK

### Testing Environment

```
Java version: 21.0.8, vendor: Eclipse Adoptium
[...]

Issue #2260, opened 2018-01-15
Enhance documentation regarding JDBCRealm and Digest auth /CUSTCOM-68


peraxel:
# Description #
----------

Most implementations of HTTP Digest auth support storing passwords as MD5(username:realm:password) a.k.a. HA1, but Payara documentation does not give any hints as how to configure the realm to support this format. I found out that there is an undocumented setting to get support this. Set encoding->HASHED.
[Stackoverflow](https://stackoverflow.com/questions/48247919/how-to-configure-payara-jdbcrealm-jdbcdigestrealm-to-store-passwords-in-httpd-st)

Please include this in documentation and/or admin console hints.

michaelranaldo:
Hi @peraxel ,

Quite right, this could be better documented within the documentation, the admin console and the asadmin command help, where it doesn't seem to be mentioned at all.

Thanks for raising this, created internal JIRA PAYARA-2397

rdebusscher:
This is now tracked as `CUSTCOM-68`

AlanRoth:
Hi, due to how long ago this issue was raised we have decided to close the issue immediately, and not consider the implementation of the fix/improvement that was requested. Please understand that this decision was taken into consideration with the resources that we have available at the moment. In case of having reported a bug, if the issue is still pressing to you, feel free to verify if it’s applicable in the current release of Payara Community edition, and proceed to raise a new issue with details of the test reproducer. Many thanks for your understanding.

PR #2213, opened 2017-12-15
PAYARA-2340 Patched Soteria making some utilities public

arjantijms:
PAYARA-2340 Also fixed some compile warnings

PR #8272, opened 2026-07-02
FISH-9740 Expose @Traced in public API and remove internal-module usage from MicroProfile tracing samples

luiseufrasio:
## Description
Follow-up to #8266 addressing @Pandrex247's post-merge review of the `payara-samples` tracing modules.

The `jaxws-tracing` and `remote-ejb-tracing` samples were depending on the **internal** server module `fish.payara.server.internal.payara-appserver-modules:microprofile-telemetry`. Samples (and user code) must only depend on public API artifacts.

The reason they reached into the internal module is Payara's `@Traced` annotation, which lived only inside that module. This PR:

- **Exposes `@Traced` publicly**: relocates the annotation to `payara-api` under a new dedicated package `fish.payara.microprofile.telemetry.tracing.api`. A dedicated sub-package (rather than the original FQN) is used deliberately to avoid an **OSGi split package** — the `microprofile-telemetry` bundle keeps its interceptor/CDI-extension classes in `fish.payara.microprofile.telemetry.tracing`, so putting `@Traced` in the same package name in a second bundle would have split that package across two bundles and broken resolution at runtime.
- **Updates the internal module** to depend on `payara-api` and import the relocated annotation (`TracedInterceptor`, `TelemetryCdiExtension`, `JaxWsContainerRequestTelemetryTracingFilter`).
- **Fixes the two samples** to use public artifacts only: `payara-api` for `@Traced`, and the public `io.opentelemetry:opentelemetry-sdk` for `ReadableSpan` (previously pulled transitively through the internal module).
- **Cleans up `samples/pom.xml`**: removes the now-stale `<!-- disabled until... -->` comment above the enabled `jaxws-tracing` module, and documents the deliberately-disabled `microprofile-rc-ft` and `opentelemetry` samples against their tracking issue **FISH-13013** (so they no longer read as incomplete work).

### Blockers
None. Depends on the MicroProfile 7.1 merge (#8266), already in `main`.

### New tests
No new tests. Behaviour of the existing `jaxws-tracing` and `remote-ejb-tracing` samples is unchanged — only the annotation's coordinates and the samples' dependencies change.

### Testing Performed
- `payara-api` builds with the relocated annotation.
- `microprofile-telemetry` module compiles against the public `@Traced`.
- Full distribution build + MicroProfile TCK + `payara-samples` to be validated by the Jenkins pipeline (the reference MP 7.1 green run for this line is https://jenkins.payara.fish/job/TCKs/job/MP-TCKs/12363/parameters/).

### Testing Environment
Azul Zulu JDK 21, Maven 3.9.9.

## Documentation
No additional documentation PR.

## Notes for Reviewers
@Pandrex247 — this addresses all five review comments left on #8266. The key reviewer decision is the **package name for `@Traced`**: it is intentionally *not* kept at the original FQN, because a same-name package in both `payara-api` and the `microprofile-telemetry` bundle would be an OSGi split package. If you'd prefer `@Traced` to keep its exact original FQN, the alternative is moving the interceptor implementation classes into a `.impl` sub-package instead — happy to switch approaches.


luiseufrasio:
## ✅ CI validation — MicroProfile 7.1 TCK green on this branch

Built this branch and ran the MicroProfile 7.1 TCK against it via the Payara Jenkins pipeline. Result matches the pre-fix baseline exactly — the `@Traced` relocation into `payara-api` does not regress MicroProfile conformance.

| Run | Payara build | Result | Tests |
|---|---|---|---|
| Baseline (main, pre-fix) | [Build #2213](https://jenkins.payara.fish/job/Build/job/Build/2213/) | SUCCESS | 1691 total · 0 fail · 15 skip — [MP-TCKs #12363](https://jenkins.payara.fish/job/TCKs/job/MP-TCKs/12363/) |
| **This branch** (`FISH-9740-samples-internal-modules-fix`) | [**Build #2260**](https://jenkins.payara.fish/job/Build/job/Build/2260/) | **SUCCESS** | **1691 total · 0 fail · 15 skip** — [**MP-TCKs #12660**](https://jenkins.payara.fish/job/TCKs/job/MP-TCKs/12660/) |

**Build params:** `repoOrg=luiseufrasio`, branch `FISH-9740-samples-internal-modules-fix`, `stream=Community`, `jdkVer=zulu-21`.
**TCK params:** `payaraBuildNumber=2260`, `testBranchCommitOrTag=microprofile-7.1`, `distribution=full`, `jdkVer=zulu-21`, suites = `Config Health JWT-Auth Rest-Client Metrics Fault-Tolerance OpenAPI OpenTelemetry-JVM-Metrics-Runtime OpenTelemetry-JVM-Metrics-Application OpenTelemetry-Tracing-Logging`.

Note: the `payara-samples` tracing modules changed here (`jaxws-tracing`, `remote-ejb-tracing`) are exercised by the samples CI, not the MP-TCK suite; this run confirms the `microprofile-telemetry` module change compiles and remains TCK-conformant.

appserver/tests/payara-samples/samples/pom.xml
```diff
@@ -80,6 +79,7 @@
         <module>ejb-http-remoting</module>
         <module>remote-ejb-tracing</module>
         <module>rolesallowed-unprotected-methods</module>
+        <!-- Disabled under FISH-13013 - re-enable once the MicroProfile REST Client Fault Tolerance sample is stabilised -->
         <!--<module>microprofile-rc-ft</module>-->
```
breakponchito: enable again this module for test, I tested locally and this will work without any problem

appserver/tests/payara-samples/samples/pom.xml
```diff
@@ -88,6 +88,7 @@
         <module>microprofile-rest-client</module>
         <module>concurrency</module>
         <module>reproducers</module>
+        <!-- Disabled under FISH-13013 - re-enable once the OpenTelemetry sample is stabilised -->
         <!--<module>opentelemetry</module>-->
```
breakponchito: still with issues if we enable again this module for tests, investigating

appserver/tests/payara-samples/samples/pom.xml
luiseufrasio: Done in 5c4cada0fa0 - the \microprofile-rc-ft\ module is enabled again and the FISH-13013 comment above it has been removed. Thanks for verifying it locally!

appserver/tests/payara-samples/samples/pom.xml
breakponchito: I re-enable the tests

breakponchito:
Jenkins test please

appserver/tests/payara-samples/samples/remote-ejb-tracing/pom.xml
```diff
@@ -101,15 +101,14 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>fish.payara.server.internal.payara-appserver-modules</groupId>
-            <artifactId>microprofile-telemetry</artifactId>
-            <version>${project.version}</version>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+            <version>${opentelemetry.version}</version>
```
Pandrex247: Defined by the BOM already
```suggestion
```

Pandrex247 (changes requested):

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -105,7 +166,13 @@
...
             <groupId>io.opentelemetry.instrumentation</groupId>
             <artifactId>opentelemetry-instrumentation-annotations</artifactId>
-            <version>${opentelemetry.version}</version>
+            <version>${opentelemetry-instrumentation-annotations.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry.semconv</groupId>
+            <artifactId>opentelemetry-semconv</artifactId>
+            <version>${opentelemetry-semconv.version}</version>
```
Pandrex247: Already defined via BOM
```suggestion
```

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -105,7 +166,13 @@
         <dependency>
             <groupId>io.opentelemetry.instrumentation</groupId>
             <artifactId>opentelemetry-instrumentation-annotations</artifactId>
-            <version>${opentelemetry.version}</version>
+            <version>${opentelemetry-instrumentation-annotations.version}</version>
```
Pandrex247: Already defined via BOM
```suggestion
```

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -52,7 +52,68 @@
     <name>Payara Samples - OpenTelemetry</name>
     <artifactId>opentelemetry</artifactId>
     <packaging>war</packaging>
+    
+    <properties>
+        <script.extension></script.extension>
+        <payara.asadmin>${payara.home}${file.separator}glassfish${file.separator}bin${file.separator}asadmin${script.extension}</payara.asadmin>
+        <payara.domain.name>domain1</payara.domain.name>
+    </properties>
```
Pandrex247: Would be better to define these in the payara-samples/pom.xml, and not to have a script.extension property

See my review here: https://github.com/payara/Payara-Enterprise/pull/2702/files#diff-a32acb967ab7571247b6c00979b3d8f4dcdcd5a163f8b7133b26b66cc67b343f

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -52,7 +52,68 @@
...
+                        <configuration>
+                            <executable>${payara.asadmin}</executable>
+                            <arguments>
+                                <argument>stop-domain</argument>
+                                <argument>--kill=true</argument>
+                                <argument>${payara.domain.name}</argument>
+                            </arguments>
+                        </configuration>
+                    </execution>
+                </executions>
```
Pandrex247, line 97: Should have skip properties for server remote vs. server managed vs. micro, and be using antrun

See https://github.com/payara/Payara-Enterprise/pull/2702/files#diff-8ebffb02404f26a1b4e23ce4bf31d3ffedebe2b3c7497617451fb53e0501df8dR104-R179 for an example to align with

Pandrex247 (changes requested):

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
Pandrex247: Would be good to make this change for all of the other opentelemetry properties that are redundant in this pom as well

pdudits:
Why not just use Open Telemetry's `@WithSpan` for the positive message? The places where the value of `trace=false` is used are already controllable via standard MP Config property setting.

Since this wasn't a public API, we don't need to make it public now, just drop the support instead.

appserver/tests/payara-samples/samples/remote-ejb-tracing/pom.xml
breakponchito: already done

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
breakponchito: already done

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
breakponchito: yes applied to all telemetry dependencies

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
pdudits: Also good news, #8262 brings in a switch in monitoring service (but requires a restart anyway).

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
breakponchito: I saw that the execution for sample-tests use the payara-server-remote. I added the configuration to interact with this

pdudits: why doesn't remote ejb tracing test clean up after itself instead?

astoica1 (dismissed):
having a look at this PR..

appserver/tests/payara-samples/pom.xml
```diff
@@ -67,6 +69,9 @@
 
         <!-- Exclude unstable tests by default-->
         <excludedGroups>fish.payara.samples.Unstable</excludedGroups>
+        <payara.home>..${file.separator}..${file.separator}..${file.separator}..${file.separator}..${file.separator}payara7</payara.home>
+        <payara.asadmin>${payara.home}${file.separator}glassfish${file.separator}bin${file.separator}asadmin</payara.asadmin>
+        <payara.domain.name>test-domain</payara.domain.name>
```
Pandrex247, line 74: Don't use test-domain
This is a legacy from many years ago which I'm trying to kill
All it does is catch people out.

```suggestion
        <payara.domain.name>domain1</payara.domain.name>
```

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -52,6 +52,52 @@
...
+                    <!-- Delete system property used to pass remote ejb tracing tests not needed for opentelemetry samples -->
+                    <execution>
+                        <id>delete-system-property</id>
+                        <phase>pre-integration-test</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skipConfig}</skip>
+                            <target if="payara.server.remote">
```
Pandrex247, line 72: This will be relevant to payara-server-managed as well

```suggestion
                            <target unless="payara.micro.managed">
```

appserver/tests/payara-samples/samples/opentelemetry/pom.xml
```diff
@@ -52,6 +52,52 @@
...
+                                <exec executable="${payara.asadmin}">
+                                    <arg value="delete-system-property"/>
+                                    <arg value="otel.sdk.disabled"/>
```
Pandrex247, line 75: The other test should be cleaning up after itself, not this one.

appserver/tests/payara-samples/samples/opentelemetry/src/main/java/fish/payara/samples/otel/manual/ManuallyTracingServlet.java
```diff
@@ -67,8 +68,8 @@ public class ManuallyTracingServlet extends HttpServlet {
     @Override
     protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
         try(Scope scope = span.makeCurrent()) {
-            span.setAttribute(SemanticAttributes.HTTP_METHOD, "GET");
-            span.setAttribute(SemanticAttributes.HTTP_URL, req.getRequestURI());
+            span.setAttribute(HttpAttributes.HTTP_REQUEST_METHOD, "GET");
+            span.setAttribute(UrlAttributes.URL_FULL, req.getRequestURI());
```
Pandrex247, line 72: See Patrik's PR: these will conflict.

https://github.com/payara/Payara/pull/8262/files#diff-116700daf3f7a1b8640d7524756104a3a12be811482e540ab63891052aa65229R70-R71

I don't know ottomh which is the correct way to do this, but it will need resolving either this way or Patrik's way.

Pandrex247 (changes requested):

appserver/tests/payara-samples/pom.xml
breakponchito: from my review of the payara-samples execution, this is very coupled with the payara-server-remote and the Jenkins configuration is creating an instance with test-domain, this is reused for all the samples

breakponchito:
@luiseufrasio @Pandrex247 @pdudits closing this PR and add remaining work on the following: https://github.com/payara/Payara/pull/8262

PR #7314, opened 2025-04-25
FISH-10049 FISH-10286 FISH-10615 Restores EAR fixes and WARlib performance enhancements

lprimak:
Restores functionality reverted from #7032 and #7097 and applies fixes from #7212

Fixes https://github.com/payara/Payara/issues/7201
Fixes https://github.com/payara/Payara/issues/7214
Fixes https://github.com/payara/Payara/issues/7215
Fixes https://github.com/payara/Payara/issues/7233

Supersedes #7212

PR #7902, opened 2026-01-26
FISH-12192 bugfix: initialize WebContainer before HK2 class scanning so Rest APIs…

lprimak:
… get loaded properly

fixes #7655

## Description
This is a bugfix.

This PR forces loading of WebContainer (if exists) prior to HK2 class scanning.
If this is not done, classes such as @RegisterRestClient that are not directly CDI beans (but are @Stereotypes) will not correctly register as valid CDI beans, and deployments will fail in postboot or when starting the server with applications already deployed.

PR #7820, opened 2025-12-12
FISH-12721 Moved InstalledLib (warlib) initialization into PostBoot, whe…

lprimak:
…re it's being installed earlier, and initializing WebContainer during application deployment to make sure it's initialized before PostBoot as well

## Description

This PR moves `InstalledLibrariesResolver` initialization into post-boot phase, because otherwise the libraries will not be registered in time for deployment of postboot applications.

Fixes #7806 
Supersedes #7808

PR #8275, opened 2026-07-04
FISH-13968 Jax-RS leak prevention also applies to Jax-RS client, not onl…

lprimak:
…y server

## Description

`weld/JaxRSJsonContextResolver.java` was previously restricted to server-only, but this issue also applies to jax-rs clients. This PR removes this restriction, and thus fixes Jax-RS clients that integrate with multiple applications in CDI.

## Notes for Reviewers
This is an enhancement to the bug fix in #7314 that applies to Jax-RS Clients.
Same ClassLoader leaks that was fixed in the above PR appears in Jax-RS clients, which this PR fixes.


lprimak:
@Pandrex247 @stian-sigvartsen I was just doing some work with #7902 and #7820 and discovered this. It's pretty simple just removes a few lines of code.

Thank you

PR #8125, opened 2026-04-30
FISH-13126 Deprecate Old Deployment Descriptors and Add New Payara Descriptors

Pandrex247:
## Description
Adds new Payara descriptors (including handling where not done before) for:
* payara-web.xml
* payara-acc.xml
* payara-ejb-jar.xml
* payara-application.xml
* payara-application-client.xml

With the exception of the payara-acc.xml (as it doesn't belong to a corresponding specification), these are meant to be versioned like so: `${productName}-${original-file-name}_${originalSpecMajor}_${originalSpecMinor}-${fileIterationVersionStartingAt0}`

For example, for Payara 7 which supports Jakarta EE11, which includes Servlet 6.1, the filename of the companion DTD would be “payara-web-app_6_1-0.dtd”, with any future iterations of this file incrementing the file number: "payara-web-app_6_1-1.dtd", "payara-web-app_6_1-2.dtd", "payara-web-app_6_1-3.dtd" etc.

We're adding support for our own payara-branded versions of these files so that they don't ever conflict with the GlassFish ones.

Deprecates the old ones, and also includes some clean up where appropriate.

This PR also currently drops the incorrectly implemented functionality of defining managed executors, managed scheduled executors, managed thread factories, and context services via the ejb-jar.xml. This was and always has been incorrectly implemented, where we added this functionality directly to the Jakarta-defined EJB jar file. As this is a vendor-specific extension to the specification, it should have either been in a Payara-forked equivalent ejb-jar.xml, or in a companion payara-ejb-jar.xml file. As far as I can tell this extended functionality was never documented.
The intent will be, time and possibility allowing, to reimplement this in our companion descriptor files.

### Blockers
None

### New tests
Converted some unit and Samples tests to use the new descriptors.
Added a simple functional test that utilises payara-application.xml to set the context root

### Testing Performed
Jenkins test please

### Testing Environment
Windows 11, Maven 3.9.15, Zulu JDK 21.0.11

## Documentation
https://github.com/payara/Payara-Documentation/pull/786

## Notes for Reviewers
None

PR #8278, opened 2026-07-06
FISH-13493 FISH-13969 FISH-13982 Forward-Port Payara 6 Deployment Descriptors

Pandrex247:
## Description
Follow-up to https://github.com/payara/Payara/pull/8125
In a similar vein to what was added in that PR, we have added Payara 6 deployment descriptors.

This PR "forward-ports" support for these older descriptors.

This PR also adjusts the raw.githubusercontent URLs in those descriptors to the new docs location in GitLab.

### Blockers
https://github.com/payara/Payara-Enterprise/pull/3084/

These PRs now block each other - they must be merged together

### New tests
Extended the payara-application-xml functional test to check the Payara 6 application XML is still valid

### Testing Performed
Jenkins test please

### Testing Environment
Jenkins

## Documentation
https://gitlab.azulsystems.com/docs/payara/-/merge_requests/1

## Notes for Reviewers
See comments

### DIFF
--- a/appserver/admin/gf_template/src/main/resources/config/payara-acc.xml
+++ b/appserver/admin/gf_template/src/main/resources/config/payara-acc.xml
@@ -49,7 +49,7 @@
    - Property security.config in message-security-config
 -->
 
-<!DOCTYPE client-container PUBLIC "-//Payara.fish//DTD Payara Application Server 7 Application Client Container Revision 1//EN" "https://gitlab.azulsystems.com/docs/payara/-/raw/main-7/content_shared/appendix/schemas/payara7-application-client-container_1_4-1.dtd">
+<!DOCTYPE client-container PUBLIC "-//Payara.fish//DTD Payara Application Server 7 Application Client Container Revision 1//EN" "https://raw.githubusercontent.com/payara/Payara/refs/heads/main/appserver/deployment/dtds/src/main/resources/glassfish/lib/dtds/payara7-application-client-container_1_4-1.dtd">
 
 <client-container>
   <target-server name="%%%SERVER_NAME%%%" address="localhost" port="%%%ORB_LISTENER1_PORT%%%"/>
--- a/appserver/admin/gf_template/src/main/resources/config/payara-acc.xml
+++ b/appserver/admin/gf_template/src/main/resources/config/payara-acc.xml
@@ -49,7 +49,7 @@
    - Property security.config in message-security-config
 -->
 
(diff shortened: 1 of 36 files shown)


Pandrex247:
Wrinkle to resolve: the payara-ejb-jar.xml has various elements removed e.g. CMP was removed in Payara 7.
By the existing versioning system, this file should technically have existed on Payara 6 first (since EJB 4.0 was a Jakarta EE 10 specification). CMP functionality still exists on Payara 6.

The current versioning system (which we inherited from GlassFish) doesn't account for us supporting multiple major versions of Payara, with differing feature sets, that implement the same version of a specification.

We either need to:
* Change the versioning system of these files to allow the above situation
* Reintroduce some level of support for the removed CMP and other entries to the DTD (and presumably alter the implementation in Payara 7 to throw a warning/error if used)
* Remove CMP from Payara 6
* Accept that the new EJB deployment descriptor doesn’t support all of the functionality that the server itself supports in Payara 6 and instruct users to keep using the old glassfish-ejb-jar descriptor (which continued support of can cause GlassFish migration issues).
* Accept that we have two DTDs with the same file name/version for each major version of Payara version (messy!). Namely the version hosted on the `main-6` branch of the docs (and bundled in Payara 6) would be different to the version hosted on the main-7 branch
* Something Else™

Pandrex247:
We are choosing option 1, and adapting the versioning system to this:

`${productName}${payaraMajor}-${original-file-name}_${payaraMajor}-${originalSpecMajor}_${originalSpecMinor}-${fileIterationVersionStartingAt0}`: e.g. `payara6-ejb-jar_4_0-0` & `payara7-ejb-jar_4_0-0`

Additionally, the ID String needed adjusting, since we didn't include a minor number (since Community and Enterprise have different versioning systems).

As such, we are adopting a system of adding a "Patch x" to the String name where appropriate:

`payara7-web-app_6_1-0.dtd`  = `-//Payara.fish//DTD Payara Server 7 Servlet 6.1//EN`
`payara7-web-app_6_1-1.dtd`  = `-//Payara.fish//DTD Payara Server 7 Servlet 6.1 Revision 1//EN`

Pandrex247:
Additionally, we are applying this versioning system to the payara-acc DTDs.
These had their own versioning system that was essentially semantic versioning, but there's a chance we'd hit similar problems so we're adopting a blanket solution.

The one exception is the `payara-application.xml` and `payara-application-client.xml` DTDs: these don't mention a spec minor number. To keep it aligned with the upstream `application.xml` and `application-client.xml` DTDs which also don't mention a minor number, we're also not going to.

kalinchan (approved):

PR #8287, opened 2026-07-10
FISH-10884 Add regex filtering to access log URIs

Viii3:
## Description

- Resolves FISH-10884
  - Adds regex filtering to request URIs. Filtered URIs are not in the access logs.

### Testing Performed

- Started the server.
- Navigated to Configurations -> server-config -> HTTP Service
- Enabled access logging.
- Moved around the pages on the admin console.
- Stopped the server (to force the access logs to update).
- Observed the URIs in the access logs.
- Started the server.
- Set the access log filter to `\/health.*`.
- Moved around the pages on the admin console, including HealthCheck.
- Stopped the server and observed the URIs logged did not include HealthCheck.

### Testing Environment

Maven version: 3.9.14
Java version: 21.0.10, vendor: Azul Systems, Inc.


appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
```diff
@@ -145,6 +146,8 @@ public final class PEAccessLogValve
      */
     private String prefix = "";
 
+    private String filter = "";
```
svendiedrichsen: Shouldn't this be a Pattern compiled from the filter string? This would avoid a performance penalty of permanently compiling the pattern when using String.matches() later on and it allows to handle the PatternSyntaxException if an illegal pattern is provided.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
```diff
@@ -433,7 +435,7 @@ public void setPrefix(String p) {
     }
 
     public void setFilter(String regex) {
-        filter = Objects.requireNonNullElse(regex, "");
+        filter = Objects.requireNonNullElse(Pattern.compile(regex), null);
```
svendiedrichsen: This may throw a PatternSyntaxException
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/regex/Pattern.html#compile(java.lang.String)

And btw. this is a funny way of using the requireNonNullElse method. The first argument cannot be NULL but throw an exception and the second is NULL.

You could probably rather not use the Objects.requireNonNullElse method but guard against NULL or empty regex, create the Pattern catching the PatternSyntaxException and log it while setting the filter to NULL.

appserver/web/web-glue/src/main/java/com/sun/enterprise/web/PEAccessLogValve.java
```diff
@@ -429,6 +435,19 @@ public void setPrefix(String p) {
         }
     }
 
+    public void setFilter(String regex) {
+        if (regex == null) {
```
svendiedrichsen: Are empty strings valid patterns? Maybe add `|| regex.trim().isEmpty()`.

stian-sigvartsen (changes requested):
@Viii3 as we just discussed it would be good to add some more things:
1) UI / asadmin CLI error reporting when RegEx pattern is invalid (admin should not have to check the server.log) 
2) Consider repeating fields to allow multiple patterns to be configured

appserver/admingui/web/src/main/resources/configuration/accessLog.jsf
```diff
@@ -104,7 +150,7 @@
         </sun:property>
 
         <sun:property id="filterProp" labelAlign="left" noWrap="#{true}" overlapLabel="#{false}" label="$resource{i18n_web.access.Filter}" helpText="$resource{i18n_web.access.FilterHelp}">
-            <sun:textField id="Filter" columns="$int{30}" maxLength="#{sessionScope.fieldLengths['maxLength.access.RotationSuffix']}" text="#{pageSession.valueMap2['filter']}" />
```
stian-sigvartsen: Incorrectly reusing `RotationSuffix` for maxLength value

stian-sigvartsen (changes requested):
This is a lot better than the first iteration I think 👍
I have sent you a PR with some additional commits that reduce the overall change. See Viii3/Payara/pull/1 

(if you like it you can merge to your PR via the GitHub merge button) 

Also note the comment I made about maxLength.

stian-sigvartsen (approved):
Looks good @Viii3 . Please review CI test failrues

lprimak (approved):

PR #8297, opened 2026-07-14
FISH-13997: Modernize OTEL propagation for async invocations

pdudits:
## Description
This is larger rewrite to fix corner cases when propagating over managed executors.

Once it started working it started to break MP telemetry TCK which does not expect an intermnediate parent for work submitted to managed workers. I therefore changed how concurrency emits telemetry:

- It will propagate the context to submitted jobs but not create span around it
- instead it will create a "waiting" span signifying the time the job spent in thread pool queue. It is child of parent span, but sibling of any span the child would create.

### Detailed changes

- Dependency on multiple internal services were reduced to two facades `InvocationFacade` and `MonitoringFacade` to improve readibility and testability
- remove ad-hoc service lookups over GlobalHabitat
- Move all opentelemetry handling into separate ThreadContextProvider
- clear local context after invocation
- change names to be conformant to OpenTelemetry conventions (the new span name is `concurrent/<executorName> wait`)
- all traces are now created with `thread.id` and `thread.name` attributes
- Fixed bug in OpenTelemetryBootstrap where global SDK customizers would be ignored
- Fixed bug in OpenTelemetryService where getting a tracer would mark application as having OTEL telemetry when it actually didn't request one.
- Reduced allocations in OpenTelemetryService by omitting Optional
- 
### New tests

- `ContextSetupProviderTest` a unit test
- `ConcurrentContextPropagationIT` Arquillian test in samples

### Testing Performed

Requires payara/Microprofile-TCK-Runners#304

- Concurrency TCK
- Telemetry TCK


pdudits:
The TCK is failing, because Concurrency spans finally consistently work and introduce additional internal span.

I'll see if I can get any compromise solution, or set the asynchronous span conditionally.

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

breakponchito:
Jenkins test please

pdudits:
Jenkins test please

breakponchito (approved):
LGTM

PR #8301, opened 2026-07-17
OpenTelemetry rework for Microprofile 7.1

pdudits:
#  FISH-14011 / FISH-13842 / FISH-13995: OpenTelemetry instrumentation improvements

## Description

This PR delivers a broad set of improvements to Payara's OpenTelemetry instrumentation, targeting correctness, standards alignment, and performance. All changes are within the OpenTelemetry/MicroProfile Telemetry layer — no functional changes to other MicroProfile specs or application APIs (except where noted as breaking changes below).

### FISH-13842 — Bridge Fault Tolerance metrics to OpenTelemetry

Previously, MicroProfile Fault Tolerance metrics were recorded twice — once through MP Metrics and once through a separate `FaultToleranceTelemetryMetricsRecorder` that replicated the same values into OpenTelemetry instruments. This duplication was fragile and hard to maintain.

The new `otel/` package (`OtelMetricRegistry`, `OtelCounter`, `OtelHistogram`, `OtelGauge`, `OtelMetricsContext`) provides a dedicated OTel-aware bridge that satisfies the same `FaultToleranceMetrics` contract used by the FT engine. The FT engine now registers metrics exactly once; the bridge translates them to OTel instruments directly, without going through MP Metrics first.

> **See breaking change note below.**

### FISH-13995 — IIOP and JAX-WS context propagation

Addressed two related propagation gaps:

- **IIOP (EJB remote)**: `OpenTelemetryIiopClientInterceptor` and `OpenTelemetryIiopServerInterceptor` now correctly inject/extract W3C `traceparent`/`tracestate` headers through the IIOP service context. The `BaseContainer` now passes the OTel context through the EJB invocation chain. Defensive null-checks guard against the interceptors being called in environments without a Payara instance (e.g., standalone test clients).
- **JAX-WS**: Unified span creation for both servlet-hosted and EJB-hosted JAX-WS endpoints in `EjbWebServiceServlet`. Removed the duplicate span-management code that was previously duplicated between `JAXWSServlet` and the filter.
- New integration test suites: `IiopSpanIT` (remote EJB tracing) and `JaxWsSpanIT` (JAX-WS span propagation).

### FISH-13999 — Error attribute cleanup in JAX-RS / client filters

Aligned error attribute recording in `WithSpanMethodInterceptor`, `OpenTelemetryRequestEventListener`, and `JaxrsClientRequestTelemetryFilter` with OTel semantic conventions. Removed dead code paths in `RestSpanHelper` (renamed from `OpenTracingHelper`).

### FISH-14000 — Document non-standard units in FT OTel metrics

Added a code comment in `OtelMetricRegistry` explaining that Fault Tolerance OTel metric units deviate from standard OTel conventions because the MP Telemetry TCK mandates specific unit strings.

### FISH-14011 — Replace Payara-specific `@Traced` with standard `@WithSpan`

Removed the Payara-specific `@Traced` interceptor binding (a migration artifact from MicroProfile OpenTracing). Applications must now use the standard `@io.opentelemetry.instrumentation.annotations.WithSpan`. Added `WithSpanEnabledLookup` to cleanly support the `@WithSpan` enabled/disabled lifecycle. Fixed `WithSpanMethodInterceptor` to correctly handle CDI interceptor chains in JAX-WS invocations.

> **See breaking change note below.**

### FISH-14040 — OpenTelemetry SDK initialized only for explicitly enabled applications

Fixed a latent bug: calling `OpenTelemetryService.getCurrentTracer()` would eagerly initialise the SDK for an application even when OpenTelemetry had not been configured. Introducing `DeferredContext` stores context until target application is identified and matching Sdk can be provided.

Additionally, the Payara Request Tracing feature flag no longer implicitly enables OpenTelemetry. The two features are now independent.

> **See breaking change note below.**

---

### 1. `@Traced` annotation removed (FISH-14011)

**What changed:** `fish.payara.microprofile.telemetry.tracing.Traced` (including `TracedInterceptor`, `OpenTracingCdiUtils`, `TracedMethodFilter`) has been deleted. This was a Payara-specific migration shim carrying a `@Deprecated(forRemoval = true)` marker since it was introduced.

**Who is affected:** Applications that import or reference `fish.payara.microprofile.telemetry.tracing.Traced` directly (i.e., code that used the old Payara OpenTracing compatibility layer rather than the standard OTel annotation).

**Migration:** Replace `@fish.payara.microprofile.telemetry.tracing.Traced` with `@io.opentelemetry.instrumentation.annotations.WithSpan`. If the annotation was used with `value = false` to suppress tracing for a method, simply remove the annotation — methods are not traced by default. The `operationName` attribute maps to the `value` attribute of `@WithSpan`.

```java
// Before
@Traced(operationName = "myOperation")
public void myMethod() { ... }

// After
@WithSpan("myOperation")
public void myMethod() { ... }

// Before — suppressing trace (value = false)
@Traced(value = false)
public void myMethod() { ... }

// After — no annotation needed
public void myMethod() { ... }
```

---

### 2. Fault Tolerance dual-metrics recording removed (FISH-13842)

**What changed:** `FaultToleranceTelemetryMetricsRecorder` has been deleted. FT metrics are no longer recorded twice (once via MP Metrics, once via OTel). The new `OtelMetricRegistry` bridge records FT metrics directly as OTel instruments.

**Who is affected:** Anyone who relied on the `FaultToleranceTelemetryMetricsRecorder` SPI (internal, but possible in custom extensions), or who observed that FT metrics appeared in both MP Metrics and OTel outputs and built tooling around that behaviour. Metric names and attribute keys remain aligned with the MP Telemetry TCK; values should be equivalent.

**Migration:** No application code changes required — the change is internal to the Payara runtime. If you have custom tooling that reads FT metrics from MP Metrics *and* OTEL simultaneously, note that from this release onward, FT metrics are published only through the OTel pipeline.

---

### 3. Request Tracing no longer implicitly enables OpenTelemetry (FISH-14040)

**What changed:** Previously, enabling Payara's built-in Request Tracing feature (`set configs.config.server-config.request-tracing-service-configuration.enabled=true`) had a side effect of activating OpenTelemetry instrumentation for all applications. This coupling has been removed.

**Who is affected:** Users who relied on enabling Request Tracing as the mechanism to get OpenTelemetry spans exported, without explicitly enabling OpenTelemetry via the dedicated configuration or MicroProfile Config properties.

**Migration:** Enable OpenTelemetry explicitly for each application via the standard MicroProfile Telemetry mechanisms (e.g., `otel.sdk.disabled=false` in MicroProfile Config / `microprofile-config.properties`, or the Payara admin console OTel configuration). Request Tracing remains fully functional as an independent subsystem.

---

### Blockers

None identified.

### New tests

- `appserver/tests/payara-samples/samples/opentelemetry/src/test/java/fish/payara/samples/otel/concurrent/ConcurrentContextPropagationIT.java` — async context propagation across `ManagedExecutorService`
- `appserver/tests/payara-samples/samples/opentelemetry/src/test/java/fish/payara/samples/otel/OpenTelemetryMonitoringCommandsIT.java` — asadmin commands for enabling OTel
- `appserver/tests/payara-samples/samples/opentelemetry/src/test/java/fish/payara/samples/otel/spanname/WithSpanEnabledIT.java` — `@WithSpan` enabled/disabled behaviour
- `appserver/tests/payara-samples/samples/remote-ejb-tracing/src/test/java/fish/payara/samples/remote/ejb/tracing/IiopSpanIT.java` — IIOP remote EJB span propagation (new test module)
- `appserver/tests/payara-samples/samples/jaxws-tracing/src/test/java/fish/payara/samples/jaxws/test/otel/JaxWsSpanIT.java` — JAX-WS span creation and propagation
- `nucleus/payara-modules/opentelemetry-tracing-starter/src/test/java/fish/payara/opentracing/PropagationHelperTest.java` — client span tracking corner case

### Testing Performed

Integration tests in `payara-samples` covering:
- HTTP span naming (route template, async requests)
- `@WithSpan` annotation processing on CDI beans and JAX-WS endpoints
- OTel context propagation across `ManagedExecutorService` task boundaries
- Span propagation for remote EJB over IIOP
- Span propagation for JAX-WS (servlet and EJB hosting)
- Fault Tolerance OTel metric recording (via TCK)
- `PropagationHelper` unit tests

- Microprofile TCKs
- **Full Jakarta EE 11 TCK**

### Testing Environment

JDK 21, Maven 3.9, macOS

## Documentation

Documentation PR to be raised separately covering:
- Removal of `@Traced` and migration to `@WithSpan`
- Request Tracing / OpenTelemetry enablement decoupling

## Notes for Reviewers

- `FISH-14011` (`@Traced` removal) is the largest single commit by line count; the net effect is a significant reduction in complexity. The key new class is `WithSpanEnabledLookup`.
- The `serialVersionUID` in `ContextSetupProviderImpl` was bumped from `-1095988075917755802L` to `1L` to reflect the non-trivial structural change; same-JVM round-trips still work but cross-version deserialization is not supported.


breakponchito (approved):
LGTM

appserver/ejb/ejb-container/src/main/java/com/sun/ejb/containers/EjbEndpointFacadeImpl.java
```diff
@@ -102,6 +106,11 @@ public ComponentInvocation startInvocation() {
         // remaining preInvoke tasks : getContext, preInvokeTx, etc.
         invManager_.preInvoke(inv);
 
+        // Application invocation context is now on the stack: promote any deferred
+        // OTel context (e.g. from EjbWebServiceServlet for JAX-WS EJB endpoints)
+        // into a live SERVER span using the correct per-application tracer.
+        openTelemetryService.applyDeferredContext();
```
breakponchito, line 112: if the openTelemetryService reference is null, then here a possible NullPointerException will be thrown
```suggestion
        if (openTelemetryService != null) {
             openTelemetryService.applyDeferredContext();
        }
```

pdudits: OpenTelemetry would be null only in constellation where the service is part of server deployment. In such case this module would fail to run due to `ClassNotFoundException`. 

Only issue we tend to get with openTelemetry lookup is when the code is used outside of properly initialized server, such as embedded payara used as remote IIOP client.

PR #8302, opened 2026-07-20
FISH-13873, FISH-14012 Fix nucleus admin and quicklook tests for Payara 7 / JDK 21

astoica1:
## Summary

- **ClusterTest** (admin + quicklook): scan for stale instances before `delete-cluster`; switch to `list-instances` (brief format) with case-insensitive regex so `"Not Running"` is correctly matched; restructure `cleanUp()` to run the scan before `delete-cluster`
- **NucleusStartStopTest**: use `stop-domain --kill=true` in `setUp`/`tearDown` to prevent 60-second timeout leaving domain running and blocking the next run
- **MiscCommandsTest** (admin + quicklook): `stop-domain --kill=true`
- **JobManagerTest**: `stop-domain --kill=true`; fix `nucleusRoot` path (`glassfish/domains/…`); disable `deleteJobsFile` as a `@Test`; remove spurious extra `start-domain`
- **OSGiCommandsTest**: throw `SkipException` when osgi shell is not available instead of failing
- **ProgressStatusComplexTest**: relax flaky scope-count assertion to `>= 5`
- **JobsResourceTest**: throw `SkipException` on non-success HTTP status
- **JvmOptionsTest**: handle Payara 7 leafList entries as `Map<String,Object>` with `"jvmOption"` key
- **NetworkListenerTest**: use `assertNull` for absent SSL fields instead of comparing to `JsonObject.NULL.toString()`
- **SystemPropertiesTest**: delete created instance and config in a `finally` block to prevent orphan configs polluting subsequent tests
- **JceCipherSuitesTest**: remove `TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256` assertion (cipher removed in JDK 21)
- **RestTest**: fix node REST API parameter name (`id=` not `name=`) and delete URL

### Tests fixed / added
See above.

### How to run:  
From Payara/nucleus/tests, run `mvn test -Dnucleus.home=appserver/distributions/payara/target/stage/payara7`

Tested locally under Macos / Zulu 21


stian-sigvartsen (changes requested):
Looks great overall. We discussed offline that the OSGiCommandsTest shoud be disabled and a new ticket raised to fix the asadmin OSGI shell integration rather than introducing @SkipException.

Besides from that the only small improvement would be to remove `assertTrue(isSuccess(response));` from 
- JobsResourceTest#testJobsListing
- JobsResourceTest#testGetJob

With the changes, that is dead code.

stian-sigvartsen (approved):
LGTM @adistoica !

PR #8325, opened 2026-07-29
FISH-14132 web.xml configured realm name is ignored on Payara 7 for JAX-WS webservices

stian-sigvartsen:
## Description
The integration of Epicyro caused a regression to JAX-WS webservice authentication. More Specifically webservices written using the *servlet* approach as implemented by the `ServletEndpointTest` . 

### Testing Performed
Re-enabled the `ServletEndpointTest` and achieved a passing result.

## Notes for Reviewers
Note that this is a regression on Payara 7 only.

### DIFF
--- a/appserver/security/webservices.security/src/main/java/com/sun/enterprise/security/jauth/jaspic/provider/config/SoapAuthenticationService.java
+++ b/appserver/security/webservices.security/src/main/java/com/sun/enterprise/security/jauth/jaspic/provider/config/SoapAuthenticationService.java
@@ -135,14 +135,13 @@ public class SoapAuthenticationService extends BaseAuthenticationService {
     @Override
     protected CallbackHandler getCallbackHandler() {
         String realmName = null;
-        WebServiceEndpoint wse = (WebServiceEndpoint) map.get(PipeConstants.SERVICE_ENDPOINT);
-        if (wse != null) {
-            Application app = wse.getBundleDescriptor().getApplication();
-            if (app != null) {
-                realmName = app.getRealm();
-            }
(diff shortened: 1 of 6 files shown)


appserver/security/webservices.security/src/main/java/com/sun/enterprise/security/jauth/jaspic/provider/config/SoapAuthenticationService.java
```diff
@@ -131,6 +132,22 @@ public SoapAuthenticationService(String layer, Map<String, Object> properties, C
         this.ejbDelegate = new EJBPolicyContextDelegate();
     }
 
+    @Override
+    protected CallbackHandler getCallbackHandler() {
+        String realmName = null;
+        WebServiceEndpoint wse = (WebServiceEndpoint) map.get(PipeConstants.SERVICE_ENDPOINT);
+        if (wse != null) {
```
luiseufrasio: Add: `&& wse.getBundleDescriptor() != null` 
to avoid NPE on line `140`

WDYT?

luiseufrasio (approved):
Just a simple suggestion, but it works and LGTM!

appserver/security/webservices.security/src/main/java/com/sun/enterprise/security/jauth/jaspic/provider/config/SoapAuthenticationService.java
stian-sigvartsen: Makes sense. I think this might happen if a request comes in during app undeployment. I've pushed a commit

PR #8337, opened 2026-08-04
Remove doc

kalinchan:


Pandrex247:
Too slow: https://github.com/payara/Payara/pull/8338

PR #8341, opened 2026-08-06
FISH-14059 OSGi Shell

Viii3:
## Description

- Resolves FISH-14059
  - Prevents the `IllegalArgumentException` and property syntax error that block the `osgi-shell` command.
  - Fixes an issue where OSGi Shell commands would execute as expected, but not output any feedback.
  - Fixes an issue where executing any command in the OSGi Shell would would return a warning message that a system terminal could not be created and a dumb terminal was in use instead.

### Testing Performed

- Started the domain.
- Ran `asadmin osgi-shell`
- Within the OSGi Shell, ran `lb` and observed the output.
- Exited the shell and stopped the domain.

Tested both normally and in multi-mode.

### Testing Environment

Maven version: 3.9.14
Java version: 21.0.10, vendor: Azul Systems, Inc.


stian-sigvartsen (approved):
LGTM!

PR #8352, opened 2026-08-17
FISH-14150 Upgrade to Hazelcast 5.7.0 whilst preserving previous AP (as in CAP) "CP" behaviour

stian-sigvartsen:
## Description
Fix three CP breakages: in 5.3.8, CPMemberCount=0 (Payara's default) caused CP to silently use AP/IMap infrastructure instead of Raft. In 5.7.0 CE, CP throws UnsupportedOperationException instead. EJB timer, CDI @Clustered, and EJB singleton locks now fall back to IMap-based equivalents explicitly, preserving the same guarantees.

Our Hazelcast 5.3.8 patches are already merged upstream and feature in Hazelcast 5.7.0.

### Testing Performed
Manual smoke testing of the server (creating multiple instances etc.)

### Testing Environment
Zulu JDK 21.0.8 on OSX with Maven 3.9.9


Pandrex247 (commented):
Overall nothing looks immediately amiss to me

I want to give the clustered singleton stuff a test locally to see if that new warning message gets thrown: I wouldn't expect a warning to be printed for the default configuration

Pandrex247 (dismissed):
OK yeah, it prints that warning along with the full stack trace.
I wouldn't have that as a warning by default, perhaps only if an attempt has been made to explicitly configure the CP subsystem.

Pandrex247 (approved):

PR #8386, opened 2026-08-28
FISH-14240 Add Agentic AI module to Payara Micro and Embedded

luiseufrasio:
## Summary

The `agentic-ai` module was bundled into the full server and featureset distributions but missing from **Payara Micro** and **Payara Embedded**. This PR adds it to both, and proves it is bundled and functional on each distribution by running the existing `agentic-ai-quickstart` sample against the corresponding Arquillian containers.

## Distribution wiring

Added the `agentic-ai` packager-fragment (`fish.payara.server.internal.packager:agentic-ai`, `type=zip`, optional) to:

- `appserver/extras/payara-micro/payara-micro-distribution`
- `appserver/extras/embedded/all`
- `appserver/extras/embedded/web`

Verified `AgenticAIExtension.class`, `agentic-ai-core`, and `jakarta.agentic-ai-api` are present in the produced artifacts.

## Tests

Rather than adding bespoke modules, the existing `agentic-ai-quickstart` sample (a single `@Agent` driving the Trigger/Decision/Action/Outcome phases behind a REST endpoint, backed by a stubbed `LargeLanguageModel`) now runs on both distributions:

- **Micro** — already covered by the `payara-micro-managed` profile.
- **Embedded** — covered by a new **`payara-server-embedded`** profile using the `arquillian-payara-server-embedded` container (roughly matching the MicroProfile-TCK-Runners setup).

The sample IT deploys the quickstart's **built war** (via `ZipImporter`) and adds a `glassfish-web.xml` with `delegate="false"`. This matters on the embedded container, which boots the server in the test JVM: the `@Agent` phase methods are package-private, and Weld can only override them in its client proxy when `QuestionAgent` is loaded by the same (webapp) classloader as that proxy. Because the module also compiles the agent into `target/classes` (on the in-JVM system classpath) and GlassFish web classloaders delegate to the parent first, the child-first override keeps Weld from loading the agent off the system classpath — which would otherwise leave the injected `LargeLanguageModel` null.

## Other

Removed the redundant `payara-micro-boot` dependency from `micro-programmatic`: its classes are already provided by the `payara-micro` uber-jar, and the explicit dependency was [removed in Payara Enterprise](https://github.com/payara/Payara-Enterprise/pull/2947). This also clears a latent POM error (it declared no version and is not managed by the bom).

## Manual reproducers

Standalone projects that independently verify the module is bundled and the `@Agent` workflow runs end to end on each distribution (attached to [FISH-14240](https://azul.atlassian.net/browse/FISH-14240)):

- **Payara Micro** — [FISH-14240-agentic-micro-reproducer.zip](https://azul.atlassian.net/rest/api/3/attachment/content/109829)
- **Payara Embedded** — [FISH-14240-agentic-embedded-reproducer.zip](https://azul.atlassian.net/rest/api/3/attachment/content/109830)

Build Payara from this branch (`mvn clean install -DskipTests`) to publish the uber-jars to `~/.m2`, then run `./run.sh` in each project and `POST` to `http://localhost:8080/agentic/api/ask`. See each project's `README.md` for details (including how to switch from the offline stub to a real local Ollama model).

[FISH-14240]: https://azul.atlassian.net/browse/FISH-14240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appserver/tests/payara-samples/micro-programmatic/pom.xml
+++ b/appserver/tests/payara-samples/micro-programmatic/pom.xml
@@ -61,10 +61,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>fish.payara.server.internal.extras</groupId>
-            <artifactId>payara-micro-boot</artifactId>
-        </dependency>
-        <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
             <scope>test</scope>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -70,6 +70,8 @@
         <payara.containerType>SERVER</payara.containerType>
         <!-- Either payara or payara-web -->
         <payara.server.artifact>payara</payara.server.artifact>
+        <!-- Either payara-embedded-all or payara-embedded-web -->
+        <payara.embedded.artifact>payara-embedded-all</payara.embedded.artifact>
 
         <!-- Exclude unstable tests by default-->
         <excludedGroups>fish.payara.samples.Unstable</excludedGroups>
@@ -312,6 +314,77 @@
                                 </configuration>
                             </execution>
                         </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>payara-server-embedded</id>
+            <!--
+            This profile runs the samples against Payara Embedded via the Arquillian
+            embedded container. The embedded distribution under test is selected by the
+            payara.embedded.artifact property (payara-embedded-all or payara-embedded-web).
+            -->
+            <dependencies>
+                <dependency>
+                    <groupId>fish.payara.arquillian</groupId>
+                    <artifactId>arquillian-payara-server-embedded</artifactId>
+                    <scope>test</scope>
+                    <optional>true</optional>
+                </dependency>
+                <dependency>
+                    <groupId>jakarta.platform</groupId>
+                    <artifactId>jakarta.jakartaee-api</artifactId>
+                    <scope>provided</scope>
+                </dependency>
+                <dependency>
+                    <groupId>fish.payara.extras</groupId>
+                    <artifactId>${payara.embedded.artifact}</artifactId>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.eclipse.parsson</groupId>
+                    <artifactId>jakarta.json</artifactId>
+                    <scope>test</scope>
+                </dependency>
+                <dependency>
+                    <groupId>org.glassfish.jersey.core</groupId>
+                    <artifactId>jersey-client</artifactId>
+                    <scope>test</scope>
+                </dependency>
+            </dependencies>
+            <build>
+                <plugins>
+                    <!--
+                    The embedded container boots GlassFish in-process, so (unlike the
+                    separate-process server/micro containers) the test JVM itself needs
+                    the JDK module openings GlassFish requires on JDK 17+.
+                    -->
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <argLine>
+                                --add-opens=java.base/jdk.internal.loader=ALL-UNNAMED
+                                --add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED
+                                --add-exports=java.base/jdk.internal.ref=ALL-UNNAMED
+                                --add-opens=java.base/java.lang=ALL-UNNAMED
+                                --add-opens=java.base/java.nio=ALL-UNNAMED
+                                --add-opens=java.base/sun.nio.ch=ALL-UNNAMED
+                                --add-opens=java.management/sun.management=ALL-UNNAMED
+                                --add-opens=java.base/sun.net.www.protocol.jrt=ALL-UNNAMED
+                                --add-exports=java.base/sun.net.www=ALL-UNNAMED
+                                --add-exports=java.base/sun.security.util=ALL-UNNAMED
+                                --add-opens=java.base/java.lang.invoke=ALL-UNNAMED
+                                --add-opens=java.desktop/java.beans=ALL-UNNAMED
+                                --add-exports=jdk.naming.dns/com.sun.jndi.dns=ALL-UNNAMED
+                                --add-opens=java.base/sun.net.www.protocol.jar=ALL-UNNAMED
+                                --add-opens=java.rmi/sun.rmi.transport=ALL-UNNAMED
+                                --add-opens=java.base/java.util=ALL-UNNAMED
+                                --add-opens=java.base/java.net=ALL-UNNAMED
+                            </argLine>
+                        </configuration>
                     </plugin>
                 </plugins>
             </build>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -445,6 +445,7 @@
             <id>web</id>
             <properties>
                 <payara.server.artifact>payara-web</payara.server.artifact>
+                <payara.embedded.artifact>payara-embedded-web</payara.embedded.artifact>
             </properties>
         </profile>
     </profiles>
--- a/appserver/tests/payara-samples/pom.xml
+++ b/appserver/tests/payara-samples/pom.xml
@@ -357,9 +357,9 @@
             <build>
                 <plugins>
                     <!--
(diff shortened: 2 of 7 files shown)


Pandrex247 (changes requested):
Please rebase to remove the merge commit, as it isn't adding anything

luiseufrasio:
> Please rebase to remove the merge commit, as it isn't adding anything

Rebased!

appserver/tests/payara-samples/pom.xml
```diff
@@ -317,6 +318,20 @@
             </build>
         </profile>
 
+        <profile>
+            <id>payara-embedded-managed</id>
```
Pandrex247: There's no such thing as payara-embedded-remote, so just call it payara-server-embedded.

Please roughly match: https://github.com/payara/MicroProfile-TCK-Runners/blob/microprofile-7.1/pom.xml#L405-L445

appserver/tests/payara-samples/micro-programmatic/pom.xml
```diff
@@ -63,6 +63,7 @@
         <dependency>
             <groupId>fish.payara.server.internal.extras</groupId>
             <artifactId>payara-micro-boot</artifactId>
+            <version>${project.version}</version>
         </dependency>
```
Pandrex247: Is this branch out of date?
Why is this dependency even here? It was removed in Enterprise

```suggestion
```

appserver/tests/payara-samples/agentic-ai-quickstart-micro/pom.xml
```diff
@@ -0,0 +1,198 @@
...
+            <groupId>fish.payara.extras</groupId>
+            <artifactId>payara-micro</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>fish.payara.server.internal.extras</groupId>
+            <artifactId>payara-micro-boot</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
Pandrex247: Internal dependency - you can't use this.

appserver/tests/payara-samples/agentic-ai-quickstart-embedded/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <!-- The real Payara Embedded distribution, booted in-process by the test -->
+        <dependency>
+            <groupId>fish.payara.extras</groupId>
+            <artifactId>payara-embedded-all</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
Pandrex247: Version should be managed by bom, and this should be a part of the samples parent.

Please see https://github.com/payara/MicroProfile-TCK-Runners/blob/microprofile-7.1/pom.xml#L405-L445 for a rough example we should adhere to

appserver/tests/payara-samples/agentic-ai-quickstart-embedded/src/test/java/fish/payara/samples/agentic/quickstart/embedded/AgenticQuickstartEmbeddedIT.java
Pandrex247, line 1: Why aren't we running this as an arquillian test?
This should just be a part of the other existing test

appserver/tests/payara-samples/agentic-ai-quickstart-micro/src/test/java/fish/payara/samples/agentic/quickstart/micro/AgenticQuickstartMicroIT.java
Pandrex247, line 1: This should just be a part of the other existing arquillian test

Pandrex247 (changes requested):

appserver/tests/payara-samples/micro-programmatic/pom.xml
Pandrex247: Removed in Enterprise here: https://github.com/payara/Payara-Enterprise/pull/2947

appserver/tests/payara-samples/pom.xml
luiseufrasio: done!

appserver/tests/payara-samples/pom.xml
```diff
@@ -70,6 +70,8 @@
         <payara.containerType>SERVER</payara.containerType>
         <!-- Either payara or payara-web -->
         <payara.server.artifact>payara</payara.server.artifact>
+        <!-- Either payara-embedded-all or payara-embedded-web -->
+        <payara.embedded.artifact>payara-embedded-all</payara.embedded.artifact>
```
Pandrex247, line 74: Please also add `<payara.embedded.artifact>payara-embedded-web</payara.embedded.artifact>` to the `web` profile in this pom, so that when that profile is activated it changes the distribution used.

appserver/tests/payara-samples/samples/agentic-ai-quickstart/src/test/java/fish/payara/samples/agentic/quickstart/AgenticQuickstartIT.java
```diff
@@ -67,19 +69,39 @@ public class AgenticQuickstartIT {
...
                 .addClass(StubLargeLanguageModel.class)
                 .addAsWebInfResource(
-                        new StringAsset("<beans xmlns=\"https://jakarta.ee/xml/ns/jakartaee\" "
-                                + "version=\"4.0\" bean-discovery-mode=\"all\"/>"),
-                        "beans.xml");
+                        new StringAsset("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+                                + "<glassfish-web-app>\n"
+                                + "    <class-loader delegate=\"false\"/>\n"
+                                + "</glassfish-web-app>\n"),
+                        "glassfish-web.xml");
```
Pandrex247: Why glassfish-web.xml?
We should be using our payara-web.xml

It would be better if we just created the file itself too.
If it's needed to run the sample it shouldn't just be generated for the test

Pandrex247 (changes requested):

appserver/tests/payara-samples/samples/agentic-ai-quickstart/src/test/java/fish/payara/samples/agentic/quickstart/AgenticQuickstartIT.java
Pandrex247: Thinking about this - is this even required? What needs overriding?

luiseufrasio: You're right; 
it was only a workaround for the embedded in-JVM classpath. 
Removed it from the sample and scoped the embedded quirk to the test.

appserver/tests/payara-samples/samples/agentic-ai-quickstart/pom.xml
```diff
@@ -103,4 +103,32 @@
...
+            The embedded Arquillian container boots the server inside the test JVM, which puts
+            this module's own target/classes on the system classpath. Because the @Agent's phase
+            methods (@Trigger/@Decision/@Action/@Outcome) are package-private, Weld can only
+            override them in a client proxy loaded by the same classloader as the bean; the
+            embedded container's parent-first delegation would instead resolve QuestionAgent from
+            the system classpath, so the proxy could not override the package-private @Action and
+            the injected LargeLanguageModel would be null. That is an artefact of the in-JVM
+            harness, not of a real deployment, so this sample is verified against the managed
+            Payara Server and Payara Micro profiles (out-of-process) and skips the embedded one.
+        -->
```
Pandrex247, line 121: Wait, so the class loader delegation is required?
I'm confused by this comment

luiseufrasio: Sorry, that comment was misleading. The class-loader delegation is not required.
It would only ever be needed to paper over the embedded Arquillian container running in the same JVM as the test (which leaks this module's target/classes onto the system classpath). Since that never happens in a real deployment, I dropped the delegation entirely and instead skip the embedded profile for this sample — it's still covered by the managed Server and Micro profiles. I've reworded the comment to lead with that so it's no longer confusing.

Pandrex247 (approved):

PR #8379, opened 2026-08-27
FISH-14186 Improve isBehindProxyEnabled method to check per instance

kalinchan:
## Description

When an instance on an SSH Node was created an error would occur due to an NPE when looking for the server config in isBehindProxyEnabled method. 

This PR resolves this issue by changing the logic to check per instance for if the isBehindProxy option is enabled.
### Blockers

n/a
### New tests

n/a
### Testing Performed

  DAS (default instance)

  1. Start the domain with debugging enabled.
  2. Set a breakpoint inside isBehindProxyEnabled() and attach the debugger.
  3. Send a request to http://localhost:4848/management/domain to trigger the breakpoint.
  4. Verify that isBehindProxyEnabled() returns false (proxy not yet configured).
  5. Enable behind-proxy on the DAS admin-listener configuration.
  6. Repeat step 3 and verify that isBehindProxyEnabled() now returns true.

  Standalone instance

  7. Create a standalone instance.
  8. Start the instance with debugging enabled.
  9. Attach the debugger to the standalone instance.
  10. Send a request to https://localhost:24848/management/domain to trigger the breakpoint.
  11. Verify that isBehindProxyEnabled() returns false (proxy not yet configured on this instance).
  12. Enable behind-proxy on the standalone instance's sec-admin-listener configuration.
  13. Repeat step 10 and verify that isBehindProxyEnabled() now returns true.

  SSH node instance

  14. Create an SSH node.
  15. Create an instance on the SSH node and start it with debugging enabled.
  16. Attach the debugger to the SSH node instance.
  17. Send a request to https://{node-host}:24848/management/domain to trigger the breakpoint.
  18. Verify that isBehindProxyEnabled() returns false (proxy not yet configured on this instance).
  19. Enable behind-proxy on the SSH node instance's sec-admin-listener configuration.
  20. Verify that isBehindProxyEnabled() now returns true.

### Testing Environment

Apache Maven 3.9.8 (36645f6c9b5079805ea5009217e36f2cffd34256)
Maven home: /Users/kalinchan/Library/apache-maven-3.9.8
Java version: 21.0.11, vendor: Azul Systems, Inc., runtime: /Users/kalinchan/.sdkman/candidates/java/21.0.11-zulu
Default locale: en_GB, platform encoding: UTF-8
OS name: "mac os x", version: "26.5.2", arch: "aarch64", family: "mac"

## Documentation

n/a
## Notes for Reviewers

n/a

PR #8389, opened 2026-09-01
FISH-14186 Wordsmith

stian-sigvartsen:
Small follow up to address https://github.com/payara/Payara/pull/8379#discussion_r3893519969


Pandrex247 (approved):
🤌

PR #8402, opened 2026-09-07
FISH-14053 Improve Corba FOLB deployment group check

kalinchan:
## Description

This PR improves the check for deployment groups by verifying that the instance is actually in a deployment group.
### Blockers

n/a
### New tests

n/a
### Testing Performed

Prerequisite for all: enable FINE logging on the instance before starting it:
  asadmin set-log-levels --target <instance> javax.enterprise.resource.corba=FINE
  Then check <domain>/nodes/<node>/<instance>/logs/server.log for isDeploymentGroupsActive: messages, and look for GMS available and enabled from GlassFishORBManager to confirm
  FOLB activated.
- - -
Test 1 — Happy path (FOLB activates)

  Setup:
  1. Create two instances, add both to one deployment group
  2. Set a custom Hazelcast member group (e.g. myFolbGroup) on both instances' configs via Admin Console → Configurations → <config> → Hazelcast → Member Group
  3. Deploy an application to the deployment group: asadmin deploy --target <dg> myapp.war 
  4. Start both instances
  
  Verify: No isDeploymentGroupsActive: failure log appears. GMS available and enabled is logged. Both instance endpoints appear in the CORBA log line iiop instance info =.

  ---

  Test 2 — Instance not in any deployment group 

  Setup:
  1. Start a standalone instance not added to any deployment group
  
  Verify: Log shows isDeploymentGroupsActive: instance is not in a deployment group. GMS available and enabled does not appear (unless it falls into traditional cluster path).

  ---

  Test 3 — Default MicroShoal group shared with DAS 

  Setup:
  1. Create two instances in a deployment group
  2. Leave their Hazelcast member group as the default (MicroShoal)
  3. DAS also uses MicroShoal by default
  4. Deploy an app to the DG and start the instances
  
  Verify: Log shows isDeploymentGroupsActive: Hazelcast instance group MicroShoal is shared with an instance outside the deployment group. GMS available and enabled does not
  appear.

  ---
  
  Test 4 — Instances have different Hazelcast member groups 

  Setup:
  1. Two instances in the same deployment group
  2. Set member group groupA on instance1's config, groupB on instance2's config
  3. Deploy app to DG and start both
  
  Verify: Log on each instance shows isDeploymentGroupsActive: not all instances in the deployment group share the same Hazelcast instance group.

  ---

  Test 5 — Instance in two deployment groups

  Setup:
  1. Create two deployment groups dg1 and dg2
  2. Add instance1 to both 
  3. Deploy app, start instance1
  
  Verify: Log shows isDeploymentGroupsActive: instance belongs to more than one deployment group.

  ---

  Test 6 — App not deployed to all DG instances 

  Setup:
  1. Two instances in one DG with a custom shared member group
  2. Deploy the app directly to only one instance (not via the DG target): asadmin deploy --target <instance1> myapp.war
  3. Start both instances

  Verify: Log shows isDeploymentGroupsActive: application myapp is not deployed to all targets of the deployment group.
### Testing Environment

Java version: 1.8.0_472, vendor: Azul Systems, Inc., runtime: /Users/kalinchan/.sdkman/candidates/java/8.0.472-zulu/zulu-8.jdk/Contents/Home/jre
OS name: "mac os x", version: "26.5.2", arch: "aarch64", family: "mac"

For the JIRA reproducer, I required a Windows VM to reproduce the original issue.
## Documentation

n/a
## Notes for Reviewers

I used remoteview to test this 
[remoteview.zip](https://github.com/user-attachments/files/31788538/remoteview.zip)

### DIFF
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
@@ -217,6 +217,9 @@ public class IiopFolbGmsClient implements ClusterListener {
 
         // Check 2: The applications on the deployment group are deployed to all its instances
         List<ApplicationRef> dgAppRefs = myDG.getApplicationRef();
+        if (dgAppRefs.isEmpty()) {
+            return false;
+        }
         for (Server instance : dgInstances) {
             for (ApplicationRef dgAppRef : dgAppRefs) {
                 if (instance.getApplicationRef(dgAppRef.getRef()) == null) {
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
@@ -215,19 +215,10 @@ public class IiopFolbGmsClient implements ClusterListener {
         DeploymentGroup myDG = myDeploymentGroups.get(0);
         List<Server> dgInstances = myDG.getInstances();
 
-        // Check 2: The applications on the deployment group are deployed to all its instances
-        List<ApplicationRef> dgAppRefs = myDG.getApplicationRef();
-        if (dgAppRefs.isEmpty()) {
+        // Check 2: Only enable if at least one application is targetted to the deployment group
+        if (myDG.getApplicationRef().isEmpty()) {
+            fineLog("isDeploymentGroupsActive: no application is deployed to the deployment group");
             return false;
-        }
-        for (Server instance : dgInstances) {
-            for (ApplicationRef dgAppRef : dgAppRefs) {
-                if (instance.getApplicationRef(dgAppRef.getRef()) == null) {
-                    fineLog("isDeploymentGroupsActive: application {0} is not deployed to all targets of the deployment group",
-                            dgAppRef.getRef());
-                    return false;
-                }
-            }
         }
 
         // Check 3: No other instance in the deployment group is in any other deployment group
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
@@ -187,7 +187,9 @@ public class IiopFolbGmsClient implements ClusterListener {
     // Implementation
     //
     private boolean isDeploymentGroupsActive() {
-        return cluster != null && cluster.isEnabled() && cluster.getClusterMembers().size() > 1;
+        return cluster != null && cluster.isEnabled()
+                && myServer != null && !myServer.getDeploymentGroup().isEmpty()
+                && cluster.getClusterMembers().size() > 1;
     }
 
     private boolean isTraditionalClusterActive() {
--- a/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
+++ b/appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
@@ -48,6 +48,7 @@ import com.sun.corba.ee.spi.folb.GroupInfoServiceObserver;
 import com.sun.corba.ee.spi.folb.SocketInfo;
 import com.sun.corba.ee.spi.misc.ORBConstants;
 import com.sun.corba.ee.spi.orb.ORB;
+import com.sun.enterprise.config.serverbeans.ApplicationRef;
 import com.sun.enterprise.config.serverbeans.Cluster;
 import com.sun.enterprise.config.serverbeans.Config;
 import com.sun.enterprise.config.serverbeans.Configs;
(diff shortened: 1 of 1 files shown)


appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
```diff
@@ -187,7 +198,73 @@ public boolean isGMSAvailable() {
...
+
+        // Check 5: The instance group is not shared with any instance outside the deployment group.
+        Set<String> dgInstanceNames = dgInstances.stream().map(Server::getName).collect(Collectors.toSet());
+        for (UUID memberId : cluster.getClusterMembers()) {
+            String memberName = cluster.getMemberName(memberId);
+            if (!dgInstanceNames.contains(memberName) && myMemberGroup.equals(cluster.getMemberGroup(memberId))) {
+                fineLog("isDeploymentGroupsActive: Hazelcast instance group {0} is shared with an instance outside the deployment group",
+                        myMemberGroup);
+                return false;
+            }
```
Pandrex247, line 249: We might need an extra check: that all instances in the deployment group use the same config.

Otherwise when instances get added to the deployment group we can hit NPEs:

```
java.lang.NullPointerException: Cannot invoke "com.sun.enterprise.config.serverbeans.Server.getConfigRef()" because "server" is null
        at org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient.getConfigForServer(IiopFolbGmsClient.java:406)
        at org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient.getClusterInstanceInfo(IiopFolbGmsClient.java:427)
        at org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient.addMember(IiopFolbGmsClient.java:307)
        at org.glassfish.enterprise.iiop.impl.IiopFolbGmsClient.memberAdded(IiopFolbGmsClient.java:519)
        at fish.payara.nucleus.cluster.PayaraCluster.memberAdded(PayaraCluster.java:118)
```

Pandrex247: OK, **that** specific error is actually from the instance config not being dynamically synced across to the instance. After restarting the first instance it resyncs with the DAS and gets the config. This is probably a bug that can be tackled separately. It's being partially addressed in another issue as we speak, see the conversation on this Enterprise PR where I talk about a similar situation but with the replication of deployment group membership across instances: https://github.com/payara/Payara-Enterprise/pull/3194

> The following behaviour I believe is incorrect, as it's implying instances are not getting their config updated until they restart: When a new deployment group is created, this is "asynchronously" replicated to all instances. Running instances don't know about the new deployment group until they've restarted and resynced with the DAS.
When an instance is added to a deployment group, this is asynchronously replicated to all instances. Running instances which aren't a part of the deployment group being edited only receive the update once they have restarted.

In this particular stack trace's case, it's with instance refs being "asynchronously" replicated.

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
```diff
@@ -187,7 +198,73 @@ public boolean isGMSAvailable() {
...
+        if (myDeploymentGroups.size() > 1) {
+            fineLog("isDeploymentGroupsActive: instance belongs to more than one deployment group");
+            return false;
+        }
+
+        DeploymentGroup myDG = myDeploymentGroups.get(0);
+        List<Server> dgInstances = myDG.getInstances();
+
+        // Check 2: The applications on the deployment group are deployed to all its instances
+        List<ApplicationRef> dgAppRefs = myDG.getApplicationRef();
```
Pandrex247: This check might need reworking.

Any application that is deployed to the deployment group itself will be present on each instance.

What if there aren't any applications deployed to the deployment group, but to the instances within the group directly?
`dgAppRefs` will be empty in this case and the check gets "skipped".

So we either need to check both the deployment group and instance applications, **or** only check the deployment group and return `false` if `dgAppRefs` is empty.

I lean towards the latter for simplicity - we can always come back to it later. So just replace this with a check for if `dgAppRefs` is empty. So this check essentially becomes: "don't enable deployment group FOLB if there aren't any deployment group applications"

Pandrex247 (changes requested):

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
kalinchan: Resolved

appserver/orb/orb-iiop/src/main/java/org/glassfish/enterprise/iiop/impl/IiopFolbGmsClient.java
```diff
@@ -187,7 +198,76 @@ public boolean isGMSAvailable() {
...
+        }
+        for (Server instance : dgInstances) {
+            for (ApplicationRef dgAppRef : dgAppRefs) {
+                if (instance.getApplicationRef(dgAppRef.getRef()) == null) {
+                    fineLog("isDeploymentGroupsActive: application {0} is not deployed to all targets of the deployment group",
+                            dgAppRef.getRef());
+                    return false;
+                }
+            }
+        }
```
Pandrex247: The for loop of this check isn't doing anything now.
An instance is in the deployment group then all applications targetted to the deployment group will be deployed to it.

The only way instances within the deployment group would have different applications is if those instances have applications deployed to them specifically, that aren't targetted at the deployment group.

```suggestion
        // Check 2: Only enable if at least one application is targetted to the deployment group
        if (myDG.getApplicationRef().isEmpty()) {
            fineLog("isDeploymentGroupsActive: no application is deployed to the deployment group");
            return false;
        }
```

Pandrex247 (changes requested):

Pandrex247 (approved):

PR #8331, opened 2026-07-30
FISH-14066 Jakarta Agentic AI: use released API 1.0.0-M1 and EE11 TCK branch

luiseufrasio:
## Description

This PR brings the Jakarta Agentic AI integration branch up to date and switches the API dependency to a released version:

- Merge latest `payara/main` into the branch.
- Use the released **Jakarta Agentic AI API `1.0.0-M1`** (available on Maven Central) instead of the `1.0.0-SNAPSHOT` version.
- Point the Jakarta Agentic AI TCK stage in the `Jenkinsfile` at the `EE11` branch of the `jakartaee-10-tck-runners` repository.

## Dependency

This PR depends on payara/jakartaee-10-tck-runners#213 and should be merged together with it.

## Testing performed

CI / Jenkins pipeline (including the Jakarta Agentic AI TCK stage).

PR #8404, opened 2026-09-07
FISH-14237 Export only the CDI extension package from agentic-ai-core

luiseufrasio:
## Description
Follow-up to review comments on #8331 (FISH-14237): review the OSGi package exports of the internal `agentic-ai-core` module.

Investigation finding: the module used an empty, self-closing `<Export-Package/>` bnd instruction with a comment claiming it "keeps every package private". That is incorrect — the maven-bundle-plugin drops the empty instruction and bnd then falls back to its default of exporting *every* package. So the module was actually exporting all four `fish.payara.ai.agent.*` packages (the same latent issue exists in `data-core`).

Not exporting is also wrong for this module specifically: unlike `data-core` (which registers its CDI extension programmatically via a HK2 deployer + `SNIFFER_EXTENSIONS`), `agentic-ai-core` has no sniffer/deployer. It registers `AgenticAIExtension` purely as a portable CDI extension via `META-INF/services/jakarta.enterprise.inject.spi.Extension`, which is discovered by the deployed application's Weld bootstrap (`DeploymentImpl.getExtensions` -> `WeldBootstrap.loadExtensions(moduleClassLoader)`). For that lookup to load the extension, its package must be exported.

Fix: export only `fish.payara.ai.agent.extension` (the extension entry point) and keep the implementation packages (`cdi`, `engine`, `llm`) private via `<Private-Package>`. The misleading comment was rewritten to document the bnd behaviour.

### Blockers
None. A related latent issue in `data-core` (same empty `<Export-Package/>`) is noted for a separate ticket.

### New tests
None. This is a build/packaging (OSGi manifest) change; behaviour is covered by the existing agentic-ai sample integration tests (`AgenticTutorialIT`, `AgenticQuickstartIT`).

### Testing Performed
- Rebuilt the module (with upstream chain): `mvn -am -pl appserver/agentic-ai/agentic-ai-core install -DskipTests` -> BUILD SUCCESS.
- Inspected the generated `META-INF/MANIFEST.MF` of `agentic-ai-core.jar`:
  - `Export-Package` = `fish.payara.ai.agent.extension` only (`uses:="jakarta.enterprise.inject.spi"`).
  - `cdi`, `engine`, `llm` are no longer exported (private) but remain packaged in the jar, so the bundle resolves them within its own class loader.
- Verified the previous behaviour for comparison: with the empty `<Export-Package/>`, the built jars of both `agentic-ai-core` and `data-core` exported all their packages.
- Dependency-surface check: no other module or the samples import `fish.payara.ai.agent.*`; the samples use only the public `jakarta.ai.agent.*` API — confirming the minimal export is sufficient.
- NOT run: the agentic-ai Arquillian sample ITs against a full distribution (recommended as a CI validation step).

### Testing Environment
Azul Zulu JDK 21.0.1 on Windows 11, Apache Maven 3.9.9.

## Documentation
N/A

## Notes for Reviewers
Start at `appserver/agentic-ai/agentic-ai-core/pom.xml`. The key point: `<Export-Package>` must name the extension package explicitly — an empty instruction is a no-op that silently exports everything.

🤖 Generated by LLM

### DIFF
--- a/appserver/agentic-ai/agentic-ai-core/pom.xml
+++ b/appserver/agentic-ai/agentic-ai-core/pom.xml
@@ -121,14 +121,25 @@
                         <configuration>
                             <instructions>
                                 <!--
-                                    This is an internal implementation module
-                                    (fish.payara.server.internal.agentic-ai), so it must
-                                    not export any packages for OSGi consumption. An empty
-                                    Export-Package keeps every package private, instead of
-                                    letting bnd fall back to exporting the module's packages.
-                                    Mirrors the Jakarta Data implementation module (data-core).
+                                    The Jakarta Agentic AI CDI portable extension
+                                    (AgenticAIExtension) is discovered by a deployed
+                                    application's Weld bootstrap via ServiceLoader on the
+                                    application module class loader (see
+                                    DeploymentImpl.getExtensions -> WeldBootstrap.loadExtensions).
+                                    For that lookup to find and load the extension, the package
+                                    that declares it must be exported for OSGi consumption.
+                                    Everything else is internal implementation and stays
+                                    private; the exporting bundle resolves those packages
+                                    within its own class loader once the extension is loaded.
+
+                                    Note: an empty/self-closing <Export-Package/> does NOT
+                                    make packages private - the maven-bundle-plugin drops the
+                                    empty instruction and bnd then exports every package by
+                                    default. Private-Package is what actually keeps a package
+                                    off the export list.
                                 -->
-                                <Export-Package/>
+                                <Export-Package>fish.payara.ai.agent.extension</Export-Package>
+                                <Private-Package>fish.payara.ai.agent.*</Private-Package>
                             </instructions>
                         </configuration>
                     </execution>


lprimak:
Hey, Luis,
Can you remove "Generated with Claude Code" with backlinks to Claude? I think they get enough free advertisement already. I just use "Generated with LLM" with no back-links, for example.

Thank you

lprimak:
Also, does this by any chance fix this warning?

```
[2026-09-06T17:28:45.940-0500] [Payara 7.2026.9] [INFO] [] [org.jboss.weld.Event] [tid: _ThreadID=153 _ThreadName=admin-thread-pool::admin-listener(1)] [timeMillis: 1788733725940] [levelValue: 800] [[
  WELD-000411: Observer method [BackedAnnotatedMethod] fish.payara.ai.agent.extension.AgenticAIExtension.processAgent(@Observes ProcessAnnotatedType<X>) receives events for all annotated types. Consider restricting events using @WithAnnotations or a generic type with bounds.]]
```

breakponchito (approved):
LGTM

Issue #8300, opened 2026-07-15
Bug Report: asadmin recorder produces wrong parameters for create-javamail-resource/FISH-14072


rrohm:
### Brief Summary

When creating a JavaMail resource, the asadmin recorder records wrong parameter names for the mail host and the mail user: 

- instead of --mailhost=... it records --host=...
- instead of --mailuser=... it records --user=...

Subsequently, the recorded command cannot be replayed. 
This has been observed on Payara Version 7.2026.6.

### Expected Outcome

The asadmin recorder should produce a command that could be replayed successfully, e.g.: 
create-javamail-resource --debug=false --storeProtocol=imap --auth=true --storeProtocolClass=org.eclipse.angus.mail.imap.IMAPStore --enabled=true --target=domain --password=${ALIAS=mail_smpt_user} --transportProtocol=smtp **--mailhost**=_your-mailhost_ --from=_your-from-address_ --transportProtocolClass=org.eclipse.angus.mail.smtp.SMTPTransport **--mailuser**=_your-mail-username_ mail/myMailSession

asadmin should return 
Command create-javamail-resource executed successfully.

### Current Outcome

Currently, the asadmin recorder procudes user and host parameters: 
create-javamail-resource --debug=false --storeProtocol=imap --auth=true --storeProtocolClass=org.eclipse.angus.mail.imap.IMAPStore --enabled=true --target=domain --password=${ALIAS=mail_smpt_user} --transportProtocol=smtp **--host**=_your-mailhost_ --from=_your-from-address_ --transportProtocolClass=org.eclipse.angus.mail.smtp.SMTPTransport **--user**=_your-mail-username_ mail/myMailSession

When executing this, asadmin returns:

Deprecated syntax, instead use:
asadmin --host _your-mailhost_ --user _your-mail-username_  create-javamail-resource [options] ...
java.net.SocketException: Connection reset
Command create-javamail-resource failed.

### Reproducer

- Enable the asadmin recorder in the Domain Administration Console
- Create a new JavaMail resource in the Domain Administration Console
- Delete the successfully created JavaMail resource in the Domain Administration Console
- Start asadmin 
- Copy the recorded create-javamail-resource command from the asadmin-commands.txt into the console and hit ENTER

### Operating System

Ubuntu 24.04 LTS

### JDK Version

OpenJDK 64-Bit Server VM (build 21.0.11+10-1-24.04.2-Ubuntu, mixed mode, sharing)

### Payara Distribution

Payara Server Full Profile

[...]

PR #8405, opened 2026-09-07
FISH-14072: Fix asadmin recorder using wrong parameter names for create-javamail-resource

raushan606:
## Description

The asadmin recorder wrote alias parameter names (e.g. --host, --user, --from) instead of the correct primary names (--mailhost, --mailuser, --fromaddress). The recorded command would fail if replayed.
### Fix
Resolve alias names to primary parameter names before passing to the recorder. The fix applies generically to any command with aliased parameters.

### Testing Performed

Created a JavaMail Session via Admin GUI with recorder enabled, confirmed correct parameter names in the recorded command, and successfully replayed it.
### Testing Environment

Zulu JDK 21.0.10 on MacOS with Maven  3.9.16 
## Notes for Reviewers

For Issue: https://github.com/payara/Payara/issues/8300


pdudits:
You demonstrate a general solution to this problem. What other commands might have been affected and did you test those?

raushan606:
> You demonstrate a general solution to this problem. What other commands might have been affected and did you test those?

# Root Issue

The root issue is specifically about commands whose `@Param` aliases collide with `asadmin`'s global `ProgramOptions`.

When the recorder writes `--host` or `--user` as a parameter name, `asadmin` interprets it as the DAS connection host/user on replay rather than as the command parameter.

## Commands with Colliding Aliases

Grep confirmed the following commands have colliding aliases:

| Command | Alias (recorded) | Canonical Name | Collides With |
|---|---|---|---|
| `create-javamail-resource` | `host`, `user` | `mailhost`, `mailuser` | `--host`, `--user` |
| `create-jms-host` | `host`, `port` | `mqHost`, `mqPort` | `--host`, `--port` |
| `create-iiop-listener` | `port` | `iiopport` | `--port` |
| `configure-jms-cluster` | `user` (via `dbuser`) | `dbuser` | `--user` |

## Why `create-jdbc-connection-pool` Is Not Affected

`create-jdbc-connection-pool` also goes through `ResourceUtil.runCommand`, but it is not affected because its aliases (`maxWaitTimeInMillis`, `poolResizeQuantity`, etc.) do not shadow any global `ProgramOptions`.

Therefore, replaying a recorded `create-jdbc-connection-pool` command works correctly.

## Why the Primary Names Are the Correct Output

The official `asadmin` documentation uses the canonical parameter names for these commands:

- `create-javamail-resource` documents:
  - `--mailhost`
  - `--mailuser`
  - `--fromaddress`

- `create-jms-host` documents:
  - `--mqhost`
  - `--mqport`
  - `--mquser`

The recorder output should therefore use the documented canonical interface rather than the potentially conflicting aliases.

## Testing

`create-javamail-resource` was verified end-to-end.

### Before the Fix

The recorder generated:

```text
--host smtp.example.com

pdudits (approved):

PR #8413, opened 2026-09-08
FISH-14056 : Scope deployment-group replication and visibility to the group's members

luiseufrasio:
## Problem

Deployment-group configuration reaches instances through two channels that were never reconciled with each other: command replication (`@ExecuteOn`), which only reaches running instances at the moment of the command, and the full `domain.xml` copied at instance startup. That mismatch produced five distinct problems.

**1. Misleading offline warnings.** `create-deployment-group` and `delete-deployment-group` replicated to `RuntimeType.ALL`, so any stopped instance in the domain produced

```
Warning: Instance X seems to be offline; command create-deployment-group was not replicated to that instance
Command create-deployment-group completed with warnings.
```

even when that instance had no relationship whatsoever with the group. This is the original report behind FISH-14056.

**2. Membership changes did not reach running members.** `add`/`remove-instance-from-deployment-group` replicated to every instance rather than to the group's members, and each member only ever learned the single delta it was told about.

**3. Every instance knew about every deployment group** — including groups with no members at all, and the membership of groups it had nothing to do with. Because the only channel feeding that information was the file copied at startup, those entries also appeared to refresh only on a restart.

**4. Silent divergence between the DAS and an instance.** Removing an instance from a group could fail on a member with `Instance X does not exist` while the DAS applied the change, leaving the two disagreeing, with the command still reporting `executed successfully`.

**5. Deleting an instance did not replicate either.** `Server.DeleteDecorator` drops the `dg-server-ref` elements with a direct configuration transaction on the DAS. That is not a command, so nothing is replicated and the group's other running members kept the deleted instance in their own `domain.xml` until their next restart.

## Solution

**`create`/`delete-deployment-group` are DAS-only.** The persisted state reaches instances through the startup sync anyway; replicating the command only produced the warning.

**`add`/`remove-instance-from-deployment-group` are scoped to the group.** They declare `@ExecuteOn({DAS, INSTANCE})` with `@TargetType(DEPLOYMENT_GROUP)`, so the cluster executor replicates to `Target.getInstances(<group>)` — offline non-members are never contacted, so the misleading warning disappears at the root rather than being suppressed by a failure policy.

On top of that:

- `prepare()` widens the replicated `instance` parameter to the group's **full** membership, and the instance side **reconciles** against it — adding what is missing *and* dropping references that are no longer members. Without the removal step, an instance rejoining a group whose membership shrank while it was outside would keep the departed members as phantom entries until its next restart.
- An instance that receives a replicated add for a group it has never seen (created after it started) **creates the group in its own config** rather than dropping the change.
- A removal is replicated to the **just-removed instances that are still running**, so they drop their own reference live. The framework would not do this, since they are no longer members at replication time. Whether an instance is running is decided by an admin-port reachability check (`ServerHelper.isRunning()`) rather than the `InstanceStateService` cache, which can report a stale non-running state right after `start-deployment-group`.
- `ifOffline` / `ifNeverStarted` stay at `FailurePolicy.Ignore` (as on the previous releases), so a genuinely offline **member** does not degrade the command to `completed with warnings`; the change is queued as a pending config change and applied on its next start. `ifFailure` is deliberately left at its default so real replication failures still surface.
- `DeploymentTargetResolver.getTarget()` exposes the group as `target` for the cluster executor **additively**, while keeping the `deploymentGroup` parameter untouched. The parameter is deliberately *not* renamed: `InstanceRestCommandExecutor` builds the replicated request from a cached command model under `~/.gfclient/cache` that is not ETag-revalidated before use, so a member still holding a previously released model would reject a renamed option.

**Deployment groups are now scoped at parse time.** `InstanceReaderFilter` already narrows `server`, `config` and `cluster` elements to what the instance needs, and its own comment claimed to cover deployment groups, but they fell through the `keep everything else` branch. They are now filtered against the full set of groups the instance belongs to — unlike a cluster, an instance can be in several. `DasReaderFilter` is untouched, so the DAS keeps the complete view.

Nothing on an instance needs a group it is not in: the runtime lookups (`MQAddressList.getDeploymentGroupForServer`/`isClustered`, `DatabaseEJBTimerService` via `getDeploymentGroupsForInstance`) all search for the group containing *this* instance.

`DomainXmlPreParser.postProcess()` also had to stop resolving deployment-group membership only for non-clustered instances: finding the cluster returned from the method outright, leaving `deploymentGroup` null and `validDG` false for a clustered instance, which was therefore already being denied the servers and configs of any deployment group it also belonged to.

**`create`/`delete-application-ref` use the `DEFAULT` operand key.** Those commands declare the application name as `@Param(primary = true)`; under the `name` key it resolves on the DAS but is not carried as the operand when the command is replicated to the target instance, which then fails with `Cannot find name in ... command model`.

**Instance deletion is routed through the command.** A new `Before` supplemental on `_unregister-instance` — the path both `delete-instance` and `delete-local-instance` go through — calls `remove-instance-from-deployment-group` for each group the instance belongs to, so the remaining running members converge live and the group's application-refs and resource-refs are cleaned up. `Server.DeleteDecorator` keeps its own cleanup, which by then has nothing left to remove; routing the removal through the command instead of teaching the decorator to replicate keeps a shared config decorator out of the replication business.

## Testing

Verified end to end against a running server with three instances and two overlapping deployment groups — `DG1 = {Insty1, Insty3}`, `DG2 = {Insty2, Insty3}`:

| | sees DG1 | sees DG2 |
|---|---|---|
| DAS | yes | yes |
| Insty1 | yes | **no** |
| Insty2 | **no** | yes |
| Insty3 | yes | yes |
| instance in no group | **no** | **no** (`<deployment-groups></deployment-groups>`) |

Each instance's `domain.xml` was checked against the DAS on five axes — visible groups, the membership of each group, the `<servers>` of its group peers, their `<configs>`, and the negative assertion that nothing leaked — with no orphan references.

Also verified:

- No offline warning for a stopped non-member on any of the four commands.
- A running instance added to a group created after it started converges live, with no `NullPointerException`.
- An application deployed to a group follows an instance in (serves HTTP without a restart) and out (404, `application-ref` removed).
- A removed running instance drops its own reference live, including immediately after `start-deployment-group`.
- An instance rejoining a group whose membership shrank converges to exactly the DAS membership, with no phantom member.
- The admin console / REST path (`POST` with the group under `deploymentGroup`) succeeds.
- Deleting an instance that belongs to groups makes the remaining running members drop it live,
  with no restart; deleting an instance outside any group is unaffected.
- Restarting an instance converges it exactly to the DAS.

`config-api` unit tests pass, including three new ones (`instanceInSeveralDeploymentGroups`, `instanceInOneDeploymentGroupOnly`, `instanceInNoDeploymentGroup`) against a new `parser/dgi1i2.xml` fixture. The functional suite gains a `TestDeploymentGroupReplicationWarning` class covering the offline-warning, live-convergence, REST and reconciliation cases.

## Known limitations

Not fixed here; happy to split out if preferred.

1. A REST client that sends only `target=` (rather than `deploymentGroup=`) gets `HTTP 500 — Invalid option: target`. `getTarget()` maps `deploymentGroup → target` but not the reverse. Neither `asadmin` nor the admin console hits this.
2. If replication to a member fails, the DAS keeps its own change and there is no rollback (`undo()` is a no-op), which also blocks a retry with `already exists`.
3. `add-instance-to-deployment-group` for an instance that is already a member fails with the raw HK2 `A DGServerRef with the same key ... already exists` rather than being a no-op. Pre-existing.
4. The `seems to be offline` text is attached by `InstanceRestCommandExecutor` whenever the connection fails, regardless of `FailurePolicy` — only the exit code is policy-driven. A stopped member still logs the line; it just no longer degrades the command, which matches the behaviour of the previous releases.

---

Same change as payara/Payara-Enterprise#3194 (Enterprise 5), where this was developed and reviewed, and payara/Payara-Enterprise#3359 (Enterprise 7). Every file involved is byte-identical across the code bases apart from `javax`/`jakarta` imports.

Verified on Payara Server 7.2026.10.

https://claude.ai/code/session_01Dnq5pS1kxheheV6Eb4PhdA


nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
```diff
@@ -0,0 +1,136 @@
...
+        ifFailure = FailurePolicy.Warn)
+@I18n("pre.unregister.instance.deployment.group")
+@PerLookup
+@ExecuteOn(value = {RuntimeType.DAS})
+@RestEndpoints({
+    @RestEndpoint(configBean = Domain.class,
+            opType = RestEndpoint.OpType.POST,
+            path = "_pre-unregister-instance-deployment-group",
+            description = "_pre-unregister-instance-deployment-group")
+})
```
Pandrex247: Is this a domain command? It seems more like an instance or deployment group command.

Does this even need a rest endpoint?

nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
Pandrex247, line 1: Why not just extend `PostUnregisterInstanceCommand`?

nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
```diff
@@ -0,0 +1,136 @@
...
+ * replicated, and the group's other members keep the deleted instance in their own
+ * configuration until they are restarted. Routing the removal through the command instead
+ * reuses its replication, so the remaining running members converge live, and it also cleans
+ * up the application-refs and resource-refs the group had pushed onto the instance.
+ * <p>
+ * This runs for both {@code delete-instance} and {@code delete-local-instance}, because both
+ * reach the configuration through {@code _unregister-instance}. By the time the decorator
+ * runs there is no reference left for it to remove, so it becomes a no-op fallback.
+ *
+ * @since 5.0
```
Pandrex247: 5.0 what? There is no version 5.0

Pandrex247 (changes requested):
There is still an outstanding bug here leaving this half-finished: instance configs aren't being synced.

* Create and start an instance: Insty1
* Create a deployment group and add Insty1 to it
* Create another instance, add it to the deployment group, and start it: Insty2

Insty1 will see in its domain.xml that it has two instances in its deployment group, but have no other knowledge of what Insty2 is until it is restarted.
Insty2 will have `<server>` entries for itself and Insty1, and `<config>` entries for its config and that of Insty1 (Insty1-config and Insty2-config assuming you left everything as default).

lprimak:
Too many comments :) Claude :)

lprimak:
Still self-advertizing Claude :)

nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
luiseufrasio: Agreed, removed in dad1778.

The `configBean = Domain.class` was copied from `PostUnregisterInstanceCommand`, which declares its endpoint the same way, but you are right that it does not earn one: this is a DAS-only `@Supplemental`, so it is only ever run in process by the supplemental command executor and is never invoked remotely. The whole `@RestEndpoints` block is gone.

nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
luiseufrasio: It cannot share that class, because of the timing.

`_unregister-instance` is the `@Delete` CRUD command of the `server` element (`Servers.java`: `@Delete(value="_unregister-instance", decorator=Server.DeleteDecorator.class)`), so anything running *after* it no longer has the `Server` bean. This command's whole job is to read `server.getDeploymentGroup()` and route the removal through `remove-instance-from-deployment-group` while that membership still exists, which is why it declares `Timing.Before`.

`PostUnregisterInstanceCommand` runs at the default (After) timing and does something unrelated: replicating the unregistration to a cluster's members. I have recorded this reasoning in the class javadoc in dad1778.

nucleus/cluster/admin/src/main/java/fish/payara/admin/cluster/PreUnregisterInstanceDeploymentGroupCommand.java
luiseufrasio: Copied from `AddInstanceToDeploymentGroupCommand`, which carries `@since 5.0` from when deployment groups were introduced. For a class being added now it should be the version it is actually being added in: changed to `@since 7.2026.10` in dad1778.

luiseufrasio:
@Pandrex247 confirmed, and fixed for the `server` element in b27a87a2.

**Cause.** When an instance parses the `domain.xml` it synced from the DAS, `InstanceReaderFilter` keeps the `server` and `config` elements of its *fellow group members*, not only its own (`handleServer` checks `getDGServerNames()`, `handleConfig` checks `isConfigFromServerInDG`). Both clauses pre-date this PR. Adding the `dg-server-ref` on its own therefore left a running member in a state it would never be in after a restart: a reference pointing at a server it held no `server` element for.

**Fix.** `add-instance-to-deployment-group` now replicates `_register-instance-at-instance` between the joining instances and the group's running members, which recreates the `server` element and the node it references. That is the same command `create-instance --deploymentgroup` already replicates to a group's members through `PostRegisterInstanceCommand`, so the two ways of joining a group now converge the same way. Only pairs where one side is joining are contacted, only running members are targeted, and offline ones are skipped silently.

**Result** on your reproducer (create and start Insty1, create the group, add Insty1, create Insty2, add it, start it), comparing Insty1, which learned live, against Insty2, which got a full sync at startup:

| | Insty1 (live) | Insty2 (startup) |
|---|---|---|
| `server` elements for Insty1 and Insty2 | both | both |
| Insty2's system-properties on Insty1 | yes | yes |
| `dg-server-ref` x2 | yes | yes |
| `config` element `Insty2-config` | **no** | yes |
| `resource-ref`s under `server` Insty2 | **no** | yes |

Insty1 now holds Insty2's `server` element including its ports, so it knows what Insty2 is without being restarted. Its `server.log` is clean, no warnings and no exceptions.

**Still open: the `config` element.** `_register-instance-at-instance` only sets the new server's `config-ref`, it never creates the config itself, because members of a *cluster* all share one config and never needed it. Group members each have their own, so Insty1 holds `config-ref="Insty2-config"` without that config until it restarts. This is not a regression from this PR: `create-instance --deploymentgroup` produces exactly the same state in shipped builds today.

**@Pandrex247 this is the one open decision, and it is yours to make.** There are two ways to make the live state and the post-restart state agree, and they go in opposite directions:

1. **Push more at runtime.** Sync the joining member's `config` element (and its `resource-ref`s) to the running members too. Fully matches the current post-restart state, but no command transports a whole config element today, so this needs a new one.
2. **Narrow the startup filter.** Stop keeping a fellow member's `config` and `resource-ref`s at parse time. An instance never acts on another member's behalf, so it arguably never needed them; it only has them because the filter copies whole elements. Much smaller change, but it alters behaviour that pre-dates this PR.

Which do you want, and do you want it in scope for this PR or as a follow-up? I have no strong preference on 1 vs 2, but I would rather not guess, since 2 changes existing behaviour.

Pandrex247 (changes requested):
Still not syncing the actual instance config

luiseufrasio:
@Pandrex247 The instance config is now synced live.

A deployment-group member's `<config>` is transported to the group's running members through a new instance-side command `_copy-config-at-instance`: the DAS serializes the config bean and each running member re-parses it into its own live configuration and persists it, mirroring how config-modularity instantiates config snippets at runtime (`ConfigurationParser#parseAndSetConfigBean`). It runs ahead of `_register-instance-at-instance`, so the `config-ref` on the newly registered `<server>` resolves immediately instead of dangling until a restart. Both join paths are covered: `add-instance-to-deployment-group` and `create-instance --deploymentgroup` (the latter guarded to the DG case, since cluster members share a single config).

`config.deepCopy(...)` / `copy-config` could not be reused on the member: they need the source `Config` bean present in the same JVM, and a member has neither the joining member's config nor `default-config` (`InstanceReaderFilter` filters it out of the synced domain.xml) — hence the serialize/transport/parse approach.

Verified on a two-instance group. With Insty1 already running, `add-instance-to-deployment-group --instance Insty2 dg1` leaves Insty1's own `domain.xml` holding both the server and the config of Insty2, without restarting Insty1:

```
<server config-ref="Insty2-config" node-ref="localhost-domain1" name="Insty2">
<config name="Insty2-config"> ... </config>
```

and symmetrically Insty2 ends up with `Insty1-config`.

One limitation worth flagging: this syncs the config at join time. If a member's config is edited on the DAS afterwards, the other running members keep the join-time snapshot until they restart. Propagating later config edits across the whole group is a larger, separate change — let me know if you'd like that tracked as a follow-up.

Pandrex247 (changes requested):
Breaks expected warning behaviour

Create an instance, start it, stop it, create a deployment group with no instances - no warnings: good.

Create an instance, start it, stop it, create a deployment group with the newly created instance - no warnings: bad. There should be a warning in this case - the command was legitimately not synced across and the instance needs to be started with sync full or sync normal as its config is no longer up-to-date.

luiseufrasio:
Fixed in eea69d1.

The offline warning was being suppressed by `ifOffline = FailurePolicy.Ignore` on `add/remove-instance-from-deployment-group`, which was carried over from when these commands replicated to `RuntimeType.ALL` (where it stopped every unrelated offline instance in the domain from warning). Now that replication is scoped to the group's members via `@TargetType(DEPLOYMENT_GROUP)`, that override is wrong: the offline instance being contacted is a genuine member that legitimately missed the change, so it should warn. I removed the override on both commands and left `@ExecuteOn` at its defaults (`ifOffline = Warn`, `ifNeverStarted = Ignore`, `ifFailure = Error`). A never-started member still doesn't warn, since it syncs its whole config from the DAS on first start anyway; and because replication no longer fans out to ALL, the warning can only ever reach an actual group member, not an unrelated offline instance.

Re-tested your scenarios via CLI:
- Create an instance, start, stop, create a deployment group with no instances → no warnings.
- Create an instance, start, stop, add it to a deployment group → `Warning: Instance Insty1 seems to be offline; command add-instance-to-deployment-group was not replicated to that instance / Command add-instance-to-deployment-group completed with warnings.`

While testing this in the admin console I hit a separate, pre-existing UI bug: clicking **Save** on the deployment-group Instances tab raised "Cannot find specified Cluster Resource". The shared instance-weight Save button (`saveInstanceWtButton.inc`) always called `gf.isClusterName`, but the deployment-group and standalone instance pages have no cluster name. Fixed in 78597bf by guarding the check on the parent each page actually has (cluster → `isClusterName`, deployment group → `isDGName`, standalone → none). It's unrelated to the replication change but reproduces on the DG workflow this PR touches, so I kept it here as its own commit.

Pandrex247 (changes requested):
Deleting a deployment group doesn't cause a warning on offline instances that were a part of it.

Instances don't lose the references to those instance's config or server entries until restart

Removing an offline instance from a deployment group doesn't cause a warning

Adding an offline instance to a deployment group via Deployment Groups > ${group} > Instances > _Select a Server to add to this Deployment Group_ doesn't cause a warning.

luiseufrasio:
Thanks for the review. All four points are addressed on the branch:

**1. Deleting a deployment group didn't warn its offline members** (`5ade280`)
A new `_pre-delete-deployment-group` supplemental (runs `Before` `delete-deployment-group`, DAS-only) routes the whole membership out through a single `remove-instance-from-deployment-group` call before the group is deleted. Its running members converge live and each offline member is warned exactly once. It is one call rather than one-per-member so an offline member is not warned twice — once as a still-present member while a fellow member ahead of it leaves, and again as the departed member.

**2. Instances kept the removed instance's config/server until restart** (`8d74007`)
Dropping the `dg-server-ref` alone left the departed member's `<server>` and `<config>` (the footprint transported when it joined) behind. `dropDepartedMemberFootprint` now removes them too — on a member that stays in the group, and on the removed instance itself (which also drops the now-inert group element). The group element is removed first, otherwise `DeploymentGroup.getReference()` reports its first member's `config-ref` and inflates that config's reference count, blocking the `<= 1` cleanup guard that decides whether the auto `<name>-config` can go.

**3. Removing an offline instance didn't warn** (`8d74007`)
The framework only replicates to *current* members, so a just-removed offline instance was silent. `warnOfflineDepartedInstances` now emits the same "seems to be offline" warning on the DAS for departed instances that are offline (never-started ones are skipped, matching `ifNeverStarted = Ignore`, since they full-sync on first start). `delete-instance` passes `notifyoffline=false` because the instance it removes is expected to be stopped.

**4. Adding/removing an offline instance via the console didn't warn** (`ac4f4ad`)
Both actions now carry the alert across the post-action redirect. Remove additionally reads the warning from the report message, because the replication framework resets the command's exit code to `SUCCESS` (it ran on the DAS only, or converged the running members), so `RestUtil.parseResponse` would otherwise take its SUCCESS branch and show nothing.

All four scenarios were verified manually via CLI and the admin console.

PR #8432, opened 2026-09-16
FISH-14642 Allow %5C (encoded backslash) in request URIs per RFC 3986

### LINKED ISSUE #8142: Bug Report: %5C in path parameter produces 400 Invalid URI
### Brief Summary

Sending the escape sequence `%5C` as part of the value of a path parameter produces the response status `400 Invalid URI` in Payara 7.2026.4, but `200 OK` in Jersey 4. As `%5C` is a valid escape sequence, Jersey is correct here, but Payara makes a mistake: While it certainly should reject a plain-text backslash `\`, it MUST NOT reject a percent-encoded one. As Jersey works well here, the bug must be somewhere in Payara itself.

I even copied the original (correctly working) `jersey-container-grizzly2-http` JAR file into `<install-dir>\payara\glassfish\modules`, but the failure persists to happen.

### Expected Outcome

Payara 7 should respond with `200 OK`, just like Jersey 4 does.

### Current Outcome

Payara 7 responds with `400 Invalid URI`; Jersey responds with `200 OK`.

### Reproducer

[reproducer.zip](https://github.com/user-attachments/files/27600300/reproducer.zip) contains a self-explanatory minimalistic JAX-RS echo application.

After building with `mvn clean package` autodeploy it on Payara 7. Then run `curl -v http://localhost:8080/reproducer-0/echo/foo%5Cbar` to see the result: `400 Invalid Uri`.

When instead starting standalone (e. g. in IDE running `Main.java`), it uses Jersey 4 (Grizzly Http Container) to bootstrap  and will answer to requests with `200 OK` for ` curl -v http://localhost:9999/echo/foo%5Cbar` (and prints the expected echo on screen: `foo\bar`).

### Operating System

Windows 11 Pro

### JDK Version

Zulu 21 (64 Bit)

### Payara Distribution

Payara Server Web Profile

lprimak:
fixes #8142

## Description
Requests whose path contains a percent-encoded backslash (`%5C`) fail with `400 Invalid URI`, because Grizzly's `HttpRequestURIDecoder` decodes the URI before normalizing it and then rejects the resulting `\`. A JAX-RS path parameter carrying JSON with escaped quotes is enough to trigger it. The same application works on standalone Jersey, which never runs Grizzly's decoder.

The Grizzly side is fixed in eclipse-ee4j/glassfish-grizzly#2316 by making the check opt-in via the `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH` system property: when set, a decoded backslash is kept as path data (RFC 3986) instead of being rejected or rewritten to `/`. A literal, unencoded `\` in the request line is still rejected.

**Changes**

- Set `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH=true` in both domain templates (`gf_template`, `gf_template_web`), for `server-config` and `default-config`, so new domains get RFC-correct behavior out of the box.
- `CoyoteAdapter.ALLOW_BACKSLASH` now also honors `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH`, in addition to the legacy `org.glassfish.grizzly.tcp.tomcat5.CoyoteAdapter.ALLOW_BACKSLASH`, so the Tomcat-compat normalization path and Grizzly's agree on one switch.

**Security note**
*Why this is safe to enable by default*

`StandardContextValve.normalize()` already treats `\` as a path separator when evaluating the `/WEB-INF` and `/META-INF` guard, and rejects any path that resolves above the context root. So `..%5C..%5CWEB-INF%5Cweb.xml` is refused before it reaches the default servlet (verified: 404), and a backslash traversal is bounded exactly like a slash one. The decoded request path itself is not rewritten — `\` still reaches the application as data. A literal, unencoded `\` in the request line remains rejected by Grizzly regardless of this setting.

**Existing domains**

Domain templates only affect newly created domains. Upgraded installations keep the old behavior unless they opt in:

    asadmin create-system-properties org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH=true

Setting the property to `false` (or removing it) restores the previous behavior.

## Important Info
**Dependency on Grizzly**

This change is inert until Payara consumes a Grizzly release containing eclipse-ee4j/glassfish-grizzly#2318, which makes `HttpRequestURIDecoder` honor `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH`. Until then, `%5C` is still rejected by Grizzly before the request reaches `CoyoteAdapter`, and the property has no effect — so merging in either order is safe, but this PR does not resolve #8142 on its own.

## Testing
Reproducer at https://github.com/flowlogix/backslash-rest-reproducer
Against a JAX-RS echo resource with `curl --path-as-is`:

| Request | Before | After |
|---|---|---|
| `/echo/foo%5Cbar` | 400 | 200 `foo\bar` |
| `/echo/a%5C..%5Cb` | 400 | 200 `a\..\b` |
| `/echo/foo\bar` (literal) | 400 | 400 |
| `/echo/..%5C..%5CWEB-INF` | 400 | 404 |
| `/echo/../WEB-INF/web.xml` | 404 | 404 |

### Testing Environment
Any

## Documentation
TBD: The new `org.glassfish.grizzly.http.util.HttpRequestURIDecoder.ALLOW_BACKSLASH` needs to be documented

## Notes for the reviewers:
This PR can be merged now, and when Grizzly is updated, the changes will take effect.
There is no specific order of merging this PR vs. Grizzly PR.


mkarg:
Thank you, @lprimak, for this fix! You are awesome! ❤️

Pandrex247:
`com.sun.enterprise.web.` is an old namespace now.
My initial thought was that this should be a fish.payara, but I see that it aligns with a Grizzly property.

I can see on the Grizzly PR a similar comment has been made though suggesting it use a `org.glassfish.grizzly` name

lprimak:
@Pandrex247 This has been done. Got an approval for Grizzly PR as well. Thank you!

> `com.sun.enterprise.web.` is an old namespace now. My initial thought was that this should be a fish.payara, but I see that it aligns with a Grizzly property.
> 
> I can see on the Grizzly PR a similar comment has been made though suggesting it use a `org.glassfish.grizzly` name

lprimak:
Grizzly PR has been merged!

PR #8443, opened 2026-09-21
FISH-14555: Fix add jvmOptions in admin console

raushan606:
## Description

The JVM Options page in the admin console (server-config → JVM Settings → JVM Options) silently drops options that share a prefix with an existing option. Adding, for example, --add-opens=java.naming/javax.naming=ALL-UNNAMED and clicking Save does nothing — the option never reaches the configuration.

Root cause: JVM options were transported from the console to the backend as a Map keyed by the option name (the text before the first =). Every --add-opens=… option produces the same key --add-opens, so they overwrite each other in the map and all but one are dropped before the REST call is made. The same happens for --add-exports=…. A modern JDK config ships with ~18 such options, so adding another one just collides with the existing ones.

Fix: Use the whole (escaped) option string as the map key with an empty value. Option strings are unique, so nothing collides, and `ResourceUtil.processJvmOptions` already reconstructs options from that shape. Applied in the three places that had the collision:
- InstanceHandler.prepareJvmOptionPayload — the Save button
- SecurityHandler — the Security Manager toggle (re-saves all options)
- CollectionLeafResource.deleteExistingOptions — the REST delete/restorepath


pdudits (approved):
Works for me

PR #8448, opened 2026-09-21
FISH-14585: Extension-based Servlet mappings produce incorrect HTTP span name (missing slash before wildcard)

raushan606:
## Description

For web applications using extension-based servlet mappings (e.g. *.jsf), the OpenTelemetry span name and http.route attribute were missing a / separator between the context path and the wildcard pattern. For example, a request to /petclinic/home.jsf would produce GET /petclinic*.jsf instead of the correct GET /petclinic/*.jsf.

This is a one-line fix that adds the missing / when the servlet pattern does not already start with one.
### Testing Performed

- Deployed petclinic (JSF app with *.jsf mapping) on Payara 7
- Sent requests and verified spans in Jaeger
- Before fix: span name was GET /petclinic*.jsf
- After fix: span name is GET /petclinic/*.jsf

### Testing Environment

Zulu JDK 21 on macOS with Payara 7
## Notes for Reviewers

The fix is in `OtelSupport.resolveServletBase()` — the single line that returns the extension pattern now prepends / if the pattern doesn't already start with one.


pdudits (approved):
If MP telemetry is happy, I'm happy as well.

This shouldn't affect JAX-RS, because this is active only when there's no route explicitly set by downstream framework.

